PR #16, opened 2015-01-27
Write the git revision for squash reporting

cbeer:


atz:
Will reorient pull request to `develop` instead of master

PR #30, opened 2015-05-04
Rescue index_queue_depth if unable to connect to the remote source

cbeer:
@atz @jmartin-sul This is the hot-fix I applied to argo-prod. It should be merged and deployed the right way.


atz:
Also picked over to develop branch.

PR #43, opened 2015-10-20
Remove initializer that opens up Standard String class, in favor for a…

mejackreed:
…lready shipping Rails helper methods

http://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-truncate


mejackreed:
Also, I'm not really sure how to test this. I can't actually access a show page on my dev environment. Halp?

atz:
:+1: to getting rid of ellipsize monkey-patch

jmartin-sul:
yes, definite :+1: 

@mejackreed, when you say you can't access a detail page... do you mean you don't have objects in fedora/solr to view, or you're getting errors when you go to a detail page for an object?

jmartin-sul:
i pulled this branch and used it on my laptop, and it works fine:  the same five tests as usual fail, and it works when i test it manually by navigating to an item with the lengthy title of "English land and English landlords: an enquiry into the origin and character of the English land system, with proposals for its reform".  `truncate` does a fine job at ellipsizing, though it seems so plain in comparison to a re-opened String class.

state of the art in testing and deployment.

PR #46, opened 2015-10-20
rename the default blacklight document handler from "document" to "/document"

jmartin-sul:
* rename the handler in solrconfig.xml (requires re-deploying configuration for the solr instance for the change to take effect)
* update CatalogController to override the default blacklight config value for document_solr_request_handler

once this code change is pulled, the solr config for the relevant instance will have to be redeployed, or the instance won't be able to view detail pages for objects, because it'll be using an undefined handler.

@atz, mind reviewing?

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -18,6 +18,8 @@ class CatalogController < ApplicationController
     BlacklightConfigHelper.add_common_default_solr_params_to_config! config
     config.default_solr_params[:rows] = 10
 
+    config.document_solr_request_handler = '/document'
+
     config.index.title_field = 'dc_title_ssi'
     config.index.display_type_field = 'content_type_ssim'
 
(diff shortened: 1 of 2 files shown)


mejackreed:
Quick question, why is this necessary? What is broken/being fixed by this?

jmartin-sul:
@cbeer says this is the new style for solr, and that the old style (without the slash) is deprecated.  tl;dr, the new style got deployed to argo-dev (and argo-prod-b, and argo-stage) with some solr configuration changes last week (none of the solrconfig.xml files in the sul-solr-configs or argo repos were appropriate as-is for deployment to SolrCloud, but that particular change made it to the final version based on the advice i got for which changes to keep).  my initial instinct was to revert the solr config change, since it goes against BL's default, but chris pushed back and convinced me to have argo override BL's default in favor of the new solr style.

fuller explanation here:  https://github.com/sul-dlss/sul-solr-configs/pull/38

i personally don't have strong feelings about this at all.  whatever gets the web code and the solr config in sync is fine by me.  overriding BL's default strikes me as being just about as objectionable as using a deprecated but supported URL style, and neither seems particularly objectionable.

mejackreed:
:+1: just wanted the context.

jmartin-sul:
@atz says all good, merging.

atz:
:8ball:

PR #50, opened 2015-10-20
point CatalogController spec to an object that is actually fixture specs

mejackreed:
Fixes 2 failing specs from:

 - `./spec/controllers/catalog_controller_spec.rb:44`
 - `./spec/controllers/catalog_controller_spec.rb:49`


jmartin-sul:
after further discussion:  deferring for now, we're going to try and figure out why the expected fixtures aren't loading in the first place (since they're there, and in the druids list).

PR #51, opened 2015-10-21
dor-services fixes for URL construction (http to https); minor coveralls and newrelic upgrades

jmartin-sul:


mejackreed:
What changes with dor-services here?

jmartin-sul:
whoops, sorry, should've given more explanation...

the dor-services changes are in this PR:  https://github.com/sul-dlss/dor-services/pull/96/files

it's actually just a 5.x port of laney's similar changes for 4.x (though those also involved rubocop'ing that'd already been performed on 5.x, which is why i didn't just cherry pick those commits):  https://github.com/sul-dlss/dor-services/pull/95

PR #53, opened 2015-10-21
Skip integration tests when running rspec command.

tingulfsen:
This is intended to get Travis to succeed, as well as skip integration tests when running the regular rspec command.

### DIFF
--- a/spec/integration/bulk_jobs_spec.rb
+++ b/spec/integration/bulk_jobs_spec.rb
@@ -17,7 +17,7 @@ describe ModsulatorJob, type: :job do
     FileUtils.rm_rf(@output_directory) if(Dir.exist?(@output_directory))
   end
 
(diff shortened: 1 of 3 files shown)


spec/integration/bulk_jobs_spec.rb
```diff
@@ -31,7 +31,7 @@
                   'xlsx',
                   'true',
                   'anote')
-
+puts "tommy: here2"
```
mejackreed: Is this `puts` statement leftover cruft?

mejackreed:
:+1: besides my comment about the `puts` statement

tingulfsen:
Thanks for catching that, I shouldn't be leaving that stuff around.

jmartin-sul:
:+1:

PR #55, opened 2015-10-22
Adding a few settings required for spreadsheet bulk upload.

tingulfsen:
These are just a few new settings we need to have for the bulk upload stuff to work.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -60,8 +60,9 @@ Argo::Application.configure do
       spreadsheet           Settings.SPREADSHEET_URL
     end
     bulk_metadata_directory Settings.BULK_METADATA.DIRECTORY # Directory for storing bulk metadata job info
-    bulk_metadata_log       Settings.BULK_METADATA.LOG # Bulk metadata log file
-    bulk_metadata_xml       Settings.BULK_METADATA.XML # Bulk metadata XML output file
+    bulk_metadata_log       Settings.BULK_METADATA.LOG       # Bulk metadata log file
+    bulk_metadata_csv_log   Settings.BULK_METADATA.CSV_LOG   # Bulk metadata log file in CSV format for end users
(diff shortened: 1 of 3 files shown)


config/environments/test.rb
```diff
@@ -60,8 +60,9 @@
       spreadsheet           Settings.SPREADSHEET_URL
     end
     bulk_metadata_directory Settings.BULK_METADATA.DIRECTORY # Directory for storing bulk metadata job info
-    bulk_metadata_log       Settings.BULK_METADATA.LOG # Bulk metadata log file
-    bulk_metadata_xml       Settings.BULK_METADATA.XML # Bulk metadata XML output file
+    bulk_metadata_log       Settings.BULK_METADATA.LOG       # Bulk metadata log file
```
mejackreed: Are these required for the tests to run?

mejackreed:
Could these two commits be squashed into one?

tingulfsen:
Yeah, I can squash them. Come to think of it, none of the bulk_metadata variables are required for the tests to run, so I can remove that stuff and squash everything into a single commit if that's preferred.

mejackreed:
Sounds good!

jmartin-sul:
@tingulfsen, re: "none of the bulk_metadata variables are required for the tests to run"

is it that the bulk metadata variables aren't needed by the tests now that integration tests are skipped (PR #53)?  specifically adding those parameters helped get rid of some failures earlier this week (see https://github.com/sul-dlss/argo/commit/85a002a64fd7265dbb748acedf87413e9a265d3f).

though, as per a message i just posted in the argo slack channel, i think what we actually want to do is incorporate the changes from the second commit in this PR, and then get rid of test.rb.example (which would require getting rid of the reference to it in .travis.yml).  even if PR #53 made those settings moot in test.rb, i think we still want to get rid of test.rb.example (via a different PR than this one).

jmartin-sul:
:+1: seems like the right thing, tests look good

PR #56, opened 2015-10-23
get rid of test.rb.example, dor.rb.example, and corresponding travis usages

jmartin-sul:
i believe these things are all now obsolete, thanks to the config refactorings from this week


mejackreed:
:+1:

PR #58, opened 2015-10-28
configuration clean up and refactoring, to support deployment to canonical instances

jmartin-sul:
* drop cap deployment type of "test" with rails env of "testing".  trade it for a deployment to "stage" (currently only argo-stage-b.stanford.edu) with a rails env of "staging".
* the "test" environment is now meant solely for unit tests.  it is possible that this will allow some clean-up and trimming.
* development.rb & test.rb: config.assets.compress is now controlled by Settings.APPLICATION.ASSETS.COMPRESS (defaults to true in root settings.yml).
* development.rb & test.rb: likewise for config.cache_classes and Settings.APPLICATION.CACHE_CLASSES (defaults to true)
* dor_development.rb & dor_test.rb: add fields from deployed canonical instances that hadn't made it to argo configs, drop obsolete fields, minor style edits for consistency.
* add dor_production.rb and production.rb, modeled off of dor_development.rb and development.rb, with changes as appropriate.
* move comment to squash_exceptions.rb (since the code it used to accompany was moved there).
* new APPLICATION section in root settings.yml, for Rails stuff.
* remove obsolete DorGenerator that would step on new config structure.

@atz @mejackreed  @tingulfsen


jmartin-sul:
thoughts on what hound's barking about?  my gut feeling is that if i change it, i should use DateTime.now.strftime.current.  or should i do something else?

config/environments/development.rb
```diff
@@ -4,12 +4,12 @@
 Argo::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
 
-  config.assets.compress = false
+  config.assets.compress = Settings.APPLICATION.ASSETS.COMPRESS
```
mejackreed: Do we need to manage this using the rails config gem?

config/environments/development.rb
```diff
@@ -4,12 +4,12 @@
...
 
   # In the development environment your application's code is reloaded on
   # every request.  This slows down response time but is perfect for development
   # since you don't have to restart the webserver when you make code changes.
-  config.cache_classes = false
+  config.cache_classes = Settings.APPLICATION.CACHE_CLASSES
```
mejackreed: Same comment as LN7

config/environments/development.rb
```diff
@@ -33,17 +44,23 @@
...
-  config.middleware.use(
-    Rack::Webauth::Test,
-    user: Settings.WEBAUTH.USER,
-    mail: Settings.WEBAUTH.MAIL,
-    ldapprivgroup: Settings.WEBAUTH.LDAPPRIVGROUP,
-    suaffiliation: Settings.WEBAUTH.SUAFFILIATION,
-    displayname: Settings.WEBAUTH.DISPLAYNAME,
-    ldapauthrule: Settings.WEBAUTH.LDAPAUTHRULE
-  )
+  if Settings.WEBAUTH
```
mejackreed: Since `Settings.WEBAUTH` is in `settings.yml` this will always evaluate true. Do we need this check?

config/settings.yml
```diff
@@ -5,6 +5,13 @@ CACHE_STORE_NAME: 'argo'
 DATE_FORMAT_STR: '%Y-%m-%d %H:%M:%S.%L'
 REINDEX_ON_FLY: false
 
+# Rails application settings
```
mejackreed: I'm a bit -1 about having these application configurations here. I think we should use the Rails default configurations per environment checked. Not a show stopper, but this feels a bit off to me.

config/environments/development.rb
mejackreed: Same as previous comment

config/environments/production.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  # the following config statement gets us two important things on each log line:
+  # 1) unique IDs per request, so that it's easier to correlate
+  # logging statements associated with a given request when multiple
+  # simultaneous requests have interleaved log statements.
+  # 2) time stamps on each log statement, because knowing when something
+  # got logged is quite nice.
+  # (declared after Argo.configure since it uses one of those params)
+  config.log_tags = [:uuid, proc {DateTime.now.strftime(Argo::Config.date_format_str)}]
```
mejackreed: :+1: to using `DateTime.now.strftime.current`... Is this going to mess with anything legacy though?

config/environments/development.rb
jmartin-sul: at present, it's a setting that differs between default laptop configurations and argo-dev's deployed configuration, though both instances go by same the rails env name ("development").  i guess we could make it false in both setups?  though i think the difference is actually reasonable (there's presumably a slight performance gain, and editing code in place on argo-dev is rare).  or, we could differentiate laptop configs from the deployed "development" environment by using a different env name, though they are largely similar.  and adding another environment config file like that if it isn't really needed feels a little worse (to me) than adding new settings entries.

i don't know... this is a basic rails setting, so it seems like an appropriate thing to do, but i'm fine switching approaches.  i think my next choice, for simplicity's sake, would be to just make it false in all cases for this env and get rid of the setting.  but i vote for leaving it as it is in this branch, unless i'm missing something.

config/environments/development.rb
jmartin-sul: same reply here too

config/environments/development.rb
mejackreed: :ok_hand:

config/environments/development.rb
mejackreed: :ok_hand:

config/settings.yml
mejackreed: :ok: per previous comments

config/environments/development.rb
jmartin-sul: ah, yeah, ok.  well, we need something along these lines, because we want to use Rack::Webauth::Test to fake webauth info on laptops, but not on argo-dev (where we want webauth to pass along real info).

i guess you could couple this with LN7 and LN12 to argue for a "laptop" environment?  i'm not sure whether a proliferation of environments is better than a proliferation of settings.  to me, one nice thing about settings is the that an "environment" can support slightly divergent deployments to specific instances.

i don't have particularly strong feelings on that tradeoff in general.  in this case, i'd vote for the approach taken so far on this branch (new settings instead of new environments).  laptops and argo-dev feel more similar than different (and talk to some of the same services).  and adding new settings feels lighter than adding new environments.  but i could also understand a distaste for having conditional statements in config files.

mejackreed: Yes agree... we should then in argo-dev sul-configs configure `config/development.yml` to have the deployed config values. the config gem will then use that as the webauth settings for the deployed app.

jmartin-sul: hmm, not sure i understand the suggestion... for argo-dev, we don't need config values for, e.g., Settings.WEBAUTH.USER, just something to keep the config from invoking config.middleware.use(Rack::Webauth::Test, ...).  i.e. the only instances with config values for things like Settings.WEBAUTH.USER are laptops.

if we can't nil out Settings.WEBAUTH, can you think of a cleaner approach than something like setting Settings.WEBAUTH.USER to empty and checking for emptiness?

i guess another approach would be to have Settings.WEBAUTH default to nil in the root settings.yml, and have laptop configs set things like USER.

mejackreed: Gotcha. Seems like we can nil out `Settings.WEBAUTH`

```ruby
Settings.WEBAUTH = nil
```

or we could just set this on the server?

```ruby
# in config/development.rb
WEBAUTH: nil
```

Thought both of these solutions seem strange to provide to essentially disable local development configurations. Happy to punt on this and come back to it.

jmartin-sul: ok... i had trouble overriding existing settings with nil.

since laptops are the only instances that set WEBAUTH info, i made it nil by default in the root settings.yml, with a comment about the expected structure (for local settings that override the default).
https://github.com/sul-dlss/argo/pull/58/files#diff-646ad6b10917e4385cbcc8c8524e24a2R88

the private repo with the vanilla laptop configs already has the WEBAUTH settings override needed to fake webauth credentials on laptops.

how's that sound?

mejackreed: :+1: :+1: great solution

jmartin-sul: thanks!

any other stuff left to clean up with this one before merging?  i think this is the last of the issues raised so far...

mejackreed: All looks good to me... Maybe squash a few commits down? Besides that.. I'll merge after squashed and tests run :)

jmartin-sul:
squashed and pushed

PR #60, opened 2015-10-28
[WAITING] Clean repo... should provide for clean repository when running in development and running tests

### LINKED ISSUE #59: git repo tweaks for db/schema.rb and workflowservice.log
On a 'clean' checkout of the develop branch, after running the specs, the git repo isn't clean, i.e.

```
$ git status
On branch develop
Your branch is up-to-date with 'origin/develop'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

    modified:   db/schema.rb

Untracked files:
  (use "git add <file>..." to include in what will be committed)

    workflow_service.log

no changes added to commit (use "git add" and/or "git commit -a")
```

1. Should `db/schema.rb` be in the .gitignore file? If not, should changes be committed?
2. Can the config settings be changed to put the `workflow_service.log` into a log path or somewhere that is already in .gitignore?

mejackreed:
Closes #59

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ public/.htaccess
 public/assets
 public/images/.dpg_pool
 tmp/
+workflow_service.log*
 
 config/settings.local.yml
 config/settings/*.local.yml
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -23,7 +23,7 @@ ActiveRecord::Schema.define(version: 20150619200031) do
     t.string   "document_type", limit: 255
   end
 
-  add_index "bookmarks", ["user_id"], name: "index_bookmarks_on_user_id", using: :btree
+  add_index "bookmarks", ["user_id"], name: "index_bookmarks_on_user_id"
 
   create_table "delayed_jobs", force: :cascade do |t|
     t.integer  "priority",   limit: 4,     default: 0, null: false
@@ -39,7 +39,7 @@ ActiveRecord::Schema.define(version: 20150619200031) do
     t.datetime "updated_at"
   end
 
-  add_index "delayed_jobs", ["priority", "run_at"], name: "delayed_jobs_priority", using: :btree
+  add_index "delayed_jobs", ["priority", "run_at"], name: "delayed_jobs_priority"
 
   create_table "indexing_exceptions", force: :cascade do |t|
     t.string   "pid",                  limit: 255
@@ -50,7 +50,7 @@ ActiveRecord::Schema.define(version: 20150619200031) do
     t.datetime "updated_at"
   end
 
-  add_index "indexing_exceptions", ["pid"], name: "index_indexing_exceptions_on_pid", using: :btree
+  add_index "indexing_exceptions", ["pid"], name: "index_indexing_exceptions_on_pid"
 
   create_table "robots", force: :cascade do |t|
     t.string   "wf",         limit: 255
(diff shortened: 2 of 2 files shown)


darrenleeweber:
Yes, that should resolve #59; :+1:

darrenleeweber:
I guess the only thing that bugs me about the .gitignore for `workflow_service.log` is the alternative solution of creating a setting (somewhere) that places the file into `./log/workflow_service.log` (then it should be ignored along with all the other log files in there).

mejackreed:
Agree @darrenleeweber that would be the more elegant solution. From my understanding there were some infrastructure concerns that expect that file to be at the root. @jmartin-sul can you confirm?

.gitignore
```diff
@@ -19,6 +19,7 @@ public/.htaccess
 public/assets
 public/images/.dpg_pool
 tmp/
+workflow_service.log*
```
darrenleeweber: The `.gitignore` already contains `log/*.log` - is there a setting that can be changed to put the `workflow_service.log` file into the `log` path?

darrenleeweber:
The only other log setting example that I could find quickly is in
```
config/initializers/delayed_job.rb
9:Delayed::Worker.logger = Logger.new(File.join(Rails.root, 'log', 'delayed_job.log'))
```

mejackreed:
I think it might be a dor services thing and not an argo thing. `¯\_(ツ)_/¯`

jmartin-sul:
redirecting the question about workflow_service.log to @tingulfsen and @atz.  maybe they know offhand?  i'm not sure what generates that log file, though i'd agree that it'd be nice to have it in log/ instead of in the root of the project.

FWIW, it does appear to be a dor-services thing, because that same log file is generated in the local dor-services project directory if you check it out and run the unit tests.

mejackreed:
Ping.. are we held up by the question to @atz and @tingulfsen ?

jmartin-sul:
i believe so.  if they don't know offhand, i guess someone just needs to look into how dor-services generates that file, and whether it's configurable, and what might expect it to live in its current location.

i was about to ssh to argo-dev and argo-prod to see where it shows up there, but i'm having VPN trouble, and so can't at the moment.

darrenleeweber:
We're also held up because unit tests are failing and it's not clear whether those failures might be related to any changes in this PR or not.  Although my guess is no, I'm not going to take responsibility for that call.

mejackreed:
@darrenleeweber no I think we are help up here because an upstream change and config change will allow us to configure location of `workflow_service.log`

mejackreed:
@darrenleeweber re: tests failing. Hoping that some of the stuff that @jmartin-sul has been working on will clear that up. and should be the subject of another pr

darrenleeweber:
It looks like argo is using a default logger, see
https://github.com/sul-dlss/dor-workflow-service/blob/75c7b46ccc499669d7c7ecfa535b7afb6bc0f70a/lib/dor/services/workflow_service.rb#L436-L442

mejackreed:
#91 now addresses the `db/schema.rb` side of this.

darrenleeweber:
https://github.com/sul-dlss/dor-workflow-service/pull/19 attempts to fix 'log/workflow_service.log' issue.

.gitignore
darrenleeweber: See also https://github.com/sul-dlss/dor-workflow-service/pull/19

jmartin-sul:
should this be closed?  comments yesterday from @mejackreed and @darrenleeweber make it sound like both the issues addressed here are being addressed separately elsewhere?

atz:
Yes, I believe all issues here are cleanly addressed elsewhere.  Closing.

PR #61, opened 2015-10-28
Adding view test for spreadsheet upload form and cleaning up links/pa…

tingulfsen:
…ths a little.

### DIFF
--- a/spec/views/catalog/bulk_upload_form.html.erb_spec.rb
+++ b/spec/views/catalog/bulk_upload_form.html.erb_spec.rb
@@ -1,0 +1,31 @@
+require 'spec_helper'
+
+RSpec.describe 'catalog/bulk_upload_form.html.erb', :type => :view do
+
+  before :each do
+    @obj = double('dor_object', id: 'druid:hv992ry2431')
+  end
+
+  it 'has the correct title' do
+    assign(:object, @obj)
+    render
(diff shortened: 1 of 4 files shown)


spec/views/catalog/bulk_upload_form.html.erb_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'spec_helper'
+
+RSpec.describe 'catalog/bulk_upload_form.html.erb', :type => :view do
+
+  before :each do
+    @obj = Dor.find 'druid:hv992ry2431'
```
mejackreed: Could this be mocked using a double? eg.
```ruby
let(:current_object) { double('object', { id: 'druid:abc123' }) }
# then in the before statement you can assign it, removing the need to do so in each `it` block
before :each do
  assign(:object, current_object)
end
```
This would remove the need to call `Dor.find` and potentially separate the view spec from the Dor.find dependency (or that object being indexed).

jmartin-sul: i'm in favor of more mocking and less actual talking to fedora and solr in unit tests.

tingulfsen:
@mejackreed good point - I was so excited that I forgot that there is no need to query the database in this test.

mejackreed:
:laughing: tests... are exciting for me too!

PR #63, opened 2015-10-30
Fixes failing specs by providing specs with objects that are actually…

mejackreed:
… able to be indexed. Previously these object druids were not indexable by the CI build

`catalog_controller_spec.rb` also cleaned up to get rid of a lot of no-op expectations.


mejackreed:
So this didn't work :0

PR #89, opened 2015-11-02
Remove no-op blacklight hierarchy install

mejackreed:


.travis.yml
```diff
@@ -10,6 +10,5 @@ before_script:
   - if [[ -r /etc/hosts ]]; then cat /etc/hosts ; fi
   - cp config/database.yml.example config/database.yml
   - bundle exec rails generate argo:solr
```
drh-stanford: isn't this line also a no-op since `ci` calls `argo:install` which calls `rails g argo:solr`?

mejackreed:
@drh-stanford good catch. Deleted that line as well.

atz:
It's only a no-op for the blacklight-hierarchy bit if you assume that there is a match between the output form the current version of that gem and what we have committed.  In such a case, you would get generate's interactive diff tool.  I'm fine with managing BL Hierarcy that way since it is basically our inherited abondonware, but it is not *impossible* for the outputs to diverge.

mejackreed:
Agreed that they can diverge, yet I'd be suspect of a gem's `install` generator that needs to be run more than once. Specifically before CI is run.

atz:
That much is obvious.  Not the original point of this issue though.

mejackreed:
:)

PR #90, opened 2015-11-02
run travis against latest Ruby 2.2.2

mejackreed:
This should speed up our CI time significantly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: ruby
 rvm:
-  - 2.1.1
-  - 2.0.0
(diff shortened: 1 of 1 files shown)


drh-stanford:
the current stable version of ruby is 2.2.3. any reason we're using 2.2.2?

darrenleeweber:
:+1: for this PR and also for @drh-stanford for suggesting 2.2.3

mejackreed:
Yes that's the version that is on the servers. When/if servers get updated then we should update the travis build.

PR #91, opened 2015-11-02
Check in default Rails database.yml file. Since ours is linked in our…

mejackreed:
… cap deploy we don't need to mess around with this example business and copying.

Removes the need for part of #60 

Addresses part of #59


darrenleeweber:
Took a quick look at this but have to defer to someone with more authority on this project.

drh-stanford:
Looks good to me.

atz:
Why do you have production becoming an sqlite DB?

drh-stanford:
That's the default sqlite `database.yml`. In production via capistrano, we have our own `database.yml` in the shared folder. 

See 

https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/railties/lib/rails/generators/rails/app/templates/config/databases/sqlite3.yml
https://github.com/sul-dlss/argo/blob/develop/config/deploy.rb#L28

mejackreed:
Exactly what @drh said. `database.yml` is managed with puppet / cap

atz:
Why would you suggest that production runs on sqlite when it doesn't?  I get that we manage our deployed configs a certain way: that doesn't mean we should obfuscate how we expect the configuration to go.  Not that anybody outside particularly cares about this repo (it would be nice!), but as represented here, in production ENV they would be including mysql gem with no configs that actually invoke it.  That makes less sense than the mysql placeholder.

mejackreed:
@atz I wouldn't say this is misleading that we run sqlite in prod, but I can see your point. I hope the cap config should clear that up by showing that we symlink `database.yml`. This is a standard practice across DLSS. Part of the intent here, is to move to standardized way that we setup and configure other rails apps. We don't run any apps in production using sqlite and I hope seeing this file we wouldn't assume this. I added a commit that adds a comment that should hopefully clear up any reasoning to future self (or developers). Let me know if that is sufficient.

jmartin-sul:
what about removing the production section from the database.yml file that's in the argo repo?  it never gets used anyway, right?  laptops (should) never see themselves as production, and deployed argo instances (should) never use this database.yml, so i can't think of a situation where the controversial part of the config is ever actually expected by running code.

mejackreed:
Ok, taking @jmartin-sul 's suggestion and removing production section. Let me know if I can doing anything else

jmartin-sul:
:+1: sounds good to me

PR #98, opened 2015-11-02
bundle update

darrenleeweber:


mejackreed:
Without a CI build that is fulling functioning or tested in many places, I'm a little skittish of running bundle updates. @atz what has been your experience with argo on this?

atz:
It should be deliberate to avoid churning on the lock file.  Dropping platform support for 1.9 and 2.0 rubies would certainly help.  The dependency tree is huge here.  Targeted updating of gems where features/fixes are known is going to be more productive than attempted wholesale updates likely to encounter platform back-compat issues.  But by all means, if you want to try it, create a topic branch and have a go.

jmartin-sul:
whoops, sorry, i didn't notice this when i submitted PR #102 yesterday.  still, i believe that one has a slightly newer version of dor-services.

since the point of PR #102 was to update to a version of dor-services that lets unit tests pass, and since that doesn't seem to be happening on travis there, probably better to discuss that before merging/closing either of these.

atz:
I think the effect of this is already accomplished by other merged changes.  Closing.

PR #102, opened 2015-11-03
various gem updates, including dor-services

jmartin-sul:
new dor-services version should fix the errors when loading test fixtures, which should allow unit tests to pass.

a variety of other incidental gem updates.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -13,7 +13,7 @@ GIT
 
 GIT
   remote: https://github.com/sul-dlss/dor-services.git
-  revision: c1174b3cd5d877a366e7658cd5ce9b523d2c0ff5
+  revision: 327b83c2ae83088502c05b5c96dc83f1083fc54a
   branch: develop
   specs:
     dor-services (5.2.0)
@@ -39,7 +39,7 @@ GIT
       retries
       rsolr-ext (~> 1.0.3)
       ruby-cache (~> 0.3.0)
-      ruby-graphviz (~> 1.0.9)
+      ruby-graphviz
       rubydora (~> 1.6.5)
       solrizer (~> 3.0)
       stanford-mods (~> 1.1)
@@ -154,7 +154,7 @@ GEM
       rack (>= 1.0.0)
       rack-test (>= 0.5.4)
       xpath (~> 2.0)
-    childprocess (0.5.6)
+    childprocess (0.5.7)
       ffi (~> 1.0, >= 1.0.11)
     coderay (1.1.0)
     coffee-rails (4.1.0)
@@ -279,7 +279,7 @@ GEM
     method_source (0.8.2)
     mime-types (1.25.1)
     mini_portile (0.6.2)
-    minitest (5.8.1)
+    minitest (5.8.2)
     moab-versioning (1.4.4)
       confstruct
       json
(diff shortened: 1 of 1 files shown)


jmartin-sul:
this fails in travis with `getaddrinfo` errors on repo loads, which are different than the 404s that occur for laptop users doing repo loads.

while we need to fix the travis build also, i'd like not to conflate problems if they're different.  would someone mind pulling this branch on their laptop and doing a clean repo load for argo and running tests?  i'd like to see if this fix at least solves that issue, even if it doesn't solve the travis problems.  i'd really appreciate it if someone else could do that, since the tests run cleanly on my laptop now, and i'd like to see whether that's something peculiar to my setup.

@atz @mejackreed

jmartin-sul:
sounds like @drh-stanford ran into the repo load issues on his laptop yesterday, and @tingulfsen will try to help resolve that by making him the victim of my above testing request.

i think the things you want to re-do are (if you don't want to do a full reinstall):
```
rake jetty:stop #if jetty is currently running
rake argo:jetty:clean
rake argo:jetty:config
rake jetty:start
rake argo:repo:load
```

doing that allowed me to successfully load all fixtures and get a clean unit test run (after switching to the version of dor-services used by this branch).  and, as expected, there were more items available in the argo UI (though the default test data still isn't the greatest).

atz:
Note, the getaddrinfo errors are NOT overall failures because I coded that to retry (up to 3 total attempts) and apparently 1 retry is enough, because they never attempt more than twice.  So all data **is** getting loaded.  At the end of that task, all specs should be expected to perform as normal.  

Side note: please take on faith that I attempted every other means available to me to diagnose the cause of those completely intermittent "cannot DNS localhost" errors without success, before I just gave up and threw the retries block in there (which it should have anyway... but shouldn't actually need).

jmartin-sul:
ah, ok.  i do remember discussing the retries thing a while back, now that you mention it.

well, now i'm especially curious as to what people find when they test this on laptops.

mejackreed:
Running through all of @jmartin-sul steps above, I'm still seeing errors when running:
`rake argo:repo:load`

Many of the same fixtures are still not being added successfully and the steps seem to also get hung up for a while. These steps that are hung up seem to coorelate to the previous fixtures that weren't correctly being added.

jmartin-sul:
@mejackreed: a bit of a shot in the dark, but after updating argo, but before running the above, did you do a `bundle update` so that the new version of dor-services would be installed?

i'll be doing some testing today to see if i can get this figured out for both travis and laptops.  i'm a bit stumped at the moment as to why my instance is behaving differently.

mejackreed:
I didn't do a `bundle update` but I did do a `bundle install`

mejackreed:
@jmartin-sul a `bundle show dor-services` shows me:

```bash
/~.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/bundler/gems/dor-services-327b83c2ae83
```
Which I think is the correct version. I'm not on Stanford network, so that may be an issue?

jmartin-sul:
ah, yes, sorry.  i'd meant `install`.  and that does appear to be the correct version of dor-services.

but you'd mentioned on slack that fixtures loaded correctly and tests passed on your laptop once you were on the VPN?  i'm mostly mentioning that here so that we don't leave this public thread hanging, though if i'm wrong, it'd be good to know that my brain is inventing new memories.

i suspect that the travis errors are different than the laptop fixture loading errors, though i'm also guessing both are related to attempts to reach other services.

PR #103, opened 2015-11-03
More efficiently handle backlog count/display

### LINKED ISSUE #66: More efficient indexer backlog count display/retrieval
right now, the indexer backlog is shown in the header bar of argo, which is on most argo pages. displaying this count requires a network call to sulstats on the backend, and if this takes a long time to return, it will delay page rendering, because the request is made in the ruby code that delivers the page. 

as per joe's suggestion the other day, one way around this would be to have the count retrieved in an AJAX page element, so that it could be fulfilled without blocking the rendering of the rest of the page. a link could be provided to a page that would block until the count returned, in case someone really specifically wanted to see the count. but it doesn't seem reasonable to hold up every page request for that count. 

another idea would be to cache the count and refresh it every so often (say any time it's requested and the cached value is more than a minute old). this would make the displayed count lag a bit, but that seems acceptable, and again, a more accurate count could be obtained by means other than argo's header, if the backlog was what was really specifically desired.

mejackreed:
Closes #66 

This pull request modifies the way indexer backlog requests are loaded into each page to be more efficient. Requests are handled via ajax now, clientside rather than a blocking serverside request.

8382b85 - refactors the way that styling is done, to be more inline with mainstream Rails/Blacklight/Bootstrap


mejackreed:
The single quote scss hound comments are now addressed by af67fe9e0dc2795c69535a04de982e5529d3d90d. Hound just isn't rerun on the files.

tingulfsen:
This looks good to me, but I'm wondering if I should expect to get this working on my laptop? Is there a setting for STATUS_INDEXER_URL that you could share, @mejackreed ? I'm currently getting "Indexer Backlog: -" in my UI, as expected when it's not getting a response.

mejackreed:
Great question @tingulfsen . I posted the URL I have been using in our slack channel based off looking at @jmartin-sul's old configs.

Although I have never been able to get a result with any objects in backlog

jmartin-sul:
@tingulfsen @mejackreed the URL jack posted in slack is, to the best of my knowledge, the right one.  unless there's an actual indexing backlog, i believe you'll get an empty array (`[]`) as a result.  i think you might also have to be on the VPN to reach that URL, though i'm not sure about that.

PR #106, opened 2015-11-04
Add link to the shared_configs repo

drh-stanford:


jmartin-sul:
considered adding that link initially, but thought it odd to include a private URL in a public README.

opinion from @eefahy?

eefahy:
Yeah, can we just refer to the shared_configs repo and branch name rather than the url of the private repo?

drh-stanford:
ok i made that change and updated the branch...

PR #107, opened 2015-11-04
add solr config task back to travis

mejackreed:
Having a non-working CI bit us (me) here. Note all of the failed travis builds after #89 was merged. This should at least get us back to our failing 5 specs.

 https://travis-ci.org/sul-dlss/argo/builds


mejackreed:
Yea! back to only failing 5

PR #109, opened 2015-11-05
When repo:load fails, it doesn't get reenabled.

atz:
That means on the second attempt, rake perceives it as having been run,
despite the fact that it failed.  The task authors are remiss for pulling
from ENV instead of accepting an actual argument, thereby disguising
from rake the variation in each execution.

### DIFF
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -212,7 +212,7 @@ namespace :argo do
       if(args.has_key?(:glob))
         file_list = glob_files(args[:glob])
       else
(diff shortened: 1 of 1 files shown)


mejackreed:
#progress ... better logging fails :+1:

jmartin-sul:
regular laptop tests pass (had to rebase off develop first, because my database.yml got eaten when i was shuffling branches just now, but not too worried about that).  CI tests fail completely on the first fixture load failure, which fails out with a nice stack trace after the third attempt.

sounds like progress to me, :+1:, merging...

PR #113, opened 2015-11-05
add_workflow needs to reindex new workflow datastream synchronously

### LINKED ISSUE #72: Initializing a workflow does not reindex the object
Adding a new workflow to the object either via Blue Button or direct service /gem call ...

To reproduce, use dor-services gem to do this: 

  i = Dor::Item.find('druid:aa111bb2222') 
  i.initialize_workflow('abcWF') 

Using a legit workflow. Goto Argo for that druid and notice that the workflow abcWF is not present in the History > Workflows section on the page. 

Click Reindex, and notice that the History > Workflows section is now correct.  Adding a new workflow should send an object update message to the DOR index queue.

drh-stanford:
commit the solr reindex for add_workflow before redirection and clear out the workflow datastream before calling to_solr. Fixes #72. Please review...

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -601,6 +601,8 @@ class ItemsController < ApplicationController
   end
 
   #add a workflow to an object if the workflow is not present in the active table
+  # note that the dor-services' initialize_workflow makes a synchronous REST API call
+  # to the workflow service.
   def add_workflow
     return unless params[:wf]
     wf = @object.workflows[params[:wf]]
@@ -610,6 +612,9 @@ class ItemsController < ApplicationController
       return
     end
     @object.initialize_workflow(params[:wf])
+    @object = Dor::Item.find @object.pid # XXX: force a reload of the workflow datastream before indexing
+    reindex @object, true
+
     if params[:bulk]
       render :text => "Added #{params[:wf]}"
     else
@@ -619,9 +624,10 @@ class ItemsController < ApplicationController
 
   private
 
-  def reindex item
-    doc=item.to_solr
-    Dor::SearchService.solr.add(doc, :add_attributes => {:commitWithin => 1000})
+  # XXX: note that this logic is replicated in DorController#reindex. probably should be moved to a helper
+  def reindex(item, synchronous = false)
+    Dor::SearchService.solr.add(item.to_solr, :add_attributes => {:commitWithin => 1000})
+    Dor::SearchService.solr.commit if synchronous
   end
 
   # Filters


app/controllers/items_controller.rb
```diff
@@ -610,6 +612,9 @@ def add_workflow
       return
     end
     @object.initialize_workflow(params[:wf])
+    @object = Dor::Item.find @object.pid # XXX: force a reload of the workflow datastream before indexing
+    reindex @object, true
+
```
drh-stanford: fixed.

mejackreed:
Is there a way that we could add a test, to test this functionality?

drh-stanford:
it would need to be an integration test between Argo and the Solr instance, or a multi-step session. i looked in `spec/integration` and didn't see a precedent for that type of test in there. I'm not fluent enough in rspec/rails to know how to implement this kind of test, but it would need to be something like this:

```
visit /catalog/:pid
click Add Workflow
select 'accessionWF' and click Add
verify that Solr has the correct data for the workflows field
redirect to /catalog/:pid
verify that 'accessionWF' is present in the Workflows section of the view
```

there's 2 problems in the issue's code path: (1)  `dor-services` doesn't refetch the workflows datastream -- it caches it, and (2) the reindex doesn't commit fast enough for the subsequence /catalog to see the updated data. i'm not sure how to implement tests for these timing issues

mejackreed:
Maybe we could break out the logic in this method into an `Argo::Workflow` class that can broker all of this stuff happening?

drh-stanford:
i'm apprehensive of refactoring until we have a discussion about our strategy for that as there's need for it.

i've opened a new issue in `dor-services` https://github.com/sul-dlss/dor-services/issues/111 that will address the underlying problem with this issue. when that is fixed we can make a better assumption here in Argo that the workflows data stream behaves correctly after the call to `initialize_workflow`. the patch in this branch right now works around the fact that `dor-services` is caching the workflows data stream but should not be. I can also implement a unit test on the `dor-services` side. stay tuned.

drh-stanford:
i've updated the Argo code to run `save_and_reindex` after `add_workflow` now, but this will not fix #72 until https://github.com/sul-dlss/dor-services/issues/111 is fixed in `dor-services`. so this PR is still a work-in-progress and needs further testing once https://github.com/sul-dlss/dor-services/issues/111 is fixed.

jmartin-sul:
i agree with @mejackreed that it'd be nice to have a test for this.  but as @drh-stanford pointed out, that doesn't seem like a straightforward thing to implement.  this also feels like a very safe change, and while i'm generally in favor of not decreasing test coverage, i'd be fine with merging this, filing an issue to come up with an approach to testing this sort of thing in general, and implementing a test for this as part of closing that issue.

open to objections, but i vote :+1: to merging this and filing an issue for the test coverage inadequacy.

app/controllers/items_controller.rb
```diff
@@ -10,7 +10,7 @@ class ItemsController < ApplicationController
   before_filter :forbid_modify, :only => [:add_collection, :set_collection, :remove_collection, :update_rights, :set_content_type, :tags, :tags_bulk, :source_id,:delete_file, :close_version, :open_version, :resource, :add_file, :replace_file,:update_attributes, :update_resource, :mods, :datastream_update ]
   before_filter :forbid_view,   :only => [:preserved_file, :get_file]
   before_filter :enforce_versioning, :only => [:add_collection, :set_collection, :remove_collection, :update_rights,:tags,:source_id,:set_source_id, :set_content_type,:set_rights]
-  after_filter  :save_and_reindex,   :only => [:add_collection, :set_collection, :remove_collection, :open_version, :close_version, :tags, :tags_bulk, :source_id, :datastream_update, :set_rights, :set_content_type, :apply_apo_defaults]
+  after_filter  :save_and_reindex,   :only => [:add_collection, :set_collection, :remove_collection, :open_version, :close_version, :tags, :tags_bulk, :source_id, :datastream_update, :set_rights, :set_content_type, :apply_apo_defaults, :add_workflow]
```
darrenleeweber: Too long for github diff to actually show the difference.

app/controllers/items_controller.rb
```diff
@@ -619,9 +622,9 @@ def add_workflow
 
   private
 
+  # XXX: note that this logic is replicated in DorController#reindex. probably should be moved to a helper
   def reindex item
-    doc=item.to_solr
-    Dor::SearchService.solr.add(doc, :add_attributes => {:commitWithin => 1000})
+    Dor::SearchService.solr.add(item.to_solr, :add_attributes => {:commitWithin => 1000})
```
darrenleeweber: Move it to a helper in this PR or create an issue for it?  The comment might get stale unless we create an action item for it.

drh-stanford:
i will create an issue to remove the duplicate `Dor::SearchService.solr.add` calls between the controllers if this PR is merged.

drh-stanford:
i will also create an issue to upgrade to use the new dor-service `create_workflow` method.

drh-stanford:
i will resubmit this PR due to the new approach in sul-dlss/dor-services#111

PR #114, opened 2015-11-05
Don't allow rake:ci on production

darrenleeweber:


mejackreed:
Agreed that you should never run CI on prod. Don't know if this is actually necessary, should you be even ssh'ing into prod servers to run this? Seems like a higher level concern?

darrenleeweber:
@mejackreed - should and could are different, this removes the 'could' by mistake but not the deliberate could with:  `RAILS_ENV=test rake ci`.  The could by mistake might happen if the ci is the default (and often it is).

lib/tasks/argo.rake
```diff
@@ -18,6 +18,7 @@ end
 task :default => [:ci]
 
 task :ci do
+  raise 'NEVER RUN CI ON PROD' if ENV['RAILS_ENV'] =~ /production/i
   ENV['RAILS_ENV'] = 'test'
```
mejackreed: This line actually never allows it be run in production, making LN21 useless

darrenleeweber: Huh?  LN21 prevents it from running on prod.  Where else is it prevented?  Where else could it be prevented?

mejackreed: It sets the RAILS_ENV to test on LN22

darrenleeweber: That's exactly why LN21 is needed.

darrenleeweber: AFAIK, the RAILS_ENV doesn't necessarily imply that production configs will use non-prod systems and services.  If they can be decoupled, as we have discussed for staging, then this safeguard helps.

jmartin-sul: i'd need to double-check (and maybe do some testing) to be sure, but...

i think that by the time RAILS_ENV gets reset here, the environment specific configs would have already been loaded as part of Rails loading (e.g. production.rb, dor_production.rb, settings/production.yml).

darrenleeweber: I prefer to be paranoid and leave the one-liner as is, regardless of the current state of affairs which could change at any later date.  The purpose is to have a safety valve because the rake task overrides an relatively important ENV value (and many similar rake spec/ci tasks do so).

mejackreed: Gotcha.. lets spin our :car: on more important stuff :ok_hand:

jmartin-sul: what about replacing the proposed exception and the existing `ENV['RAILS_ENV'] = 'test'` lines with something like `raise 'NEVER run CI outside of test!' unless ENV['RAILS_ENV'] == 'test'`?

that seems even more paranoid than the currently proposed change, and doesn't lead to a false sense of security or some weird state by resetting an environment var in a way that might not even have the intended effect.  my reasoning is that i've seen rake tasks fail to execute because the config/environments/ files that were being loaded had some error, preventing Rails itself from loading.  resetting RAILS_ENV here will not cause those configs to be reloaded, and the loading of those configs determines more meaningful environment specific behavior than anything else.

darrenleeweber: :+1: and in latest commit.

lib/tasks/argo.rake
```diff
@@ -12,15 +12,19 @@ task :app_version do
...
+    :startup_wait => 240 # seconds - 4 mins
+  }
+  Jettywrapper.load_config.merge(config)
 end
 
 task :default => [:ci]
 
 task :ci do
-  ENV['RAILS_ENV'] = 'test'
+  raise 'NEVER RUN CI OUTSIDE RAILS_ENV=="test"' if ENV['RAILS_ENV'] != 'test'
```
jmartin-sul: two minor suggestions:
1) i've seen rubocop and hound complain about using `do_stuff if some_thing != some_other_thing` and suggest `do_stuff unless some_thing == some_other_thing` instead.  maybe hound doesn't look at rake code by default?  anyway, i'd make the change just to silence future linter complaints, but this is not a thing that i personally feel strongly about.  but consistency is nice, i guess.
2) might be nice to log the actual env value (could just append `(RAILS_ENV is #{ENV["RAILS_ENV"]})` to the exception text, i think).

darrenleeweber: 1) The entire block could be rewritten to encasulate it in a positive `if ENV['RAILS_ENV'] == 'test'` and not raise an exception at all, but the exception adds emphasis and I like to read an if-dosomething block even if the condition has a `!=` in it ;-)

mejackreed:
We've done something similar before... instead of raising why not just run as ENV=test?

```ruby
if Rails.env.test?
  # run ci tests
else
  system('system('rake ci RAILS_ENV=test')
end
```
This solves the guard problem and also accomplishes the task that was originally intended.

Pattern used elsewhere in DLSS:

https://github.com/sul-dlss/SearchWorks/blob/master/Rakefile#L13-L23
https://github.com/sul-dlss/earthworks/blob/master/Rakefile#L14-L25

Seems to accomplish all goals, in an idiomatic already used pattern way.

darrenleeweber:
Does it work when the system call puts the env setting after the rake call?
 - `system('rake ci RAILS_ENV=test')`
 - `system('RAILS_ENV=test rake ci')`

In any case, I don't agree that this is right way to do it.  It gets back to the motivation for this change in the first place.  If the rake task *hijacks* the env setting, it could be run on a production config and when that config uses services that are dedicated to prod, it could wipe them out.  Paranoia is better than reindexing a prod argo?

darrenleeweber:
@mejackreed - shall we try to login to searchworks prod and run `rake ci` and see what happens?  Are you game?  I'm chicken.

mejackreed:
:fearful: I'm not up for it, because I don't login into prod boxes and run rake tasks. I'm happy to have someone else review and agree to disagree.

darrenleeweber:
I think this PR has moved too far from the original intent.  The first commit was a one liner that only excluded CI from running on prod and left the code alone otherwise, so it could set the RAILS_ENV to test and that is what we want on laptop dev.  So I'm going to revert other commits and leave it at that.

atz:
Yeah, 4 commits for a 15-line delta in one file is too noisy.  So squash or rebase interactively to clean up.  I can pair with you on interactive rebase if you want.

darrenleeweber:
@atz - rebased in 1903cda

jmartin-sul:
a few things:
* It appears that, at the very least, this needs to be rebased from develop and force pushed before it can be merged.
* I think this is an improvement over the current state of things, and so would be in favor of merging this fix once it's no longer conflicty.
* However, after further testing and consideration, I'm convinced that Jack's suggestion above is the "right" thing to do in that it's more robust (see https://github.com/sul-dlss/argo/pull/114#issuecomment-154270331).  Here's why:
 * I did some testing in Rails console.  As I expected, setting the Rails env once Rails has loaded (as would be the case once the Rake code was reached) does nothing to update the configuration that's loaded from `config/environments/` (which pulls values from `config/settings/`).  So as to not expand an already long comment, and so as not to post a bunch of internal URLs on github, I'll paste my example terminal session in the Slack channel.  But when I ran Rails console on `argo-stage-b` (started under the `staging` env), re-setting the Rails env (using a couple different approaches) did nothing but change the reported name of the Rails env.  All the important settings (e.g. Fedora URL) remained the same.  There was actually no development configuration available on that box, so had the change triggered the loading that would've been needed to do what was intended, an error would've been raised.
 * Jack's proposal above sidesteps this whole problem by checking to see if the env is `test`, and if it's not, it shells out and runs the rake task from the command line using the `test` Rails env, which will actually guarantee it's run in the proper env.
 * It's true that neither the fix in this PR nor Jack's proposal will protect against someone running CI from the command line on production with an explicitly specified RAILS_ENV of `test`.  But that should only be a problem if someone's put a `test` env settings file on production, *with settings that actually point to prod instances instead*.  That seems exceedingly unlikely, right?
  * if there's no settings file for the `test` env, the dangerous settings will be configured with meaningless placeholder values from the base `settings.yml`.
  * if there is a settings file for the `test` env, it likely contains actual test environment settings, and not prod URLs, so whatever.

i'll file an issue for shoring things up using jack's suggestion.  it'd be nice to do that, but i don't think we're at a terribly high risk of someone running CI on prod anyway.

@darrenleeweber @mejackreed

darrenleeweber:
@jmartin-sul  -  All of that makes sense, good.  To be sure, the bail-out condition(s) should check for any prod services in specific settings (instead of the Rails.env value).

Are there cases where CI might be run on -dev or -stage platforms?  If so, checking for Rails.env == 'test' is a blocker to running specs on those systems.  But, having said that, I can't think of any times that anyone has run CI on -dev or -stage platforms.  However, maybe some form of integration tests might be run (but that's usually outside the usual rspec framework).

In any case, it's possible to have prod settings inside an environment 'test' file.  Very, very unlikely, but possible.  Certainly all the higher level considerations are more important; this was thrown out there as a last resort and, perhaps, it should be an ultimate arbiter to safeguard prod.  The primary motivation was that the rake task resets the Rails.env and, as you found out, that is too late in the game.

Anyhow, this has served to raise some awareness about the implications of setting Rails.env in a rake task and, if nothing else, that's significant.  Thanks for taking the time to evaluate the issue thoroughly.  I'm somewhat ambivalent about exactly how you want this to work and suggest that you take over the details and test the solution on -prod (I hope you get it right the first time ;-)).  I'm going to put my -prod security paranoia in a box ;-)

drh-stanford:
I think we're on the same page here and we have a new issue #172. Please reopen if not.

PR #115, opened 2015-11-05
Making workflow service log be written to log/ directory by default.

tingulfsen:
This should fix issue 97.


jmartin-sul:
rspec passes on laptop, looks good, merging

PR #116, opened 2015-11-06
remove the argo:htaccess rake task

jmartin-sul:
remove the code for the argo:htaccess rake task, as well as its invocation after restart in deploy.rb.

remove the default_htaccess_directives file that the rake task relied on, as well as the reference to that file in the :linked_files list in deploy.rb.

htaccess generation is no longer needed from argo.  we expect the htaccess file to change infrequently, and do not necessarily expect changes to coincide with web code deployments (which was the expectation set by the code being removed).  this will be managed as part of the deployment code by devops (currently using puppet).

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -25,7 +25,7 @@ set :deploy_to, '/home/lyberadmin/argo'
 # set :pty, true
 
 # Default value for :linked_files is []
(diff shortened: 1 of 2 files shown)


jmartin-sul:
superseded by PR #117 (which branches off this PR's branch, and keeps its changes)

PR #117, opened 2015-11-06
various fixes to correct things preventing argo from deploying properly to stage

jmartin-sul:
(similar changes will soon be needed for all other instances).

* _deploy.rb_: `config/environments/` in `:linked_dirs` becomes `config/settings/`, since this is now where per-instance config info lives.
* _deploy.rb_: the post-restart trigger for running the `argo:htaccess` rake task has been removed, as web code deployments don't necessarily coincide with the need for .htaccess refreshing, and the task is run regularly by a cron job anyway.
* _dor_staging.rb_: full paths to cert file and key file are now specified, since they no longer necessarily share a parent directory.  NOTE: a similar change will need to be made in dor_development.rb and dor_production.rb.
* _argo.rake_: `argo:htaccess` rake task now always includes the built-in argo workgroups, and de-dupes the workgroup list.
* _argo.rake_: `argo:htaccess` rake task now has some usage explanation.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -28,7 +28,7 @@ set :deploy_to, '/home/lyberadmin/argo'
 set :linked_files, %w{config/database.yml config/solr.yml config/default_htaccess_directives}
 
 # Default value for linked_dirs is []
-set :linked_dirs, %w{log config/certs config/environments tmp/pids tmp/cache tmp/sockets vendor/bundle public/system}
+set :linked_dirs, %w{log config/certs config/settings tmp/pids tmp/cache tmp/sockets vendor/bundle public/system}
 
 # Default value for default_env is {}
 # set :default_env, { path: "/opt/ruby/bin:$PATH" }
@@ -49,16 +49,5 @@ namespace :deploy do
   end
 
   after :publishing, :restart
-
-  after :restart, :initialize_htaccess do
-    on roles(:web), in: :groups, limit: 3, wait: 10 do
-      # Here we can do anything such as:
-      within release_path do
-        with rails_env: fetch(:rails_env) do
-          execute :rake, 'argo:htaccess'
-        end
-      end
-    end
-  end
 
 end
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -278,6 +278,13 @@ namespace :argo do
     puts "#{facet.items.count} Workgroups:\n#{facet.items.collect(&:value).join(%[\n])}"
   end
 
+  # the .htaccess file lists the workgroups that we recognize as relevant to argo.
+  # if a user is in a workgroup, and that workgroup is listed in the .htaccess file,
+  # the name of the workgroup will be in a list of workgroups for the user, passed along
+  # with other webauth info in the request headers.  we use the list of workgroups a user is
+  # in (as well as the user's sunetid) to determine what they can see and do in argo.
+  # NOTE: at present (2015-11-06), this rake task is run regularly by a cron job, so that
+  # the .htaccess file keeps up with workgroup names as listed on APOs in use in argo.
   desc "Update the .htaccess file from indexed APOs"
   task :htaccess => :environment do
     directives = ['AuthType WebAuth',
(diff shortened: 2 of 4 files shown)


darrenleeweber:
Maybe add documentation in the deploy files on where, how and why the htaccess directives are managed in puppet?

config/deploy.rb
```diff
@@ -25,10 +25,10 @@
 # set :pty, true
 
 # Default value for :linked_files is []
-set :linked_files, %w{config/database.yml config/solr.yml config/default_htaccess_directives}
+set :linked_files, %w{config/database.yml config/solr.yml config/settings/staging.yml}
```
darrenleeweber: Not sure that `config/deploy.rb` should be tied to `settings/staging.yml` in particular.  Maybe that belongs in a subdirectory or another place that is specific to staging?

lib/tasks/argo.rake
```diff
@@ -279,29 +279,6 @@ namespace :argo do
...
-    unless facet.nil?
-      facets = facet.items.collect &:value
-      priv_groups = facets.select { |v| v =~ /^workgroup:/ }
-      directives += priv_groups.collect { |v|
-        ["Require privgroup #{v.split(/:/,2).last}", "WebAuthLdapPrivgroup #{v.split(/:/,2).last}"]
-      }.flatten
-      File.open(File.join(Rails.root, 'public/.htaccess'),'w') do |htaccess|
-        htaccess.puts directives.sort.join("\n")
-      end
-      File.unlink('public/auth/.htaccess') if File.exist?('public/auth/.htaccess')
```
darrenleeweber: Can puppet replace all this, esp. whatever `get_workgroups_facet` is doing?

config/deploy.rb
mejackreed: Good catch, agree

config/deploy.rb
```diff
@@ -25,10 +25,10 @@
...
 
 # Default value for linked_dirs is []
-set :linked_dirs, %w{log config/certs config/environments tmp/pids tmp/cache tmp/sockets vendor/bundle public/system}
+set :linked_dirs, %w{log config/certs tmp/pids tmp/cache tmp/sockets vendor/bundle public/system}
```
mejackreed: We should probably have `config/settings` here. See: https://github.com/sul-dlss/earthworks/blob/master/config/deploy.rb#L30

lib/tasks/argo.rake
jmartin-sul: @eefahy pointed out that it's unnecessarily circuitous.  this code queries the existing objects for permissions related info, and builds up a list from the permissions that are currently in use.  but that's unnecessary, since we know all the workgroups we'll care about supporting.  and now that i think about it, we might want to add them to the htaccess file before they're listed in an APO.  if we define new workgroups, which we do occasionally, we can manually add them to whatever config puppet uses.  and as per the commit message, updating htaccess doesn't really correspond with deploying/restarting the app, which is when this rake task is run.

it's clever, but it doesn't really gain us anything, as far as we can tell.

config/deploy.rb
jmartin-sul: yup, good catch, will fix.

config/deploy.rb
jmartin-sul: good point.  could then get rid of `config/settings/staging.yml` from `:linked_files`.

atz:
Probably should squash part or all of this?

darrenleeweber:
Some mysteries solved, but maybe logging the workgroups would help?  I'm dying to know more about `my-peeps`!

jmartin-sul:
@atz yup, this is work in progress.  will squash when done.

@darrenleeweber i do like the idea of logging, though that might get a little noisy if the task is running every few minutes.  also, it'd mostly be useful for historical comparison purposes, since the htaccess output is essentially the same as what would be logged (though i can see how the history could be useful, and it's not like the list is _that_ long).

mejackreed:
Should we also set to `set :bundle_without, %w{test development}` https://github.com/sul-dlss/argo/blob/develop/config/deploy/stage.rb#L5

rather than production.

jmartin-sul:
@atz: squashed (and rebased from develop).  diff looks good, and github says it can merge it automatically, so i think it's ok (i pulled when i shouldn't have after squashing, but i think i got myself out of that jam by resetting HEAD to my squash commit before force pushing).

@mejackreed: fixed `:bundle_without`.

@darrenleeweber: added explanation for how the .htaccess rake task is invoked.  declined to log the list used for htaccess generation.  if you feel strongly about keeping a historical record, i'd be more inclined to create a new issue/PR, and keep N backups of the .htaccess file, maybe capturing only when the new file actually differs from the old one.  but i think i'd want to discuss the how/why of doing that, and it feels outside the scope of this PR.

i think that addresses all the issues raised so far with this one?

PR #119, opened 2015-11-06
enables a negative selection for the UseLicense in APO registration

### LINKED ISSUE #69: Register/edit APO - error when no license selected
The UI represents no choice/none as "Citation only" which doesn't make sense / has no definition. Choosing it leads to a system error -- "is not a valid license code". Make the first choice something like "-- none --" and have the code skip adding license information rather than getting an error.

drh-stanford:
Fixes #69

Note that there's no unit tests for a generic APO creation:

  https://github.com/sul-dlss/argo/blob/develop/spec/controllers/apo_controller_spec.rb#L29

### DIFF
--- a/app/controllers/apo_controller.rb
+++ b/app/controllers/apo_controller.rb
@@ -87,8 +87,9 @@ class ApoController < ApplicationController
     apo.metadata_source      = md_info[:metadata_source]
     apo.agreement            = md_info[:agreement].to_s
     apo.default_workflow     = md_info[:workflow ] unless (!md_info[:workflow] || md_info[:workflow].length < 5)
-    apo.use_license          = md_info[:use_license]
     apo.default_rights       = md_info[:default_object_rights]
+    # Set the Use License given a machine-readable code for a creative commons or open data commons license
+    apo.use_license          = md_info[:use_license].blank? ? :none : md_info[:use_license]
   end
 
   def register_new_apo
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -13,7 +13,7 @@ GIT
 
 GIT
   remote: https://github.com/sul-dlss/dor-services.git
-  revision: 79bac3829225bda644875b73115527472b2fffc5
+  revision: d96fb43e08547bde54084c3e0914b4d792491d06
   branch: develop
   specs:
     dor-services (5.2.0)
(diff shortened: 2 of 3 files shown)


app/controllers/apo_controller.rb
```diff
@@ -87,8 +87,16 @@ def set_apo_metadata apo, md_info
     apo.metadata_source      = md_info[:metadata_source]
     apo.agreement            = md_info[:agreement].to_s
     apo.default_workflow     = md_info[:workflow ] unless (!md_info[:workflow] || md_info[:workflow].length < 5)
-    apo.use_license          = md_info[:use_license]
     apo.default_rights       = md_info[:default_object_rights]
+    if md_info[:use_license].blank?
+      # delete use licenses, ensure defaultObjectRights datastream is saved correctly
+      apo.defaultObjectRights.creative_commons = nil
+      apo.defaultObjectRights.open_data_commons = nil
+      apo.defaultObjectRights.creative_commons_human = nil
```
atz: Ideally, the "human" assignments should already be accomplished by the earlier setter methods, right?  This would follow the pattern set in Editable.  The reason being, it shouldn't be possible to set the code and still have an out-of-date or mismatched human readable value.  (Personally, I'd prefer not storing the "human" readable (i.e. *English*) version at all, but can punt on that.)

jmartin-sul: i'm not so opposed to storing the human readable string, but i do think that we should be using one unified setter method that takes something like a license code, and sets URI and human readable based on that.  i think we should consider setting those things individually to be deprecated, at best.

drh-stanford: thanks for the feedback. i think the next step is that i will make a PR for `dor-services` that allows `use_license=` to set a null license if `nil` is passed in. then i will change this PR to use that method.

drh-stanford:
requires https://github.com/sul-dlss/dor-services/pull/113 to work

jmartin-sul:
@drh-stanford, does this need a Gemfile.lock update before merge?

drh-stanford:
Yes, I think you're correct. I'll add that to this PR

drh-stanford:
The `Gemfile.lock` is now in this PR

PR #120, opened 2015-11-07
more verbose logging in argo:repo:load

jmartin-sul:
* print list of files being loaded (in order being loaded)
* have the error handler print the backtrace for each error it handles (instead of just the message)

### DIFF
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -214,6 +214,7 @@ namespace :argo do
       else
         puts 'No file glob was specified so file order and inclusion is determined by the load_order file'
         file_list = load_order_files(fedora_files)
+        puts "file_list=#{file_list}"
       end
 
       errors = []
@@ -224,7 +225,7 @@ namespace :argo do
 
         ENV['foxml'] = file
         handler = proc do |e, attempt_number, total_delay|
-          puts STDERR.puts "ERROR loading #{file}:\n  #{e.message}"
+          puts STDERR.puts "ERROR loading #{file}:\n#{e.message}\n#{e.backtrace.join "\n"}"
           errors << file
         end
         with_retries(:max_tries => 3, :handler => handler, :rescue => [StandardError]) { |attempt|


jmartin-sul:
points of interest discovered while digging into the constituent parts of the CI build by running rake tasks manually and playing with the uber APO (hv992ry2431) in pry...

* https://travis-ci.org/sul-dlss/argo/builds/89797899#L476
 * repo:load fails on hv992ry2431 after three attempts.  
 * however, you'll notice that the stack traces for the second two attempts differ from that of the first.
 * the second two are both the result of trying to load an object that already exists in fedora (and indeed, if you bring fedora up after a failed CI build, you can look up hv992 and find it).
 * but the first error is a result of this line executing:  https://github.com/sul-dlss/dor-workflow-service/blob/master/lib/dor/services/workflow_service.rb#L523

and that line is a call out to our favorite service:  the workflow service.  it's happening when the object is being indexed (if you run a CI build and let it fail, you can bring jetty back up, open pry, load hv992 successfully, and get the same error and backtrace as you see in the above travis link by calling `to_solr` on the item you just loaded).

so, what's keeping travis' argo build from running?  maybe other stuff too, but definitely that call to the workflow service.  that'd also explain why this isn't a problem for laptops (since they'll have a workflow service URL specified in settings, and reachable when on the stanford network).

not sure how to fix the CI problem (since this is the same as the problem that's been driving @atz crazy on hydrus).  but i propose that we at least keep the logging contained in this PR, since it was helpful in diagnosing the problem, and could be helpful in the future.

my big admission of defeat on this was renaming the branch from the speculative `ci_test_fixes` to the more realistic `argo-repo-load_verbosity` :cry: :trumpet:

drh-stanford:
one idea for the Travis CI build is we could make a trivial jetty `/workflow` service that only implemented `get_workflow_xml` REST endpoint (called here: https://github.com/sul-dlss/dor-workflow-service/blob/master/lib/dor/services/workflow_service.rb#L109) and always returned an empty `<workflows/>`. then change the default setting to`WORKFLOW_URL: 'http://localhost:8983/workflow/'`. that would at least get the Travis CI build working, with all objects having empty `workflows` data streams, so we can test other functionality.

also you can set `RESTCLIENT_LOG=stdout` in your environment to see exactly which REST calls the workflow service is making:

mejackreed:
:+1: for the solution @drh-stanford proposes here. Working on that right now

darrenleeweber:
@jmartin-sul - long shot, but maybe check for tombstones?  Maybe the loading is failing sometimes because a druid was previously loaded and deleted, leaving a tombstone footprint that can't be replaced.

darrenleeweber:
A noop WFS is an interesting idea, not sure what all the implications are.  An alternative is to use VCR configured to record WFS interactions.
  - https://github.com/vcr/vcr

mejackreed:
PR helps with logging.. :+1:

PR #122, opened 2015-11-07
add webmock to mock workflow-services & fix ci FTW

mejackreed:
This work fixes the failing travis builds by doing the following:

 - Allow fixtures to actually load, by using @drh-stanford's suggestion of [mocking wfs requests](https://github.com/sul-dlss/argo/pull/120#issuecomment-154692603)
    - needed here was the need to mock both `/dor/*` and `/workflow_archive*` paths a bit differently
    - `WebMock` is loaded in the `development` and `test` gem groups, but is disabled in `production` or when you set `Settings.DISABLE_WEBMOCK = true`
 - 3981a436035af1296905b1988ce634060d038693 removes an unnecessary setting that is set, and puts the default one in the correct format with trailing `/`
 - When `argo:repo:load` 

_Note: With these changes, I don't thing `test.local.yml` is needed for running tests._

Possible enhancements:
 - remove argo solr generator, I'm not sure what the impetus of this was, but AFAIK it's now not being used
 - Potentially delete `testcores` or reduce just to one. Fixtures can be loaded into Fedora in `development` environment and then is indexed in `development` solr core. But that same data needs to be dropped, and reloaded to run in specs against in `test` environment. Not sure about the history here either, so left it untouched.

Would like feedback from others about potential impact of mocking WFS, and the Solr cores dev vs test.

~~TODO~~ Completed during development:
 - [x] Still getting one error on `** File 48` where `to_solr` is being called for tm388wy6148, and throwing `undefined method `value' for nil:NilClass` from https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/models/workflow_object.rb#L51
 - [x] need to figure out a better way to configure WebMock before hydrajetty requests gem download
 - [x] down to only 3 failing specs now... progress

### DIFF
--- a/config/initializers/webmock.rb
+++ b/config/initializers/webmock.rb
@@ -1,0 +1,14 @@
+def mock_workflow_requests
+  escaped_url = Settings.WORKFLOW_URL.gsub('/', '\/')
+  stub_request(:get, /#{escaped_url}workflow_archive.*/).
+    to_return(body: '<objects count="1"/>')
+  stub_request(:get, /#{escaped_url}dor.*/).
+    to_return(body: '<workflows/>')
+end
+
+unless Rails.env.production?
+  require 'webmock'
+  include WebMock::API
+  WebMock.allow_net_connect!
+  mock_workflow_requests unless Settings.DISABLE_WEBMOCK
+end
--- a/config/solr.yml
+++ b/config/solr.yml
@@ -1,0 +1,16 @@
+# = jetty_path key
+# each environment can have a jetty_path with absolute or relative
+# (to app root) path to a jetty/solr install. This is used
+# by the rake tasks that start up solr automatically for testing
+# and by rake solr:marc:index.  
+#
+# jetty_path is not used by a running Blacklight application
+# at all. In general you do NOT need to deploy solr in Jetty, you can deploy it
+# however you want.  
+# jetty_path is only required for rake tasks that need to know
+# how to start up solr, generally for automated testing. 
+
+development:
+  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr" %>
+test: &test
+  url: <%= ENV['TEST_SOLR_URL'] || "http://127.0.0.1:8983/solr/test" %>
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -19,6 +19,7 @@ task :default => [:ci]
 
 task :ci do
   ENV['RAILS_ENV'] = 'test'
+  WebMock.allow_net_connect!
   Rake::Task['argo:install'].invoke
   jetty_params = jettywrapper_load_config()
   error = Jettywrapper.wrap(jetty_params) do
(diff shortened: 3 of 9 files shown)


lib/tasks/argo.rake
```diff
@@ -19,6 +19,7 @@ task :default => [:ci]
 
 task :ci do
   ENV['RAILS_ENV'] = 'test'
+  WebMock.allow_net_connect!
```
darrenleeweber: This is in `config/initializers/webmock.rb`, is it required here also?  Could it be placed in a more specific rake task to clarify the tasks that depend on it?

mejackreed: Unfortunately yep its required here too when the bundle requires the gems, it will disable all connections. Still trying to figure all of this out.

config/solr.yml
```diff
@@ -0,0 +1,18 @@
...
+# at all. In general you do NOT need to deploy solr in Jetty, you can deploy it
+# however you want.  
+# jetty_path is only required for rake tasks that need to know
+# how to start up solr, generally for automated testing. 
+
+development:
+  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr" %>
+test: &test
+  url: <%= ENV['TEST_SOLR_URL'] || "http://127.0.0.1:8983/solr/test" %>
+production:
```
jmartin-sul: as with database.yml, i'd vote for leaving this out, since production will have its own solr.yml in its private config branch.

jmartin-sul:
my, what a nice green that is.

left one minor suggestion in a line note.  but otherwise, big :+1: from me.  glad my pessimism about fixing this wasn't as warranted as i thought.  this seems like a really good solution, glad it was possible to use mocks instead of recording/replaying web requests.

config/initializers/webmock.rb
```diff
@@ -0,0 +1,14 @@
+def mock_workflow_requests
+  escaped_url = Settings.WORKFLOW_URL.gsub('/', '\/')
+  stub_request(:get, /#{escaped_url}workflow_archive.*/)
+    .to_return(body: '<objects count="1"/>')
+  stub_request(:get, /#{escaped_url}dor.*/)
+    .to_return(body: '<workflows/>')
```
jmartin-sul: i guess i agree with hound, since i don't mind long lines, but this reads fine, and i'm neutral on changing it.

jmartin-sul:
i'd leave the "possible enhancements" for another pull request and merge this as is, just because it seems like a clear improvement to the current state of things, and we can keep adding from there.

but some thoughts on the proposals:
* i agree on removing the solr generator, which seems useless now that there's a default solr.yml and instance specific solr.yml files in the private config repos.
* i like the idea of mocking solr cores, but that seems like a lot of work.  maybe not worth it at this point, since we seem to be doing OK for now with a test solr instance?  but i'd definitely be interested in revisiting that in the not too distant future.
* and i guess if solr interactions were mocked, we'd definitely be able to get rid of the test core?  but otherwise, it seems like we'd want to keep that?  not sure.

mejackreed:
@jmartin-sul I updated the PR to address your comments. I'm not really a fan of us mocking the Solr cores, just only having one to reduce developer confusion and increase happiness. Blacklight does a pretty good job of providing a pattern for testing these and as long as we follow those patterns we shouldn't have a problem _crosses fingers_

PR #139, opened 2015-11-09
only use Solr core `development` for development and testing

### LINKED ISSUE #128: laptop installation issues
* i notice that when i do a fresh installation after cloning on my laptop, things fail if i don't manually create the `log` directory.
* on my laptop, i get three unit test failures from a clean installation, even though all of the fixtures were successfully loaded:
```
Failures:

  1) CatalogController show enforces permissions without APO with user is_admin
     Failure/Error: expect(response.code).to eq("200")
       
       expected: "200"
            got: "404"
       
       (compared using ==)
     # ./spec/controllers/catalog_controller_spec.rb:38:in `block (5 levels) in <top (required)>'

  2) CatalogController show enforces permissions without APO with user is_viewer
     Failure/Error: expect(response.code).to eq("200")
       
       expected: "200"
            got: "404"
       
       (compared using ==)
     # ./spec/controllers/catalog_controller_spec.rb:43:in `block (5 levels) in <top (required)>'

  3) DiscoveryController :data should fetch json data with the gdor fields in it
     Failure/Error: expect(json_body['rows'].first).to match a_hash_including('sw_format_tesim')
       expected nil to match (a hash including "sw_format_tesim")
     # ./spec/controllers/discovery_controller_spec.rb:12:in `block (3 levels) in <top (required)>'
```
* i get the same errors running `RAILS_ENV=test bundle exec rake ci` on a clean checkout.  but the same branches where that happens for me locally (develop and laptop_installation_fixes) pass on travis.  so i'm wondering if something's screwy about my laptop setup.
* it seems that `tmp:create` was removed from the list of tasks that `argo:install` runs.  was that intentional?  relevant diff here:  https://github.com/sul-dlss/argo/commit/e2b3ee190008fa7dc58020de1650ed81cf8f66c2#diff-4c5ae9e17f273c380104e0ef14e17c7aR39

mejackreed:
Closes #128 

This PR aims to reduce confusion on "what task indexed what where". Previously when running:

 - `rake ci` would index features into `test` core
 - `rake argo:repo:load` would index features into `development` core

Just put everything into development core to reduce confusion which should provide for a more cohesive dev experience.

### DIFF
--- a/config/solr.yml
+++ b/config/solr.yml
@@ -13,4 +13,4 @@
 development:
   url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr" %>
 test: &test
-  url: <%= ENV['TEST_SOLR_URL'] || "http://127.0.0.1:8983/solr/test" %>
+  url: <%= ENV['TEST_SOLR_URL'] || "http://127.0.0.1:8983/solr" %>
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -85,7 +85,7 @@ namespace :argo do
   fedora_fileglob   = "#{Rails.root}/#{fedora_conf_dir}/data/*.xml"
   fedora_files      = File.foreach(File.join(File.expand_path('../../../fedora_conf/data/', __FILE__), 'load_order')).to_a
   live_solrxml_file = "jetty/solr/solr.xml"
-  testcores = {'development' => 'development-core', 'test' => 'test-core'}  # name => path
+  testcores = {'development' => 'development-core' } # name => path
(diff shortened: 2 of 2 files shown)


darrenleeweber:
Are there any unit tests that might infect another test because they would not use independent index data from the develop core?

mejackreed:
No idea. I doubt it though, the only 3 tests that were actually querying solr were the 3 outlined in #128.

mejackreed:
I'm actually going to withdrawl this PR. Unfortunately this approach isn't going to work because of the way hydra-jetty and jettywrapper handle all of this. The current approach is kind of already baked into the CI process, so I would say we stick with that. https://github.com/projecthydra/hydra-jetty/tree/v7.1.0/solr

PR #140, opened 2015-11-09
ignore locally deployed bundles

drh-stanford:
ignores the `vendor/` folder for locally deployed bundles (i.e., `bundle --path vendor`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ tmp/
 config/settings.local.yml
 config/settings/*.local.yml
 config/environments/*.local.yml
+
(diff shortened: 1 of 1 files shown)


mejackreed:
`vendor` is a common place to put vendor included libraries, including Javascript and css. Can we make this more specific?

drh-stanford:
changed to `vendor/ruby/`... maybe i'm the only one that does local bundle installs on my laptop?

mejackreed:
Cool, thanks. Yeah I don't do those, but maybe others do

PR #141, opened 2015-11-09
use the bundler cache on Travis

drh-stanford:
this speeds up subsequent Travis CI builds quite a bit. see http://docs.travis-ci.com/user/caching/


mejackreed:
Yep, we've used this before in some projects, but seen some strange issues. @cbeer do you remember any of those issues?

mejackreed:
Fine with merging.. we can revert if we have any issues. Looks like we are using on other projects with no problems.

cbeer:
I don't remember having any problems with bundler caching for applications.

Issue #143, opened 2015-11-09
Is solr request in page head a feature needed in argo?


mejackreed:
This legacy code added in https://github.com/sul-dlss/argo/commit/fea324130be0e0391233755c7315a82c9e0d81f7

https://github.com/sul-dlss/argo/blob/develop/app/views/catalog/index.html.erb#L1-L3

I can't figure out what this does or what might use this. @LynnMcRae Can you please provide some historical context? Can we get rid of this?

LynnMcRae:
I can't tell what this does either.  There is a"View solr document" link in the sidebar where we offer links to PURL, Fedora admin, and the FoXML as well. That is tremendously useful. Does this relate to that?

mejackreed:
Not that I can tell. That "View Solr document" link is generated using this helper:
https://github.com/sul-dlss/argo/blob/04815a9188821c0c3b090a3d60cf7c0a497ab4bf/app/helpers/argo_helper.rb#L215-L220

drh-stanford:
It looks like an attempt at linked data to me -- but it could be done by adding `.json` to the request instead, right?

mejackreed:
Yeah but why do we even need it there (there == in the <head> section)?

drh-stanford:
+1 to remove it

PR #144, opened 2015-11-09
Update to use standard Blacklight tools and navigation partials

### LINKED ISSUE #86: Add Blacklight's number-of-item-per-results-page choice feature
With choices 10, 20, 50, 100 per SW? Currently Argo defaults to 10 with no choices. SW defaults to 20. Do we want to up the Argo default to match?

### LINKED ISSUE #133: Use standard navigation links on item detail page
The Start Over and Back to Search options on the item detail page are in a different location than elsewhere and show as links rather than buttons. Move them to the empty grey bar above (the standard place where the same controls are found on other pages) and show them as buttons like they are elsewhere.

mejackreed:
Closes #133 
Closes #86 

Raises questions in #143 which may remove the need to include partial `app/views/catalog/index.html.erb`

### Show View
screen shot 2015-11-09 at 2 33 34 pm

### Index View
screen shot 2015-11-09 at 2 34 01 pm


jmartin-sul:
:+1: as far as i can tell

PR #145, opened 2015-11-10
ensure we have new workflows datastream before we index item

### LINKED ISSUE #72: Initializing a workflow does not reindex the object

drh-stanford:
This fixes #72. It requires `dor-services`. It will *not* work until PR https://github.com/sul-dlss/dor-services/pull/112 is merged.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -609,7 +609,12 @@ class ItemsController < ApplicationController
       render :status => 500, :text => "#{params[:wf]} already exists!"
       return
     end
-    @object.initialize_workflow(params[:wf])
+    @object.create_workflow(params[:wf])
+
+    # sync up the workflows datastream with workflow service and force a reindex before redirection
+    reindex Dor::Item.find(params[:id])
+    Dor::SearchService.solr.commit
+
     if params[:bulk]
       render :text => "Added #{params[:wf]}"
     else
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -374,12 +374,12 @@ describe ItemsController, :type => :controller do
       expect(@item).to receive(:workflows).and_return @wf
     end
     it 'should initialize the new workflow' do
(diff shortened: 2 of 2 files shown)


atz:
Looks like one test got broken as a side-effect: `./spec/controllers/items_controller_spec.rb:379`

drh-stanford:
i've changed my strategy on this fix and i'm going with an earlier solution. the content refresh works, but there's other cached data that i'm not seeing in `dor-services` wrt workflows. the `content` method seems to break the `ActiveFedora` model in that we're not using the built-in `content=` method to load the new workflows XML content (although the patched `dor-services` does indeed refresh it when `content(true)` is called.

jmartin-sul:
tests for this branch pass on my laptop, and travis seems happy again.  code looks reasonable.  ok to merge?  ping @atz since he pointed out the test breakage initially.

app/controllers/items_controller.rb
```diff
@@ -609,7 +609,12 @@ def add_workflow
       render :status => 500, :text => "#{params[:wf]} already exists!"
       return
     end
-    @object.initialize_workflow(params[:wf])
+    @object.create_workflow(params[:wf])
+
+    # sync up the workflows datastream with workflow service and force a reindex before redirection
+    reindex Dor::Item.find(params[:id])
```
mejackreed: Are we able to test these lines?

drh-stanford: See my comment re: testing here: https://github.com/sul-dlss/argo/pull/113 `create_workflow` is unit tested in `dor-services` but the `reindex` call isn't.

Issue #93, opened 2015-11-02
Deploy to argo-stage


atz:

darrenleeweber:
`argo-stage` doesn't exist yet, so `argo-test` might be the target for this issue.

darrenleeweber:
There may be no capistrano target for this yet.

darrenleeweber:
The vm target is `argo-stage-b` and the cap target is `stage-b` (don't ask me why it's `b`).

jmartin-sul:
my original understanding (from @eefahy ) was that there would be a load balanced argo-stage, but that the initial argo-stage deployments would only go to the argo-stage-b VM, similarly to what was done with argo-prod-b.  can't remember if there was other rationale for that besides instance similarity, or if i was possibly misunderstanding or conflating setup instructions, since nothing's using argo-stage-at at the moment (AFAIK).

indeed, argo-stage does not seem to exist, but i can ssh into argo-stage-b as lyberadmin if i've got valid kerb credentials:
`ssh lyberadmin@argo-stage-b`

darrenleeweber:
Bump as ready for further deploy testing, if @eefahy is done with config setup etc.?

jmartin-sul:
workin on it.  met w/ @eefahy earlier this afternoon, she's making some .htaccess changes in puppet, i'm pulling the htaccess generation rake task from argo (or at least keeping it from running by default on deployment).  once that's done, we'll try another deployment and regroup if there are more errors, or start testing indexing if the deployment's successful.

hopefully once this is ironed out, updating argo-dev's puppet stuff* to use the new config style will be easier and involve less iteration.

*"puppet stuff" is a technical term

jmartin-sul:
updates:
* left htaccess stuff in afterall.  needed it to keep things working as-is, so that argo can get all the workgroup info it needs from webauth to do permissions checks.  https://github.com/sul-dlss/argo/issues/138 addresses cleaning this mechanism up.
* there's currently a permission error in stage deployment that's keeping argo from opening its .key file.  this is the only current known code deployment hangup, but we shall see if another pops up after that's resolved, or if that's the last one.
* for argo to actually do permissions checking properly, it needs to be authorized to get specific workgroup info from LDAP, and that authorization might take a couple of days, according to erin.  the code could be deployed and run without that, but the web UI will not be useful until that authorization is granted.

jmartin-sul:
[...]

PR #147, opened 2015-11-10
more fixes to allow deployment to argo-stage et al

jmartin-sul:
these fixes should be relevant to other deployments (argo-dev, argo-prod), but they were first implemented to get stage deployments working.

* _Gemfile_: remove the `:assets` group distinction, since Rails 4 no longer recognizes that group (http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0-gemfile)
* _Gemfile_: take the gems that were listed in `:assets`, and move them out into the general dependencies lists.
* _Gemfile_: since we're here, clean up the lists a little (alphebetize strays, remove commented out dupe entries).
* _config/initializers/webmock.rb_:  limit webmock inititialization to the two environments where it's relevant (instead of just excluding production).
* _settings.yml_: remove `'argo.css'` from the PRECOMPILE list.
* _argo.rake_: only create the `:integration_tests` rake task in test and development, since it relies on including rspec/core/rake_task (the rspec gem is only included in development and test).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,11 +2,21 @@ source "https://rubygems.org"
 
 gem 'addressable', '=2.3.5' #>=2.3.6 breaks things w/ the following error on rails startup:  "can't modify frozen Addressable::URI"
 gem 'barby'
+gem 'bootstrap-sass'
 gem 'coderay'
+gem 'coffee-rails'
+gem 'config'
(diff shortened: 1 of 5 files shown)


jmartin-sul:
this should close issue #93 once it's merged.

PR #151, opened 2015-11-10
Update readme to do the following:

### LINKED ISSUE #128: laptop installation issues

mejackreed:
- reference current default branch for travis and coveralls badges
 - recommend installing and using Ruby 2.2.2
 - outline running CI build and what rake tasks index data into which Solr core. closes #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/sul-dlss/argo.png?branch=master)](https://travis-ci.org/sul-dlss/argo) | [![Coverage Status](https://coveralls.io/repos/sul-dlss/argo/badge.svg?branch=master&service=github)](https://coveralls.io/github/sul-dlss/argo?branch=master)
+[![Build Status](https://travis-ci.org/sul-dlss/argo.png?branch=develop)](https://travis-ci.org/sul-dlss/argo) | [![Coverage Status](https://coveralls.io/repos/sul-dlss/argo/badge.svg?branch=develop&service=github)](https://coveralls.io/github/sul-dlss/argo?branch=master)
 
 # Argo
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,6 +88,21 @@ rake argo:repo:load
 rspec
 ```
 
+The continuous integration build can also be run by:
```
jmartin-sul: would be nice to add a warning explaining that running CI (or anything that runs `argo:jetty:clean`) will reinstall your jetty instance and blow away any test data you might have setup.

mejackreed: Good catch... I'll add that in a bit.

mejackreed: Ok added!

PR #157, opened 2015-11-10
Update dev, test, stage, & prod configs

### LINKED ISSUE #152: Update staging and production configs to be consistent with Rails best practices
If we are going to use separate environments for staging and production we should probably do the following:

 - [x] have staging.rb and production.rb configs be identical (only changes should be the settings behind them referencing dor test/stage and prod
 - [x] have staging.rb and production reflect a Rails production environment setup. [Default Rails 4.2.4 production.yml](https://github.com/rails/rails/blob/v4.2.4/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt) Currently it seems like we are using a copy/pasted over development version

mejackreed:
Deployed to stage successfully

Closes #152 

Fixes up specs that were succeeding for a reason unknown

Also removes Settings for `APPLICATION` which aren't needed.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -4,17 +4,50 @@ require 'action_dispatch/middleware/session/dalli_store'
 Argo::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
 
-  config.assets.compress = Settings.APPLICATION.ASSETS.COMPRESS
+  # Code is not reloaded between requests.
+  config.cache_classes = true
 
-  # In the development environment your application's code is reloaded on
-  # every request.  This slows down response time but is perfect for development
-  # since you don't have to restart the webserver when you make code changes.
-  config.cache_classes = Settings.APPLICATION.CACHE_CLASSES
+  # Eager load code on boot. This eager loads most of Rails and
+  # your application in memory, allowing both threaded web servers
+  # and those relying on copy on write to perform better.
+  # Rake tasks automatically ignore this option for performance.
+  config.eager_load = true
+
+  # Full error reports are disabled and caching is turned on.
+  config.consider_all_requests_local       = false
+  config.action_controller.perform_caching = true
+
+  # Disable serving static files from the `/public` folder by default since
+  # Apache or NGINX already handles this.
+  config.serve_static_files = ENV['RAILS_SERVE_STATIC_FILES'].present?
+
+  # Do not fallback to assets pipeline if a precompiled asset is missed.
+  config.assets.compile = false
+
+  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
+  # yet still be able to expire them through the digest params.
+  config.assets.digest = true
+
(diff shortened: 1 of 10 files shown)


config/environments/test.rb
```diff
@@ -4,26 +4,29 @@
 Argo::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb
 
-  config.assets.compress = Settings.APPLICATION.ASSETS.COMPRESS
+  # The test environment is used exclusively to run your application's
```
jmartin-sul: i'd be inclined to put this at the top of the file, since it's a general comment on the env and not specific to the `cache_classes` setting.

config/environments/test.rb
```diff
@@ -31,24 +34,19 @@
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
 
-  # Don't care if the mailer can't send
-  config.action_mailer.raise_delivery_errors = false
+  # Randomize the order test cases are executed.
+  config.active_support.test_order = :random
```
jmartin-sul: cool, didn't know you could do this, seems like a good idea for avoiding inadvertent test interdependencies.

config/environments/production.rb
```diff
@@ -4,17 +4,50 @@
...
+  # Do not fallback to assets pipeline if a precompiled asset is missed.
+  config.assets.compile = false
+
+  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
+  # yet still be able to expire them through the digest params.
+  config.assets.digest = true
+
+  # Use the lowest log level to ensure availability of diagnostic information
+  # when problems arise.
+  config.log_level = :debug
```
jmartin-sul: i think that eventually this will be too verbose for prod (though i'd be in favor of leaving it at :debug in stage).  but maybe we should leave this at :debug for the initial release?  not that i'm expecting bugs or anything.

jmartin-sul:
:+1: (even if you pass on the one actionable but nitpicky comment i left)

config/environments/test.rb
mejackreed: I kept it this way to be more inline with what the Rails default does: https://github.com/rails/rails/blob/v4.2.4/railties/lib/rails/generators/rails/app/templates/config/environments/test.rb.tt

That way anyone stubbling upon could compare and quickly see, no special config thing.

config/environments/production.rb
mejackreed: Just left initial Rails default. Maybe `INFO` is a compromise between them?

jmartin-sul: yeah, `INFO` seems reasonable.  but i think it actually would be wise to do the initial release like this.  i figure we'll be spot checking the logs anyway, so if it does get really noisy, we can file an issue then and fix it at that point.

config/environments/test.rb
jmartin-sul: ah, ok.  consistency's nice, that seems fine.

PR #158, opened 2015-11-10
README updates for fixture loading, tests, and installation

jmartin-sul:
* modify instructions to reflect changes in test env configuration and better understanding of fixture loading for different envs
* minor typographical cleanup while we're here


atz:
:+1:

PR #161, opened 2015-11-11
Add scripts for developer setup, start and stop

darrenleeweber:
These simple scripts attempt to encapsulate the information in the README about getting started with argo.  The `dev_setup.sh` should be run once (or whenever a fresh start is required).  The `dev_start.sh` and `dev_stop.sh` are routine scripts to start and stop argo.

AFAIK, the sequence of shell commands in these utils cannot be rake tasks (unless rake calls out to the shell and that's ridiculous).


drh-stanford:
maybe we can move these to the shared_configs argo-laptop branch?

jmartin-sul:
@drh-stanford:  re: location: i'd be in favor of leaving them in `bin` in argo.  feels more appropriate than shared_configs since it's not DLSS specific.

bin/dev_setup.sh
```diff
@@ -0,0 +1,13 @@
...
+# Install components, setup DB and Solr:
+bundle install
+bundle exec rake argo:install
+
+# Run the server
+bundle exec rake jetty:stop  # Ensure it's not running
+bundle exec rake jetty:start # This may take a few minutes
+
+# Load and index records in the 'development' Solr core
+bundle exec rake argo:repo:load
```
jmartin-sul: is there a way to have `argo:repo:load` wait a bit or try to detect whether jetty's actually started?  `rake jetty:start` often returns before the server's accepting requests (i usually hit the web UI to make sure it's up before i run `repo:load`).

bin/dev_stop.sh
```diff
@@ -0,0 +1,11 @@
...
+
+if [ -e tmp/pids/server.pid ]; then
+    kill -9 $(cat tmp/pids/server.pid)
+    rm tmp/pids/server.pid
+fi
+if [ -e tmp/pids/delayed_job.pid ]; then
+    kill -9 $(cat tmp/pids/delayed_job.pid)
+    rm tmp/pids/delayed_job.pid
+fi
+
```
jmartin-sul: * take any shell scripting advice from me with a grain of salt.  but wouldn't it be better to use `kill` instead of a `kill -9`, to give things a chance to exit gracefully?  maybe echo the process IDs so that the user can easily check to see whether they're dead if they suspect something hung?
* also need to `rake jetty:stop`

darrenleeweber:
@jmartin-sul - good points and I'll try to address them.  I'll add the WIP to title while I revise this PR.

bin/dev_setup.sh
drh-stanford: it looks to me like the `dev_setup` code, without the `bundle install`, could be made into an `argo:dev:setup` rake task. you'd then be able to use ruby code to test whether jetty has booted and fedora/solr are responding before the `argo:repo:load`.

jmartin-sul: :+1: to that.

actually, what about calling it `argo:laptop:setup` and loading both the `development` and `test` repos/cores?  so basically the above, plus `RAILS_ENV=test bundle exec rake argo:repo:load`.  is there a way to invoke a rake task under a different env than the parent task without shelling out (since changing the env temporarily for the task invocation feels like a hack)?  i'd be fine with shelling out for the second load, since you wouldn't do that until after the regular repo load had run successfully, at which point you'd have already reaped the benefit of testing whether jetty was up.

darrenleeweber:
The last commit has revisions to address some comments.  There is an attempt to do the jetty start with checks on Solr and Fedora in a rake task, but I'm giving up on it because it's not working as expected (something about the rake context is preventing RestClient from behaving as expected, despite trying to work around WebMock).

So, these scripts should now encapsulate all the dependencies for getting setup and then doing routine argo start and stop.  It was more complicated that I expected, but these are handy for me.

drh-stanford:
what are the commented out sections in `argo.rake`? otherwise i'm :+1:

mejackreed:
Seems like these commits could be consolidated too, no?

bin/dev_setup.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+echo "Checking java 1.8.x"
```
jmartin-sul: this seems a little too specific, since presumably we'd want to check for java 1.8 or higher, right?  i'd almost rather leave the version check off than have to update it to include a new version or something every time java had a new version release.

bin/dev_stop.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+echo "Stopping Argo"
+if [ -e tmp/pids/server.pid ]; then
+    kill -9 $(cat tmp/pids/server.pid)
```
jmartin-sul: i'm still in favor of making this a `kill` instead of a `kill -9`, and printing the actual process ID from `server.pid`.

bin/dev_stop.sh
```diff
@@ -0,0 +1,18 @@
...
+    rm tmp/pids/server.pid
+    echo "Stopped Argo server"
+fi
+if [ -e tmp/pids/delayed_job.pid ]; then
+    kill -9 $(cat tmp/pids/delayed_job.pid)
+    rm tmp/pids/delayed_job.pid
+    echo "Stopped delayed_job"
+fi
```
jmartin-sul: i'm still in favor of making these `kill`s instead of `kill -9`s, and printing the actual process ID from `server.pid` and `delayed_job.pid`.

lib/tasks/argo.rake
```diff
@@ -79,6 +97,23 @@ namespace :argo do
...
+    #   fedora_url = 'http://localhost:8983/fedora/'
+    #   puts "Checking Solr"
+    #   if web_response_ok? 'http://localhost:8983/solr/'
+    #     puts "Success! Solr responds"
+    #   end
+    #   if web_response_ok? 'http://localhost:8983/fedora/'
+    #     puts "Success! Fedora responds."
+    #   end
+    # end
+
```
jmartin-sul: would be in favor of dropping both these commented out sections, since i generally lean towards not committing commented out code.  also, if we did keep this, i'd want to use the configured fedora and solr URLs instead of hardcoding the URL (even if we pretty much always expect localhost on laptop, the port could change).

what's the intended behavior of the commented out method and task?

darrenleeweber:
@jmartin-sul - good comments.  I'm OOO so I suggest you checkout the branch and modify as required (I can't predict what exactly the requirements are in this case and I've done my best so far).  This PR has served a purpose of trying to doc-code the commands required to get setup, start and stop argo.  I hope it makes it into develop, in some form, so I don't have to create a gist and always pull these scripts into my laptop setup.

lib/tasks/argo.rake
darrenleeweber: There seemed to be some consensus that rake tasks could startup jetty and check the Solr/Fedora instances, so this was an attempt to do that.  For whatever reason, it's not working.  Uncomment the code and see if you can get it working because I can't seem to get past some problems with webmock -- I hope so because it could save a lot of time required to hit the browser refresh button when you startup jetty.  Using the configured values is a really good idea and an excellent reason to do this in rake rather than a shell script (although in a 'development' env the URLs are usually the hydra-jetty default ports etc).  (In any case, line 107 should have been deleted anyway.)

bin/dev_stop.sh
darrenleeweber: In a 'development' env, it's always possible that something is totally screwy and you will need the `kill -9` to shutdown.  In general, it is a good idea to allow a graceful stop.  Perhaps the majority of the time it will be OK without the `-9`, but then the `rm` on the pids should be conditional on successful shutdown.  So, a heavy hammer is what we have here and your very welcome to modify it into a rubber mallet.

bin/dev_setup.sh
darrenleeweber: What we have here is specific, yes.  It follows the README, last time I checked.  Feel free to modify it.  Note that the grep is not fussy about the minor release number and I made no attempts to double check the README against any of the jetty/Solr/Fedora packages.

atz:
I don't see the value in including these at all.  They are faulty reproductions of existing rake tasks and trivial wrappers.  

Default `kill -9` is a terrible idea and prevents the server from having a chance to shutdown cleanly, including writing buffered data and logs to disk.  

The idea of checking for a running Jetty is not new or unique to Argo.  If jettywrapper tasks aren't sufficient or working, add or fix them in jettywrapper.  There is zero reason to keep the fix to ourselves in this one project.

drh-stanford:
I've gone through the README again (PR #184) and I think the steps in these scripties are all covered in the README now (whereas they weren't before this PR started). Given that developers all have their own workflows, I think our efforts should be on making the README clear and comprehensive for developer setup and ops, rather than helper scripties outside the rake tasks and whatnot. So, I'm closing this PR for now. Please reopen, and rebase/squash, if there's still a desire to have the helper scripties. Thanks.

jmartin-sul:
:+1: for @drh-stanford's suggestion.

FWIW, it should be possible to leave scripts like this in the bin directory of a personal instance and have them live through pulls and branches and such, since nothing else should conflict with the names.  the experimental rake code could live on a dormant branch (which is admittedly slightly more of a pain).  i've definitely done that in the past for utility scripts where i didn't have time at the moment to generalize enough for committing back to the repo.  

as it is, i'd prefer not to commit back to mainline development with things like `kill -9` and code that behaves in an unexpected fashion when used in a rake task instead of pry.  but i do think that these are fine ideas in principle, and would be helpful time-saving tools if they were slightly more robust, so i'm totally open to revisiting this later.  but i don't think i'll have time to fix these either at the moment, and would be unlikely to incorporate them into my workflow as-is.

PR #169, opened 2015-11-11
switch to new cert path style

jmartin-sul:
path is specified in settings, no assumption about common parent directory for key and cert.

you'll either need to update your settings (e.g. `development.local.yml`) to specify the path to the cert and key files relative to the argo app root, or you'll need to specify the full path for each.  or, you can pull changes to the private `shared_configs` version, once that PR is merged.  `dor_development.rb` and the appropriate settings file will need to be updated concurrently for things to keep working.

the relevant settings are `Settings.SSL.CERT_FILE` and `Settings.SSL.KEY_FILE`.

@atz @darrenleeweber @mejackreed @drh-stanford @tingulfsen

### DIFF
--- a/config/environments/dor_development.rb
+++ b/config/environments/dor_development.rb
@@ -1,9 +1,7 @@
-cert_dir = File.expand_path(File.join(File.dirname(__FILE__), '../certs'))
-
 Dor.configure do
   ssl do
-    cert_file File.join(cert_dir, Settings.SSL.CERT_FILE)
-    key_file File.join(cert_dir, Settings.SSL.KEY_FILE)
+    cert_file Settings.SSL.CERT_FILE
+    key_file Settings.SSL.KEY_FILE
     key_pass Settings.SSL.KEY_PASS
   end
 


darrenleeweber:
OK, better merge this to keep in sync with shared_settings; or is this a case of 'convention over configuration' that rails is so well know for?  Is this 'over configuring' something that is a convention?

mejackreed:
yep for sure :+1: ... need @jmartin-sul to rebase with :cop: changes

cbeer:
:+1: being explicit about paths, especially for something like keys, seems very healthy. This also fits nicely with other key management practices, where the cert (which could be public) is even kept in a separate directory from the key (which must be private).

jmartin-sul:
@darrenleeweber not really a change in where things are on the configuration vs convention spectrum.  just a switch from using a hardcoded assumption about cert parent directory (and a setting relative to that assumption)  to using the path as specified in the setting (with no assumption in the `config/environments` code about parent directory, since, e.g., the key and cert don't live under the argo app dir structure or share a parent directory on deployed instances such as argo-stage).

will rebase and merge since the consensus seems to be in favor, unless there's an objection shortly...

PR #177, opened 2015-11-13
Make rake:argo:install create log/ and tmp/ directories if they do no…

tingulfsen:
…t already exist.

### DIFF
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -38,8 +38,16 @@ end
 
 namespace :argo do
   desc 'Install db, jetty (fedora/solr) and configs fresh'
-  task :install => ['argo:jetty:clean', 'argo:jetty:config', 'db:setup', 'db:migrate'] do
+  task :install => ['argo:log_dir', 'argo:tmp_dir', 'argo:jetty:clean', 'argo:jetty:config', 'db:setup', 'db:migrate'] do
     puts 'Installed Argo'
+  end
+
+  task :log_dir do
+    FileUtils.mkdir_p 'log'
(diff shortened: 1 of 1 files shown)


mejackreed:
@tingulfsen are you running into issues if these dir's are not created? Rails should just create these automagically for us

atz:
What part of rails?  We aren't running `rails generate`, for example.

cbeer:
It's also conventional to just check in `log` to git (and, not unheard of to do the same to `tmp`), and gitignore the contents.

tingulfsen:
I actually really like the thought of checking in log/ and tmp/ and then ignoring the contents. Is everyone else happy with that? It seems like a very simple and foolproof way to ensure that the directories will always exist.

mejackreed:
:+1:

atz:
Yeah, you can do that but you will need the directory to have a file in it for git to care about it existing.  Convention is to put a .gitignore in the empty dir, or my preference, to put a README.md in each directory declaring its purpose.  "This is the default application temp directory, as configured by XYZ setting".

cbeer:
Rails, by default, uses `.keep` instead of `.gitignore`.

jmartin-sul:
i'm also :+1: on checking in `log/` and `tmp/`.

drh-stanford:
this PR is moot due to #179 , right?

tingulfsen:
Yeah, I was just waiting to see what the reaction to #179 would be before deleting this pull request.

drh-stanford:
PR #179 is merged so this can be closed.

PR #179, opened 2015-11-13
Adding empty log and tmp directories to the Git repository, for devel…

tingulfsen:
…oper convenience.

This adds log/ and tmp/ to source control but ignores any files inside them. I think this is a better solution than #177


drh-stanford:
:+1:

jmartin-sul:
:+1:

PR #181, opened 2015-11-13
resurrect the trigger for running 'rake argo:htaccess' on restart

jmartin-sul:
now with explanation.

(the explanation being that public gets overwritten on deployments, and so the .htaccess file with the workgroup list is lost, and then argo doesn't get the workgroup info it needs about the user to do perm checks)


mejackreed:
:+1:

Issue #97, opened 2015-11-02
Make DOR Workflow Services log be configurable.


tingulfsen:
Currently, the log file for DOR workflow services is written straight into the Argo main directory. Change so that the location can be configured, and store the file in the log/ directory by default.

tingulfsen:
Pull requests are in for DOR Services to configure this.

atz:
DOR PR's are merged.

mejackreed:
@tingulfsen is this still in progress?

atz:
`config/settings.yml` has
```
DOR_SERVICES_LOGFILE: 'log/workflow_service.log'
```
But the file still gets created in root dir with just the auto-header line:
```
# Logfile created on 2015-11-10 18:52:49 -0800 by logger.rb/1.2.8
```

So there may be an issue with timing the configuration before Dor invokes DWFS with the default config.

atz:
@mejackreed: note, sorta janky name for the setting since this is really a dor-workflow-service specific logfile and dor-services has plenty of other logs it should be generating.

drh-stanford:
i agree re: naming since it's a `workflow` service setting in `Dor.configure`. See here: https://github.com/sul-dlss/argo/blob/develop/config/environments/dor_development.rb#L18

Maybe `Settings.WORKFLOW_LOGFILE`, etc. to match `Settings.WORKFLOW_URL`.

drh-stanford:
See PR #185 for the renaming issue

mejackreed:
Is this still valid?

PR #185, opened 2015-11-14
rename workflow configuration variables

drh-stanford:
This addresses the naming issue in #97.

**NOTE** that merging this PR means that we will need to change the shared_configs branches and our local development configurations.

### DIFF
--- a/config/environments/dor_development.rb
+++ b/config/environments/dor_development.rb
@@ -15,8 +15,8 @@ Dor.configure do
 
   workflow do
     url Settings.WORKFLOW_URL
-    logfile Settings.DOR_SERVICES_LOGFILE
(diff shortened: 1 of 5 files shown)


mejackreed:
If were going to make this change, could we also put them under a different heading in the settings document rather than `#URLS`?

drh-stanford:
ok, i created a new Workflow section in `settings.yml`

mejackreed:
:+1:

mejackreed:
Looks good... A quick review of our config repo. I'm not sure what exactly needs to change there. Should these setting be different for our various env?

drh-stanford:
No, they can be the same for all envs. So I think the defaults in `config/settings.yml` should suffice.

Issue #187, opened 2015-11-15
render_buttons should not use blocking requests


mejackreed:
Currently happening with:
- [ ] wfs requests
- [ ] fedora requests

jcoyne:
This is still an issue because `!object.allows_modification?` calls the workflow service

PR #196, opened 2015-11-16
Refactor render_buttons to use non-blocking ajax requests.

mejackreed:
Buttons will be shown to user, but disabled unless api responses come back as true

Addresses part of #187 

Reduces the number of blocking requests by 6 - 9 requests per show page load, depending on the page and the logic path executed.

### DIFF
--- a/app/controllers/workflow_service_controller.rb
+++ b/app/controllers/workflow_service_controller.rb
@@ -1,0 +1,80 @@
+class WorkflowServiceController < ApplicationController
+  ##
+  # Is a document closeable?
+  def closeable
+    @status = check_if_can_close_version
+    render json: @status
+  end
+
+  ##
+  # Is a document openable?
+  def openable
+    @status = check_if_can_open_verison
+    render json: @status
+  end
+
+  ##
+  # Has an object been published?
+  def published
+    @status = check_if_published
+    render json: @status
+  end
+
+  ##
+  # Has an object been submitted?
+  def submitted
+    @status = check_if_submitted
+    render json: @status
+  end
+
+  ##
+  # Has an object been accessioned?
+  def accessioned
+    @status = check_if_accessioned
+    render json: @status
+  end
+
+  private
+
+  ##
+  # Ported over logic from app/helpers/dor_object_helper.rb#LN119
+  # @return [Boolean]
+  def check_if_published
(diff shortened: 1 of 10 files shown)


mejackreed:
Coverage up 0.8%! :dancers:

app/controllers/workflow_service_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    render json: @status
+  end
+
+  private
+
+  ##
+  # Ported over logic from app/helpers/dor_object_helper.rb#LN119
+  # @return [Boolean]
+  def check_if_published
+    return true if Dor::WorkflowService.get_lifecycle('dor', params[:pid], 'published')
```
drh-stanford: it would be nice to start having the `Dor::WorkflowService` calls put into a helper. Something like this perhaps?

```ruby
# untested example
def lifecycle_is?(druid, status, active_only = false)
  fn = (active_only ? :get_active_lifecycle : :get_lifecycle)
  Dor::WorkflowService.send(fn, 'dor', druid, status).is_a?(Time) ? true : false
end
```

over time it would help us isolate which workflow service calls are actually being made by Argo vs. dor-services, etc. maybe that's outside the scope of this PR...

jmartin-sul: :+1: for @drh-stanford's suggestion, though yeah, maybe outside the scope of this PR.

i think there are one-off methods in dor-services that do this sort of encapsulation, though i don't think we're consistent about writing or using methods like that in argo or dor-services.

mejackreed: Although I like the reuse idea, I disagree with the approach of putting it into a helper. I think helpers should be used for sharing code between controllers and views, and by adding these methods to helpers, it makes it easier for dev's to put these blocking requests into a view, causing performance issues.

Is a class under `lib` an approach that would be useful?

mejackreed: Maybe with caching / memoization too?

drh-stanford: Yes, `lib` is fine with me. caching active workflows may not be effective as they may change at anytime (although they do send JMS messages on changes which we could consume for invalidation). workflows that are reliably cacheable are those that have been archived which is a large percentage of them.

PR #202, opened 2015-11-17
remove mods editor vestigial parts

mejackreed:
Most of this added in 578c124bef85be2f65f653abf475fcc6ff41e007

According to @LynnMcRae  this isn't a feature in Argo anymore.


drh-stanford:
Can you please elaborate on why we're deleting `jquery.min.js` and the like in this PR? Is everything in `xml/**` tied only to the MODS editor feature?

mejackreed:
Yes correct. Non of these files, (as far as I can tell) are being required or executed anywhere in Argo. We require jquery from the gem here: https://github.com/sul-dlss/argo/blob/develop/app/assets/javascripts/application.js#L13

drh-stanford:
this is working fine for me. there's still a POST route for `/mods` that I'm wondering about too...

mejackreed:
Ah hah. Yeah I wasn't sure if that stuff was used.

drh-stanford:
I'm pretty sure it's broken:

```
curl -X POST --data ''  http://localhost:3000/items/asdf/mods
...
The action &#39;update_mods&#39; could not be found for ItemsController
...
```

mejackreed:
Ah.. yeah. Added here: https://github.com/sul-dlss/argo/commit/578c124bef85be2f65f653abf475fcc6ff41e007

PR #206, opened 2015-11-17
run rake ci as RAILS_ENV=test

### LINKED ISSUE #172: setting RAILS_ENV after Rails has loaded only creates misinformation
we should remove lines that set RAILS_ENV like so:
```ruby
ENV['RAILS_ENV'] = 'test'
```

e.g.:  https://github.com/sul-dlss/argo/blob/70416658b95549881c847cf01f4b06a6775ed7d1/lib/tasks/argo.rake#L20

not sure if we do this elsewhere, but we shouldn't do it anywhere.  note that doing `Rails.env = ` is no better.  neither actually reloads the environment.  all the old configuration stays in place, though `Rails.env` and `ENV['RAILS_ENV']` will misreport the loaded environment as the name that was given in the "reset".

for the specific issue seen in the rake task, we should do something like:
```ruby
if Rails.env.test?
  # run ci tests
else
  system('rake ci RAILS_ENV=test')
end
```

more explanation on that proposed fix in the comment where @mejackreed suggested it:  https://github.com/sul-dlss/argo/pull/114#issuecomment-154270331

and the comment where i explained my testing in a bit more detail: https://github.com/sul-dlss/argo/pull/114#issuecomment-156216601

in the meantime, no harm in merging PR #114, which partially addresses the situation (and which spawned this investigation).

mejackreed:
Closes #172

### DIFF
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -17,14 +17,17 @@ end
 task :default => [:ci]
 
 task :ci do
-  ENV['RAILS_ENV'] = 'test'
-  Rake::Task['argo:install'].invoke
-  jetty_params = jettywrapper_load_config()
-  error = Jettywrapper.wrap(jetty_params) do
-    Rake::Task['argo:repo:load'].invoke  # load 'em all!
-    Rake::Task['spec'].invoke
+  if Rails.env.test?
+    Rake::Task['argo:install'].invoke
+    jetty_params = jettywrapper_load_config()
+    error = Jettywrapper.wrap(jetty_params) do
+      Rake::Task['argo:repo:load'].invoke  # load 'em all!
+      Rake::Task['spec'].invoke
+    end
+    raise "test failures: #{error}" if error
+  else
+    system('RAILS_ENV=test rake ci')
   end
-  raise "test failures: #{error}" if error
 end
 
 if ['test', 'development'].include? ENV['RAILS_ENV']


lib/tasks/argo.rake
```diff
@@ -17,14 +17,17 @@ end
 task :default => [:ci]
 
 task :ci do
-  ENV['RAILS_ENV'] = 'test'
-  Rake::Task['argo:install'].invoke
-  jetty_params = jettywrapper_load_config()
-  error = Jettywrapper.wrap(jetty_params) do
-    Rake::Task['argo:repo:load'].invoke  # load 'em all!
-    Rake::Task['spec'].invoke
+  if Rails.env.test?
```
drh-stanford: why not make this:

```
fail unless Rails.env.test?
```
and then change `.travis.yml` to always set the RAILS_ENV=test? That way you cannot run the `ci` task without explicitly setting RAILS_ENV=test. That seems more consistent to me.

See http://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-.travis.yml

mejackreed: Because what if you want to run it locally? I usually run rake ci locally, and this removes the need for setting the env in travis.  ¯\_(ツ)_/¯

mejackreed: See context from #172

lib/tasks/argo.rake
```diff
@@ -17,14 +17,17 @@ end
...
+    Rake::Task['argo:install'].invoke
+    jetty_params = jettywrapper_load_config()
+    error = Jettywrapper.wrap(jetty_params) do
+      Rake::Task['argo:repo:load'].invoke  # load 'em all!
+      Rake::Task['spec'].invoke
+    end
+    raise "test failures: #{error}" if error
+  else
```
drh-stanford: ok then maybe this `else` could be `elsif Rails.env.development? ...doit... else fail` i find that more readable for the intention of the ci task.

mejackreed: I hear you on that. The intent of this PR was to implement the consensus that I thought was gained from #172. Happy to close this down, if we don't have that concensus.

drh-stanford:
ok :+1:

mejackreed:
:cool:

PR #218, opened 2015-11-18
remove dead POST route to /items/pid/mods

drh-stanford:
This route is missing its controller method. We are no longer using the MODS-specific editor (see #202)

```
curl -X POST --data ''  http://localhost:3000/items/asdf/mods
...
The action &#39;update_mods&#39; could not be found for ItemsController
...
```


mejackreed:
:+1:

PR #227, opened 2015-11-19
refactor access to authoritative APO and Collection(s) druids and titles

### LINKED ISSUE #75: Report view - drop redundant Admin Policy and Collection columns
When the facets split into DOR and Hydrus lists, the split versions were added as reporting columns and included in the default report. Only the "all" column makes sense, as the appropriate APO and Collection title should show according to the druid, with column headings "Admin Policy" and "Collection"

drh-stanford:
This fixes #75. 

The PR consolidates access to the `SolrDocument` APO druid value, for example, by providing a constant for `is_governed_by_ssim`, and a method `apo_id` to read the value. It does this for APO druid, APO title, Collection druid(s), and Collection title(s).

I've marked it work in progress because I need to test it with items that have collections (we currently don't have local fixtures for that).

@mejackreed I tried to address your comment in #75, but I'd like to pair up to review this PR. I have questions about when something is a `SolrDocument` vs. a simple `Hash`, for example.

### DIFF
--- a/app/helpers/argo_helper.rb
+++ b/app/helpers/argo_helper.rb
@@ -123,16 +123,15 @@ module ArgoHelper
   # Ideally this method should not make calls to external services to determine
   # what buttons should be rendered. These external requests are blocking and
   # will not allow the page to load until all requests are finished.
+  # @param [SolrDocument] doc
   # @return [Array]
   def render_buttons(doc, object = nil)
     pid = doc['id']
     object ||= Dor.find(pid)
-    apo_pid = ''
     # wf_stuff.include? 'accessionWF:completed:publish'
-    begin
-      apo_pid = doc['is_governed_by_ssim'].first.gsub('info:fedora/', '')
-    rescue
-    end
+
+    apo_pid = doc.apo_pid
+
     buttons = []
     if pid
       buttons << {
--- a/spec/models/solr_document_spec.rb
+++ b/spec/models/solr_document_spec.rb
@@ -74,5 +74,4 @@ describe SolrDocument, :type => :model do
       expect(versions['2']).to match a_hash_including(:tag => '1.1.0', :desc => 'Minor change')
     end
   end
-
 end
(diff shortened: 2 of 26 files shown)


mejackreed:
This looks like an awesome start. I really like how you consolidated all of the fields to the model. I'll be available online all day tomorrow, lets take a look at it.

app/helpers/argo_helper.rb
```diff
@@ -123,16 +123,15 @@ def render_document_sections(doc, action_name)
...
     pid = doc['id']
     object ||= Dor.find(pid)
-    apo_pid = ''
     # wf_stuff.include? 'accessionWF:completed:publish'
-    begin
-      apo_pid = doc['is_governed_by_ssim'].first.gsub('info:fedora/', '')
-    rescue
-    end
+
+    apo_pid = doc.apo_pid
```
mejackreed: Ahh... so much nicer!

drh-stanford:
Ok, I've made all the updates I was planning and did some testing on items with collections. For that, I also added fixtures for an item (`qq613vj0238 `) that belongs to a collection (it includes the agreement and APO).

spec/models/solr_document_spec.rb
```diff
@@ -75,4 +75,36 @@
     end
   end
 
+  describe 'SolrDocument concerns' do
```
mejackreed: We test the other concerns here: https://github.com/sul-dlss/argo/tree/develop/spec/models/concerns

Could we keep the same pattern? Or move those specs here?

I think the expectation, is that I should be able to find the respective specs in a similar location.

drh-stanford:
I didn't notice that concerns folder in spec. So I relocated the tests.

drh-stanford:
Coveralls says that the coverage went down but I have 100% coverage in all the new code I added. Is there a way to see specifically where the coverage went down on the diff? The 'details' link shows that the coverage went down in places I didn't change? https://coveralls.io/builds/4220436

mejackreed:
@drh-stanford rebase on current develop and you should get the coverage increases.

drh-stanford:
I'm still seeing the coverage drop. Am I rebasing incorrectly?

```
git checkout develop
git pull origin develop
git checkout issue-75
git rebase develop
git push -f origin issue-75
```

mejackreed:
Yep that seems correct. Looking into it.

drh-stanford:
@mejackreed what did you find on the coveralls accuracy?

mejackreed:
I think its just a fluke :confounded:

PR #230, opened 2015-11-19
move indexing code to class in lib

mejackreed:


lib/indexer.rb
```diff
@@ -0,0 +1,121 @@
...
+
+
+    def reindex_all args
+      index_log = Logger.new(File.join(Rails.root, 'log', 'reindex.log'))
+      index_log.formatter = Logger::Formatter.new
+      index_log.level = ENV['LOG_LEVEL'] ? Logger::SEV_LABEL.index(ENV['LOG_LEVEL']) : Logger::INFO
+      $stdout.sync = true
+      start_time = Time.now
+
+
```
jmartin-sul: i would vote for turning off this rule in hound and rubocop.  more often than not, when i have two lines of whitespace in a row, it's an intentional formatting choice (which i think is reasonable if its judicious).  and it's pretty rare that i commit this sort of thing on accident.

a non-issue in this specific case since this is messy experimentation code, but thought i'd mention it for consideration by others.

PR #236, opened 2015-11-20
adds All Workflows link in the header

### LINKED ISSUE #64: Add "All Workflows" link to global navigation
We eliminated the little table icon in the facet list that took you to a view of all workflows for all objects on the front page -- essentially the workflow view with all items found. Expand the row of global links at the top of pages to be 

   ...  All Workflows | Register Items | Register APO 

"All Workflows", if picked from anywhere but the home page, should act as a Start Over (results/facets cleared) with the all-workflows view.

drh-stanford:
Fixes #64.

This PR adds an `all` parameter to the `/report/workflow_grid` route. If it is passed in (no value needed), then it will reset the facets (by calling `facet_tree` with no arguments).

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -26,6 +26,8 @@
       </li>
     <% end %>
 
+    <li><%= link_to "All&nbsp;Workflows".html_safe, report_workflow_grid_path %></li>
+
     <li><%= link_to "Register&nbsp;Items".html_safe, register_items_path %></li>
 
     <%if current_user.is_admin or current_user.is_manager %>


app/helpers/workflow_helper.rb
```diff
@@ -4,7 +4,7 @@ def show_workflow_grid?
   end
 
   def render_workflow_grid
-    workflow_data = facet_tree('wf_wps')['wf_wps_ssim']
+    workflow_data = params[:all] ? facet_tree : facet_tree('wf_wps')['wf_wps_ssim']
```
mejackreed: I'm a little confused what the expected behavior here is? Could you elaborate some more?

drh-stanford: The `:all` parameter intends to reset the facets to include all items (akin to "Start Over"). Is there a better way to reset the facets in Blacklight?

mejackreed: The vanilla `report_workflow_grid_path` route should do this already. Is it not?

drh-stanford:
Revised PR so that "All Workflows" link goes to generic `/report/workflow_grid` route.

LynnMcRae:
The All Workflows link goes here in stage but gets a system error. It works in -dev though, so maybe just a question of catchup.

https://argo-stage-b.stanford.edu/report/workflow_grid

mejackreed:
@LynnMcRae That error was reported in #400 and closed by #406. Will redeploy latest to stage.

PR #243, opened 2015-11-20
make mysql a production gem

drh-stanford:
we're not using mysql except in production

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ Argo is the administrative interface to the Stanford Digital Repository. It uses
 Install Java 1.8 (or newer) JRE (and also the JDK on Mac OSX).  It is required for the version of Solr in use.
 http://java.com/en/download/
 
+Install MySQL.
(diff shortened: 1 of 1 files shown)


mejackreed:
:+1:

jmartin-sul:
actually, i use mysql on my laptop.  we once ran into a bug that didn't show up in sqlite, but did in mysql (due to some date formatting monkey patching).  so i've kept running mysql just to suss out incompatibilities earlier.  

also, we deploy to dev without production, so i think this would break dev deployments (all the canonical VM deployments use mysql):
https://github.com/sul-dlss/argo/blob/develop/config/deploy/dev.rb#L5

mejackreed:
:clap: good points @jmartin-sul

drh-stanford:
i think that would argue for the `database.yml` to use mysql for development and test. but i'm closing the PR as we have other priorities now

PR #246, opened 2015-11-20
profiler refactoring and cleanup, move indexing code to class in lib, improved indexing tools

jmartin-sul:
This is the pull request for all of the changes to profiling and indexing related code in Argo.

It introduces a lib for profiling, based on an older simpler version of the same idea.  See the PROFILING.md document for more info on how profiling works in Argo.  The short summary is that it provides convenience methods for using the RubyProf gem to profile code blocks and dump results to files with reasonable names.

It also introduces libs with a bunch of indexing code, including:
* centralized code for indexing objects and logging info about the status of indexing attempts.  This code can be reused by other parts of the app.
* methods to query Fedora and Solr, and to use those query results (or other sources of object IDs) to build lists for reindexing.
* infrastructure for managing bulk reindexing using delayed_job workers.

See INDEXING.md for more info on indexing, things to watch out for, and ways to do custom bulk indexing.

Thanks to everyone who contributed code and feedback along the way in this.


lib/indexer.rb
```diff
@@ -0,0 +1,151 @@
...
+          index_log.error("#{e.class}: #{e.message}")
+          #  if errors == 3
+          #    index_log.fatal("Too many errors. Exiting.")
+          #    raise e
+          #  end
+        end
+        pbar.inc(1)
+      end # pids.each
+      solr.commit
+      puts 'Reindexing complete'
```
jmartin-sul: since this is intended to be run from the command line, this doesn't seem like a misuse of stdout, unless there's a more appropriate way to write back to the user.  but it does seem like this would be categorized as standard output.

lib/indexer.rb
```diff
@@ -0,0 +1,151 @@
...
+
+
+      # TODO: break out pid list creation
+      $stdout.puts 'Discovering PIDs...'
+      index_log.info 'Discovering PIDs...'
+      dor_pids = []
+      solr_pids = []
+      if args[:query] != ':ALL:'
+        q = (args[:query].nil? || args[:query] == ':MISSING:') ? '*:*' : args[:query]
+        puts q
```
jmartin-sul: ditto above stdout comment

lib/profiler.rb
```diff
@@ -0,0 +1,25 @@
...
+      File.open "#{Rails.root}/log/profiler/#{out_file_id}-stack.html", 'w' do |file|
+        RubyProf::CallStackPrinter.new(results).print(file)
+      end
+
+      File.open "#{Rails.root}/log/profiler/callgrind.#{out_file_id}", 'w' do |file|
+        RubyProf::CallTreePrinter.new(results).print(file, :min_percent => 1)
+      end
+    end
+  end
+end
```
jmartin-sul: i fixed this and force pushed it (along with fixes for the parenthesis complaints).  not sure why it's not showing up on github.

mejackreed: Yeah thats weird

lib/indexer.rb
```diff
@@ -0,0 +1,151 @@
...
+    end
+  end
+
+  # TODO: DRY up the repetition btwn here and dor_controller
+  class Indexer
+    @@index_logger = nil
+    def self.index_logger
+      return @@index_logger if @@index_logger
+
+      @@index_logger ||= Logger.new("#{Rails.root}/log/indexer.log", 10, 3240000)
```
jmartin-sul: after a cursory look into "class instance variables":  they seem like needless complexity for this case.  my impression is that they're used when you want a static variable that's shared among classes, but not super-classes or sub-classes (and i guess class variables are shared by sub-classes also).  i vote for leaving this as-is for now and dealing with this when `Indexer` gets subclassed and the subclass needs its own logger.

i'd also vote for turning off this rule.

lib/indexer.rb
```diff
@@ -0,0 +1,151 @@
...
+      index_log.info 'Discovering PIDs...'
+      dor_pids = []
+      solr_pids = []
+      if args[:query] != ':ALL:'
+        q = (args[:query].nil? || args[:query] == ':MISSING:') ? '*:*' : args[:query]
+        puts q
+        start = 0
+        resp = Dor::SearchService.query(q, :sort => 'id asc', :rows => 1000, :start => start, :fl => ['id'])
+        while resp.docs.length > 0
+          solr_pids += resp.docs.collect { |doc| doc['id'] }
```
jmartin-sul: i'm inclined to not do cleanup on the moved legacy code, since there's a good chance it'll change greatly or be discarded entirely.

lib/indexer.rb
```diff
@@ -0,0 +1,151 @@
...
+        $stdout.puts msg
+        index_log.info msg
+      end
+      if dor_pids.present?
+        pids = dor_pids - solr_pids
+      else
+        pids = solr_pids
+      end
+      pids.delete_if { |pid| pid !~ /druid:/ }
+      time = Time.now - start_time
```
jmartin-sul: i'm inclined to not cleanup the moved legacy code just yet, since there's a good chance it'll change greatly or be discarded entirely.

lib/indexer.rb
mejackreed: I'm a bit :-1: on turning the rule off. I think there a lot of cases where a class instance variables aren't appropriate and having this rule may help raise those cases.

lib/indexer.rb
```diff
@@ -0,0 +1,151 @@
...
+  end
+
+  # TODO: DRY up the repetition btwn here and dor_controller
+  class Indexer
+    @@index_logger = nil
+    def self.index_logger
+      return @@index_logger if @@index_logger
+
+      @@index_logger ||= Logger.new("#{Rails.root}/log/indexer.log", 10, 3240000)
+      @@index_logger.formatter = proc do |severity, datetime, progname, msg|
```
jmartin-sul: inclined to leave this unless it's a totally standard ruby practice that i was just unaware of.

lib/indexer.rb
jmartin-sul: fair enough on the rule.

how do you feel about leaving this particular one?

mejackreed:
I think this looks great. The only reason I would say could we hold off merging because:
 - there isn't much documentation on this and what these things do, how to use them
 - it still feels like this is in flux as we don't really have our indexing solution quite yet it place, no?

Feel free to contradict me here. :smile_cat:

lib/indexer.rb
mejackreed: :ok_hand:

lib/indexer.rb
```diff
@@ -0,0 +1,151 @@
...
+      Argo::Profiler.prof(out_file_id) { Argo::Indexer.reindex_pid_list pid_list, should_commit }
+    end
+
+    def self.get_pids_for_model_type(model_type)
+      Dor::SearchService.risearch "select $object from <#ri> where $object <fedora-model:hasModel> #{model_type}"
+    end
+
+
+    def self.reindex_all(args)
+      warn '[DEPRECATION] `reindex_all` is deprecated.  Please use the bulk reindexing system instead.'
```
drh-stanford: i would say make this a `fail` and not a deprecation. this code is stale, right? we have it in the history if we need it.

lib/tasks/argo.rake
```diff
@@ -320,83 +320,6 @@ namespace :argo do
...
-        #  if errors == 3
-        #    index_log.fatal("Too many errors. Exiting.")
-        #    raise e
-        #  end
-      end
-      pbar.inc(1)
-    end # pids.each
-    solr.commit
-    puts 'Reindexing complete'
+    Argo::Indexer.reindex_all args
```
drh-stanford: i would favor not keeping this rake task to old code as a deprecation.

jmartin-sul:
yeah, documentation's a good point.  should add that before merging back.

not sure how to approach the testing.  i think `reindex_all` would be a pain to test, but i think there's a good chance it'll go away.  and the rest don't seem terrible to test, right?

i think `SimpleLogJob` will probably go away too, but i think it's ok for it to stay for now.  and i think we'll be adding a few more methods to make it easy and deterministic to launch a full reindex as delayed jobs (with optional profiling).  but i think the new methods that are there are rather stable.  there is repetition between `Indexer` and `dor_controller`, but i'd be comfortable filing an issue to refactor that and not necessarily block these changes if that refactoring doesn't happen immediately (though i don't want to defer that for too long either).

so... i'd definitely block this merge for documentation.  otherwise, i think it's a definite improvement over develop and worth bringing back in, though maybe i'm more certain than is warranted that it's generally on the right track in terms of structure and approach.  totally fine holding off on merging if you'd prefer to wait and see.

lib/tasks/argo.rake
jmartin-sul: fair point, and a thing i've complained about on other PRs.  will remove.

lib/indexer.rb
drh-stanford: i've seen `_varname` commonly used to denote local variables, but not unused ones.

lib/bulk_reindexer.rb
```diff
@@ -0,0 +1,15 @@
+module Argo
+  class BulkReindexer
+    def self.queue_pid_reindexing_jobs(pid_list, should_commit = false, should_profile = false, priority = 1, batch_size = 10)
```
drh-stanford: if I understand this correctly, you have `@param [Array<String>] pid_list`, and `get_pid_lists_for_full_reindex` returns an `Array<Array<String>>`. it's a little dense here in this important code and could use some comments.

app/jobs/indexing_job.rb
```diff
@@ -0,0 +1,18 @@
...
+# Base IndexingJob class
+class IndexingJob < ActiveJob::Base
+  queue_as :indexing
+
+  ##
+  # Perform indexing job
+  # @param [Array] pid_list
+  # @param [Boolean] should_commit
+  # @param [Boolean] should_profile
+  def perform(pid_list, should_commit = false, should_profile = false)
```
drh-stanford: The `should_profile` could be removed and a `Settings` variable used instead to indicate profiling. That way queued jobs aren't pinned to profiling... or the next line could be something like `if should_profile && Settings.ENABLE_PROFILING`

jmartin-sul:
the small modification to the same method he was commenting on seems to have disappeard @cbeer's enumerator suggestion.  but that was a fair point, and will do when cleaning this branch up (later today, hopefully).

lib/indexer.rb
```diff
@@ -0,0 +1,63 @@
...
+  class Indexer
+    ##
+    # Returns a Logger instance for recording info about indexing attempts
+    # @yield attempt to execute 'entry_id_block' and use the result as an extra identifier for the log 
+    #   entry.  a placeholder will be used otherwise. 'request.uuid' might be useful in a Rails app.
+    def self.generate_index_logger(&entry_id_block)
+      index_logger = Logger.new(Settings.INDEXER.LOG, Settings.INDEXER.LOG_ROTATION_INTERVAL)
+      index_logger.formatter = proc do |severity, datetime, progname, msg|
+        date_format_str = Argo::Config.date_format_str
+        entry_id = begin entry_id_block.call rescue '---' end
```
jmartin-sul: i'm open to better ways of trying a block to get a result, and falling back to a default if the block doesn't return something usable for any reason.  but what hound's complaining about seems ok to me.

is the objection one of style (not the most readable/maintainable way)?  or one of robustness (running whatever and just rescuing is fragile in some way i can't see)?

mejackreed: I think it's just more about generally not catching exceptions from `StandardError` https://www.relishapp.com/womply/ruby-style-guide/docs/exceptions

jmartin-sul: i get the sense that the advice there is about not being overly broad when catching exceptions, which i think is a very good rule of thumb.  but in this case, i think we want to trap anything that's `StandardError` and proceed with logging, since that seems more useful than dying.  for instance, even if the caller gave a block that worked incorrectly and always failed, i think it's better to otherwise log what that caller wants if possible, than it is to fail out of the logging call.

more catastrophic things should be `Exception`s, which aren't trapped by the default `rescue`, so those should still bubble up, which also seems right to me.

mejackreed: Agree... on all counts. Maybe explicitly calling out `StandardError` would silence the hounds? Or would it compain that its redundant :smirk_cat:

jmartin-sul:
this is, unfortunately, not going to make it out of WIP mode before winter break.  i did add a bunch of tests and comments this week, partly motivated by some indexing code consolidation.  here's what i can think of that remains to be done:
* a guide to profiling needs to be written.  how to enable profiling for indexing jobs, what the settings mean, how to do profiling of other code on an ad-hoc basis while developing.
* a guide to indexing needs to be written.  how to kick off the basic reindex_all rake task, how `PidGatherer` works, how `BulkIndexer` works, how `Indexer` works, the shortcomings of the `risearch` based stuff.  How to do more advanced things like use `BulkIndexer` to do targeted batches, or unindexed things, or how to populate `PidGatherer` with things from a file and then feed that to `BulkIndexer`.  note to self, i have saved pry sessions on my laptop from argo-prod and argo-prod-b with good example commands.
* more comments and tests could be written (with some basic overview of `PidGatherer` and `BulkIndexer` functionality in the comments).

jmartin-sul:
I am once again moving this out of the WIP state, and looking for final feedback so that these changes can be merged.

Known issues:
* this does not yet deal with the strict ordering issue, as described in the INDEXING.md doc.  i don't think that should be a deal-breaker, as this machinery is still quite useful in the absence of automated strict ordering, and the approaches we're considering for strict ordering would build on this PR rather than replace it.
* I have yet to implement Chris' generator suggestion.  Currently, the only thing that calls the `solr_pids` method intends to consume all the PIDs anyway.  I'd be happy to file an enhancement to fix that later, but I'd rather not hold this up further for that, if others wouldn't mind.
* I tried specifying `StandardError` in `begin entry_id_block.call rescue '---' end` (to shut hound up), but ruby gave me an error.  And I think think it reads better as-is anyway.
* I believe the other hound complaints were deemed acceptable after discussion in the comments.

Have at it!  For the sake of debugging the two in combination, I'd appreciate it if reviewers would at least skim INDEXING.md and PROFILING.md as an introduction to the code.  They're somewhat lengthy documents, but I tried to be as concise as possible while still explaining all of the usage details.

mejackreed:
:clap: :clap: :clap: ... great work here @jmartin-sul . I've read through the documentation you added, as well as the code here. It all seems pretty good, and we have addressed a lot of things along the way. A second set of eyes for something so large?

mejackreed:
Maybe just one minor suggestion, could we get a link from `README.md` to the new docs?

lib/pid_gatherer.rb
```diff
@@ -0,0 +1,170 @@
...
+      end
+      @logger.info "found #{@solr_pids.length} pids in solr (unfiltered)"
+
+      filter_invalid_druids_if_needed @solr_pids
+    end
+
+    ##
+    # @return [Array<String>]
+    def uber_apo_pids
+      @uber_apo_pids ||= ['druid:hv992ry2431']
```
drh-stanford: you can use SolrDocument.UBER_APO_ID here if available

lib/pid_gatherer.rb
```diff
@@ -0,0 +1,170 @@
...
+    ##
+    # @return [Array<String>]
+    def agreement_pids
+      @agreement_pids ||= pids_for_model_type '<info:fedora/afmodel:agreement>'
+    end
+
+    ##
+    # @return [Array<String>]
+    def hydrus_uber_apo_pids
+      @hydrus_uber_apo_pids || ['druid:zw306xn5593']
```
drh-stanford: we can put this druid in models/concerns/apo_concern

INDEXING.md
```diff
@@ -0,0 +1,139 @@
...
+  def index_logger
+    @index_logger ||= Argo::Indexer.generate_index_logger { request.uuid }
+  end
+```
+
+`DorController.reindex` calls `Argo::Indexer.reindex_pid` with the logger instantiated by `index_logger`, and asks Solr to commit the added document.  Since it uses the default behavior of letting errors bubble up, it specifically deals with `ActiveFedora::ObjectNotFoundError`, rendering a 404 response.
+```ruby
+  def reindex
+    solr_doc = Argo::Indexer.reindex_pid params[:pid], index_logger
+    Dor::SearchService.solr.commit # reindex_pid doesn't commit
```
drh-stanford: perhaps add a few words about why we commit here

INDEXING.md
```diff
@@ -0,0 +1,139 @@
...
+* Since you want to run a lot of workers, and they use a lot of memory, throw a lot of memory and CPU at a box that's doing heavy bulk reindexing.  For example, we've run 12 workers on a VM with 32 GB of RAM and 12 CPUs allocated.
+* At present, because of the way delayed_job works and is configured, each worker is essentially an instance of the Argo application.  So they all start pretty heavy, even before they start leaking memory.  We are interested in extricating the bulk reindexing machinery used by the delayed_job workers from Rails, but haven't started work on this yet.
+* If you're having performance issues, and you decide to turn profiling on, expect the profiling to make indexing much slower and much less stable.  Still, we found profiling a smaller but representative subset of the indexing work to be quite helpful in optimizing things like the number of workers we let run.  It's not necessary to try to profile a run of half a million objects to get useful data (it's probably not even practically feasible).
+
+
+### Settings overview
```
drh-stanford: this info could be moved into the settings.yml itself.

PROFILING.md
```diff
@@ -0,0 +1,66 @@
...
+```
+profiler = Argo::Profiler.new
+profiler.prof { Argo::Indexer.reindex_pid_list pid_list, should_commit }
+profiler.print_results_call_tree(out_file_id)
+```
+
+In the above example, we'd be profiling the call to `Argo::Indexer.reindex_pid_list` with params `pid_list` and `should_commit` (though an arbitrarily large block of code could be profiled).  The results will then be printed to a file named after the value of the string `out_file_id`, in the directory specified by `Settings.PROFILER.OUTPUT_DIR` (`log/profiler` in the app root, by default).  Unless `Settings.PROFILER.MEASURE_MODE` is edited from the default (which is not advised, as mentioned above), the stats collected will be the `WALL` time for methods ("the real-world time elapsed between any two moments", see https://github.com/ruby-prof/ruby-prof#measurements).  The output will be in the callgrind format, and will print stats for methods which consumed `Settings.PROFILER.CALLTREE_PRINTER_MIN_PERCENT` or more of the CPU time during the profiling session.
+
+### Settings
+
```
drh-stanford: this info could go into the settings.yml itself.

drh-stanford:
this PR looks great and the documentation is excellent. i made a few minor comments but I'm :+1:

jmartin-sul:
ok, i believe i've implemented all of the suggestions from @mejackreed and @drh-stanford.  was initially 7 small commits, rebased and squashed so there are only 3 new commits, which are each pretty small/discrete.  i temporarily panicked about having gotten myself into a bit of rebasing hell, but it a reword on a commit and a rebase from develop seems to have made everything well again.

config/application.rb
```diff
@@ -14,6 +14,12 @@ class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+    require 'indexer'
+    require 'pid_gatherer'
+    require 'bulk_reindexer'
+    require 'profiler'
```
darrenleeweber: `if Rails.env.development?` ?

darrenleeweber:
Generally :+1: but defer to others who have more experience with argo indexing.

config/application.rb
jmartin-sul: nope, profiler's useful in any environment.  we originally only included the gem in the development bundle, but it's always in, now.

PR #248, opened 2015-11-21
Breakout pid lists

mejackreed:
d14098b enables Resource Index to be used for development and test ... Useful for testing our indexing approach

8721a70 Moves pid gathering tasks to its own class. specs mainly test memoization functionality of collection methods

### DIFF
--- a/lib/pid_gatherer.rb
+++ b/lib/pid_gatherer.rb
@@ -1,0 +1,96 @@
+module Argo
+  ##
+  # Gatherer of the PIDS, used for Argo indexing
+  class PidGatherer
+    ##
+    # @return [Array<Array<String>>]
+    def pid_lists_for_full_reindex
+      [uber_apo_pids, workflow_pids, agreement_pids, hydrus_uber_apo_pids, apo_pids, hydrus_apo_pids, collection_pids, hydrus_collection_pids, set_pids, remaining_pids]
+    end
+
+    ##
+    # @return [Array<String>]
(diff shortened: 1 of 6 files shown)


lib/pid_gatherer.rb
```diff
@@ -0,0 +1,96 @@
...
+        set_pids)
+    end
+
+    private
+
+    ##
+    # Query the Dor SearchService for a specific model
+    def pids_for_model_type(model_type)
+      Dor::SearchService.risearch 'select $object from <#ri> where $object <f' \
+        "edora-model:hasModel> #{model_type}"
```
atz: This is pretty fugly, truncating in the middle of a word.  Go long-line instead, or at least line break before the `<f`, imho.

mejackreed: Agreed... fixed :clap: thanks!

jmartin-sul:
ha, just finished reading through this.  only had one minor suggestion, which was to rename `prioritize_pid_lists` to something indicating that it also does queuing (e.g. `prioritize_pid_lists_and_queue_for_reindexing` or `prioritize_and_queue_pid_lists`, or whatever).  but otherwise, definite :+1: from me, and i can make that change on `profiling_and_indexing_improvements`, if you guys find that suggestion agreeable.

mejackreed:
Uggg.. naming is sooo hard. :+1:

jmartin-sul:
indeed it is

Issue #215, opened 2015-11-18
Metadata source missing from display


LynnMcRae:
The Identification section in prod contans an indication of the descriptive metadata source, e.g.,  

https://argo.stanford.edu/catalog/druid:dm447zb3969
   DRUID:   druid:dm447zb3969
   Object Type:  item
   Content Type:   file
   MD Source:   Symphony

The "MD Source" is missing in Argo dev https://argo-dev.stanford.edu/catalog/druid:dm447zb3969

darrenleeweber:
Where is the metadata source field defined?  Where is this data stored and what transforms does it go through before it gets to the argo view?

darrenleeweber:
There is no 'Symphony' in the Fedora Object XML, see
https://sul-dor-prod.stanford.edu/fedora/objects/druid:dm447zb3969/objectXML

LynnMcRae:
Symphony is considered the metadata source where there is an <otherID type="catkey"> in identityMetadata. But I believe this has already been dereferenced into solr and the display gets a value from "metadata_source_ssi" to match the Metadata Source facet.

darrenleeweber:
@LynnMcRae - can you revise your last response where it looks like a typo in `where there is an in identityMetadata`.  Seems to be missing an FoXML field identifier or something?

LynnMcRae:
sorry, don't always remember to escape the brackets.

darrenleeweber:
Despite reasonable efforts and much trouble shooting discussion in slack, I cannot export this druid and load it into a laptop-dev environment.  There seems to be "complex metadata object hierarchies" in this druid.  Perhaps this explains why it's not working on dev - it's not exported and loaded correctly?

darrenleeweber:
For this druid, the identityMetadata doesn't seem to contain any information about the metadata source, i.e.
```xml
<foxml:datastream ID="identityMetadata" FEDORA_URI="info:fedora/druid:dm447zb3969/identityMetadata" STATE="A" CONTROL_GROUP="X" VERSIONABLE="false">
<foxml:datastreamVersion ID="identityMetadata.2" LABEL="Identity Metadata" CREATED="2013-03-14T12:45:56.080Z" MIMETYPE="application/xml" SIZE="381">
<foxml:xmlContent>
<identityMetadata>
  <objectId>druid:dm447zb3969</objectId>
  <objectType>item</objectType>
[...]

PR #255, opened 2015-11-24
Show md source

darrenleeweber:
Fix for https://github.com/sul-dlss/argo/issues/215

### DIFF
--- a/spec/features/item_view_metadata_spec.rb
+++ b/spec/features/item_view_metadata_spec.rb
@@ -1,0 +1,33 @@
+require 'spec_helper'
+
+feature 'Item view metadata' do
+  before do
+    @current_user = double(
+      :webauth_user,
+      login: 'sunetid',
+      logged_in?: true,
+      permitted_apos: [],
+      is_admin: true,
+      roles: []
+    )
+    allow_any_instance_of(ApplicationController).to receive(:current_user).
+      and_return(@current_user)
+  end
+
+  scenario 'MD Source is "DOR"' do
+    visit catalog_path 'druid:hj185vb7593'
+    within '#document-identification-section' do
+      expect(page).to have_css 'dt.blacklight-metadata_source_ssi', text: 'MD Source'
+      expect(page).to have_css 'dd.blacklight-metadata_source_ssi', text: 'DOR'
+    end
+  end
+
+  scenario 'MD Source is "Symphony"' do
+    visit catalog_path 'druid:kv840rx2720'
+    within '#document-identification-section' do
+      expect(page).to have_css 'dt.blacklight-metadata_source_ssi', text: 'MD Source'
+      expect(page).to have_css 'dd.blacklight-metadata_source_ssi', text: 'Symphony'
+    end
+  end
+
+end
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -148,7 +148,7 @@ class CatalogController < ApplicationController
         [SolrDocument::FIELD_APO_ID, SolrDocument::FIELD_COLLECTION_ID].map(&:to_s) + %w(project_tag_ssim source_id_ssim preserved_size_dbtsi)
       ],
       :full_identification => [
-        %w(id objectType_ssim content_type_ssim metadata_source_ssim),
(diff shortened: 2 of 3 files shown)


mejackreed:
Looks great! It would be great if we could also test this, so we know what md fields should be on the show page. Mind adding a test for this? I'm thinking a feature spec probably fits the bill.

darrenleeweber:
:+1: for tests.  This fix is the most direct solution, given the current data manipulations.  This code is slightly opaque - it depends on how the foxml gets indexed and then how the Solr index is translated to a display string.  It's not clear to me yet whether there are already tests on the index mapping (if not, need to start there).  I'm not sure whether to test the controller or the view or both (and whether blacklight has tooling to ease the testing of Solr -> view mappings).

mejackreed:
Yep, the MD solr fields are scatter shot through out the code. My first thought is that a feature spec captures that. Most likely the indexing code is not in Argo but in Dor Services. Something along the lines of:

```
visit catalog_path 'druid:somedruid'
within 'dl.document-metadata' do
  expect(page).to have_css 'dd', text: 'stuff'
  expect(page).to have_css 'dt', text: 'stuff'
end
```

jmartin-sul:
"This code is slightly opaque - it depends on how the foxml gets indexed and then how the Solr index is translated to a display string."

yeah, it's pretty roundabout, but this is how the vast majority of the search and display works in argo (and as i understand it, presenting a UI over solr results is what blacklight was designed to do).  so we index the foxml from fedora to solr (by calling `to_solr` on the fedora object and getting a hash that gets used as a solr doc for indexing).  then search and display from that solr info.  i believe user searches always go to solr, and while there are are a couple things which display directly from fedora, most display also comes from solr (i'm blanking on what comes directly from fedora at the moment).

it would actually be good to document this basic purpose/architecture info somewhere (maybe in the README, maybe a separate file or a github wiki).  it's obvious if you've worked with blacklight before.  but it took me forever to realize what role blacklight played and how everything fit together, because i hadn't worked with blacklight (or rails) at all previously.

"It's not clear to me yet whether there are already tests on the index mapping (if not, need to start there)."

there shouldn't be tests for the index mapping in argo, because that's the responsibility of dor-services (since it implements `to_solr`).  there are lots of tests in dor-services for these mappings, though they certainly aren't comprehensive (and i should turn some post-its on my desk into github tickets for new dor-services tests).  it does feel wrong that argo implicitly has to understand this mapping, even though it doesn't have tests for it.  on the other hand, since argo doesn't do the mapping, there's not really any machinery to be tested to that effect.  i think the right thing to do would be to have integration tests which check to see whether a piece of foxml gets turned into a Dor object with the expected fields, and whether the Dor object can then be solrized and retrieved as expected.  or something.  i'm not sure whether argo is the right home for that or not, but i'm sure that work is outside the scope of this particular bug (though it'd be worth filing an enhancement to come back to and discuss further, because it would be nice to have more insight into how the parts of this mapping can drift, which has obviously happened here).

for now, i'd go with something like jack's suggested test, since that specifically handles the presentation of that solr field, which is what this bug is specifically about.  one of the reasons i hadn't yet added similar tests was that i didn't know how offhand, and i hadn't gotten around to looking into it, since there was such a backlog of things that were in more dire shape.

jmartin-sul:
also, definite :+1: for the fix itself, and thanks for also cleaning up that .html.erb code (which is often even more of a mess than the ruby code, though i guess that's not surprising).

but yeah, simple tests for the UI part would be nice even if the multi-gem-spanning mapping semantics warrant further testing discussion later.

spec/features/item_view_metadata_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  scenario 'MD Source is "Symphony"' do
+    visit catalog_path 'druid:kv840rx2720'
+    within '#document-identification-section' do
+      expect(page).to have_css 'dt.blacklight-metadata_source_ssi', text: 'MD Source'
+      expect(page).to have_css 'dd.blacklight-metadata_source_ssi', text: 'Symphony'
+    end
+  end
+
```
darrenleeweber: Sorry but I like the scenario blocks delimited with white space.

PR #264, opened 2015-11-25
Commit within

atz:
Want to get this merged before I get too far afield.  Note the massive (and unbreaking!) simplification in routes.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -201,7 +201,7 @@ class ItemsController < ApplicationController
     data = @object.get_file(params[:file])
     response.headers['Content-Type'] = 'application/octet-stream'
     response.headers['Content-Disposition'] = 'attachment; filename=' + params[:file]
-    response.headers['Last-Modified'] = Time.now.ctime.to_s
+    response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
     self.response_body = data
   end
 
@@ -211,7 +211,7 @@ class ItemsController < ApplicationController
     when Net::HTTPSuccess then
       response.headers['Content-Type'] = 'application/octet-stream'
       response.headers['Content-Disposition'] = 'attachment; filename=' + params[:file]
-      response.headers['Last-Modified'] = Time.now.ctime.to_s
+      response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
       self.response_body = res.body
     else
       raise res.value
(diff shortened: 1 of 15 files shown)


spec/helpers/registration_helper_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
+
+describe RegistrationHelper do
+  it 'does something' do
```
mejackreed: Is this file a placeholder?

app/controllers/items_controller.rb
```diff
@@ -96,17 +94,12 @@ def remove_collection
   end
 
   def register
-    @perm_keys = ["sunetid:#{current_user.login}"]
-    if webauth && webauth.privgroup.present?
-      @perm_keys += webauth.privgroup.split(/\|/).collect { |g| "workgroup:#{g}" }
-    end
-    render :register, :layout => 'application'
+    @perm_keys = current_user.groups
```
drh-stanford: is there a test for this logic change?

drh-stanford:
i'm :+1: but the multiple commits could be squashed into a cleanup commit

spec/helpers/registration_helper_spec.rb
atz: Yup.  I complained about there not being even a placeholder for the stuff being untested.  So I added that.

app/controllers/items_controller.rb
atz: Covered by existed tests. The old code is logically identical (copy/pasted) from the code under `groups`.

atz:
I squashed several commits and also added a few more that are distinct (related to GMT requirement of HTTP spec).

app/views/items/file.html.erb
```diff
@@ -1,24 +1,22 @@
...
-
-  if workspace_files.include?(params[:file])
-    link='/items/'+params[:id]+'/file/?file='+params[:file]
+  # @NOTE: list_files performs remote SSH connection, may fail!
+  # @TODO: do list_files inside app and pass files to template for render, or ideally make it async
+  if @object.list_files.include?(params[:file])
+    link = "/items/#{params[:id]}/file/?file=#{params[:file]}"
     %>
-    Workspace: <a href=<%= link%>><%=params[:file]%><a/>
+    Workspace: <a href="<%= link %>"><%= params[:file] %><a/>
```
jmartin-sul: i think maybe the end should be `</a>`?

cbeer: Or, while you're at it, `#link_to`?

PR #266, opened 2015-11-25
Cleanup wf table

### LINKED ISSUE #193: wf grid view table should not use crazy custom styling
Currently not adhering to responsive table

mejackreed:
Closes #193 

screen shot 2015-11-25 at 2 27 09 pm


mejackreed:
Feel free to comment on this approach @LynnMcRae @blalbrit

blalbrit:
looks good to me, Jack.

LynnMcRae:
Good for me too. Sometime you'll have to tell me why the easter egg colors (which are truly fine) over "those colors".  Was it to avoid the white text on darker colors? Because BL has a set of named colors in its palette?

mejackreed:
Mostly for maintainability purposes. Bootstrap defines a set of colors that can be used throughout an application: http://getbootstrap.com/css/#less-variables-colors
http://getbootstrap.com/css/#helper-classes-backgrounds
We can utilize this approach throughout the entire application to provide users a standardized expectation of what colors mean various things. We can also update these colors globally rather than maintaining them in individual places. These same patterns show up in various UI elements:

 - Contextual Backgrounds http://getbootstrap.com/css/#helper-classes-backgrounds
 - Helper clases http://getbootstrap.com/css/#helper-classes-colors
 - Buttons http://getbootstrap.com/css/#buttons-options

And many more.

I guess the approach here is, for an application (like Argo) where we are not able to spend the design time on, its best to use the defaults where possible. This should set us up not only for now but also the future of the project. Using the defaults gives us:
 - less code to maintain
 - easier Blacklight upgrade path
 - easier understanding for new developers coming onto the project
 - a **large** community (Bootstrap) to gain experience from

app/views/report/_workflow_table.html.erb
```diff
@@ -1,26 +1,28 @@
 <h2>Workflow: <%= render_workflow_name(wf_name) %></h2>
-<table border="1" cellspacing="0" cellpadding="3">
+<table class='table table-striped table-hover workflow-grid-table'>
```
drh-stanford: how about adding `table-condensed`? in production there's lots of data in the workflow grid...

mejackreed:
@drh-stanford updated with `.table-condensed`. Thanks for that suggestion!

PR #268, opened 2015-11-25
reimplement CSV output for report downloads

### LINKED ISSUE #80: CSV Download of report view does not escape text fields correctly
Example: druid:ff017sr6451 has quote marks in the "Citation" field, but the CSV that you get from the Download report does not escape the quote mark correctly which causes a column shift in the report :( 

Here's the report.csv: 

Druid,Citation,Source Id,Admin Policy,Status,Reg. Date,Pub. Date,Files,Shelved Files,Resources,Preservation Size, 
"ff017sr6451","<em>APO for "The last spike, " a painting by Thomas Hill illustrating the last scene in the building of the overland railroad. With a history of the enterprise.</em>","Hydrus:adminPolicy-jejohns1-2013-11-07T18:21:30.202Z","Hydrus uber-APO","v2 Accessioned","2013-11-07","2014-08-07","","","","", 

From DarrenH: See "csv2" method here: https://github.com/sul-dlss/argo/blob/master/app/models/report.rb#L285

drh-stanford:
Fixes #80 

This PR uses the built-in Ruby `csv` package to generate the report downloads for Report and Discovery views, and adds tests. Note that the `ReportController` and `DiscoveryController` suffer from copy-n-paste code.

### DIFF
--- a/app/controllers/discovery_controller.rb
+++ b/app/controllers/discovery_controller.rb
@@ -41,6 +41,6 @@ class DiscoveryController < CatalogController
     response.headers['Content-Type'] = 'application/octet-stream'
     response.headers['Content-Disposition'] = 'attachment; filename=report.csv'
     response.headers['Last-Modified'] = Time.now.ctime.to_s
-    self.response_body = Discovery.new(params, fields).csv2
+    self.response_body = Discovery.new(params, fields).to_csv
   end
 end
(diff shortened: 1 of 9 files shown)


mejackreed:
Love the approach here. Wondering if it makes sense to pull out `to_csv` to a common module used in both classes. A quick glance looks like they are identical?

drh-stanford:
where? in a model concern?

mejackreed:
Yep that would work... `app/models/concerns/csv_concern.rb` ? Then anything that needed to respond to `to_csv` could use that?

drh-stanford:
ok I created `CsvConcern` that implements `to_csv` for the `Discovery` and `Report` models.

mejackreed:
:+1:

PR #276, opened 2015-12-01
Use the auto-generated rails schema.rb file. We should not auto chang…

mejackreed:
…e the quotes here from using Rubocop or any other tool except Rails migrations.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,66 +13,66 @@
 
 ActiveRecord::Schema.define(version: 20150619200031) do
 
-  create_table 'bookmarks', force: :cascade do |t|
-    t.integer  'user_id',       null: false
-    t.string   'document_id'
-    t.string   'title'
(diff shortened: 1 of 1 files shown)


atz:
Probably should go ahead and add a .rubocop.yml exclusion for the related rule(s) then?

mejackreed:
Yo dawg: https://github.com/sul-dlss/argo/blob/develop/.rubocop.yml#L3-L6

mejackreed:
I don't know if auto fix stuff respects that though.. @atz do you know?

atz:
Yeah, should be able to work it out.  It will look like:
```yml
Style/WhateverRuleItIs:
  Exclude:
    - 'db/schema.rb'
```

atz:
Actually, looks like schema.rb was already excluded from AllCops, so no further specification is needed.

mejackreed:
:cool:

PR #281, opened 2015-12-01
unpin addressable from 2.3.5 to allow current webmock

### LINKED ISSUE #275: can we unpin addressable in Argo?
@drh-stanford indicates that this was a ruby 1.x issue, so maybe we can drop this constraint now that argo's on 2.x?
https://github.com/sul-dlss/argo/blob/develop/Gemfile#L3
https://github.com/sul-dlss/dor-services/pull/124

drh-stanford:
Closes #275


atz:
I guess if tests are still passing, why not?

jmartin-sul:
:+1:

PR #282, opened 2015-12-01
fix Nil exception with @object, not @obj

drh-stanford:
Fixes this exception:

screen shot 2015-12-01 at 11 52 13 am


mejackreed:
This is great! Anyway a feature spec might be able to be added here?

drh-stanford:
i added a simple unit test for rendering which would have caught this exception

PR #283, opened 2015-12-01
use scss import rather than require

### LINKED ISSUE #269: Tag facet, can open only some Tag hierarchies sometimes
Cannot open “Registered by” and “Tag” values to next lower level

screen shot 2015-11-25 at 2 51 10 pm

Index seems to be built ok because I can open these on subsequent pages, but not reliably. Here I can delve into "Tag" but not "Registered by" after choosing an Object Type facet:

screen shot 2015-11-25 at 2 54 12 pm

while here I can dive down into Registered by after selecting Registered By as a facet search:

screen shot 2015-11-25 at 2 55 21 pm

mejackreed:
Closes #269 

Not a pretty approach, but will will need to suffice until TLC can be thrown on blacklight-hierarchy

Also not testable as far as I can tell.

### DIFF
--- a/app/assets/stylesheets/blacklight_hierarchy.scss
+++ b/app/assets/stylesheets/blacklight_hierarchy.scss
@@ -1,3 +1,5 @@
-/*
- *= require blacklight/hierarchy/hierarchy
- */
+@import 'blacklight/hierarchy/hierarchy';
+
+#facets ul.facet-hierarchy {
+  margin-left: -28px;
+}


app/assets/stylesheets/blacklight_hierarchy.scss
```diff
@@ -1,3 +1,5 @@
-/*
- *= require blacklight/hierarchy/hierarchy
- */
\ No newline at end of file
+@import 'blacklight/hierarchy/hierarchy';
+
```
mejackreed: Tried to mitigate this here, but we have to do this to override the defaults.

PR #284, opened 2015-12-01
[WIP] paranoid avoidance of memory leaks and use Argo::Indexer.reindex_object

drh-stanford:
…in controllers

@jmartin-sul here's some forced memory releases in the reindexing code branch... you might try them out on stage... they break ci though but should be usable


app/controllers/items_controller.rb
```diff
@@ -600,7 +595,8 @@ def create_obj
 
   def save_and_reindex
     @object.save
```
atz: save is going to call to_solr underneath anyway, right?  This is doubling it up?

jmartin-sul:
i'd vote we table this for now, since we're not sure it'll save significantly (if at all) on memory usage, and since we seem to have a handle (albeit a tenuous one) on indexing performance at the moment.

PR #294, opened 2015-12-03
Embargo facet duplication

### LINKED ISSUE #291: Embargo Release Date facets appears twice
Once as a date range search matching other date facets, and again as individual date values. Keep the date range instance, drop the list-of-timestamps instance.
screen shot 2015-12-02 at 2 56 22 pm

mejackreed:
Closes #291 

Also moves date facet configurations out of a helper. This stuff really should be in a concern or other model rather than a helper.

Also adds tests for all of the places this gets included.


mejackreed:
> Coverage increased (+0.4%) to 73.182%
:dancers:

PR #302, opened 2015-12-05
Modify catalog configuration

### LINKED ISSUE #290: Results display issue with item in multiple collections
This object, if displayed in a search result, shows its three collections as a comma delimited single string, but then repeats that string or each of the three occurrences, each time linked to one of the three individual collection objects.  
screen shot 2015-12-02 at 11 36 35 am

Either display convention would be fine -- vertical or wrapped into a single comma delimited value, but there needs to be separate links for each. Would prefer each link on a new line for clarity with long collection names that might have commas.  This applies to the results displays as well as the detail disply, though there is a problem getting the detail display for this same record #289

### LINKED ISSUE #297: What metadata fields should be shown in i results view, what should be shown in show view?
These are configured in a non standard way. Should they reflect the same fields?

mejackreed:
Closes #290 Closes #297

Among many things, this pr moves argo towards the Blacklight standard configuration pattern by doing the following:
 - use `config.add_index_field` as it is intended
 - use default Blacklight partials for rendering index results
 - index fields shown should reflect legacy argo index fields
 - show fields shown should reflect legacy argo show fields
 - remove artisanal byteparser in favor of Rails out of the box
 - comment status of render link fields which should be able to be removed on later versions of Blacklight


mejackreed:
Yea!! Increased coveralls +2.1% :dancers:

PR #305, opened 2015-12-07
[WIP] Update blacklight

mejackreed:
Updates Blacklight to latest 5.16.4. 

Will require a change to our puppet config now that we are using `blacklight.yml` in place of `solr.yml` @sul-devops-team 

Addresses all of the added Blacklight deprecation warnings that I could find.


mejackreed:
Already

PR #312, opened 2015-12-07
make bin/rails executable

mejackreed:
Deploys will fail because `bin/rails` is not executable. This PR makes the file executable. Verified working on dev


mejackreed:
An oversight on my part from: https://github.com/sul-dlss/argo/commit/42d4458f47690f92cb3e2070e8eb0b52b463d1c9

Issue #293, opened 2015-12-03
Workflow Step Re-Set Throws Error


blalbrit:
Failed response on workflow step re-set. User workflow was:
1) notice error message
2) enter workflow pop-out
3) try to set workflow step to waiting
4) error response returned

![screenshot 2015-12-02 16 05 20](https://cloud.githubusercontent.com/assets/796990/11548195/dd0d1760-990e-11e5-9028-deed590e694e.png)
![screenshot 2015-12-02 16 05 29](https://cloud.githubusercontent.com/assets/796990/11548200/e21aebec-990e-11e5-823e-aa81965ac22c.png)
![screenshot 2015-12-02 16 05 39](https://cloud.githubusercontent.com/assets/796990/11548205/e629dcfc-990e-11e5-875d-0f0a6348140b.png)

drh-stanford:
this might be a duplicate of #220 but the exception message is different in development

drh-stanford:
I have a PR that fixes the button itself, but on success it does not put the next workflow view into a modal dialog... like so

![screen shot 2015-12-07 at 1 31 07 pm](https://cloud.githubusercontent.com/assets/1861171/11640574/14be00e8-9ce8-11e5-919f-43d92ba8d4dd.png)

Issue #220, opened 2015-11-18
"set to STATUS" button in workflow detail view is no longer setting the repo value


drh-stanford:
This is what the view is outputing now...

```
<form action="http://localhost:3000/items/druid:rn653dy9317/workflows/accessionWF" accept-charset="UTF-8" method="post">
...
<input type="hidden" name="process" id="process" value="descriptive-metadata">
<input type="hidden" name="status" id="status" value="waiting">
<input type="hidden" name="repo" id="repo">
<button name="button" type="submit">set to waiting</button>
</form>
```

Here's where the set to button is found:

![screen shot 2015-11-18 at 11 21 09 am](https://cloud.githubusercontent.com/assets/1861171/11251527/8631a79a-8de6-11e5-842d-8e36df675eb4.png)

and here's the error message when you click "set to waiting" (but the error happens on all the buttons):

![screen shot 2015-11-18 at 11 21 53 am](https://cloud.githubusercontent.com/assets/1861171/11251545/9cd41848-8de6-11e5-91c3-e54d0f12afa2.png)

drh-stanford:
duplicate of #293

PR #314, opened 2015-12-07
fix broken change status button in workflow view and some cleanup

drh-stanford:
This PR addresses #293 and #220, but doesn't fix the issues completely. This fix is for the button action to work correctly.


drh-stanford:
I changed the routing behavior for this reset button so that it goes back to the catalog item show page rather than the modal dialog. I'm not seeing a technical way to reload only the `ajax-modal`

screen shot 2015-12-08 at 12 50 04 pm

PR #315, opened 2015-12-07
condense registration nav links to a dropdown

mejackreed:
# After

screen shot 2015-12-07 at 2 45 00 pm

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -27,14 +27,22 @@
     <% end %>
 
     <li><%= link_to "All&nbsp;Workflows".html_safe, report_workflow_grid_path %></li>
-
-    <li><%= link_to "Register&nbsp;Items".html_safe, register_items_path %></li>
-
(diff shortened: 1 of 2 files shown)


drh-stanford:
There seems to be a problem with the Register pull down on the Register Items page itself -- you cannot click it. 

In -dev you can click any link in the header from the Register Items page and it works. Perhaps this is a problem with the register items feature itself, but here's the behavior I'm seeing...

The header "shrinks" in font size and the Register pulldown is not clickable:

screen shot 2015-12-08 at 12 01 43 pm

There's a single error in the JavaScript console:

screen shot 2015-12-08 at 12 09 51 pm

And going to the URL itself yield the following exception (the 500 error also shows up in -dev but the header links work):

screen shot 2015-12-08 at 12 11 58 pm

mejackreed:
Thanks for reporting that @drh-stanford ... Seems to be an issue on that page where no JavaScript can run. I've created a ticket for that:
https://github.com/sul-dlss/argo/issues/320

PR #317, opened 2015-12-08
Wfs rails discoball

### LINKED ISSUE #191: Make WFS parts of argo developable and testable
Dev's should be able to develop and test WFS without connecting to external services.

mejackreed:
Closes #191 

Interested in hearing feedback/questions on this approach.


config/initializers/webmock.rb
```diff
@@ -1,23 +0,0 @@
...
-
-  # mock_workflow_requests
-  escaped_url = Settings.WORKFLOW_URL.gsub('/', '\/')
-  stub_request(:get, /#{escaped_url}workflow_archive.*/).
-    to_return(body: '<objects count="1"/>')
-  stub_request(:get, /#{escaped_url}dor.*/).
-    to_return(body: '<workflows/>')
-
-  # mock_status_requests
-  stub_request(:get, Settings.STATUS_INDEXER_URL).to_return(status: 404)
```
drh-stanford: This guy is still needed to mock the specs

drh-stanford:
My concern with the format of this PR is that the `wfs_rails` gem hasn't undergone a code review and it's not clear how to comment on that code within this PR. It also doesn't seem necessary to have `wfs_rails` as its own git repo, rather than a subdirectory of Argo. That would at least address the PR coverage to have it all within this PR.

cbeer:
One benefit of keeping `wfs_rails` in a separate project is other applications with this same issue can consume it (thinking of Hydrus, among others).

drh-stanford:
I've created some minor issues in the `wfs_rails` repo. otherwise, looks great.

atz:
:+1: Hydrus would definitely want this.

config/initializers/webmock.rb
mejackreed: Addressed in person. Prefer leaving webmock out over mocking index status.

mejackreed:
I think this is good to go now. Let me know if you have any :question: about things updated here. f460c4c
 added because I'm pretty sure our old auth pattern was masking these specs failing.

config/initializers/webmock.rb
drh-stanford: Can we change the configuration to use localhost rather than example.edu? Ideally all of the example.edu "defaults" should be localhost imho.

drh-stanford: resolved to leave as-is

README.md
```diff
@@ -79,6 +79,9 @@ bin/delayed_job start  # Necessary only for spreadsheet bulk upload
 rails server
 ```
 
+## Important note about Workflow Services integration
+The Argo code repository includes the [wfs_rails](https://github.com/sul-dlss/wfs_rails) gem, a Rails Engine, which mocks and stubs several of the Workflow Services routes within the Argo application. This engine is booted using [Capybara::Discoball](https://github.com/thoughtbot/capybara_discoball) and uses the local database as storage for Workflow Service data. This integration can be turned off by not setting Settings.WORKFLOW_URL to an instance of Capybara::Discoball and by configuring Settings.WFS_RAILS.ENBALE to `false`.
```
drh-stanford: typo of the variable name... should be `Settings.WFS_RAILS.ENABLE`

drh-stanford:
:+1: from me

README.md
mejackreed: Good catch! Updated. I can't spell enable to save my life.

spec/integration/report_spec.rb
```diff
@@ -4,7 +4,7 @@
   before :each do
     webauth = double('WebAuth', :login => 'sunetid', :attributes => {'DISPLAYNAME' => 'Example User'}, :privgroup => ADMIN_GROUPS.first)
     @current_user = User.find_or_create_by_webauth(webauth)
-    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(@current_user)
+    allow_any_instance_of(CatalogController).to receive(:current_user).and_return(@current_user)
```
atz: Shouldn't we always just do the stub in ApplicationController?  Why pull it up to the subclass?

mejackreed: Good point. Let me look into this a bit more.

mejackreed: Strange strange behavior happening here. Where `ApplicationController` will not receive the stubbed method, yet when specifying a descendant of `ApplicationController`, `ReportController` or `CatalogController` that method is received. Not sure why this is and looking into it for a bit turned up no results. Any ideas?

drh-stanford: did you try it with `ActionController::Base`? or with the fully-specified name for `ApplicationController`?

mejackreed: Just tried with the following:
 - `ActionController::Base` - :x: message not received
 - `Argo::Application::ApplicationController` :x: message not received

drh-stanford: yeah very strange -- maybe try `allow(controller)` if that's available in the integration specs -- that might be more generic.

mejackreed: Yeah.. from rspec-mocks documentation:
> - Using this feature is often a design smell. It may be that your test is trying to do too much or that the object under test is too complex.
> - It is the most complicated feature of rspec-mocks, and has historically received the most bug reports. (None of the core team actively use it, which doesn't help.)
http://www.relishapp.com/rspec/rspec-mocks/v/3-4/docs/working-with-legacy-code/any-instance

drh-stanford: how about leave as-is and put in a comment re: superclasses?

mejackreed: @drh-stanford Updated and commented. I think we may be good to go now.

PR #319, opened 2015-12-08
prefer using window.location over document.location

mejackreed:
This is what we do in GeoBlacklight, seeing if it makes a difference here in Argo.


mejackreed:
Was hoping this would work :cry:

Issue #289, opened 2015-12-02
Unable to display druid:bb004bn8654


LynnMcRae:
https://argo-stage.stanford.edu/catalog/druid:bb004bn8654

gets "Incomplete response received from application"

for reasons unknown. If caused by bad data, we want the display to be robust enough to recover/ignore, showing druids in place of missing references for instance.

LynnMcRae:
More examples ... they all come from a facet search on Collection=Hydrus or an attempt to show that collection record.  
https://argo-stage.stanford.edu/catalog/druid:bb020ng8351
https://argo-stage.stanford.edu/catalog/druid:gb045hv3517

darrenleeweber:
Displays OK on prod at https://argo.stanford.edu/catalog/druid:bb004bn8654

darrenleeweber:
I wonder if this has something to do with the REVS image.  Fedora on my laptop complains about `Object not found in low-level storage: druid:bb004bn8654` when the request hits
http://localhost:8983/fedora/objects/druid:bb004bn8654?format=xml
This is after adding the FOXML to the `fedora_conf/data/bb004bn8654.xml` and running the rake loaders for test and development.  The fedora logs contain a more details trace but it doesn't really help to identify whether it's a failure to find the FOXML or the image, i.e.
```
WARN 2015-12-09 10:57:42.046 [qtp152005629-14 - /fedora/objects/druid:bb004bn8654?format=xml] (BaseRestResource) Resource not found: Object not found in low-level storage: druid:bb004bn8654; unable to fulfill REST API request
org.fcrepo.server.errors.ObjectNotInLowlevelStorageException: Object not found in low-level storage: druid:bb004bn8654
	at org.fcrepo.server.storage.lowlevel.akubra.AkubraLowlevelStorage.retrieve(AkubraLowlevelStorage.java:423) ~[fcrepo-server-3.7.1.jar:na]
	at org.fcrepo.server.storage.lowlevel.akubra.AkubraLowlevelStorage.retrieveObject(AkubraLowlevelStorage.java:180) ~[fcrepo-server-3.7.1.jar:na]
	at org.fcrepo.server.storage.lowlevel.akubra.AkubraLowlevelStorageModule.retrieveObject(AkubraLowlevelStorageModule.java:83) ~[fcrepo-server-3.7.1.jar:na]
[...]

PR #336, opened 2015-12-10
Issue289

darrenleeweber:
First attempt to address one error in the display of a Revs object, to address #289.  Unable to get the argo-stage object data into a fixture to completely test this fix.  The additional spec does test just this helper method.  Although travis CI, hound, coveralls etc. can evaluate this PR to some extent, the PR branch might have to be pushed to argo-stage to better evaluate how it behaves with the data noted in #289.

### DIFF
--- a/app/helpers/value_helper.rb
+++ b/app/helpers/value_helper.rb
@@ -69,9 +69,14 @@ module ValueHelper
     end.join('<br>').html_safe
   end
 
+  # Discussed removing this method because the field should not be modified.
   def value_for_originInfo_date_created_tesim(args)
-    val = Time.parse(args[:document][args[:field]].first)
+    field_value = args[:document][args[:field]].first
+    # Try to parse the string as a time value
+    val = Time.parse(field_value)
     val.localtime.strftime '%Y.%m.%d %H:%M%p'
+  rescue
+    field_value
   end
 
   def value_for_identifier_tesim(args)
--- a/app/helpers/value_helper.rb
+++ b/app/helpers/value_helper.rb
@@ -69,14 +69,14 @@ module ValueHelper
     end.join('<br>').html_safe
   end
 
-  # Discussed removing this method because the field should not be modified.
-  def value_for_date(args)
+  # @return [String]
+  def value_for_date_as_localtime(args)
     date_value = args[:document][args[:field]].first
     # Try to return the date in local time zone;
     # assume it can be parsed according to app time zone.
     return '' if date_value.nil?
     val = Time.zone.parse(date_value)
-    return date_value if val.nil?
+    return date_value.to_s if val.nil?
     val.localtime.strftime '%Y.%m.%d %H:%M%p'
   rescue ArgumentError
     date_value.to_s
--- a/app/helpers/value_helper.rb
+++ b/app/helpers/value_helper.rb
@@ -70,13 +70,16 @@ module ValueHelper
   end
 
   # Discussed removing this method because the field should not be modified.
-  def value_for_originInfo_date_created_tesim(args)
-    field_value = args[:document][args[:field]].first
-    # Try to parse the string as a time value
-    val = Time.parse(field_value)
+  def value_for_date(args)
+    date_value = args[:document][args[:field]].first
+    # Try to return the date in local time zone;
+    # assume it can be parsed according to app time zone.
+    return '' if date_value.nil?
+    val = Time.zone.parse(date_value)
+    return date_value if val.nil?
     val.localtime.strftime '%Y.%m.%d %H:%M%p'
-  rescue
-    field_value
+  rescue ArgumentError
+    date_value.to_s
   end
 
   def value_for_identifier_tesim(args)
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -261,9 +261,12 @@ namespace :argo do
   end
 
   def load_order_files(fedora_files)
+    data_path = File.expand_path('../../../fedora_conf/data/', __FILE__)
     file_list = []
     fedora_files.each do |file|
-      file_list.push(File.join(File.expand_path('../../../fedora_conf/data/', __FILE__), file.strip))
+      file_name = file.strip
+      next if file_name.empty?
+      file_list.push(File.join(data_path, file_name))
     end
     file_list
   end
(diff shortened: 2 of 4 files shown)


spec/helpers/value_helper_spec.rb
```diff
@@ -31,4 +31,25 @@
...
+      field = 'originInfo_date_created_tesim'
+      value = ['1966', 'anything_after_first_element_ignored']
+      document = SolrDocument.new({ field => value })
+      args = { document: document, field: field }
+      expect(helper.value_for_originInfo_date_created_tesim(args))
+        .to eq(value.first)
+    end
+    it 'does normalize a time stamp string' do
+      now_utc = Time.now.utc.to_s
+      now_loc = Time.parse(now_utc).localtime.strftime '%Y.%m.%d %H:%M%p'
```
darrenleeweber: How are you supposed to know the offset when this code is run and could be run on any machine anywhere?  The whole point of not specifying the offset is to use the local machine time.  Sorry, don't agree with houndci on this one.

app/helpers/value_helper.rb
```diff
@@ -69,9 +69,14 @@ def links_to_collections(args)
     end.join('<br>').html_safe
   end
 
+  # Discussed removing this method because the field should not be modified.
   def value_for_originInfo_date_created_tesim(args)
-    val = Time.parse(args[:document][args[:field]].first)
+    field_value = args[:document][args[:field]].first
+    # Try to parse the string as a time value
+    val = Time.parse(field_value)
```
darrenleeweber: The very next line of this code does use the `localtime` zone of the machine running the code.  Good enough?  (Guess houndci doesn't look ahead?)

spec/helpers/value_helper_spec.rb
cbeer: I believe `Time#in_time_zone` is the environment-aware local time zone.

app/helpers/value_helper.rb
cbeer: Is it? I think the behavior rubocop is trying to discourage here is using `Time.parse` with a poorly-specified timezone reference, to catch differences like:

```ruby
# Using the system time zone (PT)
irb(main):003:0> Time.parse('2015-01-01')
=> 2015-01-01 00:00:00 -0800

# Using the application time zone (UTC)
irb(main):004:0> Time.zone.parse('2015-01-01')
=> Thu, 01 Jan 2015 00:00:00 UTC +00:00
```

See also this pretty nice guide to time handling in rails:  http://danilenko.org/2012/7/6/rails_timezones/

lib/tasks/argo.rake
```diff
@@ -261,9 +261,12 @@ namespace :argo do
   end
 
   def load_order_files(fedora_files)
+    data_path = File.expand_path('../../../fedora_conf/data/', __FILE__)
```
mejackreed: Seems like this might be part of another PR or a different commit. One looking at this commit down the road might wonder why this task was changed as part of this bug fix.

app/helpers/value_helper.rb
```diff
@@ -69,9 +69,14 @@ def links_to_collections(args)
...
     val.localtime.strftime '%Y.%m.%d %H:%M%p'
+  rescue
```
mejackreed: If we are going to use the `rescue` approach, it might be good to rescue from the specific error that is raising rather than `StandardError`.

app/helpers/value_helper.rb
```diff
@@ -69,9 +69,14 @@ def links_to_collections(args)
     end.join('<br>').html_safe
   end
 
+  # Discussed removing this method because the field should not be modified.
   def value_for_originInfo_date_created_tesim(args)
```
mejackreed: Since this method isn't doing anything specifically related to `orginInfo_date_created_tesim` it would be nice to see it renamed and added explicitly using `helper_method` in `catalog_controller.rb` https://github.com/sul-dlss/argo/blob/develop/app/controllers/catalog_controller.rb#L49 https://github.com/projectblacklight/blacklight/wiki/Configuration---Results-View#helper-method

spec/helpers/value_helper_spec.rb
darrenleeweber: Can't find that one, i.e.
```
$ ruby -v
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-linux]
$ pry
[1] pry(main)> Time.in_time_zone
NoMethodError: undefined method `in_time_zone' for Time:Class
[2] pry(main)> Time.now.in_time_zone
NoMethodError: undefined method `in_time_zone' for 2015-12-10 08:40:18 -0800:Time
```

mejackreed: `#in_time_zone` is an ActiveSupport thing, I believe. Try within the context of the rails console.

```sh
$ rails c
[1] pry(main)> Time.now.in_time_zone
=> Thu, 10 Dec 2015 16:46:40 UTC +00:00
```

lib/tasks/argo.rake
mejackreed: See now this is in another commit... :ok_hand:

mejackreed: Also was this causing an error?

darrenleeweber: Yes, understood.  I didn't have the energy to be perfect after a less than perfect day at the keyboard ;-)  This file was touched because I wanted a blank line in the `load_order` file to separate out the fixtures included for working on #289, which have been subsequently removed for several reasons, primarily because they are not representative of the argo-stage data and secondarily because it may not be good to add a new fixture data file for every bug and lastly because the spec code may suffice to test the method modifications in this PR.  But thanks for encouraging perfection and I'm open to suggestions on how to work around this if it's really important.

app/helpers/value_helper.rb
darrenleeweber: Good point.  Without knowing what the data is, it's not obvious what the exceptions may be.  Will explore this in revised specs and add specific exception classes.

lib/tasks/argo.rake
mejackreed: Nope just curious about the intent here as I didn't understand. :ok:

app/helpers/value_helper.rb
darrenleeweber: It looks like argo does not explicitly set the app time zone, so `Time.zone` is the default UTC.  Can we assume that this data should be parsed as a UTC value?

cbeer: I think it's safe to assume the application should be set to the correct time zone (and whether it is or not  is a separate question that, indeed, should be answered..)

app/helpers/value_helper.rb
darrenleeweber: Thanks for the tip and refs.  This PR should use this!  Will revise it.

darrenleeweber:
After further discussion, this solution will be replaced with an alternative that removes the helper method entirely.

PR #347, opened 2015-12-11
remove metaprogramming from ApplicationController initialization. Hel…

mejackreed:
…per modules can be explicitly included in controllers if needed (probably a smell though)

This was initially added in the infamous Padrino conversion commit 50cf101286104310f970041b867ddb38aee69500

### DIFF
--- a/app/controllers/registration_controller.rb
+++ b/app/controllers/registration_controller.rb
@@ -1,4 +1,5 @@
 class RegistrationController < ApplicationController
+  include RegistrationHelper
 
   def form
     redirect_to register_items_url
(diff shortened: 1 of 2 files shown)


mejackreed:
I think the coverage drop here is a result of unexpected craziness.

app/controllers/registration_controller.rb
```diff
@@ -9,7 +10,7 @@ def tracksheet
     name     = params[:name] || 'tracksheet'
     sequence = params[:sequence] || 1
     response['content-disposition'] = "attachment; filename=#{name}-#{sequence}.pdf"
-    pdf = help.generate_tracking_pdf(druids)
+    pdf = generate_tracking_pdf(druids)
```
cbeer: Does this method really belong in a helper, or maybe it should be its own class/model?

mejackreed: Oh.. totally agree here. Just /shrug.... so much of that

mejackreed: And... the tests aren't quite there... https://github.com/sul-dlss/argo/blob/develop/spec/helpers/registration_helper_spec.rb

mejackreed: Created #348

PR #350, opened 2015-12-11
Use alternative function to wait for applied params

darrenleeweber:
Attempting to work around intermittent travis failures with a wait function that delays a bit longer each time it waits.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,11 +13,8 @@ class ApplicationController < ActionController::Base
 
   attr_reader :help
 
-  if Rails.env.production?
-    require 'squash/rails'
-    include Squash::Ruby::ControllerMethods
-    enable_squash_client
-  end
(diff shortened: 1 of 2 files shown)


darrenleeweber:
The wait time in each loop of this function are `[*1..45].map {|n| 2 * Math.log(n)}` -- if the wait loop runs to completion it should fail the spec.

darrenleeweber:
When these specs fail, it doesn't seem to matter how long it waits.  It seems the JS gets jammed.  Maybe relevant:
  - http://stackoverflow.com/questions/16657744/phantomjs-hanging-when-called-from-cli-or-web

mejackreed:
I've had success with this approach https://github.com/sul-dlss/argo/pull/353 which seems to be consistent with other community errors. Data on tries/success recorded here: https://github.com/sul-dlss/argo/pull/352#issue-121720520

darrenleeweber:
#353 is a better solution.

Issue #304, opened 2015-12-05
ci build fails on develop sporadically


drh-stanford:
The `develop` build is failing sporadically on Travis. See https://travis-ci.org/sul-dlss/argo/builds/95020769 for a failure example and https://travis-ci.org/sul-dlss/argo/builds/95022296 for a success example

The build errors are these:

```
  1) Date range form with no after date
     Failure/Error: find('#appliedParams')
     Capybara::ElementNotFound:
       Unable to find css "#appliedParams"
     # ./spec/features/date_range_form_spec.rb:45:in `block (2 levels) in <top (required)>'
  2) Date range form with no before date
     Failure/Error: find('#appliedParams')
     Capybara::ElementNotFound:
       Unable to find css "#appliedParams"
     # ./spec/features/date_range_form_spec.rb:60:in `block (2 levels) in <top (required)>'
```

I tried increasing the Capybara timeout but that hasn't worked on my laptop. It seems hung on the visit of the page.

drh-stanford:
I turned on debugging for poltergeist and I get the following for the first failure:

```json
{"name":"visit","args":["http://127.0.0.1:49331/"]}
{"response":{"status":"success"}}
{"name":"find","args":["css","[data-target=\"#facet-object_modified_date\"]"]}
{"response":{"page_id":2,"ids":[0]}}
{"name":"visible","args":[2,0]}
{"response":true}
{"name":"click","args":[2,0]}
{"response":{"position":{"x":128,"y":749.5}}}
{"name":"find","args":["css","#facet-object_modified_date"]}
{"response":{"page_id":2,"ids":[1]}}
{"name":"visible","args":[2,1]}
{"response":true}
[...]

PR #351, opened 2015-12-11
use travis_retry on the build so it retries a few times on failure

drh-stanford:
This PR attempts to address #304 by using `travis_retry`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ jdk:
 email: false
 before_script:
   - if [[ -r /etc/hosts ]]; then cat /etc/hosts ; fi
-script: bundle exec rake ci
+script: travis_retry bundle exec rake ci
 sudo: false
 cache: bundler


mejackreed:
This gem might also be worth checking out. Will only retry the failing specs:

https://github.com/NoRedInk/rspec-retry

mejackreed:
I've had success with this approach https://github.com/sul-dlss/argo/pull/353 which seems to be consistent with other community errors. Data on tries/success recorded here: https://github.com/sul-dlss/argo/pull/352#issue-121720520

darrenleeweber:
Although #353 looks like a good solution, these additional options in this PR and the `rspec-retry` sound good too.  Maybe 3 retries is too much?  I'd rather have one retry to avoid overloading travis when the specs are actually failing.

drh-stanford:
That `rspec-retry` looks good as it has fine-granularity down to the specific test that needs retries. Closing this for now.

PR #353, opened 2015-12-11
trigger click rather than click_button

mejackreed:
See data about tries/success recorded here: https://github.com/sul-dlss/argo/pull/352#issue-121720520

This approach is consistent with other's experiences:
 - https://github.com/teampoltergeist/poltergeist/issues/530
 - http://www.soulcutter.com/blog/2013/click-events-in-poltergeist-vs-capybara-webkit/
 - http://stackoverflow.com/questions/20071160/capybara-poltergeist-click-not-working


darrenleeweber:
:+1:

PR #357, opened 2015-12-11
fix JavaScript errors in Register Items modal dialog

### LINKED ISSUE #320: Register Items page has broken javascript
Reported originally by @drh-stanford here: https://github.com/sul-dlss/argo/pull/315#issuecomment-163004597

JavaScript functionality in the items registration page is broken, thus causing additional JavaScript to run.

drh-stanford:
This PR fixes #320. The JavaScript errors were from AJAX queries that were throwing 500 errors, and from a request timeout to an old URL. I removed that lyberapps-prod URL and deleted the MDSource pulldown (see comments). I also made the `collection_list` route more fault-tolerant when collections referenced in the APO do not exist in Solr and/or DOR. It also now truncates long collection titles. I also made the Workflow pulldown have unique values. Finally, I added spec tests for collection_list and workflow_list routes, neither of which existed prior.

### DIFF
--- a/app/controllers/registration_controller.rb
+++ b/app/controllers/registration_controller.rb
@@ -14,16 +14,6 @@ class RegistrationController < ApplicationController
     render :text => pdf.render, :content_type => :pdf
   end
 
-  def form_list
-    docs = Dor::SearchService.query(%(id:"#{params[:apo_id]}")).docs
-    md_format = docs.collect { |doc| doc['metadata_format_ssim'] }.flatten.first.to_s
-    forms = JSON.parse(RestClient.get('http://lyberapps-prod.stanford.edu/forms.json'))
-    result = forms[md_format.downcase].to_a.sort { |a, b| a[1].casecmp(b[1]) }
-    respond_to do |format|
-      format.any(:json, :xml) { render request.format.to_sym => result }
-    end
-  end
-
   def workflow_list
     docs = Dor::SearchService.query(%(id:"#{params[:apo_id]}")).docs
     result = docs.collect { |doc| doc['registration_workflow_id_ssim'] }.compact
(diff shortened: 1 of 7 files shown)


app/controllers/registration_controller.rb
```diff
@@ -34,24 +24,37 @@ def workflow_list
...
-      solr_doc = Blacklight.solr.find({:q => "id:\"#{col['id']}\"", :rows => 1, :fl => 'id,tag_ssim,dc_title_tesim'}).docs
-      if solr_doc.first['dc_title_tesim'] && solr_doc.first['dc_title_tesim'].first
-        res[col['id']] = "#{solr_doc.first['dc_title_tesim'].first}(#{col['id']})"
+    truncate_limit = (params[:truncate] || 60).to_i
+    collections = { '' => 'None' }
+    apo_object = Dor.find(params[:apo_id])
+    apo_object.administrativeMetadata.ng_xml.search('//registration/collection/@id').each do |node|
+      col_id = node.to_s
+      col_druid = col_id.gsub(/^druid:/, '')
+      col_title_field = :dc_title_tesim # TODO: why is this :dc_title_tesim and not :dc_title_ssi?
```
jmartin-sul: this actually probably should be `dc_title_ssi`.  both should exist.  `_tesim`'s there for stemmed search, `_ssi`'s there for faceting and exact search and display.  i think i tended to default to `_tesim` when in doubt when i was converting fields early on, so it's possible that i just picked the wrong variant for that or some other reason (without looking back through commit history, i'm assuming this was one of the many field names i updated earlier this year).

drh-stanford:
Ok, I changed it to the standard title field name `SolrDocument::FIELD_TITLE`

PR #376, opened 2015-12-15
avoid commitWithin when updating index for items controller

drh-stanford:
This PR makes explicit whether to execute a synchronous Solr update (with a `.commit`) or an asynchronous Solr update (without a `.commit` and using the default `commitWithin` value). The default Solr behavior for commiting adds is dependant on its configuration. See https://wiki.apache.org/solr/CommitWithin for details.

In this PR, the default behavior for the `reindex` route is a synchronous index update. This mimics the previous very short `commitWithin` value (1 second) for reindexing. I've added an `asynchronous=true` flag that you can pass `reindex` to skip the commit. Same with the Item routes, it uses synchronous updates since it was using a very short `commitWithin` value.

I've also changed the status value for objects not found to `404`, and added synchronous vs asynchronous tests.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -593,7 +593,8 @@ class ItemsController < ApplicationController
     end
     @object.create_workflow(params[:wf])
 
-    # sync up the workflows datastream with workflow service and force a reindex before redirection
+    # We need to sync up the workflows datastream with workflow service (using #find)
+    # and then force a committed Solr update before redirection.
     reindex Dor::Item.find(params[:id])
     Dor::SearchService.solr.commit
 
@@ -607,8 +608,7 @@ class ItemsController < ApplicationController
   private
 
   def reindex(item)
-    doc = item.to_solr
-    Dor::SearchService.solr.add(doc, :add_attributes => {:commitWithin => 1000})
+    Dor::SearchService.solr.add item.to_solr
   end
 
   # Filters


atz:
I think it might be better to just stop using commitWithin here since we can't really vouch for downstream logic tolerating even a 1 second window of asynchronicity, unfortunately.

drh-stanford:
@atz I've updated the PR (see first comment) to make the updates to the Solr index explicit. Anytime a `commitWithin` of 1 second was make, I assumed that to be synchronous (it's intent at least).

The outstanding question I have is whether to make the reindex route an asynchronous update. The Solr core's configuration determines the default commit behavior and in looking at our Solr configuration it doesn't appear to be using Auto Commit. That should mean that it does a commit in the default manner which I believe is to commit after n seconds or the add buffer is full.

Anyway, the current reindex route behavior is commit after every add.

Let me know what you think.

app/controllers/dor_controller.rb
```diff
@@ -47,24 +47,36 @@ def label
...
   end
 
+  ##
+  # reindex will make a synchronous index update to the Solr core
+  #
+  # @option params [String] `:pid` the druid for the object to reindex
+  # @option params [String] `:asynchronous` if present does an asynchronous transaction with solr index (i.e., no commit)
   def reindex
-    obj = Dor.load_instance params[:pid]
+    obj = Dor.load_instance params[:pid] # could be any type of Dor object
```
jmartin-sul: FWIW:  when i was trying to wrap up https://github.com/sul-dlss/argo/tree/profiling_and_indexing_improvements last night, i ran across my note to self about consolidating the dor_controller.rb code and the indexer.rb code.  and i think i got that working, though i need to fix and add some unit tests.  i'll push my WIP, since the commits need more grooming anyway, and you can take a look and tell me what you think... i assumed synchronous commit in my code, but it'd be easy enough to make that optional...

jmartin-sul: here's the part in the other branch that seems relevant here:
https://github.com/sul-dlss/argo/pull/246/files#diff-881a03fbeec456e041aa84741b3b7dd9R45
https://github.com/sul-dlss/argo/pull/246/files#diff-881a03fbeec456e041aa84741b3b7dd9R7
https://github.com/sul-dlss/argo/pull/246/files#diff-44392392ca7d32d0aa276eac0fa5e7a8R32
https://github.com/sul-dlss/argo/pull/246/files#diff-44392392ca7d32d0aa276eac0fa5e7a8R7

atz: Looks like an improvement. :+1:

drh-stanford: Ok, this PR should wait until your PR is merged.

jmartin-sul:
profiling and indexing PR has been merged, so this branch can be rebased off develop.

drh-stanford:
This PR has been rebased which removed much of the original PR content. What's left is in the `save_and_reindex` logic for items controller. I've changed it such that we omit the `commitWithin` parameter altogether.

jmartin-sul:
:+1:

i don't understand why test coverage would go down, since this appears to be the same number of operations, with one line of code fewer.  regardless, the drop seems negligible.

PR #377, opened 2015-12-15
validate whether the XML can be parsed before form submission

### LINKED ISSUE #78: Datastream edit - system error on bad XML input
A mistake when editing a datastream's XML in the lightbox will lead you to a system error screen -- RuntimeError in ItemsController#datastream_update. Better if you could tell the user they had invalid XML without leaving the edit context, so they don't lose their work.

drh-stanford:
This PR fixes #78. It uses the jQuery validation plugin and jQuery `parseXML` to verify that the XML in the data stream edit modal dialog is parseable before form submission.

### DIFF
--- a/app/assets/javascripts/jquery.persistentModal.js
+++ b/app/assets/javascripts/jquery.persistentModal.js
@@ -49,6 +49,13 @@
 
         // show the current modal
         modalForTarget().modal('show');
+
+        // Fire an event when the modal is loaded
+        var e = $.Event('loaded.persistent-modal');
+        $('body').trigger(e);
+        if (e.isDefaultPrevented()) {
+          return;
+        }
       }
 
       function receiveAjax(data) {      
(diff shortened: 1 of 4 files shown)


mejackreed:
:+1:  for the fix. Though is it possible we could move this out of the partial into a javascript file? Happy to help with that if needed.

drh-stanford:
Can you point me to an example? I was working off of https://github.com/sul-dlss/argo/blob/develop/app/views/apo/register.html.erb

mejackreed:
Yeah.. not great examples in argo.. Let me conjure something up.

drh-stanford:
the form is running validation on every keystroke in the text area. need to debug further.

drh-stanford:
ok the form validation is running by default, and it's triggered on changes to the textarea. The message is below the textarea. I've also changed the text area font to your default monospace, rather than a specific monospace font (I did this on the view datastream XML too).

screen shot 2015-12-16 at 12 50 31 pm

PR #382, opened 2015-12-16
Fixes Register Items exception for missing method object_location

### LINKED ISSUE #320: Register Items page has broken javascript

drh-stanford:
This PR fixes the Register Items exceptions, does some cleanup, and adds a development fixture for the `digitizationWF`. The cleanup is that we were only implementing the create (POST) to `/dor/objects` that's used by the Register Items UI, so I removed the unimplemented routes. I also moved `object_location` from ApplicationHelper to a private method inside the controller. There was also no spec file corresponding to the objects controller, so I created a stub spec.

Fixes #320

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -162,11 +162,11 @@ Argo::Application.routes.draw do
     match 'reindex/:pid',   :action => :reindex, :as => 'reindex', :via => [:get, :post]
     match 'delete_from_index/:pid', :action => :delete_from_index, :via => [:get, :post]
     get 'index_exceptions'
-    resources :objects
+    resources :objects, :only => :create # we only implement create for object registration
   end
 
   namespace :legacy do
-    resources :objects
+    resources :objects # TODO: do we know what this does?
   end
 
   get 'index_queue/depth', to: 'index_queue#depth'
--- a/spec/controllers/dor/objects_controller_spec.rb
+++ b/spec/controllers/dor/objects_controller_spec.rb
@@ -1,0 +1,5 @@
+describe Dor::ObjectsController, :type => :controller do
+  it 'tests nothing' do
+    skip
+  end
+end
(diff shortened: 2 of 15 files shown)


config/routes.rb
```diff
@@ -162,11 +162,14 @@
     match 'reindex/:pid',   :action => :reindex, :as => 'reindex', :via => [:get, :post]
     match 'delete_from_index/:pid', :action => :delete_from_index, :via => [:get, :post]
     get 'index_exceptions'
-    resources :objects
+    namespace :objects do
```
mejackreed: If we are implementing create only, maybe:
`resources :objects, only: :create`
?

That way it keeps the resourceful structure without just tacking it on a namespace?

spec/controllers/dor/objects_controller_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
```
mejackreed: Is this an autogen file? If we aren't testing anything do we need this?

drh-stanford:
I tested this on -dev and it worked as expected. It's a bit difficult to test but I did add the `digitizationWF` fixtures so I was able to test it on my laptop.

drh-stanford:
The coverage went down but only on files I didn't touch?

mejackreed:
Strange on the coverage. Now that I look at this again, I'm wondering if a quick controller spec testing `create` would have caught that helper include regression. Is it possible to add in?

drh-stanford:
this should be ready to go...

mejackreed:
:+1: :shipit:

PR #388, opened 2015-12-17
[WIP] cleans up the robot index. Removing svg images in favor of classes

mejackreed:
This page still needs some work. Robot queries are inefficient and we should probably use active record groupings rather than the thing we are doing in the view. Also removes routes that I can't verify are being used anywhere.

screen shot 2015-12-17 at 11 39 45 am


drh-stanford:
@LynnMcRae wants to know what's the entry point to this page? Is it linked in anywhere in the UI?

mejackreed:
No idea. I would prefer to rip out all of this if possible. No idea where the entry point is. It feels like a 1/4 complete feature.

mejackreed:
setting this to WIP until we can get some clarification on this from @LynnMcRae . I think most likely we could probably just rip it all out.

mejackreed:
Close in favor of #390 killing it.

PR #393, opened 2015-12-17
strip HTML from error messages in bulk update log

### LINKED ISSUE #335: The error messages during bulk update really stand out
Perhaps a little more than desired ...

screen shot 2015-12-09 at 4 12 00 pm

drh-stanford:
This PR fixes #335 by stripping HTML from the error messages and updating the styling to bootstrap.

screen shot 2015-12-17 at 2 50 20 pm
screen shot 2015-12-17 at 2 49 52 pm

### DIFF
--- a/app/assets/javascripts/bulk.js
+++ b/app/assets/javascripts/bulk.js
@@ -85,7 +85,7 @@ function fetch_druids(fun) {
 	if(pids_txt.length > 5) {
 		var druids = extract_pids_list(pids_txt);
 
-		log.innerHTML = "Using " + druids.length + " user supplied druids.\n<br>";
+		log.innerHTML = "Using " + druids.length + " user supplied druids.\n<br/>";
 		job_count = [];
 		for(i=druids.length; i>0; i--) {
 			job_count.push(i);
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/bulk.js
```diff
@@ -279,7 +274,7 @@ function upd_values_for_druids(upd_req_url, upd_textarea_id, row_processing_fn,
 		
 		//skip bad rows
 		if(is_invalid_row_fn(upd_info)) {
-			log.innerHTML = "<span class=\"error\"> "+job_count.pop()+" "+object_link+" : "+invalid_row_err_msg+" '"+upd_info['upd_data']+"'</span><br>\n"+log.innerHTML;
+			log.innerHTML = "<span class=\"text-danger\"> "+job_count.pop()+" "+object_link+" : "+invalid_row_err_msg+" '"+upd_info['upd_data']+"'</span><br>\n"+log.innerHTML;
```
atz: Can add a slash here to `<br/>` if we're going to do it 30 lines earlier too.

drh-stanford:
i corrected the rest of the broken `<br>` tags

app/assets/javascripts/bulk.js
mejackreed: :+1: for consistency, but seems like `<br/>` is a holdover from XHTML, no? Though I'm sure I've committed code with it in the past year :)

drh-stanford:
The `bulk.js` code writes out HTML without using any styling. Ideally it should be using lists or a table to do the logging rather than using `<br/>` tags.

mejackreed:
:+1:

Issue #100, opened 2015-11-02
Upgrade development/Travis versions of Fedora 3 and Solr 4


drh-stanford:
We're running Fedora 3.8.0 and Solr 4.10.4 in our deployment environment, but the development / Travis version of jetty runs Fedora 3.7.1 and Solr 4.9.  This issue it to upgrade to Fedora 3.8.0 (or 3.8.1) and Solr 4.10.4 for the development/ci environments.

See https://github.com/projecthydra/hydra-jetty/issues/40

PR #394, opened 2015-12-17
Upgrade to v7.2.0 of hydra-jetty

drh-stanford:
This PR upgrades our hydra-jetty instance to the tip of the `7.x-stable` branch which has Fedora 3.8.1 and Solr 4.10.2. See https://github.com/projecthydra/hydra-jetty/issues/40  Addresses #100

### DIFF
--- a/lib/tasks/argo.rake
+++ b/lib/tasks/argo.rake
@@ -68,7 +68,7 @@ namespace :argo do
   end
 
   namespace :jetty do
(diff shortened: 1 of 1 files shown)


mejackreed:
Let's see if we can cut a release and use the tag?

drh-stanford:
That's fine with me.. I have a question into them about cutting a new release...

drh-stanford:
Got confirmation that `v7.2.0` is the latest release on `7.x-stable`. It includes Fedora 3.8.1 and Solr 4.10.2.

Issue #85, opened 2015-11-01
Encoded newline characters disappear when editing XML


LynnMcRae:
Data entered into long text fields in our metadata, e.g., MODS notes or abstract, is treated as unformatted text. This means it will ignore whitespace (essentially by presenting it as html), including linefeeds in the data. Desired linefeeds must be encoded in the data as a `&#x10;` or `&#xA;`. When these are added to XML edited in Argo (when editing a datastream), they get translated as linefeeds in the resulting XML, that is ... 

   `<note>paragraph1&#xA;&#xA;paragraph2</note>`

becomes 

```
<note>paragraph1 

paragraph2</note>
```

and disappear if updated again.

Other encoded characters seem to be preserved, e.g., `&gt;` or `&amp;`

Newline encoding can be either `&#xA;` or `&#10;`

atz:
I don't think this came through entirely.  Please identify the expected encoded linefeeds again using markdown syntax, i.e. backticks like:
```
`encoded thing`
```
Or give a more descriptive attribution.

LynnMcRae:
Sorry for the multiple updates.  It needed a combination of backticks for the encoded characters and escaped \<note> tags to properly show how it displays.

drh-stanford:
The FoXML does not have the `&#xA;` entities in it. They are getting stripped at some point...

![screen shot 2015-12-17 at 3 38 50 pm](https://cloud.githubusercontent.com/assets/1861171/11885179/501178fc-a4d4-11e5-97b8-185352ce0d8e.png)

drh-stanford:
It looks like ActiveFedora is not taking the content verbatim when it saves it to Fedora. Instead it calls `Nokogiri::XML::Document.parse(new_content).to_s`. So it not preserve the linefeeds.

https://github.com/projecthydra/active_fedora/blob/v6.7.3/lib/active_fedora/om_datastream.rb#L110

```
[4] pry(main)> doc = Nokogiri::XML.parse('<speak>Hello&#xA;World</speak>')
=> #(Document:0x3ff1e2acc0c4 { name = "document", children = [ #(Element:0x3ff1e27f94f0 { name = "speak", children = [ #(Text "Hello\nWorld")] })] })
[5] pry(main)> doc.to_s
=> "<?xml version=\"1.0\"?>\n<speak>Hello\nWorld</speak>\n"
[6] pry(main)> 
```

So this is an issue with ActiveFedora.

drh-stanford:
[...]

PR #395, opened 2015-12-18
set the XML to be verbatim as posted from the client after validation

drh-stanford:
Tries to address #85 by setting the datastream in the XML editor to the exact content. Unfortunately ActiveFedora filters this content through `Nokogiri::XML.to_s` so the newline entities disappear. However, this is our part of the fix.

This PR also does major cleanup of the `datastream_update` method and specs to handle empty and malformed arguments.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -203,15 +203,20 @@ class ItemsController < ApplicationController
     end
   end
 
+  ##
+  # @option params [String] `:content` the XML with which to replace the datastream
+  # @option params [String] `:dsid` the identifier for the datastream, e.g., `identityMetadata`
+  # @option params [String] `:id` the druid to modify
   def datastream_update
-    ds = @object.datastreams[params[:dsid]]
-    # check that the content is valid xml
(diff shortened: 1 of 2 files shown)


app/controllers/items_controller.rb
```diff
@@ -204,14 +204,14 @@ def embargo_update
   end
 
   def datastream_update
-    ds = @object.datastreams[params[:dsid]]
-    # check that the content is valid xml
     begin
-      content = Nokogiri::XML(params[:content]){ |config| config.strict }
+      # check that the content is well-formed xml
+      Nokogiri::XML(params[:content]) { |config| config.strict }
     rescue
```
mejackreed: Is it possible to rescue the specific error here? Is it always going to be `Nokogiri::XML::SyntaxError`?

drh-stanford:
I've updated the PR to handle both empty and malformed cases -- Nokogiri throws a RuntimeError if the document is empty so I check for that beforehand so I can rescue on SyntaxError only.

The specs were also incomplete and off a bit so i restructured them all

mejackreed:
:+1:

PR #396, opened 2015-12-18
Change source id

### LINKED ISSUE #84: The "Change Source ID" button should let you set the source ID if it's missing
... but this situation currently gets you an error on submit

mejackreed:
Closes #84 

Adds in DatabaseCleaner around specs for test cleaning up (specifically works now around WFS actions). If the test is a JS spec it will use `truncation` strategy, for rack based specs it will use `transaction`.

Also removes legacy custom Argo logic for version enforcing in favor of standard DorServices call. Note: this change effects all `ItemController` specs as it removes the outdated `#on_hold` method from that controller.

557abf7b6e6184c1e8951bbf44d4af77e680f188 also adds more specific expectations for for our `allow_any`'s . These greedy mocks seem to cause chaos and don't always decompose between specs.

### DIFF
--- a/spec/features/item_view_metadata_spec.rb
+++ b/spec/features/item_view_metadata_spec.rb
@@ -1,17 +1,20 @@
 require 'spec_helper'
 
 feature 'Item view metadata' do
-  before do
-    @current_user = double(
+  let(:current_user) do
+    double(
       :webauth_user,
       login: 'sunetid',
       logged_in?: true,
       permitted_apos: [],
       is_admin: true,
-      roles: []
+      roles: [],
+      groups: []
     )
-    allow_any_instance_of(ApplicationController).to receive(:current_user)
-      .and_return(@current_user)
+  end
+  before do
+    expect_any_instance_of(CatalogController).to receive(:current_user)
+      .at_least(1).times.and_return(current_user)
   end
 
   scenario 'MD Source is "DOR"' do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -43,7 +43,20 @@ RSpec.configure do |config|
   config.fixture_path = "#{::Rails.root}/spec/fixtures"
 
   # Run each example in an ActiveRecord transaction
-  config.use_transactional_fixtures = true
+  config.use_transactional_fixtures = false
+
+  config.before :each do
+    if Capybara.current_driver == :rack_test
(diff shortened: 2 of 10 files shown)


spec/features/item_view_metadata_spec.rb
```diff
@@ -1,17 +1,20 @@
 require 'spec_helper'
 
 feature 'Item view metadata' do
-  before do
-    @current_user = double(
+  let(:current_user) do
```
drh-stanford: why all the changes to use `let` instead of a `@` variable?

spec/spec_helper.rb
```diff
@@ -43,7 +43,20 @@ def instantiate_fixture(druid, klass = ActiveFedora::Base)
...
 
   # Run each example in an ActiveRecord transaction
-  config.use_transactional_fixtures = true
+  config.use_transactional_fixtures = false
+
+  config.before :each do
+    if Capybara.current_driver == :rack_test
+      DatabaseCleaner.strategy = :transaction
+    else
+      DatabaseCleaner.strategy = :truncation
```
drh-stanford: does the `:transaction` strategy fail with the `wfs_rails` inclusion?

mejackreed: Anything outside of the rack context will fail (I believe). So also poltergeist javascript tests will most likely also fail.

spec/features/item_view_metadata_spec.rb
mejackreed: I'm concerned that the instance variable approach is also part of the problem here. Setting an instance variable is not an rspec convention (that I can find) and hope to move away from this practice:

http://betterspecs.org/#let
http://stackoverflow.com/questions/5359558/when-to-use-rspec-let/5359979#5359979

drh-stanford: They use instance variables in the doc, though it may not be a best practice. 

http://www.relishapp.com/rspec/rspec-core/docs/hooks/before-and-after-hooks

mejackreed: Ah.. yes. I think the issue might be here with using in "Feature" spec while giving using the "any_instance_of" mock.

mejackreed:
@drh-stanford The default behavior is exactly one time. This allows it receive multiple times.

PR #422, opened 2016-01-07
use show page size for index thumb, also move away from legacy "thumb…

### LINKED ISSUE #404: Thumbs on the result page are rather small
They seem disproportionately tiny to me ... we have room to show larger?
screen shot 2016-01-04 at 11 20 24 am

compare thumb on detail page
screen shot 2016-01-04 at 11 22 55 am

mejackreed:
…" endpoint to iiif way

Grabs the 400px on the long edge image and displays it using the same width/height limitations as show page (240px/240px).

Closes #404 

screen shot 2016-01-07 at 11 44 10 am


cbeer:
:+1:

PR #426, opened 2016-01-07
remove broken render_facet_value helper

### LINKED ISSUE #402: Selecting a facet via "more" interaction doesn't take you to a search result
Open a facet that requires "more", e.g. SW Region.
screen shot 2016-01-04 at 10 45 14 am
Open up "more' and pick a value, e.g. Scotland.  You are left on a screen that is an expansion of the tag layout, not an actual search result page.
screen shot 2016-01-04 at 10 46 23 am

drh-stanford:
This PR fixes #402. It removes the overridden method `render_facet_value`. This method was added circa 2011 and isn't needed anymore as far as I can tell. To test, I changed the catalog controller to use very short `limit:` parameters (e.g., 2). The APO and Collection facet values are titles (which the original comment said they weren't).

screen shot 2016-01-07 at 11 42 20 am

screen shot 2016-01-07 at 11 41 55 am

### DIFF
--- a/app/helpers/argo_helper.rb
+++ b/app/helpers/argo_helper.rb
@@ -82,12 +82,6 @@ module ArgoHelper
     render_thumbnail_helper doc, 'index-thumb', '', 'max-width:80px;max-height:80px;'
   end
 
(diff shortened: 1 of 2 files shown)


mejackreed:
:+1:

PR #432, opened 2016-01-08
reindex button should trigger a flash message, not a modal

### LINKED ISSUE #415: Response modal from Reindex has display issues
The response from reindexing blue button is to show status plus the entire resulting new solr doc.  It has some formatting issues, cosmetic only, but perhaps including this big wad of data seems pointless given we can view the solr document directly from the detail page now.  Omit it? Just give status=ok vs error?

screen shot 2016-01-05 at 3 44 46 pm

mejackreed:
Closes #415 

Rather than using modals to display messages, reindexing now uses flash messaging.

screen shot 2016-01-08 at 11 36 49 am
screen shot 2016-01-08 at 11 36 30 am


app/controllers/dor_controller.rb
```diff
@@ -51,11 +51,15 @@ def reindex
...
+    message = "Successfully updated index for #{params[:pid]}"
+    index_logger.info message
+    flash[:notice] = message
+    redirect_to :back
   rescue ActiveFedora::ObjectNotFoundError # => e
-    index_logger.info "failed to update index for #{params[:pid]}, object not found in Fedora"
-    render :status => 500, :text => 'Object doesnt exist in Fedora.'
-    return
+    message = "Failed to update index for #{params[:pid]}, object not found in Fedora"
+    index_logger.info message
```
drh-stanford: maybe this log output should be `.warn` or `.error` call? @jmartin-sul how do we parse this log output, if at all?

drh-stanford:
i'm :+1: with a minor comment above

app/controllers/dor_controller.rb
mejackreed: Good idea. Kept as `.info` to leave previous functionality, but probably does make sense in `.warn`?

jmartin-sul: this has been refactored in the profiling indexing work; i agreed with @drh-stanford's suggestion, apparently:
```ruby
# DorController.reindex
  def reindex
    @solr_doc = Argo::Indexer.reindex_pid params[:pid], index_logger
    Dor::SearchService.solr.commit # reindex_pid doesn't commit
  rescue ActiveFedora::ObjectNotFoundError # => e
    render :status => 404, :text => 'Object does not exist in Fedora.'
    return
  end
```

```ruby
Argo::Indexer.reindex_pid
    def self.reindex_pid(pid, index_logger = nil, should_raise_errors = true)
      index_logger ||= default_index_logger
      obj = Dor.load_instance pid
      solr_doc = reindex_object obj
      index_logger.info "updated index for #{pid}"
      solr_doc
    rescue StandardError => se
      if se.is_a? ActiveFedora::ObjectNotFoundError
        index_logger.warn "failed to update index for #{pid}, object not found in Fedora"
      else
        index_logger.warn "failed to update index for #{pid}, unexpected StandardError, see main app log: #{se.backtrace}"
      end
      raise se if should_raise_errors
    rescue Exception => ex
      index_logger.error "failed to update index for #{pid}, unexpected Exception, see main app log: #{ex.backtrace}"
      raise ex # don't swallow anything worse than StandardError
    end
```

mejackreed:
Ok, updated using `.warn`.

jmartin-sul:
well, since dor_controller and dor_controller_spec were refactored as part of the indexing changes, i'd hold off on merging this until that's merged.  the conflict should be pretty straight-forward to resolve, you'd drop the catch for StandardError, and you wouldn't have to handle logging at all anymore.

or, we could merge this, and i could rebase profiling_and_indexing_improvements off develop, and fix conflicts there.

darrenleeweber:
:+1:

drh-stanford:
i merged #246 so this PR needs to rebase to merge in the changes Jon made already to the logging (see above)

drh-stanford:
The message in the error screenshot will be "Object does not exist in Fedora", not as shown. It's a very rare case.

PR #435, opened 2016-01-09
Revert "reindex button should trigger a flash message, not a modal"

drh-stanford:
Reverts sul-dlss/argo#432


mejackreed:
If we decide to revert as the method of solving this problem, I'm fine with it. Should we add an explicit test for this so that we don't change later?

I think there may be other options to consider as well.

mejackreed:
See #439 for my proposed solution.

Issue #76, opened 2015-11-01
Allow users with APO roles access to the APO itself


LynnMcRae:
A person with any role in an APO should have access to view the APO. 

A person with a dor-apo-manager role in an APO should be able to edit the APO object itself as well as create a collection that is governed by that APO. 

Example Jon Manton in the ARS APO https://argo.stanford.edu/catalog/druid:bz845pv2292

blalbrit:
pending conversation with @eefahy

mejackreed:
Still pending?

blalbrit:
@LynnMcRae - where are we with this?

LynnMcRae:
I know of nothing blocking this unless there are more questions about what it means.  What was the conversation with Erin about? ... I don't see how that relates.

blalbrit:
I think that was the conversation about workgroups vs. ldap, wasn't it?

LynnMcRae:
I don't think this has anything to do with workgroup/ldap ...  An APO manager can get at any item governed by the APO but not the APO itself, which is an odd omission.  I’m hoping it’s as simple as augmenting the background filter that is constructed on roles to be  “objects governed by [list of my APOs] OR objects = [list of my APOs]", but I may be wrong about the mechanisms at play.

darrenleeweber:
@LynnMcRae - what is wrong with https://argo.stanford.edu/catalog/druid:bz845pv2292 ?  It displays fine.  What is the display or the interaction(s) that cause problems for this druid?

LynnMcRae:
A DEV test case ...

the Jarndyce Single-Volume Nineteenth-Century Novel Collection has one item, one collection object, and one APO ...

you can see all three with your super powers, but impersonate user llam813 (enter sunetid:llam813)and you can see the item and colleciton only. the APO disappears from your view.

darrenleeweber:
- "A person with a dor-apo-manager role in an APO should be able to edit the APO object itself as well as create a collection that is governed by that APO."  This seems to be controlled by logic that displays the UI buttons at:
  - https://github.com/sul-dlss/argo/blob/develop/app/helpers/argo_helper.rb#L130-L164

darrenleeweber:
The logic patterns uses in controllers and views are:
```
app/controllers/items_controller.rb
619:    if current_user.is_admin || @object.can_manage_content?(current_user.roles @apo)
625:    if current_user.is_admin || @object.can_view_content?(current_user.roles @apo)

[...]

PR #437, opened 2016-01-09
[WIP] Modifications to user permissions

darrenleeweber:
Work in progress on #76 - needs integration tests to verify the changes work as expected when viewing an APO.


jmartin-sul:
strange...  it seems coveralls is still being misbehaving on this branch?  was the simplecov/coveralls change a red herring on that front?  still reasonable to put it in a separate PR, but it looks like the same changes were made on PR #443, and its reporting on the state of that branch is still ok.

also, the redundant CI status for both PR and push is gone, and now there's just push.  that is redundant, right, both of those seem to run the same tests?  not a bad thing to get rid of the redundancy, just confusing as to what did it in these changes.

darrenleeweber:
The specs failure is a mystery.  When all the specs are run, i.e. `rake ci` there are capybara errors on the apo_spec, but they work fine outside the context of all other specs, i.e.
```
bundle exec rspec spec/integration/apo_spec.rb
...
Finished in 7.45 seconds (files took 3.96 seconds to load)
5 examples, 0 failures
```

So, this is beyond me at the moment and I have no idea how, and not enough time, to fix it.  Despite some helpful discussion on slack and trying a few things, this spec failure persists.

app/controllers/application_controller.rb
```diff
@@ -20,6 +20,25 @@ class ApplicationController < ActionController::Base
...
   layout 'application'
 
+  # A common method for controllers to find a Dor object, whether the
+  # input is a DRUID, a PID or a Dor object.
+  # @param druid [String|ActiveFedora::Base]
+  #              DruidTools::Druid is used to parse String, which is
+  #              then input for Dor.find to obtain a Dor object
+  #              Dor objects inherit from ActiveFedora::Base
+  # @raises ArgumentError
+  def find_druid(druid)
```
mejackreed: Why is this an action in the controller? If we need this I would advocate for adding a class that contains this logic.

darrenleeweber:
Closing this to clean up the comments, will reopen a new PR now that this is ready for review.

PR #438, opened 2016-01-10
Re-enabling three of the buttons that were erroneously disabled.

### LINKED ISSUE #428: Several bulk update buttons are inert
In dev and stage, these buttons do nothing
- Set source id
- Set object rights
- Set content type

tingulfsen:
Fixes #428

### DIFF
--- a/spec/features/bulk_screen_spec.rb
+++ b/spec/features/bulk_screen_spec.rb
@@ -1,0 +1,43 @@
+require 'spec_helper'
+
+# Feature/view tests for the (old) bulk actions view.
+feature 'Bulk actions view', js: true do
+  before :each do
+    @current_user = double(
+      :webauth_user,
+      login: 'sunetid',
+      logged_in?: true,
+      permitted_apos: [],
+      is_admin: true,
+      permitted_collections: [ [ "string" ] ]
+    )
+
+    expect_any_instance_of(ApplicationController).to receive(:current_user).exactly(8).times.and_return(@current_user)
+  end
+
+  scenario 'basic page renders ok', :focus => true do
+    visit report_bulk_path
+
+    expect(page).to have_css('h1', 'Bulk update operations')
+    expect(find(:xpath, "//span[@class='bulk_button'][text()='Get druids from search'][not(@disabled)]")).to be_truthy
+    expect(find(:xpath, "//span[@class='bulk_button'][text()='Paste a druid list'][not(@disabled)]")).to be_truthy
+
+    find(:xpath, "//span[@class='bulk_button'][text()='Get druids from search'][not(@disabled)]").click
+    
+    expect(find_button('Prepare objects')).to be_truthy
+    expect(page).to have_button('Prepare objects', disabled: false)
+    expect(page).to have_button('Refresh MODS', disabled: false)
+    expect(page).to have_button('Set source Id', disabled: false)
+    expect(page).to have_button('Set object rights', disabled: false)
+    expect(page).to have_button('Set content type', disabled: false)
+    expect(page).to have_button('Set collection', disabled: false)
+    expect(page).to have_button('Add collection', disabled: false)
+    expect(page).to have_button('Apply APO defaults', disabled: false)
+    expect(page).to have_button('Add a workflow', disabled: false)
+    expect(page).to have_button('Close versions', disabled: false)
+    expect(page).to have_button('Reindex', disabled: false)
+    expect(page).to have_button('Republish', disabled: false)
+    expect(page).to have_button('Tags', disabled: false)
+    expect(page).to have_button('Purge', disabled: false)
+  end
+end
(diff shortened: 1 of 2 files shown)


mejackreed:
Since we have already had a regression here, does it make sense to add a view spec for the expected buttons and attributes?

tingulfsen:
Yes, I will make sure to add tests before this is merged.

tingulfsen:
I discussed turning the test into a view test with @mejackreed but found that the view test became rather cumbersome to code since the .html.erb file in question calls a number of Blacklight and other assorted tools in order to do its work. In addition, JavaScript is actually required to make most of the buttons show up.

mejackreed:
:+1:

tingulfsen:
Bumping this pull request - can someone merge? I don't know why coveralls says that the coverage decreased by 0.0%...

spec/features/bulk_screen_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  before :each do
+    @current_user = double(
+      :webauth_user,
+      login: 'sunetid',
+      logged_in?: true,
+      permitted_apos: [],
+      is_admin: true,
+      permitted_collections: [["string"]]
+    )
+
```
darrenleeweber: Replace lines 6-13 with `admin_user` after we merge the APO PR.

spec/features/bulk_screen_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect_any_instance_of(ApplicationController).to receive(:current_user).exactly(8).times.and_return(@current_user)
```
darrenleeweber: This is way too specific and it's not really the point of this spec to check how many times a controller is called or not.

darrenleeweber:
:-1:  because it uses a spurious `expect` on AppController to receive `current_user`.  It doesn't matter how many times it's called.  A simple `allow` and `return` for `current_user` will suffice.

mejackreed:
I disagree with @darrenleeweber here. I think the expect is warranted, if something was removed downstream of the controller logic unexpectedly this spec would catch it. I'd much rather be more cautious than opening up allows on `ApplicationController`. I saw :shipit:

jmartin-sul:
though i tend to err on the side of using expect, i think i agree with @darrenleeweber in this particular case.  it doesn't look like the code being tested by the test in question calls `current_user` directly, right?  `ReportController#bulk` is what gets called, and that calls `get_search_results`?  and neither seems to call `current_user`.  it seemed to me like our team was generally in favor of testing a thing, but not testing both that thing as well as the things it relies on.  or at least, it seems to me like that can introduce unnecessary test fragility, and i'd gotten the impression that others felt similarly.  hopefully there are tests in the appropriate places for the things that actually call `current_user`.

since it seems like allow has caused more trouble than expect, maybe `expect(...).to receive(...).at_least(:once)` would be better if it were agreed that this sort of change should be made?

i also don't feel strongly about this and would be totally :ok_hand: to merge it as-is.  seems like it'd be easy enough to fix even if it did get broken inadvertently.

mejackreed:
@jmartin-sul :+1: ... The code in question is a feature spec so I might argue that the feature requires those calls. I'd prefer an `expect...at_least(:once)` over an allow. That would allow us (no pun intended) to catch any changes that happen in the request chain. 

I'd prefer that (expect once) than having no-op allow mocks being leftover in specs without anyway to find out they need to be removed. Historically, Argo has been shot through with legacy code that isn't ever executed (or doesn't need to be). I'm hoping these approaches (prefer `expect` over `allow`) can help future maintainers. That's another reason for my heavy resistance against `allow` (when possible). The danger of overmocking > undermocking.

jmartin-sul:
@mejackreed:  fair enough, i agree with all of that.

darrenleeweber:
See, for example, https://github.com/sul-dlss/argo/blob/issues76-apo-access/spec/features/bulk_jobs_spec.rb#L6

darrenleeweber:
@mejackreed - when the controller or view just needs an admin user, it should use a User object, not a mock of a user.  The APO PR encountered many strange failures that were finally resolved by using a common user that is hooked into `ApplicationController#current_user` by an `allow`.  Unless a spec requires manipulation of a user attributes and/or methods, a stub/mock is not required.

mejackreed:
This PR may not be the best place to discuss our current overall test strategies. Maybe we should meet about that? I'm glad that we are having the discussion and hopefully will facilitate a better overall codebase.

jmartin-sul:
@tingulfsen:  preferences on the part of the implementer after all that discussion?

would anyone seriously object to `expect...at_least(:once)`, which it at least @mejackreed and i are comfortable with, and which seems to address @darrenleeweber's concern about over-specification?

mejackreed:
:+1:

tingulfsen:
My personal preference is to specify an exact number if possible, because that actually tells us something about how the application behaves. In the future, if the behavior changes, this might help flag issues that might otherwise be missed. I think specifying the exact number also flags the insaneness of Argo to future developers.

That being said, I don't feel strongly enough about this to keep the discussion going.

I do think that having a separate meeting regarding testing practices might be beneficial.

darrenleeweber:
I'll merge this and see what happens with the APO PR, where all the feature specs simply call `admin_user` from the spec_helper and that's it.

PR #439, opened 2016-01-11
enable non-referrer HTTP requests to return appropriate status and body

### LINKED ISSUE #433: Handle `DorController#reindex` for non argo clients
Should probably be moved to a `POST`. Updates need to happen with other services that use this an indexing endpoint.

### LINKED ISSUE #434: /dor/reindex no longer works as a REST endpoint
it used to be that you could do something like `/dor/reindex/druid:hv992ry2431` to reindex an object via REST call (and indeed, this is what fedora ultimately does when it sends a message indicating that an object was updated, though karaf acts as a queueing intermediary).

however, a regression was introduced as part of PR #432, and an error occurs unless the URL is hit from a link in the app:
`No HTTP_REFERER was set in the request to this action, so redirect_to :back could not be called successfully. If this is a test, make sure to specify request.env["HTTP_REFERER"].`
http://localhost:3000/dor/reindex/druid:hv992ry2431
https://argo-prod-b.stanford.edu/dor/reindex/druid:hv992ry2431
https://sul-squash.stanford.edu/projects/argo/environments/production/bugs/61
https://github.com/sul-dlss/argo/blob/develop/app/controllers/dor_controller.rb#L45-L53

@mejackreed points out that we can get the referrer info via `request.referrer` (or `request.referer`, if you're into preserving RFC typos).  if it's acceptable to differentiate behavior based on that, i'd propose that we use the new behavior whenever `request.referrer` indicates that the request was a result of clicking on a reindex link in the app.  otherwise, default to the old behavior.

mejackreed:
Closes #434 
Closes #433 

Ports in a new Rails 5 method which gives us a redirect back with a fallback path.

This PR now handles the following cases:

 - Within Argo (or another application) following a link to the reindex endpoint will return a user to `:back` or `catalog_index` (show page) (fallback case)
 - An external application hitting that endpoint (no "Referer" header) will get a 200 with text body or a 404 with text body

### DIFF
--- a/app/controllers/dor_controller.rb
+++ b/app/controllers/dor_controller.rb
@@ -46,10 +46,16 @@ class DorController < ApplicationController
     @solr_doc = Argo::Indexer.reindex_pid params[:pid], index_logger
     Dor::SearchService.solr.commit # reindex_pid doesn't commit, but callers of this method may expect the update to be committed immediately
     flash[:notice] = "Successfully updated index for #{params[:pid]}"
-    redirect_to :back
+    render status: 200, text: flash[:notice] and return unless request.headers['Referer']
+    redirect_back(
+      fallback_location: proc { catalog_path(params[:pid])}
+    )
   rescue ActiveFedora::ObjectNotFoundError # => e
     flash[:error] = 'Object does not exist in Fedora.'
-    redirect_to :back
+    render status: 404, text: flash[:error] and return unless request.headers['Referer']
+    redirect_back(
+      fallback_location: proc { catalog_path(params[:pid])}
+    )
   end
 
   def delete_from_index
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -78,4 +78,15 @@ class ApplicationController < ActionController::Base
     true
   end
 
+  ##
+  # A ported over Rails 5 enhancement to ActionPack
+  # @see https://github.com/rails/rails/commit/13fd5586cef628a71e0e2900820010742a911099
+  def redirect_back(fallback_location:, **args)
+    if (referer = request.headers['Referer'])
+      redirect_to referer, **args
+    else
+      redirect_to fallback_location, **args
+    end
+  end
+
 end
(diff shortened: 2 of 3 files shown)


app/controllers/dor_controller.rb
```diff
@@ -46,10 +46,16 @@ def reindex
...
     flash[:notice] = "Successfully updated index for #{params[:pid]}"
-    redirect_to :back
+    render status: 200, text: flash[:notice] and return unless request.headers['Referer']
+    redirect_back(
+      fallback_location: proc { catalog_path(params[:pid])}
+    )
   rescue ActiveFedora::ObjectNotFoundError # => e
     flash[:error] = 'Object does not exist in Fedora.'
-    redirect_to :back
+    render status: 404, text: flash[:error] and return unless request.headers['Referer']
```
mejackreed: `and` here is appropriate as it is a Rails thing to use `and return` in controllers.

app/controllers/dor_controller.rb
```diff
@@ -46,10 +46,16 @@ def reindex
     @solr_doc = Argo::Indexer.reindex_pid params[:pid], index_logger
     Dor::SearchService.solr.commit # reindex_pid doesn't commit, but callers of this method may expect the update to be committed immediately
     flash[:notice] = "Successfully updated index for #{params[:pid]}"
-    redirect_to :back
+    render status: 200, text: flash[:notice] and return unless request.headers['Referer']
```
mejackreed: `and` here is appropriate as it is a Rails thing to use `and return` in controllers. https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/actionpack/lib/action_controller/base.rb#L157

app/controllers/dor_controller.rb
jmartin-sul: it turns out i have an even worse understanding of how `and` works than i thought it did.  anyway, this seems fine based on the documentation i found and the testing i did in pry (and the unit tests passing).

mejackreed: Yeah.. tis a strange one indeed :smiley_cat:

PR #442, opened 2016-01-12
Catching error when an invalid APO is entered into the RELS-EXT datas…

### LINKED ISSUE #418: [WIP] Governing items by non-existing APOs causes Rails exception display
In an attempt to reproduce #334 using instructions from @drh-stanford (update the RELS-EXT datastream so a regular item is governed by a non-existing APO) I get the dreaded Rails exception display. This is likely related to #405 - while I realize that Argo cannot always be robust in the face of willfully or accidentally invalid input, I don't think this error is even remotely useful to anyone.

rels-ext-bug

tingulfsen:
…tream.

Resolves #418

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -8,7 +8,7 @@ class ItemsController < ApplicationController
   before_filter :forbid_modify, :only => [:add_collection, :set_collection, :remove_collection, :update_rights, :set_content_type, :tags, :tags_bulk, :source_id, :delete_file, :close_version, :open_version, :resource, :add_file, :replace_file, :update_attributes, :update_resource, :mods, :datastream_update ]
   before_filter :forbid_view,   :only => [:get_preserved_file, :get_file]
   before_filter :enforce_versioning, :only => [:add_collection, :set_collection, :remove_collection, :update_rights, :tags, :source_id, :set_source_id, :set_content_type, :set_rights]
-  after_filter  :save_and_reindex,   :only => [:add_collection, :set_collection, :remove_collection, :open_version, :close_version, :tags, :tags_bulk, :source_id, :datastream_update, :set_rights, :set_content_type, :apply_apo_defaults]
+  after_filter  :save_and_reindex,   :only => [:add_collection, :set_collection, :remove_collection, :open_version, :close_version, :tags, :tags_bulk, :source_id, :set_rights, :set_content_type, :apply_apo_defaults]
 
   def purl_preview
     @object.add_collection_reference @object.descMetadata.ng_xml
@@ -209,6 +209,15 @@ class ItemsController < ApplicationController
       fail ArgumentError, 'XML is not well formed!'
     end
     @object.datastreams[params[:dsid]].content = params[:content] # set the XML to be verbatim as posted
+
+    # Catch reindexing errors here to avoid cascading errors
+    begin
+      save_and_reindex
+    rescue ActiveFedora::ObjectNotFoundError => e
+      render text: 'The object was not found in Fedora. Please recheck the RELS-EXT XML.', status: :not_found
+      return
+    end
+
     respond_to do |format|
       format.any { redirect_to catalog_path(params[:id]), :notice => 'Datastream was successfully updated' }
     end
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -256,6 +256,7 @@ describe ItemsController, :type => :controller do
   end
   describe '#datastream_update' do
     let(:xml) { '<contentMetadata/>' }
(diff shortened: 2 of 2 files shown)


mejackreed:
Does it make sense to catch this error in the way that is done in `ApplicationController`? Is that not catching it?

https://github.com/sul-dlss/argo/blob/develop/app/controllers/application_controller.rb#L10

I think the error I'm seeing with #418 (the double render) is because the rescue in https://github.com/sul-dlss/argo/blob/develop/app/controllers/application_controller.rb#L10 is not returning.

Try adding `and return` to the end of the `rescue_from` statement.

tingulfsen:
Good idea, but when I tried it I ended up with the same AbstractController::DoubleRenderError as before. FYI I changed the ApplicationController line to:

rescue_from ActiveFedora::ObjectNotFoundError, with: -> { render(text: 'Object Not Found', status: :not_found) and return }

I also added a "and return" statement to the redirect() call within ItemsController itself, but that also failed to fix the issue.

If anyone has any other ideas on how to somehow catch this error without causing DoubleRenderError to appear, let me know, although I suspect that the ideal solution is to fix the problem in indexing so that the exception is not thrown in the first place.

jmartin-sul:
the issue with catching it in `ApplicationController` is that it can look like the object that was requested wasn't found, as opposed to the object referenced by the requested object (and it's often nice for that distinction to be clear up front).

but it does feel messy to wait till the change is made and the object is reindexed.  is there any easy way to check that problem and give an error without saving?

app/controllers/items_controller.rb
```diff
@@ -209,6 +209,15 @@ def datastream_update
       fail ArgumentError, 'XML is not well formed!'
     end
     @object.datastreams[params[:dsid]].content = params[:content] # set the XML to be verbatim as posted
+
+    # Catch reindexing errors here to avoid cascading errors
+    begin
+      save_and_reindex
+    rescue ActiveFedora::ObjectNotFoundError => e
+      render text: 'The object was not found in Fedora. Please recheck the RELS-EXT XML.', status: :not_found
```
jmartin-sul: are there other situations (besides the one addressed by this PR) where `ActiveFedora::ObjectNotFoundError` could get thrown here?

also, could the wording be a bit more specific?  maybe "An object was not found in Fedora. Possible bad object reference in RELS-EXT XML?"

or, if a bad reference really is the only realistic way this path gets hit, maybe "An object was not found in Fedora. Please check for invalid references in RELS-EXT XML."

mejackreed:
It seems to me that the 404 is happening after the redirect to catalog_path... No? https://github.com/sul-dlss/argo/pull/442/files#diff-2eb2bf110a48bee6c79d3bbb377a549eR222

The RELSEXT in question is being updated in Fedora, the catalog view just can't find the apo on the next page.

jmartin-sul:
yeah, the APO lookup during solrization is what's failing, i believe.  which is why the check after RELS-EXT save seems less than ideal.

"I suspect that the ideal solution is to fix the problem in indexing so that the exception is not thrown in the first place." - https://github.com/sul-dlss/argo/pull/442#issuecomment-171127391

i'd vote for that, if it's not too arduous.  haven't looked at that code in a while, so not sure one way or the other.

atz:
```
but it does feel messy to wait till the change is made and the object is reindexed. is there any easy way to check that problem and give an error without saving?
```

You are basically asking "can I get transactionality?", or "can I dry run the second half of our transactionless kludge first and then maybe run it again third?"

atz:
The exception is sorta legitimate to the model though.  In the model we have guaranteed that we always have an APO.  Here the problem is from manually screwing with the authoritative place that stores that APO's identifier.

jmartin-sul:
`You are basically asking "can I get transactionality?"`

but it sounds so reasonable when you phrase it that way.

mejackreed:
I'm back to :+1: Tommy's original approach.

@atz plz get on adding transactionality to hydra in a box cc @mjgiarlo so he can architect it

mjgiarlo:
:architected:

LynnMcRae:
The ability to "manually screw up" the integrity of the APO or collection reference comes in part from the direct datastream XML editing we support, which is intended for those repository acolytes tasked with managing the quality of data and effecting repairs at the core when other means fail. These are the same folks who can write/run remediation scripts doing essential repairs, and we trust these processes to do good rather than evil. Access to this level of object management should not grow with an expanding Argo audience, it is basically direct Fedora manipulations through the lens of Argo based authorization (and a nicer interface :)

Also remember that the data can go bad without manual intervention -- a collection or event APO reference could go stale because those objects are abandoned or deleted. In either case, we want to be able to get to the item display in order to SEE that some data in the object is bad, which is the thrust of this fix.

mejackreed:
Please someone create a slack emoji for "repository acolyte",,, that is all

jmartin-sul:
i propose that the repository acolyte look something like a https://en.wikipedia.org/wiki/Druid

PR #443, opened 2016-01-13
Enable simplecov with coveralls

darrenleeweber:
This is designed to enable SimpleCov coverage reports in ./coverage on laptop development, so that the coverage data can be viewed immediately, regardless of travis/coveralls functionality.  The coveralls functionality should be preserved, according to their website docs, see https://coveralls.zendesk.com/hc/en-us/articles/201769485-Ruby-Rails

A similar example of this working is in the triannon project at:
https://github.com/sul-dlss/triannon/blob/master/spec/spec_helper.rb#L12-L29

And the coveralls reports are working, see:
https://coveralls.io/github/sul-dlss/triannon

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -168,12 +168,13 @@ GEM
       activesupport (>= 3.0)
       deep_merge (~> 1.0.0)
     confstruct (0.2.7)
-    coveralls (0.8.3)
+    coveralls (0.8.10)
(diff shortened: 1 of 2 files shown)


darrenleeweber:
ping @jmartin-sul @mejackreed - thanks for comments on the #437

jmartin-sul:
:+1:

PR #455, opened 2016-01-14
override solr facet.limit configuration for hierarchical facets

### LINKED ISSUE #371: argo-prod-b -- Tags facet is incomplete?
Compare lists from prod and current prod-b 
screen shot 2015-12-15 at 11 29 04 am
screen shot 2015-12-15 at 11 29 20 am

In case this is some interaction with the facet count limit, this should be another facet that shows all entries rather than goes through "more" since I imagine the hierarchical behavior is incompatible with more.

drh-stanford:
This PR fixes #371 by overriding the `facet.limit` Solr configuration for hierarchical facets. The decision to make the limit's 9999 here is that we're not able to guarantee that the Solr `facet.limit` parameter is appropriately large enough, and on our Solr cluster core it's set to 10.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -72,7 +72,7 @@ class CatalogController < ApplicationController
     # exploded_tag_ssim indexes all tag prefixes (see IdentityMetadataDS#to_solr for a more exact
     # description), whereas tag_ssim only indexes whole tags.  we want to facet on exploded_tag_ssim
     # to get the hierarchy.
-    config.add_facet_field 'exploded_tag_ssim',               label: 'Tag',                 partial: 'blacklight/hierarchy/facet_hierarchy'
+    config.add_facet_field 'exploded_tag_ssim',               label: 'Tag',                 limit: 9999, partial: 'blacklight/hierarchy/facet_hierarchy'
     config.add_facet_field 'objectType_ssim',                 label: 'Object Type',         limit: 10
     config.add_facet_field 'content_type_ssim',               label: 'Content Type',        limit: 10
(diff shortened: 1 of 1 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -72,7 +72,7 @@ class CatalogController < ApplicationController
     # exploded_tag_ssim indexes all tag prefixes (see IdentityMetadataDS#to_solr for a more exact
     # description), whereas tag_ssim only indexes whole tags.  we want to facet on exploded_tag_ssim
     # to get the hierarchy.
-    config.add_facet_field 'exploded_tag_ssim',               label: 'Tag',                 partial: 'blacklight/hierarchy/facet_hierarchy'
+    config.add_facet_field 'exploded_tag_ssim',               label: 'Tag',                 limit: 9999, partial: 'blacklight/hierarchy/facet_hierarchy'
```
cbeer: > facet.limit
>
> This param indicates the maximum number of constraint counts that should be returned for the facet fields. A negative value means unlimited.

Should this be -1, instead of simply a large number?

jmartin-sul: if that works, that seems reasonable, and feels a bit less hacky.

on the other hand, maybe we want to draw a line somewhere.

drh-stanford: That's fine with me to make it `-1`. I haven't tried it out though yet. I was thinking that the tags facet in particular could have more than 9,999 values. Is there a way to count the number of facet values easily?

LynnMcRae: Production currently has 2998 tag values (counting headings and terminal nodes).

PR #461, opened 2016-01-16
use limit -1 for unlimited facets

drh-stanford:
This PR changes how we do unlimited facets from 9999 to -1.


drh-stanford:
you have to use `limit = -2` or `limit = -9999` for this to work. -1 doesn't work. tabling for now

drh-stanford:
See https://github.com/projectblacklight/blacklight/blob/858f5324f6344e07fa1c490c542764834fd78e43/lib/blacklight/solr/search_builder_behavior.rb#L135

Issue #462, opened 2016-01-19
Consolidate constants


darrenleeweber:
There is a relatively new file in `lib/constants.rb`, so how does that practice compare with other `config/initializers/*.rb` such as https://github.com/sul-dlss/argo/blob/develop/config/initializers/permissions.rb?  Is there a consensus on using the `lib/constants.rb` in favor of the initializers?  Should the permissions initializer move into `lib/constants.rb`?  Does the same argument apply to a one liner like https://github.com/sul-dlss/argo/blob/develop/config/initializers/indexer.rb?

What’s the general rails-y way to do this?  Is it `lib/constants.rb` or `config/initializers/*.rb`?  (I wonder if the permissions initializer was created by a rails generator or if it belongs to a rails gem.  I don’t see any evidence of the latter.)

darrenleeweber:
The sulpub app uses a similar pattern to the `lib/constants.rb`, see
https://github.com/sul-dlss/sul-pub/blob/develop/config/application.yml
https://github.com/sul-dlss/sul-pub/blob/develop/lib/settings.rb
https://github.com/sul-dlss/sul-pub/blob/develop/app/helpers/authentication_helper.rb#L6-L7

darrenleeweber:
There are also Argo settings in
https://github.com/sul-dlss/argo/blob/develop/config/settings/test.yml

mejackreed:
Argo stuff in settings are something different than the constants being discussed consolidated.

jmartin-sul:
agreed on constants vs settings.  settings are environment specific or likely to be tweaked over time, constants should be the constant things that we imagine will change very slowly or not at all.  i can see how there could be gray area, but i don't think any of the things we're looking at are in it.

jmartin-sul:
which is to say, the permissions initialization clearly seems like a constant to me, and the log file thing clearly seems like a setting.  but the log file constant appears to be unused, and is misleading now that we've got other log file stuff in settings, so i just pushed a branch that drops it, and will create a PR shortly...

darrenleeweber:
https://github.com/bbatsov/rails-style-guide#configuration

mejackreed:
Yep, we are also using the config gem here and that has a specific pattern of approach to follow for configuration.

jmartin-sul:
[...]

PR #464, opened 2016-01-19
drop unused LOG_FILE constant

jmartin-sul:
this LOG_FILE constant doesn't appear to be referenced anywhere in the application.  additionally, there are log file related things in the settings, so this could be confusing.  seems like the right thing to do is to remove it.


darrenleeweber:
:+1: per https://github.com/sul-dlss/argo/issues/462, if the checks are all passing.

mejackreed:
:+1:

PR #466, opened 2016-01-21
Status controller handles exception

darrenleeweber:
Fixes a spec failure, i.e.
```
  1) StatusController log with test_obj should 404 instead of 500 on bad IDs
     Failure/Error: get 'log', :test_obj => 'junk'
       See logger for details
     # ./app/controllers/status_controller.rb:17:in `log'
     # ./spec/controllers/status_controller_spec.rb:52:in `block (3 levels) in <top (required)>'
```

### DIFF
--- a/app/controllers/status_controller.rb
+++ b/app/controllers/status_controller.rb
@@ -23,7 +23,7 @@ class StatusController < ApplicationController
       render :status => 500, :text => "Nothing indexed recently. Even after saving '#{params[:test_obj]}'."
       return
     end
-  rescue ActiveFedora::ObjectNotFoundError
+  rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest
     render :status => 404, :text => "No object '#{params[:test_obj]}' found"
   end
 


mejackreed:
Seems like a good idea. Where was that test failing in the test suite? It doesn't seem to be failing on Travis. Is there a failing test that we can write to ensure the case is tested?

darrenleeweber:
@mejackreed - did you read the description?

mejackreed:
I did. I just don't see our Travis build failing on the referenced spec. Maybe I'm missing the context of this?

darrenleeweber:
Well, I get the failure somehow and I don't care why I get it and travis doesn't.  This fixes it.

mejackreed:
Ah I see. That can be frustrating. Are you getting the error when running `bundle exec rake ci` locally? I wonder if there is something else we are missing that might be causing this larger issue.

darrenleeweber:
If it's an invalid request, then so be it.  Who knows what parameters could be input in place of a properly formed DRUID?

jmartin-sul:
IMO:
* the error being trapped feels more like it should be presented as a 500.  an invalid request was made to a backend service, which i think is better described by "internal server error" than "object not found".
* i think in the course of doing all this cleanup in argo, being assiduous about not merging fixes that we don't fully understand has served us well.  i'm leery of getting lax on that habit to get specs to pass for one developer when the specs aren't breaking for anyone else or for travis, and no one's sure why the specs are breaking on that one laptop.

not to say that specs seeming to break randomly isn't frustrating.  it is quite a pain (one i've lost plenty of time to), and can really throw a wrench in one's ability to get things done.  but i don't think this is the right solution for the problem at hand.

jmartin-sul:
or, we could pass a 400 ("invalid request") back to the caller.

atz:
I'm for closing this.

darrenleeweber:
See snippets in slack, where the problem might be in dor-services at Dor.find.

darrenleeweber:
The exeption handling can be dealt with in the application controller if it's not dealt with in the dor-services gem.  Either way, the change in this PR will be obsolete.

PR #467, opened 2016-01-21
Merge master jan 2016

jmartin-sul:
merges master as of now (2016-01-20) into develop, so that master can be fast-forwarded to develop, barring any unforeseen production hotfixes between now and the deployment of argo 3.0.  i just branched off develop, merged master into that, and resolved conflicts.

this was actually pretty straight-forward, i'd worried it'd take hours, but it was more like an hour and a half.

the merge commit message makes for a good summary:

Merge branch 'master' into merge-master-jan-2016

manually resolved conflicts:
* .travis.yml: kept develop version, all desirable changes from master already included
* Gemfile: kept develop version, all desirable changes from master already included
* Gemfile.lock: kept develop version as-is since Gemfile didn't change
* app/helpers/argo_helper.rb: kept develop version, only desirable change already incorporated, index_queue_depth was removed
* app/views/catalog/_home_text.html.erb: kept develop version, https changes already incorporated
* spec/models/user_spec.rb: refactoring on develop thwarted auto-merge, re-implemented sensible minor updates to mock workgroup and user names from master


jmartin-sul:
for reference, comparison of master with develop as it stands without this change:  https://github.com/sul-dlss/argo/compare/master

jmartin-sul:
the admin version bump is weird, but i think it's fine, and we can bump the version to 3.x properly in a later commit.

darrenleeweber:
:+1: for merge

PR #471, opened 2016-01-22
fix Create Collection's duplicate default rights pulldowns

### LINKED ISSUE #436: Create Collection blue button does not work
It can't get to the form ...
screen shot 2016-01-08 at 6 23 44 pm

I tried it using the Chit Chat Club APO https://argo-dev.stanford.edu/catalog/druid:qh637kf2429

drh-stanford:
This PR fixes #436. It has a major cleanup of the register collection view and tests that the `:get` renders the form and `:post` creates the collection.


mejackreed:
Rather than adding all of the `<br>` tags, can we wrap each set of form elements in a `<div>` with the `.form-group` class?

Then all input controls can use `.form-control`. This should remove the need for some of our inline styling.

mejackreed:
An example for the first form input:
```
<div class="form-group">
  <label>
    Create a collection that can be assigned to items governed by this APO -- <%= @object.label %>
  </label>
  <input checked type="radio" class="form-control" name="collection_radio" value="create" onclick="$('.collection_div').hide();$('#create-collection').show();"/>
</div>
```

drh-stanford:
Revised styling looks like this:

screen shot 2016-01-22 at 12 39 56 pm
screen shot 2016-01-22 at 12 39 47 pm

drh-stanford:
the new modal dialogs look like this:

screen shot 2016-01-22 at 12 43 57 pm
screen shot 2016-01-22 at 12 43 46 pm

drh-stanford:
The coverage should have increased but it "decreased" by 0.0% -- lame that it fails the build...

mejackreed:
:+1:

PR #472, opened 2016-01-22
bump to version 3 beta

jmartin-sul:
simple change, as per the title


jmartin-sul:
i like how coveralls seems to flip a coin to decide whether it thinks "0.0%" is a coverage increase or decrease.  maybe over time its rounding up and down of zero ends up averaging to zero.

jmartin-sul:
which is to say, this build isn't broken

PR #474, opened 2016-01-24
Modifications to user permissions for APO access.

### LINKED ISSUE #76: Allow users with APO roles access to the APO itself

darrenleeweber:
This fixes #76.

The user permissions are now almost entirely consolidated in the User model.  The one exception is in `CatalogController#valid_user?` where the response messages provide granular information about access restrictions.

Changes to user permissions have wide-spread implications for specs, so this PR includes some modifications for more consistent behavior, including:
  - unified user mocks in spec_helper; many specs are much simpler than before
  - unified `param[:id]` parsing for `Dor.find` in `ApplicationController#find_druid`, so that any `expect` on `Dor.find` in controllers can rely on it to get a consistent DRUID/PID input string pattern and more consistent exception handling.  This also ensures that `param[:id]` is a valid DRUID/PID.


spec/spec_helper.rb
```diff
@@ -109,3 +109,68 @@ def item_from_foxml(foxml, item_class = Dor::Base, other_class = ActiveFedora::O
...
+end
+
+def admin_user
+  webauth = double(
+    'WebAuth',
+    :login => 'sunetid',
+    :attributes => {'DISPLAYNAME' => 'Admin'}
+  )
+  @admin_user = User.find_or_create_by_webauth(webauth)
+  allow(@admin_user).to receive(:groups).and_return(User::ADMIN_GROUPS)
```
mejackreed: Why not just add `groups` to the double?

spec/spec_helper.rb
```diff
@@ -109,3 +109,68 @@ def item_from_foxml(foxml, item_class = Dor::Base, other_class = ActiveFedora::O
...
+  # admin privilege supercedes all others, no need to mock anything else.
+  allow(@admin_user).to receive(:is_admin).and_return(true)
```
mejackreed: Why not just add `is_admin` to the double?

spec/spec_helper.rb
darrenleeweber: Tried that and it didn't work.

mejackreed: Do you need to actually create the user object? (if it just needs to respond to a certain method)

darrenleeweber:
This is ready for review.  I've done the best I can to quiet hound, the remaining complaints are reasonable but I think they have to be this way (unless a better way is available).

app/controllers/application_controller.rb
```diff
@@ -20,6 +20,25 @@ class ApplicationController < ActionController::Base
...
+  # @see https://github.com/sul-dlss/druid-tools
+  # @param druid [String]
+  # @return object [ActiveFedora::Base] any Dor object
+  # @raises ArgumentError
+  def find_druid(druid)
+    druid = DruidTools::Druid.new(druid).druid
+    Dor.find(druid)
+  rescue ArgumentError
+    flash[:error] = "Invalid DRUID: #{druid}"
+    render status: 400, text: flash[:error] and return
```
darrenleeweber: I think rails requires the `and` behavior for `render ... and return`, no?

app/controllers/catalog_controller.rb
```diff
@@ -345,24 +342,29 @@ def load_bulk_jobs(druid)
...
-    rescue
-      return false
-    end
-
+    # First check the workgroups defined for admin, manager or viewer; see
+    # User::ADMIN_GROUPS, User::MANAGER_GROUPS and User::VIEWER_GROUPS
+    return true if @user.is_admin || @user.is_manager || @user.is_viewer
+    # If that fails, check an APO for permitted roles of the current_user.
+    valid = false
+    @apo = dor_object.admin_policy_object rescue nil
```
darrenleeweber: Yea, yea, I know, but this is how I want this to behave and I prefer this succinct pattern.

app/controllers/application_controller.rb
```diff
@@ -20,6 +20,25 @@ class ApplicationController < ActionController::Base
...
+  rescue ActiveFedora::ObjectNotFoundError # => e
+    flash[:error] = 'Object does not exist in Fedora.'
+    render status: 404, text: flash[:error] and return
```
darrenleeweber: I think rails requires the `and` behavior for `render ... and return`, no?  This is not a boolean operation here.

spec/spec_helper.rb
```diff
@@ -72,8 +59,21 @@ def instantiate_fixture(druid, klass = ActiveFedora::Base)
   config.infer_spec_type_from_file_location!
 end
 
-def log_in_as_mock_user(subject)
-  allow(subject).to receive(:webauth).and_return(double(:webauth_user, :login => 'sunetid', :logged_in? => true))
+
```
darrenleeweber: Hound is nuts, this it outdated code.

spec/spec_helper.rb
darrenleeweber: I did try your suggestion and I don't know why it doesn't work, but it didn't.  So, I'm kinda tired of this PR after working all day on my Sat.  So my mantra now is "If it ain't broke, don't fix it".  If you got a better idea, checkout the branch and try it.  If it works and it's a much better approach, I'm all ears.  FWIW, this consolidation of user mocks into the spec helper has removed a lot of duplication and the weird behavior across feature/integration specs.  It's not applied everywhere, because some specs need very particular user mocks, but most specs just need an `admin_user` to be the `current_user`.  This serves that purpose.

atz:
Rather against the setting of special `@named_vars` inside spec helper methods.  It is unnecessary, since the author of the spec already has to know the name of the method to invoke, it should just return the mocked/stubbed/setup user and let them attach it to a variable, if they want.  As proposed, the author has to know the method names and the variable names.  Also the methods have directly conflicting effects.  It prevents their use to test impersonation, for example.  

You touch dozens of lines to convert everything through `find_druid`, intending to ease a legitimate problem of targeting expect statements (`Dor::Item.find` vs. `Dor::AdminPolicyObject.find`, etc.).  But then you end up using the new convergence only once.  

Mainly, I don't understand why this PR has a 2,000 line delta in one commit, including refactoring, new tests, presentation markup, a new 400 page template, deletion of 3 test files, config changes, lots of tidying up whitespace and somewhere in there actual core permissions-related logic.  This is not structured for logical review or acceptance, so even if I liked all the changes, I would be reluctant to merge it.

app/models/user.rb
```diff
@@ -25,80 +29,165 @@ def to_s
...
-      toret << 'dor-apo-manager' if resp[key] && (resp[key] & my_groups).length > 0
-    end
-
+    return @role_cache[druid] if @role_cache[druid]
+    # Try to retrieve a Solr doc for the DRUID
+    obj_doc = Dor::SearchService.query('id:"' + druid + '"')['response']['docs'].first || {}
+    return [] if obj_doc.empty?
+    druid_roles = Set.new
+    # Determine whether user has 'dor-apo-manager' role
+    apo_roles = %w(apo_role_group_manager_ssim apo_role_person_manager_ssim)
```
jmartin-sul: i'd be inclined to call this something like `apo_role_keys` or `apo_role_fields` to make it clear that it's the field name in a solr doc and not the name of the role itself.

app/models/user.rb
```diff
@@ -25,80 +29,165 @@ def to_s
...
+      dor-viewer
+      sdr-administrator
+      sdr-viewer
+    )
+  end
+
+  # Validate a user belongs to a workgroup allowed to access a DOR object
+  # @param dor_doc [Hash] A Solr document for a DOR object
+  # @param roles [Array<String>] Roles to be validated for the DOR object
+  def role_allowed?(solr_doc, role)
```
jmartin-sul: same sort of thing here, seems like `role_field_name` or similar would be clearer.  also, the comment about `roles` is slightly off on both name and expected type, since what's expected is a single string for use as a key (right?).

jmartin-sul: i'd also be inclined to rename `role_allowed?` to something like `has_role_in_apo?`, and i'd make the solr_doc name more specific to go along with that.  also, the comments for the current names seem a bit off, and might need to be updated from an earlier iteration of the method?

anyway, maybe something like:
```ruby
# Validate that the current User is part of the given role in the given APO
# @param apo_solr_doc [Hash] A Solr document for a DOR APO object
# @param role_def_field_name [String] Name of field containing workgroups/users for a role, for comparison to the current user's workgroups
def has_role_in_apo?(apo_solr_doc, role_def_field_name)
  # apo_solr_doc[role_def_field_name] returns an array of groups and users permitted to adopt the role
  !(apo_solr_doc[role_def_field_name] & groups).blank?
end
```

logically the same in terms of what's executed, but i'd submit that this version is a bit clearer about its intent.

app/models/user.rb
```diff
@@ -25,80 +29,165 @@ def to_s
...
+  # @return [Boolean]
+  def can_manage_object?(dor_object)
+    dor_object = parse_object(dor_object)
+    pid = dor_object.pid
+    apo_pid = dor_object.admin_policy_object.pid
+    manager_access = is_admin || is_manager
+    manager_access ||= dor_object.can_manage_item?(roles(apo_pid))
+    manager_access ||= dor_object.can_manage_content?(roles(apo_pid))
+    manager_access ||= dor_object.is_a?(Dor::AdminPolicyObject) && dor_object.can_manage_item?(roles(pid))
+    manager_access ||= dor_object.is_a?(Dor::AdminPolicyObject) && dor_object.can_manage_content?(roles(pid))
```
jmartin-sul: i think the above two lines might be a bug.

my understanding of the above is as follows:
* if the user is a repo wide admin or manager (`is_admin` or `is_manager` is true), return true.
* if the user can manage items or manage content for things governed by `apo_pid`, return true.
* this is the part that i believe is a bug:  if the user can manage items or content for things governed by `pid`, in the case that `pid` represents an APO, return true.  which is to say, if you can manage things governed by an APO, you can manage the APO itself.

if this is the case, i can see how it is analogous to what was asked for in the issue for which this PR was created:  the desired behavior was that if you can see an item, you should be able to see the APO which governs it.  more accurately relative to this implementation, if an APO might allow you to see an object, you should be able to see that APO.

however, that's quite different from saying that if an APO allows you to manage an object, you should be able to manage that APO.  i don't think that was implied in the issue description, and i don't think it's desirable behavior, but that's how this code currently reads to me.

in looking at the tests, these lines appear not to be exercised.  nothing seems to specifically test the case where `dor_object.is_a?(Dor::AdminPolicyObject)` evaluates to true.

app/models/user.rb
```diff
@@ -25,80 +29,165 @@ def to_s
...
+    dor_object = parse_object(dor_object)
+    return true if can_manage_object?(dor_object)
+    pid = dor_object.pid
+    apo_pid = dor_object.admin_policy_object.pid
+    view_access = is_viewer
+    view_access ||= dor_object.can_view_content?(roles(apo_pid))
+    view_access ||= dor_object.is_a?(Dor::AdminPolicyObject) && dor_object.can_view_content?(roles(pid))
+    view_access
+  end
+
```
jmartin-sul: in the above two methods (`can_view_object?` and `can_manage_object?`) it'd be nice if there were in-line comments explaining the particulars of the logic.  as it is, there are top-level comments that are exactly the same for both, though the two methods should do slightly different things.

app/models/user.rb
```diff
@@ -25,80 +29,165 @@ def to_s
...
+  #                                      any subclass of Dor::Base
+  # @return [Boolean]
+  def can_view_object?(dor_object)
+    dor_object = parse_object(dor_object)
+    return true if can_manage_object?(dor_object)
+    pid = dor_object.pid
+    apo_pid = dor_object.admin_policy_object.pid
+    view_access = is_viewer
+    view_access ||= dor_object.can_view_content?(roles(apo_pid))
+    view_access ||= dor_object.is_a?(Dor::AdminPolicyObject) && dor_object.can_view_content?(roles(pid))
```
jmartin-sul: it appears that this line implements the behavior of allowing a user to view an APO which grants them access to view other objects, even if they wouldn't otherwise have the right to view that governing APO.  however, this logic also appears to be implemented in `allow_view` in items_controller.rb.  why not just call `can_view_object` over there?  this doesn't actually appear to be called at all.

Gemfile
```diff
@@ -65,6 +65,7 @@ group :test, :development do
   gem 'rack-test', :require => 'rack/test'
   gem 'simplecov', :require => false
   gem 'pry-byebug'
+  gem 'pry-doc'
```
mejackreed: Maybe a different PR for adding this gem?

app/controllers/application_controller.rb
```diff
@@ -20,6 +20,25 @@ class ApplicationController < ActionController::Base
...
 
+  # A common method for controllers to find a Dor object.  It accepts any valid
+  # PID (druid:aa111aa1111) or DRUID (aa111aa1111).  DruidTools::Druid parses
+  # input and returns a uniform PID, which is input to Dor.find to obtain a Dor
+  # object.
+  # @see https://github.com/sul-dlss/druid-tools
+  # @param druid [String]
+  # @return object [ActiveFedora::Base] any Dor object
+  # @raises ArgumentError
+  def find_druid(druid)
```
mejackreed: Can this method be private?

app/models/user.rb
```diff
@@ -25,80 +29,165 @@ def to_s
...
+
+  private
+
+  # Find a Dor object, whether the input is a DRUID, a PID or a Dor object.
+  # @param druid [String|ActiveFedora::Base]
+  #              DruidTools::Druid is used to parse String, which is
+  #              then input for Dor.find to obtain a Dor object
+  #              Dor objects inherit from ActiveFedora::Base
+  # @raises ArgumentError
+  def parse_object(druid)
```
jmartin-sul: what about something like:
`def get_dor_object_object(obj)`

since you're not analyzing something that's definitely an object, `parse_object` doesn't seem quite right.  ditto for `druid`, which may be either a druid or a Dor object.

also:  i actually think this method is unnecessary, and would advocate for its removal.  `parse_object` is called by two methods:  `can_view_object?` and `can_manage_object?`.  `can_view_object?` doesn't appear to be called anywhere besides the tests.  `can_manage_object?` gets called in three places (in `allow_modify` and `allow_view` in items_controller, and in `render_buttons` in argo_helper).  in all three cases, it appears that it already gets what we'd expect to be a Dor object.  if there's no real reason to believe that it might be valid to get a druid, i'd prefer to remove the extra indirection.  i can see how this might be a useful thing to have in some cases, but it doesn't seem to actually be used at present, and it's straight-forward enough that it should be easy to write again later if it is needed.

app/controllers/application_controller.rb
jmartin-sul: while i think hound's complaint is indeed off the mark, i'm actually concerned about the use of `flash` and `render` at all in this method.

`find_druid` gets called from `parse_object`, which gets called from `can_manage_object?`.  `can_manage_object?` is a method on the User model, and as such, it seems wrong for it to do controllerish things like rendering.  more practically, and less philosophically, `User#can_manage_object?` is the sort of thing you might want to call from the console, or a delayed job, possibly indirectly, and there's nothing to indicate that calling it outside the context of a web request is inappropriate.

it's true that it's mostly called directly from controllers, but maybe there should be a smaller method to do the druid tools thing, and a wrapper method that does error handling for use by the controllers that call `find_druid` directly.  though i don't think this extra level of in-line error handling is actually necessary either.  i think it'd be preferable to just let `ArgumentError` or `ObjectNotFoundError` bubble up and get handled by Rails.  maybe `ArgumentError` could get handled in application_controller in the same fashion that application_controller currently handles `ObjectNotFoundError`.  something like:
```ruby
rescue_from ArgumentError, with: -> { render text: 'Bad Request.  Please check the url and post data for syntax errors.  For item requests, ensure that a valid DRUID is specified.', status: 400 }
```

darrenleeweber: All good points, but the actual resolution to this is unclear.  The evolution of this method was (a) a simple wrapper on `Dor.find` to use DruidTools and then, due to specs failing, (b) a method with consistent exception handling that is specific to any request to get a DOR object.

The separation of controller and model methods makes sense.  Maybe the DruidTools parameter parsing belongs in dor-services within the Dor.find method?

app/models/user.rb
darrenleeweber: Next commit will use a `solr` prefix for these, e.g. `solr_apo_roles` and `solr_role_allowed?` (to address comment below).

app/models/user.rb
darrenleeweber: The data in my `coverage/index.html` file indicates these lines are tested already.  What makes you think they are not tested or how can they be tested better?  I seem to recall the specs did hit these lines of code because there was a typo in one line (it was calling `can_manage_content` instead of `can_manage_content?` on the last line).

app/models/user.rb
darrenleeweber: Additional comments will be added in the next commit (although they may not really add much beyond reading the code itself).

app/models/user.rb
darrenleeweber: That's what this bug is about and this is a method to help fix it.  This method complements the `can_manage_object` method.  The reason it is not called in ItemController is vexing, because the specs require the item controller to return specific status codes and messages when this authorization fails.  As much as I hate it, the specs require a more detailed, specific response from the item controller.  If those specs are wrong or should be modified, then item controller could call this method (and I think that would be preferable).

Gemfile
darrenleeweber: Too trivial for an additional PR, unless we really need to discuss it and review it etc?

app/controllers/application_controller.rb
darrenleeweber: This method is vexing and requires more discussion.  It has evoled this way to provide (a) consistent behavior and (b) satisfy existing specs.  My gut feeling is that dor-services could use DruidTools in the Dor.find method and that would help to resolve this, but it wouldn't resolve the consistent handling of the exceptions.

app/models/user.rb
darrenleeweber: OK, `get_dor_object` it is.  There are two methods that require the same code to determine if the argument is a string (and what kind of string) or a dor object and there are places where either can be given in the specs.  Removing the call to `find_druid` and using the code there directly (despite the duplication, to maintain separation of model and controller and avoid the complications of that method).

darrenleeweber:
Comments to address concerns for @atz above, with quotes from above here for easier reading.

@atz says, "Rather against the setting of special @named_vars inside spec helper methods. It is unnecessary, since the author of the spec already has to know the name of the method to invoke, it should just return the mocked/stubbed/setup user and let them attach it to a variable, if they want. As proposed, the author has to know the method names and the variable names."

All the User spec_helper methods now return a User object.  The existing specs that refer to it have it assigned to `@current_user`.

@atz says, "Also the methods have directly conflicting effects. It prevents their use to test impersonation, for example."

Please explain, this is not clear to me.  These are convenience methods.  If a spec requires very specific User manipulation, they may not use these methods to get a user.  See the argo_helper_spec, for example.

@atz says, "You touch dozens of lines to convert everything through find_druid, intending to ease a legitimate problem of targeting expect statements (Dor::Item.find vs. Dor::AdminPolicyObject.find, etc.). But then you end up using the new convergence only once."

There are multiple calls to this method, i.e.
```
$ ack-grep 'find_druid' app/
app/controllers/items_controller.rb
673:    reindex find_druid(params[:id])
713:    @object = find_druid(params[:id])

app/controllers/application_controller.rb
31:  def find_druid(druid)

app/controllers/apo_controller.rb
300:    @object = find_druid(params[:id])

app/controllers/status_controller.rb
14:    test_item = find_druid(params[:test_obj])

app/controllers/catalog_controller.rb
183:    @obj = find_druid(params[:id])
206:    @obj = find_druid(params[:id])
211:    @apo = find_druid(params[:id])
227:    @obj = find_druid(params[:id])
279:    @obj ||= find_druid(params[:id])
```

@atz says, "Mainly, I don't understand why this PR has a 2,000 line delta in one commit, including refactoring, new tests, presentation markup, a new 400 page template, deletion of 3 test files, config changes, lots of tidying up whitespace and somewhere in there actual core permissions-related logic. This is not structured for logical review or acceptance, so even if I liked all the changes, I would be reluctant to merge it."

The modification to User permissions had consequences beyond the User model.  The most difficult part of this PR was fixing all the specs that were broken for code that was not immediately relevant to this PR, on the face of it.  But, to get all the checks passing, they had to be fixed.

The extent of the changes is also partly to appease the numerous hound.ci complaints and to improve some code coverage.  As DLSS best practices warrant following the advice, I did.  As they were specific to this PR, they got addressed in this PR.

Gemfile
mejackreed: How does it relate to the rest of the commit? If I was coming across this commit later, it may be a red herring to see this gem change with the rest of the changes. That's why I would put this with another PR, unless it has direct relevance to the commit?

spec/integration/report_spec.rb
```diff
@@ -1,29 +0,0 @@
...
-  describe 'bulk' do
-    it 'should return a page with the expected elements' do
-      visit '/report/bulk'
-      expect(page).to have_content('Bulk update operations')
-      expect(page).to have_css('.bulk_button', text: 'Get druids from search')
-      expect(page).to have_css('.bulk_button', text: 'Paste a druid list')
-      # expect(page).to have_css('.bulk_button', text: 'Reindex')
-    end
-  end
-end
```
darrenleeweber: Moved to feature specs because it is `:type => :feature` and not `:request` as the other integration specs are.

spec/integration/item_displays_spec.rb
```diff
@@ -2,23 +2,11 @@
...
-
-    ##
-    # Stubbing of both ItemsController and CatalogController which are the
-    # controllers used in this spec, needing a valid return from current_user. A
-    # higher level up the inheritance chain stub of `ApplicationController` is
-    # insufficient here, due to a possible bug in rspec-mocks or our overuse of
-    # `allow_any_instance_of`.
-    allow_any_instance_of(ItemsController).to receive(:current_user).and_return(@current_user)
-    allow_any_instance_of(CatalogController).to receive(:current_user).and_return(@current_user)
   end
```
darrenleeweber: Most of this `before` code is handled in the `admin_user` helper method.  There is no point in defining `is_manager`, `roles` or `permitted_apos` because the `is_admin` value overrides all other privileges, which can be clearly seen in the User model.

spec/integration/apo_register_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      expect(find_field('copyright').value).to eq('New copyright statement')
+      expect(find_field('use').value).to eq('New use statement')
+      expect(find_field('managers').value).to eq('dlss:dpg-staff')
+      expect(find_field('viewers').value).to eq('sunetid:anyone')
+      expect(find_field('desc_md').value).to eq('MODS')
+      expect(find_field('use_license').value).to eq('by-nd')
+      expect(page).to have_no_field('collection')
+    end
+  end # with admin user
+end
```
darrenleeweber: renamed the spec, it was just `apo_spec` (see below).

spec/integration/about_page_spec.rb
```diff
@@ -2,11 +2,7 @@
 
 describe 'about_page', :type => :request do
   before :each do
-    @current_user = double(:webauth_user, :login => 'sunetid', :logged_in? => true, :privgroup => ADMIN_GROUPS.first)
-    allow(@current_user).to receive(:is_admin).and_return(true)
-    allow(@current_user).to receive(:roles).and_return([])
-    allow(@current_user).to receive(:is_manager).and_return(false)
-    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(@current_user)
+    admin_user # see spec_helper
```
darrenleeweber: All of this before code is handled in the `admin_user` helper method. There is no point in defining is_manager, roles etc. because the is_admin value overrides all other privileges, which can be clearly seen in the User model.

spec/helpers/argo_helper_spec.rb
```diff
@@ -24,25 +24,27 @@
...
       allow(Dor).to receive(:find).with(@item_id).and_return(@object)
+      allow(Dor::WorkflowService).to receive(:get_active_lifecycle).and_return(true)
+      allow(Dor::WorkflowService).to receive(:get_lifecycle).and_return(true)
+      # User mocks
+      @usr = double()
+      allow(@usr).to receive(:is_admin).and_return(true)
+      allow(@usr).to receive(:groups).and_return([])
+      allow(@usr).to receive(:is_manager).and_return(false)
+      allow(@usr).to receive(:roles).with(@apo_id).and_return([])
+      allow(helper).to receive(:current_user).and_return(@usr)
```
darrenleeweber: Simply rearranged the code to clarify what is going on here.  Although some of this might be replaced with an `admin_user` from spec_helper, I did not want to mess with this spec too much.  I get the impression a lot of this might have been done before `instantiate_fixture` arrived on the scene, but I didn't want to change it.

spec/features/workflow_service_creation_spec.rb
```diff
@@ -1,25 +1,22 @@
...
+
+  scenario 'redirect and display on show page - with JS', js: true do
+    visit add_workflow_item_path(pid)
+    expect(page).to have_content('Add workflow')
+    expect(page).to have_button('Add')
+    find('#wf').find(:option, 'accessionWF').select_option
+    find('#add_wf_button').trigger('click')
     within '.flash_messages' do
       expect(page).to have_css '.alert.alert-info', text: 'Added accessionWF'
     end
```
darrenleeweber: This spec was giving weird results with or without JS.  Through trial-n-error, the resulting spec passed and consistently passes now.  It was a revelation to realize that it did not require an item mock, because the fedora-test already contains the item.

spec/features/view_switcher_spec.rb
```diff
@@ -1,55 +1,40 @@
...
         expect(page).to have_css 'li.active a', text: 'Discovery Report View'
       end
     end
+
     scenario 'workflow grid' do
-      expect_any_instance_of(ReportController).to receive(:current_user)
-        .at_least(1).times.and_return(current_user)
       visit report_workflow_grid_url f: { objectType_ssim: ['item'] }
       within '.report-toggle' do
         expect(page).to have_css 'li.active a', text: 'Workflow Grid View'
```
darrenleeweber: All these `current_user` specs appear to be spurious to the focus of this view spec.  All it requires is a vanilla admin user and it should not be putting any expectations on it.  These `expect_any_instance` are a reaction to potential problems with `allow_any_instance` and we don't really understand whether this `expect` pattern is better.  Given weird failures in the feature specs, I found it was better to remove all these and use just the admin_user from the spec helper for more consistent behavior.

spec/features/search_results_spec.rb
```diff
@@ -1,21 +1,10 @@
...
-      permitted_apos: [],
-      is_admin: true,
-      roles: [],
-      groups: []
-    )
-  end
   before :each do
-    allow_any_instance_of(CatalogController).to receive(:current_user).
-      and_return(current_user)
+    admin_user # see spec_helper
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/report_view_spec.rb
```diff
@@ -1,23 +1,22 @@
...
-      is_admin: true,
-      roles: [],
-      groups: []
-    )
-  end
   before :each do
-    allow_any_instance_of(ReportController).to receive(:current_user).
-      and_return(current_user)
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/item_view_metadata_spec.rb
```diff
@@ -1,20 +1,8 @@
...
-      roles: [],
-      groups: []
-    )
-  end
-  before do
-    expect_any_instance_of(CatalogController).to receive(:current_user)
-      .at_least(1).times.and_return(current_user)
+  before :each do
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/item_source_id_change_spec.rb
```diff
@@ -1,21 +1,10 @@
...
-      groups: []
-    )
-  end
-  before do
-    allow_any_instance_of(ItemsController).to receive(:current_user)
-      .and_return(current_user)
+feature 'Item source id change' do
+  before :each do
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/item_source_id_change_spec.rb
```diff
@@ -27,10 +16,9 @@
         'its current state.'
     end
   end
+
   feature 'when modification is allowed' do
     scenario 'changes the source id' do
-      expect_any_instance_of(CatalogController).to receive(:current_user)
-        .at_least(1).times.and_return(current_user)
       expect_any_instance_of(Dor::Item)
```
darrenleeweber: This is taken care of in the `admin_user`.

spec/features/item_registration_spec.rb
```diff
@@ -1,21 +1,10 @@
...
-      roles: [],
-      groups: []
-    )
-  end
-  before do
-    expect_any_instance_of(ItemsController).to receive(:current_user)
-      .at_least(7).times.and_return(current_user)
+  before :each do
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/indexer_backlog_spec.rb
```diff
@@ -2,17 +2,9 @@
...
-      login: 'sunetid',
-      logged_in?: true,
-      permitted_apos: [],
-      is_admin: true,
-      can_view_something?: true
-    )
-    allow_any_instance_of(ApplicationController).to receive(:current_user).
-      and_return(@current_user)
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/full_width_spec.rb
```diff
@@ -2,17 +2,9 @@
...
-      logged_in?: true,
-      permitted_apos: [],
-      is_admin: true,
-      can_view_something?: true
-    )
-    allow_any_instance_of(ApplicationController).to receive(:current_user).
-      and_return(@current_user)
+    admin_user # see spec_helper
   end
+
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/enable_buttons_spec.rb
```diff
@@ -1,25 +1,10 @@
...
-      admin_policy_object: false,
-      datastreams: {},
-      can_manage_item?: true
-    )
-    allow_any_instance_of(ApplicationController).to receive(:current_user).
-      and_return(@current_user)
-    allow(Dor).to receive(:find).and_return(@obj)
+  before :each do
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/apo_view_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    it 'allows a user of an authorized workgroup to view an APO' do
+      # Ensure all user authority methods return false and the
+      # user belongs to at least one authorized workgroup.
+      dor_viewer_roles = %w(dor-viewer sdr-viewer)
+      dor_viewer_roles.each do |role|
+        allow(@view_user).to receive(:roles).and_return([role])
+        allow(item).to receive(:can_view_content?).with([role]).and_return(true)
+        visit "/catalog/#{druid}"
```
mejackreed: Can we use the named path here?

spec/features/enable_buttons_spec.rb
```diff
@@ -1,25 +1,12 @@
...
+
```
darrenleeweber: Use a vanilla admin user and a regular fixture (a Dor fixture will throw a spanner at `User#get_dor_object`).

spec/features/date_range_form_spec.rb
```diff
@@ -1,22 +1,14 @@
...
-      can_view_something?: true
-    )
-    allow_any_instance_of(ApplicationController).to receive(:current_user).
-      and_return(@current_user)
+feature 'Date range form', js: true do
+  before :each do
+    admin_user # see spec_helper
     visit root_path
     find('[data-target="#facet-object_modified_date"]').click
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/consistent_titles_spec.rb
```diff
@@ -1,26 +1,17 @@
...
-      roles: [],
-      groups: []
-    )
-  end
-  before do
-    allow_any_instance_of(ApplicationController).to receive(:current_user)
-      .and_return(current_user)
+  before :each do
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/bulk_screen_spec.rb
```diff
@@ -3,16 +3,7 @@
...
-      login: 'sunetid',
-      logged_in?: true,
-      permitted_apos: [],
-      is_admin: true,
-      permitted_collections: [["string"]]
-    )
-
-    expect_any_instance_of(ApplicationController).to receive(:current_user).at_least(:once).and_return(@current_user)
+    admin_user # see spec_helper
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/bulk_jobs_spec.rb
```diff
@@ -3,31 +3,25 @@
...
-      permitted_apos: [],
-      is_admin: true
-    )
-    allow(@current_user).to receive(:roles).and_return([])  # necessary to determine if current user has right permissions
-    allow_any_instance_of(ApplicationController).to receive(:current_user).
-                                                    and_return(@current_user)
+    admin_user # see spec_helper
+    @druid = 'hv992ry2431'  # a Dor object fixture
+    @pid = DruidTools::Druid.new(@druid).druid
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/features/apo_view_spec.rb
darrenleeweber: It would be more helpful if you include the named path in the comment.  I think you mean
`visit catalog_path druid` ?

spec/controllers/status_controller_spec.rb
```diff
@@ -61,7 +68,7 @@
     end
     describe 'with reindexing' do
       before :each do
-        expect(Dor::Item).to receive(:find).with(@druid).and_return(@item)
+        expect(Dor).to receive(:find).with(@druid).and_return(@item)
   #     expect(Dor::Item).to receive(:find).with(@druid).and_call_original
         expect(@item).to receive(:save)
       end
```
darrenleeweber: This spec was driving me nuts with inconsistent behavior and a failure in Dor.find to handle an unusual Fedora exception.  The revisions to `find_druid` that intercept anything that's not a valid DRUID before calling Dor.find have resolved the spec failures on this one.

spec/controllers/registration_controller_spec.rb
```diff
@@ -2,10 +2,8 @@
 
 describe RegistrationController, :type => :controller do
   before :each do
+    @current_user = admin_user # see spec_helper
     @item = double(Dor::Item)
-    @current_user = double(:webauth_user, :login => 'sunetid', :logged_in? => true, :privgroup => ADMIN_GROUPS.first)
-    allow(@current_user).to receive(:is_admin).and_return(true)
-    allow(controller).to receive(:current_user).and_return(@current_user)
     allow(Dor::Item).to receive(:find).and_return(@item)
   end
```
darrenleeweber: Another case for a vanilla admin user.

spec/controllers/items_controller_spec.rb
```diff
@@ -331,15 +299,16 @@
...
-      allow(@item).to receive(:datastreams).and_return({'contentMetadata' => double(Dor::ContentMetadataDS)})
-      get 'resource', :id => @pid, :resource => '0001'
-      # XXX : isn't actually testing what it says!
-    end
+    it 'should set the object and datastream, then call the view'
+    #   expect(Dor::Item).to receive(:find)
+    #   allow(@item).to receive(:datastreams).and_return({'contentMetadata' => double(Dor::ContentMetadataDS)})
+    #   get 'resource', :id => @pid, :resource => '0001'
+    #   # XXX : isn't actually testing what it says!
+    # end
```
darrenleeweber: Well if ` # XXX : isn't actually testing what it says!` is true, make it pending!

spec/controllers/items_controller_spec.rb
```diff
@@ -1,60 +1,30 @@
...
-    allow(@item).to receive(:workflows).and_return(wf)
-    allow(Dor::SearchService.solr).to receive(:add)
+    @current_user = admin_user # see spec_helper
+    # Item
+    @druid = 'rn653dy9317'
+    @pid = DruidTools::Druid.new(@druid).druid
+    @item = instantiate_fixture(@druid, Dor::Item)
+    allow(Dor).to receive(:find).with(@pid).and_return(@item)
+    @apo = @item.admin_policy_object
   end
```
darrenleeweber: Excessive mocking.  Replaced with a fixture.

public/400.html
```diff
@@ -0,0 +1,27 @@
...
+
+<body>
+  <!-- This file lives in public/400.html -->
+  <div class="dialog">
+    <h1>Your request could not be handled.</h1>
+    <p>Please check the url and post data for syntax errors.</p>
+    <p>For item requests, ensure that a valid DRUID is specified.</p>
+  </div>
+</body>
+</html>
```
darrenleeweber: This was included in case the app needs to use one.  At present, it uses flash messages instead.  No harm in leaving it in, along with other public pages that might not be used anyway.

lib/tasks/argo.rake
```diff
@@ -293,13 +293,13 @@ namespace :argo do
...
-      priv_groups.uniq! # no need to repeat ourselves (mostly there in case the builtin groups are already listed in APOs)
-
-      directives += priv_groups.collect { |v|
+      # we always want these built-in groups to be part of .htaccess
+      priv_groups += User::ADMIN_GROUPS
+      priv_groups += User::VIEWER_GROUPS
+      priv_groups += User::MANAGER_GROUPS
+      directives += priv_groups.uniq.map { |v|
         ["Require privgroup #{v.split(/:/, 2).last}", "WebAuthLdapPrivgroup #{v.split(/:/, 2).last}"]
       }.flatten
```
darrenleeweber: We consolidated the config/initialize/permissions.rb into the User.

config/initializers/permissions.rb
```diff
@@ -1,3 +0,0 @@
-ADMIN_GROUPS = ['workgroup:sdr:administrator-role', 'workgroup:dlss:dor-admin']
-VIEWER_GROUPS = ['workgroup:sdr:viewer-role', 'workgroup:dlss:dor-viewer']
-MANAGER_GROUPS = ['workgroup:sdr:manager-role', 'workgroup:dlss:dor-manager']
```
darrenleeweber: All moved into User model.

app/controllers/items_controller.rb
```diff
@@ -4,11 +4,40 @@ class ItemsController < ApplicationController
   include ItemsHelper
   include DorObjectHelper
   include ModsDisplay::ControllerExtension
-  before_filter :create_obj, :except => [:register, :open_bulk, :purge_object]
-  before_filter :forbid_modify, :only => [:add_collection, :set_collection, :remove_collection, :update_rights, :set_content_type, :tags, :tags_bulk, :source_id, :delete_file, :close_version, :open_version, :resource, :add_file, :replace_file, :update_attributes, :update_resource, :mods, :datastream_update ]
-  before_filter :forbid_view,   :only => [:get_preserved_file, :get_file]
-  before_filter :enforce_versioning, :only => [:add_collection, :set_collection, :remove_collection, :update_rights, :tags, :source_id, :set_source_id, :set_content_type, :set_rights]
-  after_filter  :save_and_reindex,   :only => [:add_collection, :set_collection, :remove_collection, :open_version, :close_version, :tags, :tags_bulk, :source_id, :set_rights, :set_content_type, :apply_apo_defaults]
+  before_filter :create_obj, :except => [
```
darrenleeweber: Lots of hound.ci complaints about long lines in this controller, so some short util methods were created and various long lines shortened to make for more readable code.

app/controllers/items_controller.rb
```diff
@@ -585,38 +659,64 @@ def apply_apo_defaults
...
     if params[:bulk]
-      render :text => "Added #{params[:wf]}"
+      render :text => msg
     else
-      redirect_to catalog_path(params[:id]), :notice => "Added #{params[:wf]}"
+      redirect_to catalog_path(params[:id]), notice: msg
     end
   end
 
   private
```
darrenleeweber: Only reordered some methods into alphabetical order and changed before filter permission method names to the positive sense (like forbid_modify to allow_modify).

app/controllers/items_controller.rb
```diff
@@ -585,38 +659,59 @@ def apply_apo_defaults
...
+    render :status => :forbidden, :text => 'forbidden'
+    false
+  end
+
+  def allow_view
+    return true if current_user.can_view_object?(@object)
+    render :status => :forbidden, :text => 'forbidden'
+    false
+  end
+
```
darrenleeweber: The `allow_modify` and `allow_view` methods now make use of the methods consolidated in the User model.

app/controllers/items_controller.rb
```diff
@@ -585,38 +659,59 @@ def apply_apo_defaults
...
 
-  # Filters
   def create_obj
-    raise 'missing druid' unless params[:id]
-    @object = Dor::Item.find params[:id]
+    @object = find_druid(params[:id])
+    return if @object.nil?
     @apo = @object.admin_policy_object
-    @apo = ( @apo ? @apo.pid : '' )
+  end
```
darrenleeweber: Use the `find_druid` method for consistency across controllers.

app/controllers/items_controller.rb
```diff
@@ -624,23 +719,22 @@ def save_and_reindex
...
-    render status: :forbidden, text: 'Object cannot be modified in its current state.'
-    false
+  def dor_lifecycle(object, stage)
+    Dor::WorkflowService.get_lifecycle('dor', object.pid, stage)
+  end
+
+  def set_dor_accession_status(object, task, status)
+    Dor::WorkflowService.update_workflow_status('dor', object.pid, 'accessionWF', task, status)
   end
 end
```
darrenleeweber: Various utils to simplify the calls to `Dor::WorkflowService` and shorten some long lines in this controller.

app/controllers/catalog_controller.rb
```diff
@@ -345,24 +342,26 @@ def load_bulk_jobs(druid)
...
+      msg = 'Item has no APO that allows access'
+    end
+    unless valid
+      render :status => :forbidden, :text => msg
     end
-    true
+    valid
   end
 
   def get_leafdir(directory)
```
darrenleeweber: The `valid_user?` method could be replaced with calls to `User#can_manage_object?` or `User#can_view_object?` but the specs require the 403 (forbidden) response to be accompanied by a text message (`msg` here).  So, not sure how to do that without duplicating code between `User` and this method.

app/controllers/catalog_controller.rb
```diff
@@ -278,8 +276,7 @@ def bulk_jobs_delete
   private
 
   def show_aspect
-    pid = params[:id].include?('druid') ? params[:id] : "druid:#{params[:id]}"
-    @obj ||= Dor.find(pid)
+    @obj ||= find_druid(params[:id])
     @response, @document = get_solr_response_for_doc_id pid
   end
```
darrenleeweber: Several places where `ApplicationController#find_druid` can be called to validate `params[:id]`.

app/controllers/apo_controller.rb
```diff
@@ -3,10 +3,25 @@
 
 class ApoController < ApplicationController
```
darrenleeweber: Mostly changes to quite hound.ci long line complaints.

app/controllers/apo_controller.rb
```diff
@@ -3,10 +3,25 @@
...
 
-  before_filter :create_obj, :except => [:register, :is_valid_role_list_endpoint, :spreadsheet_template]
-  after_action :save_and_index, :only => [:delete_collection, :delete_collection, :add_collection, :update_title, :update_creative_commons, :update_use, :update_copyright, :update_default_object_rights, :add_roleplayer, :update_desc_metadata, :delete_role, :register_collection]
```
darrenleeweber: Note that `:delete_collection` was specified twice.

app/controllers/apo_controller.rb
```diff
@@ -3,10 +3,25 @@
...
+    :update_default_object_rights, :update_desc_metadata,
+    :update_title, :update_use,
+    :delete_role,
+    :register_collection
+  ]
+
+  # These manager workgroups are specific to APO permissions and they are
+  # distinct from the repository-wide permissions in User.MANAGER_GROUPS
+  # This is used in app/views/apo/register.html.erb
+  DEFAULT_MANAGER_WORKGROUPS = %w(sdr:developer sdr:service-manager sdr:metadata-staff).freeze
```
darrenleeweber: Hound.ci recommended the `%w()` style and the `.freeze` for a constant with a mutable object.

Gemfile.lock
```diff
@@ -575,6 +579,7 @@ DEPENDENCIES
   prawn-table
   progressbar
   pry-byebug
+  pry-doc
   pry-rails
   pry-remote
   rack-test
```
darrenleeweber: I like to use `show-doc` when debugging with pry.

Gemfile
darrenleeweber: Well I guess I just haven't got the right balance between separation of concerns and `git rebase` to consolidate or simplify the history.  There's a ton of commits that have been consolidated into this PR.
Anyhow, I get the point and I'll keep it in mind for future PR.  I have bigger fish to fry rather than figure out how to fix this one for this PR.

app/controllers/application_controller.rb
jmartin-sul: All understandable reasons for it to get to the state it's in right now.

But all I'm saying is, I don't think there's a need to do error handling at all in this method.  I'd propose just removing both of the `rescue` statements and otherwise leaving things as they are.

My justification is that I think there are two situations where it's right to trap an exception:
1) You can recover from it.  Sometimes people see that as using exceptions for flow control, and some people don't like that.  I don't particularly mind it if it's not being used as an ersatz `GOTO`, but that's not what's happening here anyway.
2) You can log something in greater detail or show a better error message, then you still propagate the error up the stack.  But I don't think the error messages here are enough of an improvement over the default Rails error page to justify having controller behavior in a method that shouldn't be restricted to being called from controllers.  `ApplicationController` already recognizes `ActiveFedora::ObjectNotFoundError` and displays a special error message in that case, so handling it above seems redundant.  And you could do the same for `ArgumentError`, as suggested in my prior comment (essentially using the same error message as in your 400 template).

With that approach, you'd still be presenting specific error messages for those conditions, and you'd do it without repeating the error handling code throughout the individual controllers, and you'd do it without including error handling that assumes web presentation in code that might be used in a wider variety of situations.  That seems like an all around improvement to me.

Additionally, the methods that call `find_druid` are somewhat inconsistent as to whether they check the result and return if `find_druid` returns `nil`.  If `find_druid` just exited on the spot with the exception it ran into, which isn't recovered from anyway, that check would become unnecessary and could be removed (because that code wouldn't be reached in the case of an unretrievable druid).

Maybe the `DruidTools` thing should be in dor-services.  I'd be fine with it staying here, I'd be fine with having a more robust `find` method there.  I do agree with your concern that using `DruidTools` in `Dor.find` might be too presumptuous.  But regardless of what happens with the DruidTools parsing, I'm very much in favor of pulling the error handling from this method, even if no other changes were made, since i don't that removal would result in a real functional deficiency.

Gemfile
jmartin-sul: @mejackreed's worry about future devs figuring out what relation this has to the commit might be assuaged somewhat if the commit message had more explanation?

i'm a big fan of good commit messages, where there's a concise subject line, and the body has a good detailed summary of the changes (but not verbatim repetition of the diff, in the same way that good code comments should be more of an overview than `a += 2   # adds 2 to a`).

@darrenleeweber, you've left a bunch of helpful explanatory notes as comments on this PR, which i think is much appreciated, but it'd be helpful to future users of the repo (who might be slogging through commits to investigate a bug) if similar helpful info was in the commit message itself.  though it can be a pain, i think it's worth spending some time crafting good commit messages, especially for a change as big as this.  and i've found that it can help me understand the changes i made if i'm forced to summarize them, in the same way that i find it good to be forced to write a quality explanatory comment about a method.

app/controllers/application_controller.rb
darrenleeweber: I suggest we merge this, as is, and create a new issue/PR to resolve this debate.  The existing specs required flash messages and I don't know why and I don't think this PR is the place to change that.  So I think the current code satisfies the current specs and your suggestions would break them.  That is not to say that an alternative is better or worse, but simply that it's out of scope here and the current code was crafted to satisfy current specs with robust and reliable behavior.

app/controllers/application_controller.rb
darrenleeweber: @mejackreed -- Now it's protected.

app/models/user.rb
```diff
@@ -25,80 +29,178 @@ def to_s
...
+      dor-apo-depositor
+      dor-apo-manager
+      dor-apo-metadata
+      dor-apo-reviewer
+      dor-apo-viewer
+      dor-viewer
+      sdr-administrator
+      sdr-viewer
+    )
+  end
```
darrenleeweber: These known_roles may need to be revised to be consistent with comments in
https://github.com/sul-dlss/argo/issues/76#issuecomment-171129115

app/controllers/application_controller.rb
```diff
@@ -7,7 +7,13 @@ class ApplicationController < ActionController::Base
...
   before_filter :fedora_setup
 
-  rescue_from ActiveFedora::ObjectNotFoundError, with: -> { render text: 'Object Not Found', status: :not_found }
+  rescue_from ActiveFedora::ObjectNotFoundError do
+    render text: "Object Not Found", status: :not_found
+  end
+
+  rescue_from Rubydora::FedoraInvalidRequest do
+    render text: "Fedora Rejected Object Request", status: :bad_request
+  end
```
darrenleeweber: This exception is incurred by the status_controller_spec and raised by Dor.find; it's not handled anywhere else.

app/controllers/application_controller.rb
```diff
@@ -49,6 +55,28 @@ def default_html_head
...
+  rescue ArgumentError
+    # Allow Argo to find anything using Dor.find,
+    # in case there is bad IDs that need to be remediated.
+    # The Dor.find method should be read-only access.
+    msg = "Trying to find an invalid DRUID: #{druid}"
+    logger.warn msg
+    flash[:alert] = msg
+    Dor.find(druid)
+  end
+
```
darrenleeweber: This method is now protected and it allows Dor.find requests for invalid DRUIDs in case they need to be remediated.

app/models/user.rb
```diff
@@ -2,6 +2,11 @@ class User < ActiveRecord::Base
   # Connects this user object to Blacklights Bookmarks and Folders.
   include Blacklight::User
 
+  # The code base requires Arrays for these constants, for intersection ops.
+  ADMIN_GROUPS = %w(workgroup:sdr:administrator-role).freeze
+  MANAGER_GROUPS = %w(workgroup:sdr:manager-role).freeze
+  VIEWER_GROUPS = %w(workgroup:sdr:viewer-role).freeze
+
```
darrenleeweber: Deprecated all the `workgroup:dlss:dor-*` groups, per Lynn's requirements.

app/models/user.rb
```diff
@@ -25,80 +30,176 @@ def to_s
...
+      dor-apo-creator
+      dor-apo-depositor
+      dor-apo-manager
+      dor-apo-metadata
+      dor-apo-reviewer
+      dor-apo-viewer
+      sdr-administrator
+      sdr-viewer
+    )
+  end
```
darrenleeweber: Deprecated the `dor-administrator` and  `dor-viewer` roles, per Lynn's requirements.  They are superceded by equivalent  `sdr-administrator` and  `sdr-viewer` roles.

spec/spec_helper.rb
darrenleeweber: hound.ci is somehow out of sync with current code here.

app/models/user.rb
```diff
@@ -25,80 +30,184 @@ def to_s
...
+  # APO role permissions.  If the object is an APO, first check permissions on
+  # the governing APO of this APO object. As a last resort, check the object
+  # roles when this object is an APO.
+  #
+  # At present, there are a few methods that all do the same thing;
+  # see all the `can_manage_*` methods in Dor::Governable, i.e.
+  # https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/models/governable.rb
+  #
+  # The upshot of this is that it’s not yet clear how Argo should query a Dor
+  # object about management rights. That is, it’s not clear how granular the
```
darrenleeweber: hound.ci is out of sync with current code.

app/models/user.rb
```diff
@@ -25,80 +30,184 @@ def to_s
...
+  # Validate access permissions using Argo user roles. If that fails, check the
+  # APO role permissions.  If the object is an APO, first check permissions on
+  # the governing APO of this APO object. As a last resort, check the object
+  # roles when this object is an APO.
+  #
+  # At present, there are a few methods that all do the same thing;
+  # see all the `can_manage_*` methods in Dor::Governable, i.e.
+  # https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/models/governable.rb
+  #
+  # The upshot of this is that it’s not yet clear how Argo should query a Dor
```
darrenleeweber: hound.ci is out of sync with current code.

app/controllers/application_controller.rb
```diff
@@ -49,6 +65,31 @@ def default_html_head
...
+  rescue ArgumentError
+    # Allow Argo to find anything using Dor.find,
+    # in case there is bad IDs that need to be remediated.
+    # The Dor.find method should be read-only access.
+    msg = "Trying to find an invalid DRUID: #{druid}\n"
+    msg += params.inspect
+    logger.warn msg
+    flash[:alert] = msg
+    Dor.find(druid, options)
+  end
```
jmartin-sul: i'm :ok_hand: on this version of `find_druid`

app/controllers/application_controller.rb
```diff
@@ -7,7 +7,23 @@ class ApplicationController < ActionController::Base
...
+  rescue_from ActiveFedora::ObjectNotFoundError do
+    msg = "Object Not Found: #{params}"
+    logger.error msg
+    render text: msg, status: :not_found
+  end
+
+  rescue_from Rubydora::FedoraInvalidRequest do
+    msg = "Fedora Rejected Object Request: #{params}"
+    logger.error msg
+    render text: msg, status: :not_found
```
jmartin-sul: this seems more like a 400 to me:
"The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing)."
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error
https://tools.ietf.org/html/rfc7231#section-6.5.1

app/controllers/apo_controller.rb
```diff
@@ -268,16 +286,21 @@ def update_desc_metadata
...
+    )
   end
 
   private
 
   def create_obj
     raise 'missing druid' unless params[:id]
-    @object = Dor.find params[:id] # , :lightweight => true
+    @object = find_druid(params[:id])
+    return if @object.nil?
```
jmartin-sul: now that `find_druid` would throw an `ObjectNotFoundError` if the requested object isn't found, i think these `return if @object.nil` lines are no longer necessary.

app/controllers/catalog_controller.rb
```diff
@@ -180,11 +180,11 @@ def default_solr_doc_params(id = nil)
...
 
   def show
-    params[:id] = 'druid:' + params[:id] unless params[:id].include? 'druid'
-    @obj = Dor.find params[:id]
-
+    @obj = find_druid(params[:id])
+    return if @obj.nil?
```
jmartin-sul: now that `find_druid` would throw an `ObjectNotFoundError` if the requested object isn't found, i think these `return if @obj.nil` lines are no longer necessary.

app/controllers/items_controller.rb
```diff
@@ -585,38 +659,59 @@ def apply_apo_defaults
...
+    render status: :forbidden, text: 'Object cannot be modified in its current state.'
+    false
   end
 
-  # Filters
   def create_obj
-    raise 'missing druid' unless params[:id]
-    @object = Dor::Item.find params[:id]
+    @object = find_druid(params[:id])
+    return if @object.nil?
```
jmartin-sul: now that `find_druid` would throw an `ObjectNotFoundError` if the requested object isn't found, i think these `return if @object.nil` lines are no longer necessary.

app/controllers/status_controller.rb
```diff
@@ -11,7 +11,8 @@ def log
       render :status => 500, :text => 'Nothing indexed recently.'
       return
     end
-    test_item = Dor::Item.find(params[:test_obj])
+    test_item = find_druid(params[:test_obj])
+    return if test_item.nil?
```
jmartin-sul: same as the other `find_druid` calls, seems like the return if nil can go

app/controllers/catalog_controller.rb
jmartin-sul: alright, i just spent a bit trying to figure out a way to reimplement `valid_user?` with `can_view_object?`.  here's what i came up with:

* though the two methods are effectively the same since they both use the same role list, `valid_user?` calls `can_view_metadata?`, but `can_view_object?` calls `can_view_content?`.  i have a vague memory of @darrenleeweber and @LynnMcRae discussing this recently?  anyway, if it'd be ok to have `valid_user?` rely on `can_view_content?` for its logic instead, i think the below code would be a more concise reimplementation with the same messages and logic and less duplication of the basic permission checking logic (though an assumption about the cause of access denial is made for message construction purposes):
```ruby
def valid_user?(dor_object)
  valid = @user.can_view_object? dor_object
  return valid if valid

  if dor_object.is_a? Dor::AdminPolicyObject
    msg = 'Item is an APO that forbids access'
  elsif dor_object.admin_policy_object
    msg = 'Governing APO forbids access'
  else
    msg = 'Item has no APO that allows access'
  end
  render :status => :forbidden, :text => msg
end
```

jmartin-sul: @LynnMcRae @atz:  there are known instances of objects not having an APO, right?  i think `valid_user?` accounts for that (`@apo = dor_object.admin_policy_object rescue nil`) in its current incarnation.

it looks like `can_view_object?` doesn't account for that, so my proposed code above does, but maybe `can_view_object?` should?  e.g.:
`apo_pid = dor_object.admin_policy_object.pid rescue nil`

app/models/user.rb
```diff
@@ -25,80 +30,204 @@ def to_s
...
+  #
+  # @param dor_object [String|Dor::Base] Accepts a DRUID string or
+  #                                      any subclass of Dor::Base
+  # @return [Boolean]
+  def can_view_object?(dor_object)
+    dor_object = get_dor_object(dor_object)
+    # Any object administrator or manager can also view the object.
+    return true if can_manage_object?(dor_object)
+    pid = dor_object.pid
+    apo_pid = dor_object.admin_policy_object.pid
```
jmartin-sul: maybe:
`apo_pid = dor_object.admin_policy_object.pid rescue nil`

in case the object in question doesn't have an APO, as with the current incarnation of `CatalogController#valid_user?`

app/controllers/catalog_controller.rb
jmartin-sul: also, what if the perm check was done as a `before_filter` on `bulk_jobs_index` and `show`, instead of inline?  that feels a bit cleaner, and is comparable to what's done in other places in argo.

LynnMcRae: Google Scanned books lack an APO since they pre-dated the concept.  That means in principle only sdr-* roles can find/see them since no APO filtering is applied to their views, and it's a condition we would like to see in Argo in order to fix it there.  However, new Argo still has a problem trying to display the item page for these (true, despite all the "let's not implode if a druid can't be de-referenced" work?) so we plan to add in the APO references for those before release.

app/controllers/application_controller.rb
darrenleeweber: On my laptop a request for `find_druid('junk')` will hit the `Rubydora` exception and, yes, I agree it should be a 400.  On travis, it hits an `ActiveFedora::ObjectNotFoundError` and that should be a 404.  So, the specs are somehow platform dependent and I chose to coerce them into the same response code to avoid the platform specific inconsistency.

app/controllers/catalog_controller.rb
darrenleeweber: Applied all of these suggestions.  We discussed what constitutes a `valid_user?` and decided it is any admin/manger/viewer or a user in a workgroup who `can_view_metadata?`.  The idea of a before filter is good, but the catalog controller requires the `@obj` if it is a valid user and I'm not sure a before filter would provide it.  It could be done and, actually, the catalog controller should fall into line with other controllers and use `@object` instead of `@obj`, but these are larger changes and devs are already complaining about this PR.  Keeping it simpler for now, but let me know if these are changes worth making now.

app/controllers/apo_controller.rb
darrenleeweber: :ok_hand: -- removed them all.

app/models/user.rb
darrenleeweber: Good point; see revised code for how this is handled.

app/controllers/apo_controller.rb
```diff
@@ -268,16 +286,19 @@ def update_desc_metadata
 
   def spreadsheet_template
     binary_string = RestClient.get(Argo::Config.urls.spreadsheet)
-    send_data(binary_string, :filename => 'spreadsheet_template.xlsx', :type => 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet')
```
mejackreed: Is the formatting change here required for this PR?

app/controllers/items_controller.rb
mejackreed: Agree with the hound here, are all of these before_filter's being changed though? Personal preference would only touch the ones in this commit needed for the feature that this PR addresses.

app/controllers/items_controller.rb
```diff
@@ -19,7 +48,11 @@ def purl_preview
   def prepare
     if can_open_version? @object.pid
       begin
-        vers_md_upd_info = {:significance => params[:severity], :description => params[:description], :opening_user_name => current_user.to_s}
```
mejackreed: Similar comment as before. Prefer not touching lines in a commit that addresses a bug, unless the lines touched relate to the issue at hand.

app/controllers/items_controller.rb
```diff
@@ -54,10 +87,11 @@ def set_collection
         end
       else
         if can_set_collection
-          format.html { redirect_to catalog_path(params[:id]), :notice => 'Collection successfully set' }
```
mejackreed: Any functional change that relates to the issue here?

app/controllers/items_controller.rb
```diff
@@ -120,7 +157,11 @@ def workflow_view
       format.any(:png, :svg, :jpeg) do
         graph = @workflow.graph
         fail ActionController::RoutingError.new('Not Found') if graph.nil?
-        send_data graph.output(request.format.to_sym => String), :type => request.format.to_s, :disposition => 'inline'
+        send_data(
```
mejackreed: Similar to other comments related to necessity for the issue at hand. I'm all for conformity to a style, just not at the expense of confusing a future reviewer in a single commit or PR.

app/controllers/items_controller.rb
```diff
@@ -138,18 +179,16 @@ def workflow_update
...
       # update the status for the step and redirect to the workflow view page
       Dor::WorkflowService.update_workflow_status params[:repo], *args
-
       respond_to do |format|
         if params[:bulk].present?
           render :status => 200, :text => 'Updated!'
         else
-          format.any { redirect_to catalog_path(params[:id]), :notice => "Updated #{params[:process]} status to '#{params[:status]}' in #{params[:wf_name]}" }
+          msg = "Updated #{params[:process]} status to '#{params[:status]}' in #{params[:wf_name]}"
+          format.any { redirect_to catalog_path(params[:id]), notice: msg }
```
mejackreed: Any functional changes here? Prefer not to make changes in this PR that aren't related to current issue.

app/controllers/items_controller.rb
```diff
@@ -248,9 +287,15 @@ def update_attributes
     [:publish, :shelve, :preserve].each do |k|
       params[k] = (params[k].nil? || params[k] != 'on') ? 'no' : 'yes'
     end
-    @object.contentMetadata.update_attributes(params[:file_name], params[:publish], params[:shelve], params[:preserve])
+    @object.contentMetadata.update_attributes(
```
mejackreed: Same questions around formatting changes that could be addressed outside of this commit/pr

app/controllers/apo_controller.rb
darrenleeweber: hound.ci would not pass the PR (this comment applies to all similar concerns for Jack).

app/controllers/items_controller.rb
darrenleeweber: Appease the hound.ci or the PR doesn't get the green light.  (Discussed in slack and can be discussed further in person.  Understand the pain of reviewing something like this where it's difficult to focus on the main thing.  Will learn from this experience and raise the learning experience in the DLSS best-practices forum.)

app/controllers/apo_controller.rb
```diff
@@ -313,6 +334,13 @@ def redirect
 
   # check that the user can carry out this object modification
   def forbid
+    # TODO: this could be a more granular permission request, see the
+    # Dor::Permissable#can_create_apo? and Dor::Permissable#can_manage_apo? in
+    # https://github.com/sul-dlss/dor-services/blob/permissable/lib/dor/models/permissable.rb
+    #return if current_user.can_manage?(@object, 'apo')
+    #
+    # Note that this is different from User#can_manage? because it
+    # excludes current_user.is_manager
```
jmartin-sul: i would be a little hesitant to leave this comment for two reasons:
1) it refers to a branch in dor-services for example code, but because that branch is unmerged and in flux, it's possible that the comment will fall out of date as the referenced branch changes (and it seems unlikely that changes on an unmerged branch will trigger a developer to go look for comments or code that reference it, esp outside of the codebase in question).  not that it's never right to reference other code, just that it seems like the code being referenced should be related, committed, and relatively stable (expected to stick around for a while).
2) i think it's often a good idea to try and create an actual issue rather than leave a TODO comment, though i'm certainly not opposed to the latter in all cases.  this one would be a tough call for me, and i think i'd lean towards filing a ticket for discussion (since it appears to suggest a change in functionality, rather than just an optimization or functionally identical refactoring).

also, i'm not sure whether this is an example of how you're suggesting things could look, or what, since there's not much explanation as to what this comment line means: `#return if current_user.can_manage?(@object, 'apo')`

i also just noticed that there's at least one other similar comment in a totally different location in the code.  i think that'd push me even more towards filing an issue and getting rid of these specific TODOs.

app/controllers/items_controller.rb
```diff
@@ -184,6 +223,9 @@ def release_hold
   end
 
   def embargo_update
+    # TODO: this could be a more granular permission request, i.e.
+    #apo = @object.admin_policy_object
+    #@object.can_manage_embargo?(current_user.roles(apo.pid))
```
jmartin-sul: can we lump all the TODOs of this sort into a github issue with references to the code?  these all seem like suggestions for functional changes.  they might be minor changes that would be easily agreed upon, but they seem to warrant discussion (both for whether they're desirable in principle, and for whether we want to bother implementing them even if they are).

app/controllers/items_controller.rb
```diff
@@ -373,15 +437,18 @@ def self.normalize_source_id(new_source_id)
   def source_id
     new_id = ItemsController.normalize_source_id(params[:new_id])
     @object.set_source_id(new_id)
-    # TODO: the content= and content_will_change! calls belong in dor-services, for this method and other similar methods.
-    # can then clean up allow(idmd).to receive(:"ng_xml") (and "content_will_change!") in tests.  getting rid of "Metadata\.content\s*=\s*"
+    # TODO: the content= and content_will_change! calls belong in dor-services,
+    # for this method and other similar methods. can then clean up
+    # allow(idmd).to receive(:"ng_xml") (and "content_will_change!") in tests.
+    # getting rid of "Metadata\.content\s*=\s*"
```
jmartin-sul: a personal gripe that i've made plenty of times already:  i would love to turn off the line-length complaint in hound and rubocop.  i don't find this re-written paragraph more readable, and i think most of the times that hound (or whatever) encourages short lines, we'd be better off ignoring the suggestion.  there are times when a line should be shortened.  if it's that egregious, someone can call it out during code review.

app/models/user.rb
jmartin-sul: actually, hound might be right, i think you're using an apostrophe instead of a single quote:  `it’s` instead of `it's`

app/models/user.rb
jmartin-sul: ditto the quote here

app/controllers/items_controller.rb
darrenleeweber: Well, as you can see, one set of lines fits in this github view and one doesn't.  Guess which one fits and can be read in it's entirety without scrolling the view?

app/controllers/apo_controller.rb
darrenleeweber: Disagree with (1).  The link will always exist as long as the branch is not deleted.  The granularity of permission might be important and it is a TODO.  There is a clear suggestion on how the code can function, so (2) is much clearer in the code that in some disassociated ticket or issue that might or might not get closed without acting on it.  I'm not taking out the TODO comments, but feel free to file a ticket/issue and note these TODO lines of code in that issue.  It's likely to add to a long list of role/permission issues that linger in a backlog and don't get resolved.

app/controllers/items_controller.rb
darrenleeweber: Sure.  See comments below.

darrenleeweber:
Closing this out of date PR to replace it with cleaner commits in a new PR that is up to date with the develop branch.

PR #476, opened 2016-01-26
Update dependencies

cbeer:


mejackreed:
Thanks for the contribution. I just tried out a deploy to dev, and there might be some updates needed to our deploy. Will investigate further... but getting:

```sh
INFO [deploy:migrate] Run `rake db:migrate`
(Backtrace restricted to imported tasks)
cap aborted!
NoMethodError: undefined method `verbosity' for "/usr/bin/env deploy:migrating\n":String

Tasks: TOP => deploy:migrate
(See full trace by running task with --trace)
The deploy has failed with an error: undefined method `verbosity' for "/usr/bin/env deploy:migrating\n":String
```

cbeer:
Hm. I haven't seen that before, I'll try to dig into it some today.

config/deploy.rb
```diff
@@ -39,17 +38,6 @@
...
-  desc 'Restart application'
-  task :restart do
-    on roles(:app), in: :sequence, wait: 5 do
-      # Your restart mechanism here, for example:
-      execute :touch, release_path.join('tmp/restart.txt')
-    end
-  end
-
-  after :publishing, :restart
-
```
tingulfsen: Chris, what was the reason for removing the restart task?

cbeer: capistrano-passenger handles it for us now.

jmartin-sul:
tommy and i each tried deploying this branch (to dev and stage, respectively).  but we ran into the same error that @mejackreed mentioned above.

as such, figuring we should get the rails update in, and that we might as well do a full bundle update, i created a new branch that just did a bundle update on Gemfile.lock.  interestingly, it ran into the same problem as this.

i noticed that my branch that just updated rails and didn't have deployment issues was missing an upgrade for capistrano-rails from 1.1.5 to 1.1.6.  when i pinned capistrano-rails to 1.1.5 in my otherwise complete update branch, deployment worked fine.
https://github.com/sul-dlss/argo/pull/478 (full upgrade, but pin capistrano-rails)
https://github.com/sul-dlss/argo/pull/477 (the more limited Rails-only upgrade)

i would propose:
* merge 478 if there are no objections to the pin
* file a bug to alleviate that pinning
* rebase this off 478 and see if that resolves the issue experienced with this PR

jmartin-sul:
well, i rebased this on the branch from PR #478, and that seems to work fine.  PR #479 represents that work.  i vote for merging that one and closing 476 through 478.

tingulfsen:
After some discussion with @cbeer and @jmartin-sul we decided to go with #479 instead, so closing this pull request.

PR #478, opened 2016-01-26
[WIP] Bundle update

jmartin-sul:
do a full bundle update (with one exception) to get the latest versions of dependencies.  spurred by a rails vulnerability that bundle-audit complained about, updated everything since we're still testing heavily and might as well use the latest versions we expect to be compatible with.

pin capistrano-rails to 1.1.5 so we don't get this error:
`NoMethodError: undefined method `verbosity' for "/usr/bin/env deploy:migrating\n":String`

i was able to deploy this successfully to stage with this branch.


jmartin-sul:
i vote for merging PR #479, then closing this and PR #476

Issue #380, opened 2015-12-16
fix risearch queries for full reindexing


jmartin-sul:
some background:
- As I noticed earlier, and as @rosy1280 pointed out to me today, the argo-prod-b index is currently about 200k records short of the argo-prod index.  Rosy says these are mostly Google books, but she can clarify the characteristics of the shortfall whenever she gets a chance (she had other observations I've forgotten).
- IIRC from when we had to fully reindex Argo last fall, the druids returned by Fedora's risearch interface did not provide a complete list for what needed to be indexed.  I have vague memories of Willy saying that he needed to query Fedora's underlying MySQL database to get the full list of PIDs for reindexing, but there was no explanation as to why.
- Fedora's risearch functionality uses Fedora's triple store index, which we turned off some time ago (prior to last fall's full reindex).  I believe it was for performance reasons, but @LynnMcRae can clarify there.  So risearch will return stale results on our Fedora instances.
- AFAICT, Fedora's underlying relational database does not contain any usable info on object type.  The available tables in the Fedora DB are:

```
mysql> show tables;
+---------------------+
| Tables_in_fedora    |
+---------------------+
| datastreamPaths     |
| dcDates             |
| doFields            |
| doRegistry          |
| fcrepoRebuildStatus |
| modelDeploymentMap  |
| objectPaths         |
| pidGen              |
+---------------------+
```

```
mysql> describe doRegistry;
+---------------+--------------+------+-----+---------+-------+
| Field         | Type         | Null | Key | Default | Extra |
+---------------+--------------+------+-----+---------+-------+
| doPID         | varchar(64)  | NO   | PRI | NULL    |       |
| systemVersion | int(11)      | NO   |     | 0       |       |
| ownerId       | varchar(64)  | YES  |     | NULL    |       |
| objectState   | varchar(1)   | NO   |     | A       |       |
| label         | varchar(255) | YES  |     |         |       |
+---------------+--------------+------+-----+---------+-------+
```

```
mysql> describe doFields;
+---------------+--------------+------+-----+---------+-------+
| Field         | Type         | Null | Key | Default | Extra |
+---------------+--------------+------+-----+---------+-------+
| pid           | varchar(64)  | NO   | MUL | NULL    |       |
| label         | varchar(255) | YES  |     | NULL    |       |
| state         | varchar(1)   | NO   |     | A       |       |
| ownerId       | varchar(64)  | YES  |     | NULL    |       |
[...]

PR #482, opened 2016-01-29
add solr_pid_dumper.rb

jmartin-sul:
add basic methods for dumping pids from solr by object type and writing them to files

### DIFF
--- a/lib/solr_pid_dumper.rb
+++ b/lib/solr_pid_dumper.rb
@@ -1,0 +1,50 @@
+# a thrown-together (but useful and not overly hacky) one-off script for dumping
+# object IDs from solr, to files (json lists), so that other instances can use them for
+# reindexing.  good for things like taking known lists from an existing
+# prod instance and using them for indexing in an upgraded soon-to-be prod instance.
+#
+# to use:
+# * make sure that BASE_DIR exists relative to where this script is executed
+# * make sure OBJ_TYPE_FIELD_NAME is right for the schema of the index you're dumping
+# * make sure OBJ_TYPES is the full list of object types you're interested in
+# * open rails console on the instance you want to dump from
+# * require 'solr_pid_dumper.rb'
+# * call Argo::SolrPidDumper.write_solr_pids_by_obj_type
+# * use your pids for indexing (e.g., as per https://github.com/sul-dlss/argo/blob/develop/INDEXING.md)
+
+module Argo
+  class SolrPidDumper
+    OBJ_TYPE_FIELD_NAME = 'objectType_facet'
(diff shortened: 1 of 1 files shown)


jmartin-sul:
this is a new util class i implemented against the 2.x branch to (as the name implies) dump pids by object type from solr.  i deployed it by hand on old argo 2.x prod.  i ran the version from the first commit in the PR to gather the pid lists used for the january and february full reindexing runs in preperation for the recent argo 3 deployment.

there's actually a good chance this'll never get used again, and so silencing the hound complaints was probably unnecessary (and i didn't bother with the line length ones, and i missed a couple others).  still:  i'd like to merge this for posterity's sake, in case we have to come back and reconsider our old work, since this was a very handy tool and was used twice in a recent deployment process.

ndushay:
@jmartin-sul should this be closed, as travis didn't pass?   Or should it be ported to argo 3?  Or ???

jmartin-sul:
@ndushay:  definitely shouldn't be ported to 3, since argo 3 already has an equivalent `solr_pids` method, and since `write_solr_pids_by_obj_type` could be ported there if/when the need arises (i suspect it will eventually, and i'd probably add the ability to take a target output dir, but i'd rather wait and write more query/indexing code as the need arises).

this branch was pretty much a backport of `solr_pids` from argo 3 to argo 2, plus a method on top of that to dump pids by object type.  this was needed because fedora's `risearch` is unreliable (it returned incomplete results when querying by object type, see issue #380).  but we trusted argo 2's solr index for getting IDs by object type, so we used that to generate the lists used for re-indexing fedora to the argo 3 solr schema (so that collections, APOs, etc would get indexed prior to regular objects).

really, i mostly created this as a branch at the time to make sure i didn't lose work i thought i'd re-use if i lost my laptop or something, since we anticipated having to build the index multiple times.  but i'd also wanted to preserve the logic that was used for the reindexing, in case we care about it when figuring something out in the future.

my guess is that the 2.x build is failing in general, and that this PR isn't failing it specifically.  and as per my prior comment, i don't really anticipate using this code again (or deploying argo 2 again, knock on wood).  this is mostly for posterity's sake, at this point, though i understand/appreciate the hesitancy to merge a failing build.

ndushay:
ok, if you want to merge this, you've got my approval, FWIW.

jmartin-sul:
just so there's a bit more transparency, here are the identically failing builds for 2.x-stable and this branch:
https://travis-ci.org/sul-dlss/argo/builds/105543786
https://travis-ci.org/sul-dlss/argo/builds/113310985

merging, but totally happy to revert if others who haven't chimed in yet object.

PR #483, opened 2016-01-29
add bin/dump_fedora_pids.rb

jmartin-sul:
simple script that can be copied to a machine running a fedora instance, to be used to dump all the pids in the instance, since risearch is unreliable for that.  results can be used for custom indexing according to instructions in https://github.com/sul-dlss/argo/blob/develop/INDEXING.md

### DIFF
--- a/bin/dump_fedora_pids.rb
+++ b/bin/dump_fedora_pids.rb
@@ -1,0 +1,51 @@
+## assumptions:
+# * you're running this on the same machine as the fedora instance you want to dump from
+# * the user you're running it as can execute the 'mysql' command, and can write to '/tmp'
+# * the user can temporarily write a sql script to the place from which their running this script
+# * .fedora.my.cnf is a config file with a line like "  password = supersecretcredential"
+# * the password in .fedora.my.cnf will get you into fedora sql db for it's user (which defaults to "fedora")
+# * .fedora.my.cnf exists in the same directory from which you're running this script, and can be read by the user running it
+#
+# if all that's true, you should be able to do "ruby dump_fedora_pids.rb"
+# and it should tell you where it wrote the pids
+
+require 'date'
+
+# a list of the files read and written by this script
+cur_datetime_str = DateTime.now.strftime('%Y-%m-%d_%H:%M:%S')
+fedora_conf_file_name = '.fedora.my.cnf'
+out_file_name = "/tmp/all_pids_#{cur_datetime_str}"
+script_file_name = "fedora_mysql_pid_dump_#{cur_datetime_str}.sql"
+
+def get_fedora_password(conf_file_name)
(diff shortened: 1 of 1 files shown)


bin/dump_fedora_pids.rb
```diff
@@ -0,0 +1,46 @@
...
+
+cur_datetime_str = DateTime.now.strftime('%Y-%m-%d_%H:%M:%S')
+fedora_conf_file_name = '.fedora.my.cnf'
+out_file_name = "/tmp/all_pids_#{cur_datetime_str}"
+script_file_name = "fedora_mysql_pid_dump_#{cur_datetime_str}.sql"
+
+
+def get_fedora_password conf_file_name
+  fedora_conf_file_contents = open(conf_file_name, 'r').read
+  return (/^\s*password\s*=\s*(.*)$/.match(fedora_conf_file_contents))[1]
```
jmartin-sul: this would be even less readable without parens

bin/dump_fedora_pids.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+
+
+db_user_name = 'fedora'
+db_password = get_fedora_password(fedora_conf_file_name)
+db_name = 'fedora'
+
+
```
jmartin-sul: also seems more readable as-is

bin/dump_fedora_pids.rb
```diff
@@ -0,0 +1,46 @@
...
+out_file_name = "/tmp/all_pids_#{cur_datetime_str}"
+script_file_name = "fedora_mysql_pid_dump_#{cur_datetime_str}.sql"
+
+
+def get_fedora_password conf_file_name
+  fedora_conf_file_contents = open(conf_file_name, 'r').read
+  return (/^\s*password\s*=\s*(.*)$/.match(fedora_conf_file_contents))[1]
+end
+
+
```
jmartin-sul: ditto

jmartin-sul:
since this is more of a one-off script to be run from the command line (not even classes to use from rails console), i think the extra blank lines for separation are fine, and more readable than getting rid of them.  the only complaint i'm inclined to agree with hound on is parens in the def for `get_fedora_password`.  i can push a version with that if people feel that's important.

cbeer:
Instead of just blank lines, maybe a simple comment that labels the following section would be helpful?

bin/dump_fedora_pids.rb
cbeer: Instead of just removing the parens, maybe this is a good time to refactor that method further?

It seems like this line could be split into multiple, highly readable lines (and might benefit from it -- it seems like some error handling if the was no match might be useful).

jmartin-sul:
fair point, done

PR #486, opened 2016-02-01
Adding separate Bulk Updates button to the search results page.

### LINKED ISSUE #417: Bulk update as its own thing rather than a view?
Do we need to treat bulk update as if it was a peer search result view to other displays?  Having it accessible under the View drop down and calling it "Bulk update view" is not intuitive and has already tripped me up a couple of times.  At a minimum, can it be presented as its own button?

[Sort by druid v]  [10 per page v]  [Select view v]  [ Bulk update]

(an aside, I'd like this also so that it can become a dropdown to other Bulk actions unrelated to the results display, such as the downloading of descMetadata)

An on the Bulk action page, this top heavy screen could be simplified by eliminating the Select View button taking up a whole slice of the screen, and we could simply have "Back to Search" next to "Start Over".  
screen shot 2016-01-06 at 11 44 34 am

This would make it comparable to the Spreadsheet upload screen:
screen shot 2016-01-06 at 11 47 46 am

tingulfsen:
Closes #417

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -12,6 +12,7 @@ class CatalogController < ApplicationController
 
   before_filter :reformat_dates, :set_user_obj_instance_var
   before_action :show_aspect, only: [:dc, :ds]
+  before_action :sort_collection_actions_buttons, only: [:index]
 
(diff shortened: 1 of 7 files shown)


mejackreed:
:+1:

PR #487, opened 2016-02-01
reduce @page_title setting duplication and make sure it is stripped o…

### LINKED ISSUE #481: There's an <em> in my tab; and an inverted argo shell
The tab label for detail display, e.g., https://argo-stage.stanford.edu/catalog/druid:bb004bn8654 has an odd \<em> insertion .... see old/new
screen shot 2016-01-27 at 2 05 50 pm

In both the favicon shows the argo shell upside down from the way it is now shown in our new header bar. Can it be flipped? I can provide an image if needed.

Marking this as blocker only as an indication of its anticipated inclusion in the release, at least for the unwanted tag.

mejackreed:
…f html

Closes #481


mejackreed:
Also now updated to flip the favicon

screen shot 2016-02-03 at 3 36 18 pm

PR #489, opened 2016-02-03
use globally defined content types, and add in support for video and …

### LINKED ISSUE #68: Add "media" to bulk update and blue button item update for content type changes
These are presented in a drop-down menu currently. Accompanying resource types in an associated drop-down are "audio" and "video"

mejackreed:
…audio

Closes #68 

screen shot 2016-02-03 at 1 06 11 pm
screen shot 2016-02-03 at 1 05 59 pm

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -547,7 +547,7 @@ class ItemsController < ApplicationController
   end
   # set the content type in the content metadata
   def set_content_type
-    unless %w(book file image map manuscript).include? params[:new_content_type]
+    unless Constants::CONTENT_TYPES.include? params[:new_content_type]
       render :status => :forbidden, :text => 'Invalid new content type.'
       return
     end
--- a/spec/lib/constants_spec.rb
+++ b/spec/lib/constants_spec.rb
@@ -1,0 +1,14 @@
+require 'spec_helper'
+
+RSpec.describe Constants do
+  it 'has correct CONTENT_TYPES defined' do
+    expect(Constants::CONTENT_TYPES).to include(
+      'image', 'book', 'file', 'manuscript', 'map', 'media'
+    )
+  end
+  it 'has correct RESOURCE_TYPES defined' do
+    expect(Constants::RESOURCE_TYPES).to include(
+      'image', 'page', 'file', 'audio', 'video'
+    )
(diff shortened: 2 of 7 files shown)


LynnMcRae:
The new content type should be "media" which then offers one of two resource types -- audio or video

LynnMcRae:
The "suggested mapping" list above the input would then have "**media** audio video" added

mejackreed:
Thanks @LynnMcRae for keeping me honest. Updated screenshots of the enhancement with the correct (hopefully) changes.

app/controllers/items_controller.rb
```diff
@@ -547,7 +547,7 @@ def set_rights
   end
   # set the content type in the content metadata
   def set_content_type
-    unless %w(book file image map manuscript).include? params[:new_content_type]
+    unless Constants::CONTENT_TYPES.include? params[:new_content_type]
       render :status => :forbidden, :text => 'Invalid new content type.'
```
drh-stanford: shouldn't this be a 400 Bad Request rather than 403?

spec/lib/constants_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+RSpec.describe Constants do
```
drh-stanford: i'm not sure it's necessary to have a spec test on the _data_ for the constants. imho the existence and value type of the constant is fine for the spec. otherwise you have to maintain the data in two places.

app/controllers/items_controller.rb
mejackreed: Not sure... the intent here is that if someone updated the constant they would also need to update the spec. Making sure that someone didn't remove one and know that it was actually expected to be there.

spec/lib/constants_spec.rb
mejackreed: Not sure... the intent here is that if someone updated the constant they would also need to update the spec. Making sure that someone didn't remove one and know that it was actually expected to be there.

app/controllers/items_controller.rb
drh-stanford: I created a new PR for it as I found a few cases in items controller: https://github.com/sul-dlss/argo/issues/500

Issue #414, opened 2016-01-05
Republish fail affects subsequent modals


LynnMcRae:
Pick a record in stage ... https://argo-stage-b.stanford.edu/catalog/druid:bb420md3291

Try to Republish .... you will currently get a system error
![screen shot 2016-01-05 at 3 24 36 pm](https://cloud.githubusercontent.com/assets/1824732/12130586/7655d9e0-b3c0-11e5-9057-407b172cefd8.png)

Now open up descMetadata ... it's all centered, with a bad title
![screen shot 2016-01-05 at 3 25 26 pm](https://cloud.githubusercontent.com/assets/1824732/12130620/a41ae29e-b3c0-11e5-8d9a-6e851af8f8c5.png)

And the underlying screen is wonky too 
![screen shot 2016-01-05 at 3 26 30 pm](https://cloud.githubusercontent.com/assets/1824732/12130632/ba007dc6-b3c0-11e5-85b1-e66b3a89c88c.png)

mejackreed:
Republish should move to using flash messages (bootstrap alerts) similar to #415

drh-stanford:
This also happens when clicking to download one of the content files.

From stage, druid:gs846fm9652:

before:

![screen shot 2016-01-07 at 12 09 39 pm](https://cloud.githubusercontent.com/assets/1861171/12181516/8cf8806c-b537-11e5-9aa6-9396dfb70780.png)

click on file:

![screen shot 2016-01-07 at 12 08 17 pm](https://cloud.githubusercontent.com/assets/1861171/12181488/5aa2d7b6-b537-11e5-9c58-22218f958bd4.png)

after:

![screen shot 2016-01-07 at 12 08 50 pm](https://cloud.githubusercontent.com/assets/1861171/12181505/75b95c78-b537-11e5-971d-b664d86cfe0e.png)

drh-stanford:
re: Net:SSH (expected) errors in development, you get a stack trace in a modal dialog without a close.

![screen shot 2016-02-02 at 4 27 13 pm](https://cloud.githubusercontent.com/assets/1861171/12769127/f5168896-c9c9-11e5-9498-74dd913207c8.png)

drh-stanford:
Here's what the flash message looks like for Republish:

![screen shot 2016-02-03 at 3 33 40 pm](https://cloud.githubusercontent.com/assets/1861171/12800654/8c3bb4dc-ca8b-11e5-9e67-d49994c32097.png)

[...]

PR #493, opened 2016-02-03
change Republish button to use flash message

drh-stanford:
This PR addresses part of #414. It changes the Republish button to use flash messages on the subsequent `/catalog` page via the standard redirect pattern.


mejackreed:
:+1:

PR #494, opened 2016-02-04
Bulk actions have editable pids and can query from the last search

mejackreed:
This pull request adds the ability to kick off a job from a previous search


mejackreed:
Duplicate

PR #496, opened 2016-02-04
Argo::Indexer functionality mostly refactored up into dor-services (Dor::IndexingService)

jmartin-sul:
* config/environments/dor_*.rb: config params for Dor::IndexingService
* indexer.rb: Indexer now extends Dor::IndexingService, which implements all the old methods from Indexer, except self.reindex_pid_list_with_profiling, which remains in Argo, since it uses Argo::Profiler, which is staying put.
* indexer_spec.rb: remove all the tests except the one related to reindex_pid_list_with_profiling.  remove old TODO (turned into github issue).
* dor_controller.rb: get rid of index_logger method; rails logger creation centralized to Dor::IndexingService.rails_index_logger, use that instead (via Argo::Indexer inheritance).
* dor_controller_spec.rb: reflect switch from generate_index_logger (called by removed index_logger method) to rails_index_logger.


Gemfile
```diff
@@ -41,7 +41,7 @@ gem 'active-fedora'
 gem 'blacklight', '~> 5.9.0' # TODO: BL >= 5.10.x has new deprecation warnings vs <= 5.9.x, will investigate and unpin after current upgrade stuff has settled
 gem 'blacklight-hierarchy'
 gem 'blacklight-marc'
-gem 'dor-services', '~> 5.0', :git => 'https://github.com/sul-dlss/dor-services.git', :branch => 'develop'
+gem 'dor-services', '~> 5.0', :git => 'https://github.com/sul-dlss/dor-services.git', :branch => 'indexing-service'
```
jmartin-sul: once https://github.com/sul-dlss/dor-services/pull/153 is merged, this can go back to develop.  probably still just a bit too early to start pointing to released versions.

Gemfile.lock
```diff
@@ -7,8 +7,8 @@ GIT
 
 GIT
   remote: https://github.com/sul-dlss/dor-services.git
-  revision: 24c77cdc06b562bc6d3e61cd46d2cf0597453f4b
-  branch: develop
+  revision: c09d4887445b663556c4823bcf3775d7768043be
+  branch: indexing-service
```
jmartin-sul: this'll get changed along with Gemfile before merge

jmartin-sul:
d'oh.  centralizing the rails logging as i've done in these two PRs doesn't work, because the `request` object used in the block is from the context of `Dor::IndexerService`, which does not have the Rails `request` object that we care about.  will have to leave that code in the controllers, unless others have better ideas.  not ideal, but not a huge deal, pretty short chunk to repeat.

jmartin-sul:
out of WIP, ready for evaluation.

mejackreed:
I've been keeping an eye on this, looks good to me! :+1:

PR #502, opened 2016-02-09
Bulk actions framework

### LINKED ISSUE #136: Bulk Descriptive Metadata Download
1. from a list of Druids, fetch and output the MODS from the descMetadata stream
     - as a single \<xmlDocs> document (see https://consul.stanford.edu/display/chimera/MODS+bulk+loading)
2. Offline process managed as background task and tracked, delivered via a history screen similar to that developed for Replayable Spreadsheets 
3. Authorization checked for access with "Not authorized" error" on a per object basis
3. First instance of data output in bulk ... offer "druids from search" and "paste a druid list" choice like Bulk Update
4. Change name of "Bulk Update" to "Bulk actions" ... or implement separate "Bulk output" for this and (future) tracking sheet generation.

mejackreed:
Closes #136 

This pull request lays the foundation for the future of Argo's bulk actions by adding a framework and implementing descriptive metadata download.

screen shot 2016-02-08 at 5 21 52 pm
screen shot 2016-02-08 at 5 21 21 pm
screen shot 2016-02-08 at 5 21 38 pm

### DIFF
--- a/app/jobs/generic_job.rb
+++ b/app/jobs/generic_job.rb
@@ -1,6 +1,22 @@
 ##
 # A GenericJob used as a super class for Argo Bulk Jobs
 class GenericJob < ActiveJob::Base
-  def perform(_a, _b, _c)
+  # A somewhat easy to understand and informative time stamp format
+  TIME_FORMAT = '%Y-%m-%d %H:%M%P'.freeze
+
+  around_perform do |_job, block|
+    bulk_action.update_attribute(:status, 'Processing')
+    block.call
+    bulk_action.update_attribute(:status, 'Completed')
+  end
+
+  ##
+  # @param [Integer] _a GlobalID for a BulkAction object
+  # @param [Hash] _b additional parameters that an Argo job may need
+  def perform(_a, _b)
+  end
+
+  def bulk_action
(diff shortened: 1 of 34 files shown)


app/jobs/generic_job.rb
```diff
@@ -0,0 +1,22 @@
+##
```
mejackreed: I think this is a miscue by Hound.

drh-stanford:
Re: schema. Since the `action_type` and `status` columns both seem to be enumerated types, it'd be nice to see somewhere what all the possible values are for them. Ideally with a check constraint on the schema.

mejackreed:
Thanks for the comment @drh-stanford . In 8db6f29 I've added the validation for `action_type` at the model level and added a test for this. Is that sufficient?

drh-stanford:
What about `status`? Or do you never query on status?

mejackreed:
Ah.. yeah.. we don't query by status but we do check if "Completed". Hmm.. I feel like that is ok with being left more free form as we aren't executing any user provided params here. Thoughts?

drh-stanford:
If it's meant to be free-form text then no need to validate it. :+1:

drh-stanford:
I'm :+1: on this

PR #505, opened 2016-02-09
add flush_index after filter for embargo_update and add_workflow

### LINKED ISSUE #416: Update Embargo modal problems
https://argo-dev.stanford.edu/catalog/druid:bs277kf6370

Press the Update Embargo button .... the button goes dark (it's processing) but nothing happens for quite a long time for me.  When the form finally shows up, it looks like this:
screen shot 2016-01-05 at 3 53 01 pm

The Cancel button does nothing. The update works but the underlyiong screen should be refreshed to show the new date. (note new date of 4/1/17 in upper right metadata, old date still showing in lower left)
screen shot 2016-01-05 at 3 56 21 pm

drh-stanford:
This PR partially fixes #416. It adds an after filter Solr commit for embargo update so that the redirect has the correct information in the index to display the left sidebar Embargo information. It also migrates the add_workflow action to this Solr commit after filter. I updated the specs to expect the after filter.


jmartin-sul:
i would propose adding a simple test for the flush_index method, but other than that, :+1:

drh-stanford:
I think that I've addressed all of the comments.

PR #516, opened 2016-02-12
[WIP] add externalFile support to Contents section

### LINKED ISSUE #485: Show external file references in the Contents section of detail display
For regular contents, the display looks like this 
screen shot 2016-02-01 at 10 30 30 am

based on this kind of content metadata:
```
<contentMetadata type="image" objectId="bg730rr6720">
  <resource type="image" sequence="1" id="bg730rr6720_1">
    <label>Image 1</label>
    <file id="bg730rr6720_00_0001.tif" mimetype="image/tiff" size="31691632" preserve="yes" publish="no" shelve="no">
      <checksum type="md5">c3c03bf5e14d0726a53dce2e1e9b6d2a</checksum>
      <checksum type="sha1">c29a20b265844c7c053321903bf2dc8e4456b5df</checksum>
      <imageData width="2639" height="4000"/>
    </file>
    <file id="bg730rr6720_00_0001.jp2" mimetype="image/jp2" size="2001433" preserve="no" publish="yes" shelve="yes">
      <checksum type="md5">b9c58466d95922be4b86c2e778ee6eb7</checksum>
      <checksum type="sha1">33936e8e9c1f47e99ac73666ab0ec49d18d0e032</checksum>
      <imageData width="2639" height="4000"/>
    </file>
  </resource>
  <resource type="image" sequence="2" id="bg730rr6720_2">
```

For virtual resources, that are defined by an `<externalFile>` tag instead of a `<file>` tag, we get this:
screen shot 2016-02-01 at 10 31 53 am

Here's the start of content metadata:
```
<contentMetadata objectId="hj097bm8879" type="image">
  <resource id="hj097bm8879_1" sequence="1" type="image">
    <externalFile fileId="2542A.jp2" objectId="druid:cg767mn6478" resourceId="cg767mn6478_1"/>
    <relationship objectId="druid:cg767mn6478" type="alsoAvailableAs"/>
  </resource>
  <resource id="hj097bm8879_2" sequence="2" thumb="yes" type="image">
    <externalFile fileId="2542B.jp2" objectId="druid:jw923xn5254" resourceId="jw923xn5254_1"/>
    <relationship objectId="druid:jw923xn5254" type="alsoAvailableAs"/>
  </resource>
  <resource id="hj097bm8879_3" sequence="3" type="image">
```

Suggest a formatted display of the external file information under resource like so:
```
Resource (1) image 
    External file: 2542B.jp2 (from druid:jw923xn5254, resource "jw923xn5254_1)
```
Optionally the external object reference could be a link to that object in Argo. The file itself does not need to be linked for access like a regular file (might be nice someday, but an unnecessary complication to start with).

drh-stanford:
This PR fixes #485. It adds External File information into the Contents section of the catalog show page. The `_default_contents.html.erb` has ZERO tests, and all the data parsing, etc. is done in the view that is far from DRY. I did a major overhaul of the code formatting for the view to make it more readable hopefully. The only real part I added was the support for the `externalFile` elements. To test it, you need to create a contentMetadata datastream that has data as described in #485. If the manual testing is not acceptable, then I'd prefer to refactor how the view works (recommendations?). I also made some minor adjustments to the CSS styling.

Here's what the new display looks like:

screen shot 2016-02-12 at 12 17 25 pm

### DIFF
--- a/app/assets/stylesheets/argo/document.sass
+++ b/app/assets/stylesheets/argo/document.sass
@@ -75,16 +75,13 @@ h3.section-head
   clear: both
 
 ul.resource-tree
-  margin: 0
+  padding-left: 0px
   li
-    padding-left: 8px
     list-style-type: none
     span.label
       color: gray
       &:after
         content: ':'
-    ul
-      margin: 0
 
 table.detail
   thead
--- a/app/views/catalog/_show_sections/_default_contents.html.erb
+++ b/app/views/catalog/_show_sections/_default_contents.html.erb
@@ -1,100 +1,142 @@
-<% if object.datastreams.keys.include?('contentMetadata') &&
-  (not object.datastreams['contentMetadata'].new?) && object.datastreams['contentMetadata'].metadata? %>
-  <% content_ds = object.datastreams['contentMetadata'] %>
-  <% admin_policy_object = @apo %>
-  <div style="clear:left">
-    <ul class="resource-tree">
-      <li>
-        <% if document.preservation_size.present? %>
+<%# TODO: DRY up code here - major duplication due to nested resources support %>
+<div style="clear:left">
+  <ul class="resource-tree">
+  <% if object.datastreams.keys.include?('contentMetadata') && (not object.datastreams['contentMetadata'].new?) && object.datastreams['contentMetadata'].metadata? %>
+    <li>
+      <% if document.preservation_size.present? %>
+        <div>
           <span class="label">Preservation Size</span>
-          <%= number_to_human_size(document.preservation_size) %>
-          <br/>
-        <%end%>
+          <%= number_to_human_size(document.preservation_size) -%>
+        </div>
+      <% end %>
+      <div>
         <span class="label">Type</span>
-        <%= content_ds.ng_xml.root['type'] %>
-        <ul>
-          <% apo_pid = nil
-          if admin_policy_object
-            apo_pid = admin_policy_object.pid
-          end
-          
-          can_manage = current_user.is_admin || current_user.is_manager || (apo_pid && object.can_manage_content?(current_user.roles(apo_pid)))
-          can_view = current_user.is_admin || current_user.is_manager || (apo_pid && object.can_view_content?(current_user.roles(apo_pid)))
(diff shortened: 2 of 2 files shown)


mejackreed:
Sigh... so your work looks great here. The sigh is for the view stuff already in there. A few thoughts on how to approach.

 - We really shouldn't be assigning variables in the view like this. This is an actual use case for moving logic to a helper.

 - There is a big opportunity to break this out into multiple partials. I could see something like `_default_contents_resource.html.erb` and  `_default_contents_file.html.erb`. Each of those individual partials could be passed the locals it needs to render properly.

 - `ContentMetadata` resources could be broken out into their own classes. This would be useful for testing purposes and maintainability. sul-embed does something similar here with its `Resources` and `ResourceFile` classes generating objects based off of purlxml: https://github.com/sul-dlss/sul-embed/blob/master/app/models/embed/purl.rb#L138-L204

This is a tough one, not sure what the best way moving forward is on all of this or how much we can actually get done.

drh-stanford:
Because there's so much refactoring to be done, perhaps we should open a new technical debt issue for it?

mejackreed:
I'm :ok_hand: with that.

drh-stanford:
ok i created https://github.com/sul-dlss/argo/issues/536

drh-stanford:
@mejackreed is there anything else I should address in this PR?

atz:
can_manage, can_view, etc. really aren't defined anywhere else in the application?  I swear I've seen the same logic before.  Helper would be reasonable.

Side note: I prefer the long-line version of those assignments instead of the ugly paragraphs.

darrenleeweber:
This PR will have conflicts with the APO PR because of the user permission calls like `can_manage` or `can_view`.  So if this is go ahead, it's a blocker for the APO PR and that will have to be rebased on this.

drh-stanford:
ok, i marked this WIP until #532 clears

darrenleeweber:
The #532 changes that conflict with this can be seen clearly in ea1b53b43831ebc006588dcc253d13a20d380cb3

drh-stanford:
I will resubmit this later...

PR #517, opened 2016-02-13
add div container for bulk update buttons

drh-stanford:
@tingulfsen this fixes the missing left edge padding problem for the Bulk Update buttons. This PR adds `<div class="container">` around the content and inserts `col-xs-offset-1` between the buttons.

screen shot 2016-02-12 at 4 06 14 pm

vs.

screen shot 2016-02-12 at 4 01 54 pm

### DIFF
--- a/app/views/report/bulk.html.erb
+++ b/app/views/report/bulk.html.erb
@@ -50,7 +50,7 @@ ul li{
 }
 </style>
 <% end %>
-
+<div class="container">
 <% @page_title = "#{application_name} Search Results" %>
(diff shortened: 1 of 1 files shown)


mejackreed:
I think the column offsets aren't needed here. @tingulfsen set them up to be the way that @LynnMcRae wanted and i think the offsets break that.

screen shot 2016-02-12 at 7 13 09 pm

atz:
So we should close this?

drh-stanford:
yeah

PR #527, opened 2016-02-17
Prevent crash when retrieving roles for an APO identifier that is nil.

### LINKED ISSUE #334: Detail display, don't die on bad APO or Collection reference
Objects shouldn't exist with a bad APO or collection reference, but they do, esp with test data. And this is exactly what the UI should show us should it actually happened for real, rather than getting a system error.  In the case of these lookups, show the druid in place of an unresolved lookup, perhaps highlighted somehow (in red?) to signal the problem?

tingulfsen:
Fixes #334

### DIFF
--- a/spec/helpers/argo_helper_spec.rb
+++ b/spec/helpers/argo_helper_spec.rb
@@ -100,6 +100,13 @@ describe ArgoHelper, :type => :helper do
           expect(buttons).to include(button)
         end
       end
+      it 'should not generate errors given an object that has no associated APO' do
+        allow(@doc).to receive(:apo_pid).and_return(nil)
+        allow(@usr).to receive(:roles).with(nil).and_return([])
+        buttons = helper.render_buttons(@doc)
+        expect(buttons).not_to be_nil
+        expect(buttons.length).to be > 0
+      end
     end
   end
   describe 'render_facet_value' do
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1,5 +1,3 @@
-# encoding: UTF-8
-
 require 'spec_helper'
 
 describe User, :type => :model do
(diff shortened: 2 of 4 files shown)


jmartin-sul:
assuming they don't leave us waiting forever, it'd be nice to see coveralls come back online and report coverage before merging (they seem to be down for maintenance).  but :+1: from looking at the code.

spec/helpers/argo_helper_spec.rb
```diff
@@ -100,6 +100,13 @@
           expect(buttons).to include(button)
         end
       end
+      it 'should not generate errors given an object that has no associated APO', :focus => true do
```
mejackreed: Why the focus?

spec/models/user_spec.rb
```diff
@@ -1,4 +1,4 @@
-# encoding: UTF-8
+# encoding: utf-8
```
mejackreed: Should we just kill this line entirely?

jmartin-sul: ah, yeah, that is unnecessary now, right?

mejackreed: Yep Ruby > 2.0

tingulfsen:
Fair comments, just altered and rebased.

tingulfsen:
This happens when we have bad data, which we have. I don't believe that a nil input is a truly exceptional case - returning an empty array was carefully chosen to ensure that the code behaves predictably and gracefully.

LynnMcRae:
and to close the loop, what we're trying to prevent is Argo from going kablooey on such bad data so we can see it and fix it. I know we have code that expects good data because regular apps should depend on it to behave correctly, while Argo is the app that needs to look under the hood. I imagine that's a dicey balance at times.

darrenleeweber:
We don't always expect good data and we handle general exceptions in the ApplicationController:
https://github.com/sul-dlss/argo/blob/issue76-apo-nice/app/controllers/application_controller.rb#L10-L26

The way things should behave depends on whether the item itself is an APO or not.  So this problem intersects with work going on to deal with that in https://github.com/sul-dlss/argo/tree/issue76-apo-nice.  I'll try to better understand how the issue in this PR related to the APO PR to evaluate whether this issue is already resolved by changes in the APO PR or if this additional fix is required.  I would have preferred this issue be evaluated after we merge the APO PR to see whether or not changes there already resolve this.

PR #529, opened 2016-02-17
Remove cruft from tests.

tingulfsen:
This removes the "focus" attribute from a couple of tests, which was accidentally left in.


mejackreed:
Coveralls seems paralyzed. :+1:

PR #530, opened 2016-02-17
Add tests for get_metadata_source().

tingulfsen:


spec/helpers/dor_object_helper_spec.rb
```diff
@@ -48,4 +48,31 @@
...
+
+      allow(id_metadata).to receive(:otherId).with('mdtoolkit').and_return(['1'])
+      expect(get_metadata_source(mock_obj)).to eq('Metadata Toolkit')
+    end
+
+    it 'should return Symphony when identityMetadata.otherId contains catkey' do
+      id_metadata = double('identity_metadata')
+      mock_obj = double('Dor::Item', { :identityMetadata => id_metadata })
+
+      allow(id_metadata).to receive(:otherId).with('mdtoolkit').and_return(['1'])
```
atz: You might position these as `expect` statements to ensure that the value you are injecting is actually getting called!

spec/helpers/dor_object_helper_spec.rb
```diff
@@ -48,4 +48,31 @@
       expect(render_status_style(nil, nil)).to eq('')
     end
   end
+
+  describe 'get_metadata_source' do
+    it 'should return Metadata Toolkit when identityMetadata.otherId contains mdtoolkit' do
+      id_metadata = double('identity_metadata')
+      mock_obj = double('Dor::Item', { :identityMetadata => id_metadata })
```
atz: First 2 lines are shared in each `it` block.  Probably could go into a `before` block, setting `@id_metadata` and `@mock_obj`.

PR #539, opened 2016-02-18
Remove outdated dependencies

mejackreed:
- Selenium-Webrick is no longer needed, as we use poltergeist
 - Unicorn is not needed, we use the default Rails server Webrick


mejackreed:
Quick smoke test on dev and everything seems fine.

Issue #523, opened 2016-02-16
Remove "edit" link from Content portion of detail display


LynnMcRae:
This shows up for any object that is editable -- registered or opened -- with content.  See https://argo-dev.stanford.edu/catalog/druid:qs553rb8234

![screen shot 2016-02-16 at 10 35 07 am](https://cloud.githubusercontent.com/assets/1824732/13086653/fcffaec8-d498-11e5-84f5-f91a627f128c.png)

drh-stanford:
does that also mean removing all the associated functionality which is available from the (edit) link? There's a bunch of stuff in there (See https://github.com/sul-dlss/argo/blob/develop/app/views/items/resource.html.erb)

LynnMcRae:
That's a call for you guys to make. There is a useful set of functionality represented by that code, but it was never quite completed. Who knows when and under what circumstances we'll get back to it.  Jack had some opinions I think, about benefits/expedience of keeping/removing inert code this week.

mejackreed:
I'd vote for removing the dead route, and dead partials associated with it. If there is a github ticket for adding this functionality later, we can link the removal commit to that ticket. That way later on down the line we can easily find the old code. 

I prefer not maintaining code that isn't executed or used. And that way we don't have to have this discussion, or a future argo dev doesn't have to spend the time figuring out what the heck that thing is.

drh-stanford:
I prefer removing the code too, but I wanted to make sure that dialog is no longer needed. 

There's at least a few routes I see that would be candidates for removal: `resource`, `add_file`, `replace_file`, `delete_file`.

mejackreed:
I think `add.png` can also be removed.

PR #541, opened 2016-02-18
item controller - resource spec should be fixed

darrenleeweber:
Marked a spec as pending, so this should fail to pass unit testing until it is fixed.

### DIFF
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -340,10 +340,10 @@ describe ItemsController, :type => :controller do
   end
   describe 'resource' do
     it 'should set the object and datastream, then call the view' do
-      expect(Dor::Item).to receive(:find)
-      allow(@item).to receive(:datastreams).and_return({'contentMetadata' => double(Dor::ContentMetadataDS)})
-      get 'resource', :id => @pid, :resource => '0001'
-      # XXX : isn't actually testing what it says!
+      skip 'test unimplemented'
+      # expect(Dor::Item).to receive(:find)
+      # expect(@item).to receive(:datastreams).and_return({'contentMetadata' => double(Dor::ContentMetadataDS)})
+      # get 'resource', :id => @pid, :resource => '0001'
     end
   end
   describe 'add_collection' do


jmartin-sul:
well... my understanding is that "pending" is for tests that are expected to fail.  but the test claims it "isn't actually testing what it says" (which doesn't necessarily mean it's expected to fail, since it indicates that the test, and not the code being tested, needs the fix).

as such, seems like "skip" is more appropriate than "pending".

darrenleeweber:
The use of `pending` is intentional.  This spec is not testing what it should and, yet, it is passing.  Unless it is marked `pending`, it will never get attention to fix it and make sure it is doing what it should be doing.  So, no, it should not be `skip` unless you never want it fixed?

mejackreed:
@darrenleeweber You have a good point. Would you mind updating to add the failing expectation? That way we know what the expected behavior is and when it is fixed our build should break because the pending spec will pass.

atz:
There is no point running this test as is, so it is a `skip` unless you can actually specify a test condition that should pass but doesn't yet.  It will still show in rspec results with the other pending tests.

darrenleeweber:
Respectfully disagree with @atz on skipping it.  Someone who knows something about this should fix it or remove it entirely if it's never going to be fixed or explain why it will not be fixed in a `skip('explanation')`.

atz:
The explanation is that the test was never fully written (likely what is there was copied).  There are many other "Test unimplemented" specs marked skip.  You are trying to use the skip/pending as a **motivational** tool or an expression of intent.  But the point is that the test isn't written, so its execution is worthless.  Pending tests are expected to fail until some anticipated moment they start passing, at which time the pending line is removed, thereby graduating into regular test status *otherwise unchanged*.  Since this wasn't written and needs to be changed it cannot be pending.  

Note: the semantics of this did change between rspec 2 and 3.  Pending tests must fail for the build to pass now.  

I'd be happy to remove/comment what is there leaving only a skip line.

atz:
Pushed a reworked version of the branch.

jmartin-sul:
my thinking in leaving the original comment is well summarized by @atz above.  i agree that skip/pending shouldn't be a way to force issue prioritization, and that the state of this test does not line up with my semantic understanding of what pending means.  as such, :+1: for the recently pushed skip version.  i'm not usually a fan of committing commented out code, but :ok_hand: with it here, because i haven't taken the time to evaluate whether the old code is even a reasonable starting point for the described test.

jmartin-sul:
filed issue #556 to revisit pending/skipped tests in general

darrenleeweber:
So I took a look at this thing and the `ItemController#resource` only pulls out content metadata DS.  It doesn't seem to do anything with it.  So, this issue might be bigger than just a skipped spec.  What the heck is the resource doing anyway?

mejackreed:
Not sure why we didn't catch this sooner, but looks like we decided to remove all of this anyways in https://github.com/sul-dlss/argo/issues/523#issuecomment-184919607

PR #549, opened 2016-02-19
Merge develop back into master

mejackreed:
This pull request aims to leave develop behind in favor of master being the primary branch of work.

*Note after this is done* 
 - existing PR's to develop should be changed to master.
 - default branch should be changed back to master


jmartin-sul:
https://en.wikipedia.org/wiki/Point_of_no_return

drh-stanford:
we should probably tag b9383debc27e6932da817a8db0eb0dad6d75535e with v2.x or something so we can look back at the good ol days... or is there already a tag that marks current prod Argo?

jmartin-sul:
@drh-stanford:  branched https://github.com/sul-dlss/argo/tree/2.x-stable when we merged master into develop so that this merge would be a fast-forward.  so should be all good.

PR #550, opened 2016-02-19
Use Time.zone.parse to parse user supplied date string.

### LINKED ISSUE #545: Update embargo date-time with zone?
Rubocop complains about:
```
app/controllers/items_controller.rb:243:37: C: Do not use DateTime.parse without zone. Use one of Time.zone.parse, DateTime.current, DateTime.parse.in_time_zone, DateTime.parse.utc, DateTime.parse.getlocal, DateTime.parse.iso8601, DateTime.parse.jisx0301, DateTime.parse.rfc3339, DateTime.parse.to_i, DateTime.parse.to_f instead.
    @object.update_embargo(DateTime.parse(params[:embargo_date]))
```
How is this to be resolved?  What date-time should be used for an embargo update?  Also, how does this interact with the datastore settings (if any could modify the ruby code assumptions)?

mejackreed:
Parsing the date this way gives us the same timezone that dor-services does.

Closes #545

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -236,7 +236,7 @@ class ItemsController < ApplicationController
       return
     end
     fail ArgumentError, 'Missing embargo_date parameter' unless params[:embargo_date].present?
-    @object.update_embargo(DateTime.parse(params[:embargo_date]))
+    @object.update_embargo(DateTime.parse(params[:embargo_date]).utc)
     @object.datastreams['events'].add_event('Embargo', current_user.to_s , 'Embargo date modified')
     respond_to do |format|
       format.any { redirect_to catalog_path(params[:id]), :notice => 'Embargo was successfully updated' }


darrenleeweber:
Is rubocop OK with this form?  If so, it's :+1: for merge.

drh-stanford:
Isn't this redundant with what dor-services is already doing with the embargo date?  

  https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/models/embargoable.rb#L72

mejackreed:
@drh-stanford Yes it it.

atz:
Looks somewhat redundant, but I'm not against it.  But I'm all for converting to UTC as early as possible and only ever passing dor-services UTC dates.

jmartin-sul:
unless there's some double-encoding-esque corruption that this could cause, i'd be fine with this redundancy.  the caller wants to make sure it parses to utc.  dor-services wants to make sure it saves utc.  doesn't seem like a particularly costly bit of repetition.

PR #553, opened 2016-02-19
reindex and commit newly registered objects before returning

### LINKED ISSUE #531: Continuing Indexing Errors Post-Registration in -stage
1) User registers x number of objects
2) Objects appear in Fedora
3) Objects do not appear in Argo
4) User is forced to manually reindex all objects in order to get them to appear in Argo

drh-stanford:
This PR fixes #531 by using the dor-services IndexingService to synchronously update the index, without waiting for the reindexing pipeline to engage. @jmartin-sul can you please review this?

### DIFF
--- a/app/controllers/dor/objects_controller.rb
+++ b/app/controllers/dor/objects_controller.rb
@@ -9,6 +9,16 @@ class Dor::ObjectsController < ApplicationController
     response = Dor::RegistrationService.create_from_request(params)
     pid = response[:pid]
 
+    #
+    # we need to reindex and *commit* this new object so that source id checks
+    # (which read from the index) work in future registrations.
+    #
+    # TODO: thus note that this create endpoint will have race conditions if the client
+    # sends registration requests concurrently -- ideally this registration should be
+    # a serialized process
+    #
+    Dor::IndexingService.reindex_pid_list([pid], true)
+
     respond_to do |format|
       format.json { render :json => response, :location => object_location(pid) }
       format.xml  { render :xml  => response, :location => object_location(pid) }
--- a/spec/controllers/dor/objects_controller_spec.rb
+++ b/spec/controllers/dor/objects_controller_spec.rb
@@ -10,6 +10,9 @@ describe Dor::ObjectsController, :type => :controller do
       expect(Dor::RegistrationService)
         .to receive(:create_from_request)
         .and_return(dor_registration)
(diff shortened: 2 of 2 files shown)


jmartin-sul:
weird... what happened to the coveralls check?

re:  the meat of the PR...

i'm totally :ok_hand: on this... but the `reindex_pid_list` call on a single pid got me thinking... it's weird that `reindex_pid_list` has a commit option, but `reindex_pid` doesn't.  but adding more options to `reindex_pid` feels like it'd make it too unwieldy, and it seems like you might want to wrap other operations or sets of operations in a commit anyway.

so... i think i'm going to file an issue in dor-services to create something like a `solr_commit` method that takes a block, executes it, and does a `Dor::SearchService.solr.commit`.  then i'll update callers who called `reindex_pid_list` with commit=true to use the new commit wrapper (and get rid of the option in `reindex_pid_list`).  as well as callers who called `reindex_pid` and did a commit inline afterwards.

in the meantime, i'm totally fine to merge this.  if it were me, i'd have probably called `reindex_pid` and called commit inline after that, but that doesn't seem any more semantically correct than what you did here, and i would not advocate changing to that approach unless you also think it reads better.

well, totally fine to merge pending explanation of the coveralls report's disappearance.

jmartin-sul:
created https://github.com/sul-dlss/dor-services/issues/174 for the above pedantry

drh-stanford:
maybe it's gone coz we're on `master` now? Strange.

drh-stanford:
i deleted and reuploaded the branch and that kicked off both the Travis builds. i guess it was github weirdness

atz:
The semantics on the Solr commit block will not be able to be strictly maintained.  Solr has no sessions.  I.E., it is entirely possible that in the middle of a block posting 1000 solr documents individually, that somebody *else* accessing solr will demand a synchronous update/commit.  Whatever you have done up to that point will be committed even though your block has not yet completed.  If the block later fails, there is no rollback.  This isn't an argument against using the block syntax, just a proviso that you aren't getting DB-like transactionality out of it.

jmartin-sul:
ah, yes, good clarification.  i wasn't expecting transactional semantics, just to run a commit after running this random collection of stuff.  i guess the only burden that's relieved is that the caller doesn't have to know how to commit, just that it wants to.  i don't know, maybe that risks confusion because people expect transactions from commit blocks.

PR #532, opened 2016-02-17
[WIP] Issue76 apo nice

darrenleeweber:
This fixes #76 

This is blocked by https://github.com/sul-dlss/argo/pull/563, which will replace the "Consolidate Argo user groups in the User model" (commit 347e7fd8f1fd2f2b02788c9bed83b3f7cc2e1a66).  That PR also has some overlap with the work in "General review of user roles and groups" (commit 653b702868a9ec8939ce47276e5b7ce09b68cd33, mostly code style/efficiency changes).  @jmartin-sul and I have previously reviewed most of those changes and they might get pulled out of this PR.

The core of this PR is in `User#can_admin?`, `User#can_manage?` and `User#can_view?`
(commit 02e697c50439ff0c50f2a8d2926a47b4d2af672c).  These public methods call a very important private method.  The essence of the permission refactoring is in the new logic in `User#request_object_permission`.  It should be possible to make `User#roles` a private method also, but it is left public and the further analysis and discussion required for that can be another PR.

The use of these methods in the application is in "Apply User#can_admin?, User#can_manage? and User#can_view?" (commit ea1b53b43831ebc006588dcc253d13a20d380cb3).

The last part of this PR is the use of the new User model in specs.  The main part of this is to create methods in the spec_helper and use them consistently in specs that require a certain type of user (mostly `admin_user`).  All of this is in "Revise specs after application of User#can_admin?, User#can_manage? ..." (commit 676f8b48b0efdbf844ef8d004cebe5b6c0675d7a).  Part of this work motivated changes for more consistent `id` parsing in the `find_druid` work, has been isolated in https://github.com/sul-dlss/argo/pull/555.  So the WIP on this PR is trying to separate the changes from those, otherwise this PR is blocked by #555 because the commits there need to be disentangled from those in this PR.

PR #555, opened 2016-02-19
[WIP] Find druid

darrenleeweber:
Opening this PR for commentary.  @mejackreed, @jmartin-sul and @drh-stanford will want to take a look.  I'm expecting a couple of tweaks for this, but it's close.

Blocks #532 

Part of the focus of this PR is on consistent parsing of `params[:id]`.  For example, catalog controller contains various places when this parameter is parsed, including some inconsistency in form, e.g.:
```
params[:id] = 'druid:' + params[:id] unless params[:id].include? 'druid'
```
and
```
pid = params[:id].include?('druid') ? params[:id] : "druid:#{params[:id]}"
```
If routes were modified to use, say `pid` instead of `id`, then all these lines from every controller that parses this param need to be modified.  So, for example, if the pattern really should be:
```
params[:id] = 'druid:' + params[:id] unless params[:id] =~ /^druid:/
```
Then how many lines of code need to be touched to make that change?  And should all these changes be assigned back into `params[:id]` or to a new variable, like `pid`?

Also, if Dor is supposed to handle any `id`, then these patterns completely mess up anything that has `druid` *somewhere* in the string, like `google-scanned-druid` could be an object ID that would become `druid:google-scanned-druid` and that would be a mistake.  The DruidTools project has consolidated the regex matching and string munging for druids, so Argo should use it.


darrenleeweber:
There are some additional places where `Dor.find` is used, but it's not clear whether they can use the `ApplicationController#find_druid`, i.e.
```
 ack-grep 'Dor.find' app/
app/jobs/modsulator_job.rb
74:        dor_object = Dor.find current_druid

app/jobs/descmetadata_download_job.rb
24:              dor_object = Dor.find current_druid

app/controllers/application_controller.rb
70:  # input and returns a uniform DRUID, which is input to Dor.find to obtain a Dor
72:  # a UI alert and log a warning, but allow Dor.find to try to get it.
75:  # @param options [Hash] options for Dor.find
76:  # @return object [ActiveFedora::Base] anything returned by Dor.find
78:  # @raises exceptions from Dor.find
84:              # Allow Argo to find anything using Dor.find,
86:              # The Dor.find method should be read-only access.
92:    Dor.find(druid, options)

app/helpers/value_helper.rb
10:        item ||= Dor.find(druid)

app/helpers/dor_object_helper.rb
54:    Dor.find(obj_id)

app/helpers/argo_helper.rb
86:    dor_object = @obj # Dor.find doc['id'].to_s, :lightweight => true
104:    object ||= Dor.find(pid)
```

darrenleeweber:
The changes in `app/controllers/registration_controller.rb` might have to be pulled out of this.

darrenleeweber:
The change from `@object` to `@obj` in `app/views/catalog/bulk_upload_form.html.erb` is just because most of the code in the controller that calls that view is using `@obj`, so it's consistent with the controller.  However, that controller is not consistent with others that use `@object`.

darrenleeweber:
The change to an item fixture in `spec/controllers/items_controller_spec.rb` is working, but might have to come out of this changeset.

darrenleeweber:
@drh-stanford notes in https://github.com/sul-dlss/argo/pull/532#discussion_r53531149 that:
`Dor::Item.find` and `Dor.find` do not follow the same logic, so calling `find_druid` (which uses `Dor.find`) in this case changes the behavior.

I would like to understand how it changes the behavior and why the Argo specs should differ depending on whether `Dor.find` is used or `Dor::Item.find`.  It's not obvious from reading code in dor-services, because it seems that `Dor.find` tries to use Solr to identify the object type and ensure the object returned has the correct type.  So, I would expect `Dor.find` to return the same object as `Dor::Item.find` when the object is a `Dor::Item`.  It could be a mistake to use `Dor::Item.find` when the object is not a `Dor::Item` or the object type is ambiguous.  Given that Argo can receive anything in `params[:id]`, the object type should be assumed to be ambiguous.  If an action requires a specific object type, it should use `Dor.find` first and then verify that the object requested is of the required type for the action.  It should not simply assume the `params[:id]` identifies an object of the required type.  So, `Dor.find` appears to be a unified API for retrieving a Dor object and it would make sense for Argo to use it consistently.

atz:
A repository object can have potentially many models.  Calling `Dor::Item.find` should guarantee that the returned object is a `Dor::Item`, even if that is not its primary model.  It should fail if the object cannot be instantiated as a `Dor::Item`.  `Dor.find`, on the other hand says "I'll take whatever class the data says the object prefers".  It would be used in places where the same code applies to both Collections and Items, for example.  Or Collections and Sets.  Or places where the thing is, at the very least, expected to not necessarily be a Dor::Item.  We don't have a great variety of models, so it might seem like the two `find`s are the same, but the difference between them is important.  

I am not in favor of any druidtools validation checking for retrievals beyond appending "druid:" where it might be wanted.  I favor strong validation for **writes**, but this is mostly a non-issue since druids are minted externally.  In particular, I don't see what makes this a blocker.  This is not a must-have.

darrenleeweber:
So, is there an option to `Dor.find` that can specify that it should use a specific model loader, such has `Dor::Item.find`?

This is disentangling APO work that is a blocker, so this is blocking a blocker.  We discussed the use of `find_druid` in slack where you and others were in favor of using it.  I don't understand the change of tune.  There are a lot of places where controllers are using regex patterns to parse `param[:id]`, so any change to that pattern has to touch a lot of the same lines.  It makes sense to consolidate that into a single location that can provide consistent behavior in both parsing and exception handling.

darrenleeweber:
Something really weird is going on in travis recently; specs are all passing on a laptop.

darrenleeweber:
Maybe this thing needs to morph into simply a parser for the `id` param, without the additional call to `Dor.find`?

drh-stanford:
While I appreciate the confusion around the `.find` in dor-services, I don't think using `Dor.find` instead of `Dor::Item.find` is the right solution. Instead I recommend we fix the dor-services API itself before changing how Argo uses it. As I understand it, `Dor::Item.find` calls `ActiveFedora::Base.find` directly whereas `Dor.find` does a Solr call to determine the class (and why it doesn't just fetch the Fedora datastream for that info I dunno) then loads the instance. So, I'd be in favor of clarifying the `.find` API in a dor-services 6.x or something and updating Argo (and importantly other downstream apps too) at that time.

darrenleeweber:
Curious about your thoughts on sul-dlss/dor-services#177 @drh-stanford ...

jmartin-sul:
i would be more in favor of what @darrenleeweber proposed a couple comments up in response to observations by @atz and @drh-stanford:  it seems more appropriate to just make parsing of the id parameter more consistent, but to leave the current choices of `Dor.find` vs `Dor::Item.find` as-is in Argo for now.  i agree that the current behavior is confusing and needs refinement, but it doesn't seem like this is the time or place to do so.  i think we could also add the functionality @darrenleeweber describes in https://github.com/sul-dlss/dor-services/issues/177 in dor-services v5 if the new functionality was added w/o changing existing behavior, but i think we can defer discussing implementation details of that sort of overhaul till later.

i can personally corroborate differing behavior between `Dor.find` and `Dor::Item.find`.  i recall investigating a bug @rosy1280 found, and initially failing to reproduce it in the console, because i was getting subtly different behavior (different available datastream methods, IIRC).  this was because i was using `Dor.find` in my initial testing, when the code with the bug was using `Dor::Item.find`.  bug's in jira, was fixed a while back, i can look it up if people are really curious.

re: my prior opinion on the `find_druid` method:  i wouldn't really say that i was ever *in favor* of using it... i was initially really opposed to having it handle the `Dor.find` call as well as the web UI error messaging when something wasn't found.  that seemed like a conflation of disparate concerns, and trouble for future callers who weren't aware of the UI part, which wasn't apparent from the name.  i was more ok with the overhauled `find_druid` that only found druids, esp once it didn't limit itself to valid druids.  but it never felt like something that was necessary or a clear improvement, and in light of the reminder about differing `find` method behaviors, i'd definitely be opposed to refactoring this code in the way that this PR proposes.

actually, i'm looking over this again, and it appears that the flash message setting is back?  i thought that was the one thing we initially agreed didn't belong in `find_druid`?

darrenleeweber:
Marking this PR as [WIP] and it might be replaced with something similar that only deals with parsing the `params[:id]` to consolidate how that should work.

jmartin-sul:
since the direction of this PR is uncertain, and since it'd have to be rebased anyway, i'd propose closing it, so as to get rid of outstanding PRs against develop (so that we can delete the develop branch).

darrenleeweber:
Move to #610 on master

jmartin-sul:
closing since this has been moved to #610

PR #557, opened 2016-02-20
cap for stage-a & prod-a

eefahy:
Adds capistrano for new load balancer ready nodes. After I rebuild `argo-stage-b` and `argo-prod-b` to use the remote mysql machine I will clean up these cap environments to deploy both `-a` and `-b` nodes at the same time.

### DIFF
--- a/config/deploy/prod-a.rb
+++ b/config/deploy/prod-a.rb
@@ -1,0 +1,8 @@
+server 'argo-prod-a.stanford.edu', user: 'lyberadmin', roles: %w(web db app)
+
+Capistrano::OneTimeKey.generate_one_time_key!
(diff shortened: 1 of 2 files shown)


config/deploy/prod-a.rb
```diff
@@ -0,0 +1,11 @@
+server 'argo-prod-a.stanford.edu', user: 'lyberadmin', roles: %w(web db app)
+
+## argo-prod-b will be used to generate a new (dev) Solr index from PRODUCTION fedora.
```
cbeer: These comments seem out of place, or maybe unnecessary?

eefahy:
Good catch @cbeer - removed comments.

PR #562, opened 2016-02-21
Consistent use of @obj in catalog controller

darrenleeweber:
Pulled out of https://github.com/sul-dlss/argo/pull/532

The motivation for this is simply consistent use of `@obj` in the catalog controller.  Although other controllers may use `@object`, changing the catalog controller and all it's views to match may be a much bigger task than making the bulk action(s) and views use `@obj`.

This is ready for review and I'm guessing @tingulfsen could review this one.


mejackreed:
:+1:

PR #563, opened 2016-02-22
Consolidate Argo user groups/roles in the User model

darrenleeweber:
Extracted from https://github.com/sul-dlss/argo/pull/532.  This is a minimal change to the Argo permissions to remove an obscure init file and put these values in the User model, where they belong.  This was discussed in slack where a consensus agreed this was a good approach.

The file removed was `config/initializers/permissions.rb` and the values were transferred to
`app/models/user.rb`.  Apart from `lib/tasks/argo.rake`, the `User` model is the only place these values are used in the application.  All other changes are spec files that need the new `User::` namespace to use the values.

As discussed in slack and confirmed by @LynnMcRae , the `dor-*` roles are deprecated (`dor-admin`, `dor-viewer` and `dor-manager`).  Lynn confirmed that these user roles are removed from APO metadata.

This blocks #532 and it is ready for review.  @mejackreed or @jmartin-sul is likely already familiar with this code from other APO reviews.

### DIFF
--- a/spec/integration/about_page_spec.rb
+++ b/spec/integration/about_page_spec.rb
@@ -2,10 +2,15 @@ require 'spec_helper'
 
 describe 'about_page', :type => :request do
   before :each do
-    @current_user = double(:webauth_user, :login => 'sunetid', :logged_in? => true, :privgroup => ADMIN_GROUPS.first)
+    @current_user = double(
+      :webauth_user,
+      :login => 'sunetid',
+      :logged_in? => true,
+      :privgroup => User::ADMIN_GROUPS.first
+    )
     allow(@current_user).to receive(:is_admin).and_return(true)
-    allow(@current_user).to receive(:roles).and_return([])
     allow(@current_user).to receive(:is_manager).and_return(false)
+    allow(@current_user).to receive(:roles).and_return([])
     allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(@current_user)
   end
 
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1,5 +1,8 @@
 require 'spec_helper'
 
+# General documentation about roles and permissions is on SUL Consul at
+# https://consul.stanford.edu/display/chimera/Repository+Roles+and+Permissions
+
 describe User, :type => :model do
   describe '.find_or_create_by_webauth' do
     it 'should work' do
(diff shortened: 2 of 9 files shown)


spec/models/user_spec.rb
```diff
@@ -151,30 +206,31 @@
...
       mock_webauth = double('webauth', :login => 'asdf', :logged_in? => true, :privgroup => webauth_privgroup_str)
       user = User.find_or_create_by_webauth(mock_webauth)
-      expect(user.is_admin  ).to be_truthy
-      expect(user.is_manager).to be_truthy
-      expect(user.is_viewer ).to be_truthy
-      user.set_groups_to_impersonate(['workgroup:sdr:not-an-administrator-role', 'workgroup:sdr:not-a-manager-role', 'workgroup:sdr:not-a-viewer-role'])
-      expect(user.is_admin  ).to be_falsey
-      expect(user.is_manager).to be_falsey
-      expect(user.is_viewer ).to be_falsey
+      expect(user.is_admin).to be true
```
mejackreed: So it seems like we have gone back in forth between `be_truthy` and `be true` and `eq true` a few times. Could we discuss the reasons for the historical and current changes?

@atz 

https://github.com/sul-dlss/argo/commit/bc56d4039d37ade6cc2aa86b9c7cba2cd03164be

darrenleeweber: These methods are all boolean and should not be used to extract any data, so they must return simply `true` or `false` and they do.  It would be a mistake for them to return any other values.  If the `truthy` and `falsy` tests persist, the specs are not so precise.  They would allow these methods to return other kinds of values.  I use these forms when I want a method that can be used in logical tests and used to return some form of data or nil.

atz: I prefer truthiness in the same way I prefer `be_nil` to `be(nil)`.  At some point it was recommended by transpec or one of the tools. It isn't a big deal though.  

This test was previously testing before/after impersonation.  Now it isn't.  That seems weird.

mejackreed: :+1: makes a lot of sense @darrenleeweber

darrenleeweber: "This test was previously testing before/after impersonation."  Right, that must be something lost in `git add -p` while trying to restrict the scope of this PR.  I'll double check this and fix it ASAP.

darrenleeweber:
OK, this was pulled out of this, where the impersonation is consolidated, see
https://github.com/sul-dlss/argo/blob/issue76-apo-nice/spec/models/user_spec.rb#L239-L309
I'll have to review the details of what goes into this PR.

spec/models/user_spec.rb
```diff
@@ -40,54 +40,109 @@
...
+      allow(subject).to receive(:groups).and_return(User::VIEWER_GROUPS)
+      expect_admin false
     end
   end
 
   describe 'is_manager' do
+    def expect_manager(value)
+      expect(subject.is_manager).to be value
+      expect(subject.is_manager?).to be value
+    end
```
jmartin-sul: to me, this seems like overkill, and i find that it makes the below tests a bit harder to read.  the equivalence of `is_manager` and `is_manager?` is due to an alias.  it seems fine to me to trust that `alias` works as advertised.  if you wanted to be really paranoid, it seems preferable to test once that calling `is_manager?` just calls `is_manager`, rather than calling both each time in a wrapper method.

spec/models/user_spec.rb
```diff
@@ -40,54 +40,109 @@
...
+      allow(subject).to receive(:groups).and_return(User::VIEWER_GROUPS)
+      expect_manager false
     end
   end
 
   describe 'is_viewer' do
+    def expect_viewer(value)
+      expect(subject.is_viewer).to be value
+      expect(subject.is_viewer?).to be value
+    end
```
jmartin-sul: ditto `expect_manager`

spec/models/user_spec.rb
```diff
@@ -40,54 +40,109 @@
   end
 
   describe 'is_admin' do
+    def expect_admin(value)
+      expect(subject.is_admin).to be value
+      expect(subject.is_admin?).to be value
+    end
```
jmartin-sul: ditto `expect_manager`

spec/models/user_spec.rb
jmartin-sul: yeah, also agreed w/ @darrenleeweber's reasoning here.  seems fine to prefer truthiness in general, but it also seems reasonable to expect actual boolean values in this specific test.

"This test was previously testing before/after impersonation" would be worrisome... but i don't see it?  i might be missing something, but it still appears to be testing before/after impersonation, as far as i can tell...

darrenleeweber:
Will address points from @jmartin-sul

jmartin-sul:
overall, :+1: on the general direction of this refactoring, other than the testing suggesting i had, and the coverage worry @atz had.

spec/models/user_spec.rb
```diff
@@ -40,94 +44,182 @@
   end
 
   describe 'is_admin' do
+    it 'is aliased to is_admin?' do
+      allow(subject).to receive(:groups).and_return(User::ADMIN_GROUPS)
+      expect(subject).to respond_to('is_admin?')
+      expect(subject.is_admin).to be true
+      expect(subject.is_admin?).to be true
+      expect(subject.is_admin).to be subject.is_admin?
+    end
```
darrenleeweber: This spec addresses concerns of @jmartin-sul and the same pattern is applied for the other aliases.

spec/models/user_spec.rb
```diff
@@ -151,30 +243,69 @@
...
+      expect(user.is_admin).to be true
+      expect(user.is_manager).to be true
+      expect(user.is_viewer).to be true
+      user.set_groups_to_impersonate(%w(workgroup:sdr:not-an-administrator-role workgroup:sdr:not-a-manager-role workgroup:sdr:not-a-viewer-role))
+      expect(user.is_admin).to be false
+      expect(user.is_manager).to be false
+      expect(user.is_viewer).to be false
     end
   end
+
```
darrenleeweber: The specs below for `set_groups_to_impersonate` might be what @atz is looking for?

spec/models/user_spec.rb
```diff
@@ -1,5 +1,9 @@
+# encoding: UTF-8
```
mejackreed: Why is this added back? Think it can be removed.

spec/integration/about_page_spec.rb
```diff
@@ -2,10 +2,15 @@
...
+    @current_user = double(
+      :webauth_user,
+      :login => 'sunetid',
+      :logged_in? => true,
+      :privgroup => User::ADMIN_GROUPS.first
+    )
     allow(@current_user).to receive(:is_admin).and_return(true)
-    allow(@current_user).to receive(:roles).and_return([])
     allow(@current_user).to receive(:is_manager).and_return(false)
+    allow(@current_user).to receive(:roles).and_return([])
```
mejackreed: Why the changed ordering here? (and other places?)

jmartin-sul: i also found this a bit unnecessary, but am :ok_hand: with it, in that i wouldn't stop the PR over it.

spec/models/user_spec.rb
jmartin-sul: :+1:

spec/models/user_spec.rb
atz: Agree, removed.

spec/models/user_spec.rb
atz: As revised, I think that is fixed.

spec/integration/about_page_spec.rb
mejackreed: Agreed... just curious about why the change?

spec/models/user_spec.rb
```diff
@@ -40,94 +44,216 @@
...
+    end
+    it 'should return an empty array for any blank object identifer' do
+      ['', nil].each do |pid|
+        expect{subject.roles(pid)}.not_to raise_error
+        expect(subject.roles(pid)).to be_empty
+      end
+    end
+    it 'should build a set of roles from groups' do
+      doc_groups = %w(workgroup:dlss:groupF workgroup:dlss:groupA)
+      doc_roles = %w(sdr-administrator sdr-viewer).sort
```
jmartin-sul: i'd submit that these would be more appropriately named as `user_groups` and `user_roles`, since the former is returned by the mock user, and the latter is an expected result based on the intersection of the mock user's groups and the groups listed for roles on the mock `solr_doc`.

jmartin-sul:
pretty much :+1:.  would be nice to ditch the stray `# encoding: UTF-8` that crept back in.  and the `doc_groups`/`doc_roles` thing if i'm not off-base there and that's not too much trouble.  otherwise, looks good to me.

spec/integration/about_page_spec.rb
darrenleeweber: Oh, I like to see all the `is_admin` or `is_manager` stuff precede any `roles` stuff because they are closely related and higher precedence than roles.  It's the mathematician in me, I guess.

darrenleeweber:
OK, thanks for reviews.  About to squash in requested changes, i.e.
```
-# encoding: UTF-8
 require 'spec_helper'
 
 # General documentation about roles and permissions is on SUL Consul at
@@ -228,10 +227,10 @@ describe User, :type => :model do
       end
     end
     it 'should build a set of roles from groups' do
-      doc_groups = %w(workgroup:dlss:groupF workgroup:dlss:groupA)
-      doc_roles = %w(sdr-administrator sdr-viewer).sort
-      expect(subject).to receive(:groups).and_return(doc_groups).at_least(:once)
-      expect(subject.roles(druid)).to eq(doc_roles)
+      user_groups = %w(workgroup:dlss:groupF workgroup:dlss:groupA)
+      user_roles = %w(sdr-administrator sdr-viewer)
+      expect(subject).to receive(:groups).and_return(user_groups).at_least(:once)
+      expect(subject.roles(druid)).to eq(user_roles)
     end
```
After that, the user specs pass with 100% coverage on the user model.

spec/models/user_spec.rb
darrenleeweber: :ok_hand:

spec/models/user_spec.rb
darrenleeweber: :ok_hand: -- it had a `doc` prefix because I was thinking of it related to Solr content, but I agree with the change.

mejackreed:
:+1:

PR #564, opened 2016-02-22
Fix strange JS failures in workflow_service_creation_spec.rb

darrenleeweber:
This is pulled out of https://github.com/sul-dlss/argo/pull/532.  This arose from mysterious JS errors while running feature specs that were failing to find a UI button.  This change adds a CSS #id to the button and adds some checks in the specs.  The strange behavior was discussed with @mejackreed in slack and he may want to review this.

While working on this, discovered that the feature spec is very sensitive to the format of the `id`.  That is, it accepts `druid:qq613vj0238` but not `qq613vj0238`.  With the latter, my system raises:
```
  1) Workflow Service Creation redirect and display on show page - with JS
     Failure/Error: @object = Dor::Item.find params[:id]
     
       See logger for details
     # ./app/controllers/items_controller.rb:716:in `create_obj'
```
This sensitivity (failure) is the motivation for more consistent parsing of the `id`, which is in the PR at
https://github.com/sul-dlss/argo/pull/555

### DIFF
--- a/spec/features/workflow_service_creation_spec.rb
+++ b/spec/features/workflow_service_creation_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.feature 'Workflow Service Creation' do
+feature 'Workflow Service Creation' do
   let(:current_user) do
     double(
       :webauth_user,
@@ -11,15 +11,17 @@ RSpec.feature 'Workflow Service Creation' do
       roles: []
     )
   end
-  before do
-    allow_any_instance_of(ItemsController).to receive(:current_user)
-      .and_return(current_user)
-    allow_any_instance_of(CatalogController).to receive(:current_user)
-      .and_return(current_user)
+  before :each do
+    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(current_user)
+    @druid = 'druid:qq613vj0238' # a fixture Dor::Item record
   end
-  scenario 'redirect and display on show page' do
-    visit add_workflow_item_path 'druid:qq613vj0238'
-    click_button 'Add'
+
+  scenario 'redirect and display on show page - with JS', js: true do
+    visit add_workflow_item_path(@druid)
+    expect(page).to have_content('Add workflow')
+    expect(page).to have_button('Add')
+    find('#wf').find(:option, 'accessionWF').select_option
+    find('#add_wf_button').trigger('click')
     within '.flash_messages' do
       expect(page).to have_css '.alert.alert-info', text: 'Added accessionWF'
     end
--- a/app/views/items/_add_workflow.html.erb
+++ b/app/views/items/_add_workflow.html.erb
@@ -2,7 +2,7 @@
   <div class='form-group'>
     <%= select_tag :wf, options_for_select(workflow_options, default_workflow_option), class: 'form-control' %>
   </div>
-  <button class='btn btn-primary'>
+  <button id='add_wf_button' class='btn btn-primary'>
     Add
   </button>
 <% end %>


spec/features/workflow_service_creation_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.feature 'Workflow Service Creation' do
+feature 'Workflow Service Creation' do
```
mejackreed: Curious, why the change here?

spec/features/workflow_service_creation_spec.rb
```diff
@@ -11,15 +11,17 @@
...
-      .and_return(current_user)
+  before :each do
+    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(current_user)
+    @druid = 'druid:qq613vj0238' # a fixture Dor::Item record
   end
-  scenario 'redirect and display on show page' do
-    visit add_workflow_item_path 'druid:qq613vj0238'
-    click_button 'Add'
+
+  scenario 'redirect and display on show page - with JS', js: true do
```
mejackreed: Why do we need to test this with JavaScript on?

mejackreed:
I really like how you were able to pull this out!

A few questions inline, and one overall. Why did we make this a js enabled spec? Is there additional functionality that is tested by that?

spec/features/workflow_service_creation_spec.rb
darrenleeweber: The Rspec configure loads the namespace already.

spec/features/workflow_service_creation_spec.rb
darrenleeweber: That's part of the puzzle.  Weirdness stopped when it was turned on.  In this version of the spec, I think it is required for `find` to get the CSS # for button.  That's a guess, I don't know for sure.  I've had problems with `click_button` before.  I'm not entirely convinced that this revision is the right way or the best way to do things, I just know that it gave more consistent spec results.

spec/features/workflow_service_creation_spec.rb
mejackreed: :ok_hand:

spec/features/workflow_service_creation_spec.rb
mejackreed: Hmm. Yeah that stuff can be strange some times. I think the general consensus most times is to prefer non-js specs unless js is needed for them. Maybe let's put a pin in this PR until the other PR's shake out?

darrenleeweber:
Why does the spec test both the flash message and some other page content as well?  The controller only returns a flash message.
```
    within '.flash_messages' do
      expect(page).to have_css '.alert.alert-info', text: 'Added accessionWF'
    end
    expect(page).to have_css 'tr td a', text: 'accessionWF'
```

It seems like it should only test the flash message, like:
```
    within '.flash_messages' do
      expect(page).to have_css '.alert.alert-info', text: 'Added accessionWF'
    end
```

darrenleeweber:
I'm now running into some different problems with `RestClient` exceptions in this spec and the solution that is working now is:
```
feature 'Workflow Service Creation' do
  let(:item) do
    object = instantiate_fixture('hj185vb7593', Dor::Item)
    allow(Dor::Item).to receive(:find).with(object.pid).and_return(object)
    object
  end
  before :each do
    @current_user = admin_user # see spec_helper
  end
  scenario 'redirect and display on show page' do
    expect(item).to receive(:create_workflow).and_return(true)  # don't do the HTTP stuff
    visit add_workflow_item_path item.pid
    click_button 'Add'
    within '.flash_messages' do
      expect(page).to have_css '.alert.alert-info', text: 'Added accessionWF'
    end
    #only check the flash message because that's all the controller returns
    #and redirects (after the redirect, it no longer controls the page content).
  end
end
```

jmartin-sul:
can this be marked as [WIP] and rebased on master, so that we can delete the develop branch?

darrenleeweber:
Moving this #609 on master.

jmartin-sul:
closing since this has been moved to #609

PR #568, opened 2016-02-22
Removes defunct resource editing functionality (backported to develop)

mejackreed:
This commit removes legacy and non-working resource editing functionality. This functionality is broken and if needed, should be rewritten accordingly with updated patterns.

Is this helpful @darrenleeweber ?


cbeer:
:clap:

PR #569, opened 2016-02-22
Fix logical operators for user permission in views

darrenleeweber:
Simple PR to correct a couple of logical operators.

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -39,7 +39,7 @@
         <li>
           <%= link_to 'Register Items', register_items_path %>
         </li>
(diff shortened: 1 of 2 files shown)


mejackreed:
Seems good to me. Should we submit this to `master`?

darrenleeweber:
After merge to develop, also merge to master?  Yes, I suppose it belongs there too.  It's another piece out of the APO that is based on develop and I don't know if it's OK on master.

jmartin-sul:
i assumed we'd merge develop back to master en-masse again this week once develop's settled down.  i'd be inclined to let this get merged to master via that process (rather than a specific backport commit for this).

darrenleeweber:
:+1:

PR #570, opened 2016-02-22
revision of  app/models/argo/access_controls_enforcement.rb

darrenleeweber:
There is no substantial functional change in this commit, it just adopts better ruby patterns (probably easier to maintain).


jmartin-sul:
:+1: from me.  PR is focused, code looks functionally equivalent, reads better, tests pass (and thankfully this actually appears to be tested already).

would be nice to get a second set of eyes since it's access control, though.

darrenleeweber:
Confirming that simplecov report still has this at 100% coverage.

PR #576, opened 2016-02-23
Remove unused dependencies

cbeer:
- `solrizer` and `kaminari` are transitive dependencies and don't need to be explicitly declared
- `kgio` was introduced in https://github.com/sul-dlss/argo/commit/c18bd49d4c8e8a548d458a78fa4bf308617eb329, but wasn't removed when `dalli` was.
- `progressbar` was supporting the `reindex_all` rake task, which has since been rewritten without the progress bar.
- `unicode` was introduced when upgrading to blacklight 4 (https://github.com/sul-dlss/argo/commit/924a3a8092af917216df61b770f43a39282eb98e) but we don't seem to be using it
- `rsolr-client-cert` isn't needed with our Solr Cloud setup.
- `daemons` was added at the same time as delayed job, but seems unused (https://github.com/sul-dlss/argo/commit/a7d1a094543ed0b54887baf45e40ecf111dbe6bc)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,6 @@ gem 'bootstrap-sass'
 gem 'coderay'
 gem 'coffee-rails'
 gem 'config'
-gem 'daemons'
 gem 'delayed_job'
 gem 'delayed_job_active_record'
 gem 'equivalent-xml', '>= 0.6.0'   # For ignoring_attr_values() with arguments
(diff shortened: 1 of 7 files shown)


mejackreed:
Thanks for the contribution! :+1: Tested a deploy to dev successfully

PR #578, opened 2016-02-24
catch develop up to master

jmartin-sul:
so that master can be fast-forwarded to develop when develop settles down.

the only thing i had to resolve by hand involved these commits:
https://github.com/sul-dlss/argo/commit/bd9a87d1b7d07f78b86678fabedcd5d1456d04fd#diff-594c25bef59b1b4465c75aaaf8eebcf4
https://github.com/sul-dlss/argo/commit/c06311bf7425d7a040c6d9a44056439c001d24e7#diff-594c25bef59b1b4465c75aaaf8eebcf4
https://github.com/sul-dlss/argo/commit/d6cf12538c5e3544bbfb9bda6147c6f5498d83e1#diff-594c25bef59b1b4465c75aaaf8eebcf4

note that the first two share the same diff against what appears to be the same code, on master and develop, respectively (though they don't share a parent commit).  the third commit builds on the first one, on master.  it seemed like a reasonable refactoring that we'd want to keep, so i kept what was on master.  my guess is that the only reason it didn't apply automatically is because the two commits had differing hashes even though they had the same content?

@tingulfsen and @atz, since you guys wrote and reviewed the commits mentioned above, and should check my work to make sure i did the right thing.


mejackreed:
:+1:

mejackreed:
Should we also do the opposite?

jmartin-sul:
probably.  if we do that, we should probably tag a new release candidate too?

mejackreed:
Yes of course.

PR #585, opened 2016-02-25
Create User#can_admin?, User#can_manage? and User#can_view?

darrenleeweber:
This PR is just about creating additional `User` methods that can be used in authorization.  Examples of it's use are in https://github.com/sul-dlss/argo/commit/214acb7598d532a0f6b49e4675a28966b5259713, but this PR does not use it anywhere in the app (yet).  The core of this PR is in the ability of the user model to check permissions on any DOR object, which is consolidated in this method:
https://github.com/sul-dlss/argo/blob/enhance-user-permissions/app/models/user.rb#L229-L248

This is an extract from a larger goal that was previously proposed in https://github.com/sul-dlss/argo/pull/532, but that work is too complex to evaluate.  Smaller chunks of that work can proceed and gradually accomplish the same outcome.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -360,6 +360,351 @@ describe User, :type => :model do
     end
   end
 
+  describe '#can_admin?' do
+    before :each do
+      expect(subject).not_to receive(:is_manager)
+      expect(subject).not_to receive(:is_viewer)
+    end
+
+    shared_examples 'Argo grants permission to manage object' do
+      it 'permits admin' do
+        expect(subject).to receive(:is_admin).once.and_return(true)
+        expect(subject.can_admin?(item)).to be true
+      end
+      it 'denies manager' do
+        expect(subject).to receive(:is_admin).once.and_return(false)
+        expect(subject.can_admin?(item)).to be false
+      end
+    end
+
+    context 'checks object permissions' do
+      let(:druid) { 'druid:hv992ry2431' }
+      let(:item) { instantiate_fixture(druid, Dor::AdminPolicyObject) }
+      context 'when user has no Argo permissions' do
+        before :each do
+          expect(subject).to receive(:is_admin).once.and_return(false)
+        end
+        it 'using default "can_manage_item?"' do
+          expect(item).to receive(:can_manage_item?).at_least(:once)
+          subject.can_admin?(item)
+        end
+        it 'using custom "can_manage_content?"' do
+          expect(item).to receive(:can_manage_content?).at_least(:once)
+          subject.can_admin?(item, 'content')
+        end
+        it 'raises ArgumentError for invalid permissions' do
+          expect{subject.can_admin?(item, 'WTF')}.to raise_error(ArgumentError)
+        end
+      end
+    end
+
+    context 'DOR object is an APO with a governing APO' do
+      let(:druid) { 'druid:hv992ry2431' }
+      let(:item) { instantiate_fixture(druid, Dor::AdminPolicyObject) }
+      let(:apo) { item.admin_policy_object }
+      it_behaves_like 'Argo grants permission to manage object'
+      context 'when user has no Argo permissions' do
+        before :each do
+          expect(subject).to receive(:is_admin).once.and_return(false)
+        end
+        it 'allows user with authorized role in governing APO' do
+          # This checks the governing APO and grants permission.
+          role = 'sdr-administrator'
+          expect(subject).to receive(:roles).with(apo.pid).once.and_return([role])
+          expect(subject).not_to receive(:roles).with(item.pid)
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -360,6 +360,376 @@ def groups_to_impersonate
...
+  describe '#can_admin?' do
+    before :each do
+      expect(subject).not_to receive(:is_manager)
+      expect(subject).not_to receive(:is_viewer)
+    end
+
+    shared_examples 'Argo grants permission to manage object' do
+      it 'permits admin' do
+        expect(subject).to receive(:is_admin).once.and_return(true)
+        expect(subject).not_to receive(:is_manager)
```
mejackreed: This is repeated already in the `before` block... Can it be removed from the before block or is it needed in both places?

spec/models/user_spec.rb
```diff
@@ -360,6 +360,376 @@ def groups_to_impersonate
...
+        expect(subject).not_to receive(:is_viewer)
+        expect(subject.can_admin?(item)).to be true
+      end
+      it 'denies manager' do
+        expect(subject).to receive(:is_admin).once.and_return(false)
+        expect(subject).not_to receive(:is_manager)
```
mejackreed: Similar comment to above.

cbeer:
In many of our other projects, we're using [cancancan](https://github.com/CanCanCommunity/cancancan/) for authorization. If nothing else, it might provide better facilities for testing the various permutations.

spec/models/user_spec.rb
```diff
@@ -360,6 +360,376 @@ def groups_to_impersonate
...
+  describe '#can_manage?' do
+    before :each do
+      expect(subject).not_to receive(:is_viewer)
+    end
+
+    shared_examples 'Argo grants permission to manage object' do
+      it 'permits admin' do
+        expect(subject).to receive(:is_admin).once.and_return(true)
+        expect(subject).not_to receive(:is_manager)
+        expect(subject).not_to receive(:is_viewer)
```
mejackreed: Possibly repeated from `before` block?

darrenleeweber:
Jack originally suggested we look at cancancan and we did take a look at it.  I don't recall the details of that discussion, but we concluded that it would not work in Argo.  Let's not discuss that here in this PR, but please open an issue for it and point to examples of how it can be used with webauth and workgroup permissions.

spec/models/user_spec.rb
darrenleeweber: The shared examples are supposed to be entirely self contained so they can ensure that they behave the right way.  So even if a surrounding context sets things up differently for one or a few specs, the shared example needs to explicitly override some specific values on each spec.

mejackreed: Possible duplication of expectation already in `before` block?

spec/models/user_spec.rb
```diff
@@ -360,6 +360,376 @@ def groups_to_impersonate
...
+        expect(subject.can_manage?(item)).to be true
+      end
+      it 'permits manager' do
+        expect(subject).to receive(:is_admin).once.and_return(false)
+        expect(subject).to receive(:is_manager).once.and_return(true)
+        expect(subject).not_to receive(:is_viewer)
```
mejackreed: Similar to comment on LN 487

cbeer:
Thanks, I'm happy to move a discussion about cancancan to https://github.com/sul-dlss/argo/issues/586. As I mention in the ticket, I'm surprised that Argo has such divergent needs, and before we get too far in implementing this authorization framework, we might want to consider our other options. The code you've written here strikes me as amenable to converting to `cancancan` fairly easily, but once we start consuming this API elsewhere, I suspect it would be harder to change later.

darrenleeweber:
Addressed Jack's comments about duplicate specs from before context in shared examples and removed them.

spec/models/user_spec.rb
darrenleeweber: No it's not.

spec/models/user_spec.rb
darrenleeweber: No repetition here; no before block in this context.

darrenleeweber:
Chris proposes using cancancan and that might work, but I don't know.  One advantage of my solution is that it can adapt to different access permissions requested for a DOR object, like 'item', 'content', 'metadata' etc. that all depend on methods in Dor::Governable. My solution is somewhat flexible with changes to Dor::Governable. That's about all I can say about this at this time. Given deadlines involved right now, I can't consider this seriously until later and I don't know how difficult it would be to change. It's already difficult to apply the current methods to Argo, but I have explored code that can do it and work as expected.

darrenleeweber:
Closing this as the work is moving to other PRs on master.

PR #590, opened 2016-02-26
User permission applied only to catalog controller

darrenleeweber:
This applies the enhanced user only to the catalog controller and specs for it are passing, but the app as a whole might or might not pass specs (travis will tell us).  This is a potential solution to https://github.com/sul-dlss/argo/issues/76

This branch is based on https://github.com/sul-dlss/argo/compare/enhance-user-permissions

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -345,26 +345,18 @@ class CatalogController < ApplicationController
     sorted_info.reverse!
   end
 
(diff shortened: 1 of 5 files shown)


darrenleeweber:
In the travis results, we can see that some of the `spec/features/*` are failing because the new user model has not been applied to those specs.

darrenleeweber:
Closing this to move it to a new PR on master.

PR #592, opened 2016-02-26
Replace rest-client with faraday

cbeer:


drh-stanford:
The test coverage went down due to the new code added to modulator_job (client method).

cbeer:
Maybe @tingulfsen can add tests that invokes `generate_xml`? It looks like the tests there are all marked pending.

tingulfsen:
I will look into adding more tests. What is the reason for using faraday instead of rest-client?

cbeer:
I believe the primary motivation is consistency; it was readily apparent in `Dor::WorkflowService` which used both rest-client and faraday within the same class. The interest in Faraday, in particular, is its support for persistent HTTP connections and more direct support for testing.

In this case, I hope the improved testing support will make it possible to write the pending modsulator tests here.

app/jobs/modsulator_job.rb
```diff
@@ -256,4 +258,14 @@ def status_ok(dor_object)
...
     [1, 6, 7, 8, 9].include?(status)
   end
+
+  private
+
+  def client
+    Faraday.new do |f|
+      f.request :multipart
+      f.request :url_encoded
+      f.adapter :net_http
```
darrenleeweber: Can we look into using `http-client` here for better performance and persistent connections?
https://rubygems.org/gems/httpclient/
Compare with https://gist.github.com/darrenleeweber/89efe92d43449c07c729

darrenleeweber:
Generally :+1: with a move to Faraday, depending on the actual adapter used.

app/jobs/modsulator_job.rb
cbeer: Sure, looking into alternative HTTP adapters makes a lot of sense. Could we do that separate from switching to Faraday in the first place?

cbeer:
I've added happy-path tests for `generate_xml`, and hope that @tingulfsen could back-fill the others (perhaps in a future pull request?)

mejackreed:
:+1:

PR #593, opened 2016-02-26
Remove legacy argo rake tasks.

mejackreed:
With new implementations and added support on setup, these tasks are no longer used.

@atz what do you say? I'm pretty sure all of this is not used ever now. Also removes a nuke solr task.

Also makes the Blacklight 5.18 upgrade easier :smile_cat:


jmartin-sul:
looks fine to me, but defer to @atz, since he's done much more w/ solr-related rake tasks

atz:
Yeah, these shouldn't be strictly necessary anymore.  They should be spun off to some discreet spot like jettywrapper or a solr.

Issue #586, opened 2016-02-25
Use cancancan for managing authorization


cbeer:
With #585, it seems like we're rolling our own authorization framework. Many other DLSS projects are successfully using cancancan (including several examples with `dor-rights-auth`-based rights), and it seems a shame to have divergent implementations. I'm concerned, too, that our solution is perhaps less robust or reusable than other community solutions.

Before we get too far with using our home-grown solution, maybe this is a good time to evaluate the various costs. In #585, @darrenleeweber suggested that we had previously concluded cancancan was inappropriate, and I'm curious how and why we came to that decision.

The solution in #585 strikes me as remarkably amenable to converting to cancan-style ability checking, e.g. just taking `can_admin?` and turning it into something like:

```ruby
# can? :admin, object

can :admin, Dor::item if user.is_admin
can :admin, Dor::Item do |item|
  apo = dor_object.admin_policy_object
  true if apo && dor_object.can_manage_item(roles(apo.pid))
 end

can :admin, Dor::AdminPolicyObject do |item|
  dor_object.can_manage_item(roles(dor_object.pid))
 end
```

(compare to https://github.com/sul-dlss/argo/blob/enhance-user-permissions/app/models/user.rb)

I'm surprised, especially given our other successful implementations, that Argo's needs may be so divergent. For something as important as authorization, it might behoove us to try harder to find convergent solutions.

Here are some examples of various cancan implementations in our group:
https://github.com/sul-dlss/hydrus/blob/master/app/models/ability.rb
https://github.com/sul-dlss/digital_stacks_rails/blob/master/app/models/ability.rb
https://github.com/sul-dlss/spotlight/blob/master/app/models/spotlight/ability.rb

darrenleeweber:
One advantage of my solution is that it can take adapt to different access permissions requested for a DOR object, like 'item', 'content', 'metadata' etc. that all depend on methods in Dor::Governable.  My solution is somewhat flexible with changes to Dor::Governable.  That's about all I can say about this at this time.  Given deadlines involved right, I can't consider this seriously until later and I don't know how difficult it would be to change.  It's already difficult to apply the current methods to Argo, but I have explored code that can do it and work as expected.

LynnMcRae:
[...]

PR #595, opened 2016-02-26
Create User#can_admin?, User#can_manage? and User#can_view?

### LINKED ISSUE #76: Allow users with APO roles access to the APO itself

darrenleeweber:
This code, when applied, is a possible solution for #76.  An alternative solution is sketched in https://github.com/sul-dlss/argo/pull/596 (discussed in https://github.com/sul-dlss/argo/issues/586).

This PR is just about creating additional User methods that can be used in authorization.  The core of this PR is in the ability of the user model to check permissions on any DOR object.

This is an extract from a larger goal that was previously proposed in #532, but that work is too complex to evaluate. Smaller chunks of that work can proceed and gradually accomplish the same outcome.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -135,4 +135,96 @@ class User < ActiveRecord::Base
   def can_view_something?
     is_admin || is_manager || is_viewer || permitted_apos.length > 0
   end
+
+  # Notes on definitions for known roles in dor-services:
+  #
+  # At present, there are a few methods that all do the same thing;
+  # see all the `can_manage_*` methods in Dor::Governable, i.e.
+  # https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/models/governable.rb
+  #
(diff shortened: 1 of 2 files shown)


blalbrit:
@darrenleeweber - Just want to make clear that work is on hold on this and related tickets as we go through the deploy & release process. We will meet as a team after new Argo is safely in the field and then take the time to revisit this issue and fully plan the approach.

jmartin-sul:
doing a bit of argo housekeeping, and closing this PR under the assumption that it's been superseded by the cancan work.  feel free to reopen if that's not the case.

PR #596, opened 2016-02-26
Shameless green for cancan authorization

cbeer:


app/controllers/items_controller.rb
```diff
@@ -693,16 +692,16 @@ def save_and_reindex
   # Permissions
 
   # check that the user can carry out this item modification
-  def forbid_modify
-    return true if current_user.is_admin || @object.can_manage_content?(current_user.roles(@apo))
-    render :status => :forbidden, :text => 'forbidden'
-    false
+  def authorize_manage_obj!
+    authorize! :manage_item, @object
   end
```
darrenleeweber: We did discuss the defaults with regard to `manage_content` vs. `manage_item` and resolved the latter is preferred.  I could be important that the prior logic only includes `is_admin` and not `is_manager`.

app/controllers/items_controller.rb
```diff
@@ -693,16 +692,16 @@ def save_and_reindex
...
 
-  def forbid_view
-    return true if current_user.is_admin || @object.can_view_content?(current_user.roles(@apo))
-    render :status => :forbidden, :text => 'forbidden'
-    false
+  def authorize_view_obj!
+    authorize! :view_content, @object
+  end
```
darrenleeweber: It seems odd there is no equivalent `view_item` so it is `view_content` here.  This seems like an omission in `Dor::Governable`.

app/helpers/argo_helper.rb
```diff
@@ -102,11 +102,11 @@ def render_buttons(doc, object = nil)
     end
 
     # if this is an apo and the user has permission for the apo, let them edit it.
-    if (object.datastreams.include? 'roleMetadata') && (current_user.is_admin || current_user.is_manager || object.can_manage_item?(current_user.roles(apo_pid)))
+    if object.datastreams.include?('roleMetadata') && can?(:manage_item, object)
```
darrenleeweber: If the `can?(:manage_item, object)` is specific to an object that has 'roleMetadata' (generally and perhaps only a `Dor::AdminPolicyObject`), then the logic may change here and  `can?(:manage_item, object)` may suffice.

app/helpers/argo_helper.rb
```diff
@@ -102,11 +102,11 @@ def render_buttons(doc, object = nil)
...
       buttons << {:url => url_for(:controller => :apo, :action => :register, :id => pid), :label => 'Edit APO', :new_page => true}
       buttons << {:url => url_for(:controller => :apo, :action => :register_collection, :id => pid), :label => 'Create Collection'}
     end
-    if object.can_manage_item?(current_user.roles(apo_pid)) || current_user.is_admin || current_user.is_manager
+    if can?(:manage_item, object)
       buttons << {
```
darrenleeweber: Notice that the object roles preceded the Argo roles here before and we concluded that was a mistake.  The `is_admin || is_manager` should precede the object role permission request for `can_manage_item`.

app/helpers/argo_helper.rb
```diff
@@ -147,7 +147,7 @@ def render_buttons(doc, object = nil)
       # date=embargo_data.split.last
       if text != 'released'
         # TODO: add a date picker and button to change the embargo date for those who should be able to.
-        buttons << {:label => 'Update embargo', :url => embargo_form_item_path(pid)} if current_user.is_admin
+        buttons << {:label => 'Update embargo', :url => embargo_form_item_path(pid)} if can?(:manage_embargo, object)
```
darrenleeweber: Assuming that `can?(:manage_embargo, object)` includes `User#is_admin` as it's highest precedence, this will behave the same way and may be preferred, but if it falls into requesting object permission it could behave differently.  It may be that this should not change and only Argo admins can do this, regardless of object permissions.

app/controllers/items_controller.rb
cbeer: Thanks, `manage_item` vs `manage_content` was a mistake on my part. I'm interested in understanding more about when `is_manager` implies `manage_content` and when it does not. Maybe there's something deeper going on. @LynnMcRae ?

app/controllers/items_controller.rb
cbeer: As you noted above, it should have been `manage_content` above.

app/helpers/argo_helper.rb
```diff
@@ -179,7 +179,7 @@ def render_foxml_link(document, link_text = 'FoXML', opts = {:target => '_blank'
   end
 
   def render_solr_link(document)
-    return '' unless current_user.is_admin
+    return '' unless can? :read, document
```
darrenleeweber: Not clear whether this will behave the same way and is less clear about it's intent.  It could be that this should not change.

app/helpers/argo_helper.rb
cbeer: My understanding of the following lines is that it only applies to objects that have a `roleMetadata` datastream. The comment suggests, though, an obvious refactor would be to check if the object is an APO.

app/models/ability.rb
```diff
@@ -0,0 +1,71 @@
+class Ability
```
darrenleeweber: Rather than comment on every piece of this, it would be better to discuss it to understand how it works.

app/helpers/argo_helper.rb
cbeer: Indeed, and cancan makes this distinction in the `Ability` class, instead of forcing us to replicate it in all other contexts.

cbeer:
I should also be clear, this should not be considered the final state of this code, merely an attempt to quickly get to green with fairly mechanical replacements. There are obvious future refactors to further clarify the code, some of which may require more domain knowledge.

app/helpers/argo_helper.rb
cbeer: `cancan` offers us an opportunity to clearly separate attributes of the user (like group membership) from the actions they can perform. The underlying behavior isn't changed, we're simply decoupling user attributes from the view.

app/models/ability.rb
```diff
@@ -0,0 +1,71 @@
...
+      dor_item.can_view_item? user.roles(dor_item.pid)
+    end
+
+    can :view_content, Dor::AdminPolicyObject do |dor_item|
+      dor_item.can_view_content? user.roles(dor_item.pid)
+    end
+
+    can :view_desc_metadata, Dor::AdminPolicyObject do |dor_item|
+      dor_item.can_view_desc_metadata? user.roles(dor_item.pid)
+    end
```
darrenleeweber: An important part of the APO logic needs to be clarified here.  As I understand it, any APO object should first query it's governing APO for permission and then fall-back to query itself if that fails.

app/models/ability.rb
```diff
@@ -0,0 +1,71 @@
+class Ability
+  include CanCan::Ability
+
+  def initialize(user)
+    user ||= User.new
+
+    can :manage, :all if user.is_admin?
+    can [:manage_item, :manage_content, :manage_desc_metadata, :register_apo], :all if user.is_manager?
+    can [:view_item, :view_metadata, :view_content], :all if user.is_viewer?
```
darrenleeweber: I need to understand how this provides the precedence for Argo `is_admin`, `is_manager` and `is_viewer` over the object permissions.  Does the `can?` logic apply `a || b` or is it `a && b`?  We mostly want `a || b` and when it's something like `is_admin || object.can_manage_item(user.roles(apo.pid))` we don't want it to evaluate the object permission.

We should discuss to clarify the differentiation of `is_admin` from `is_manager`.

app/views/catalog/_ds.html.erb
```diff
@@ -1,13 +1,7 @@
...
   apo = @obj.admin_policy_object
   if(params[:edit])
-    if @user &&
-      (@user.is_admin ||
-        (
-          @obj.can_manage_content?(@user.roles(apo.pid)) &&
-          @obj.can_manage_desc_metadata?(@user.roles(apo.pid))
-        )
-      )
+    if can?(:manage_content, @obj) && can?(:manage_desc_metadata, @obj)
```
darrenleeweber: Assuming this has not lost the `is_admin` but it looks like any rspec `expect` on this will call `is_admin` twice instead of once.

app/views/catalog/_show_sections/_default_contents.html.erb
```diff
@@ -13,14 +12,7 @@
         <span class="label">Type</span>
         <%= content_ds.ng_xml.root['type'] %>
         <ul>
-          <% apo_pid = nil
-          if admin_policy_object
-            apo_pid = admin_policy_object.pid
-          end
-          
-          can_manage = current_user.is_admin || current_user.is_manager || (apo_pid && object.can_manage_content?(current_user.roles(apo_pid)))
-          can_view = current_user.is_admin || current_user.is_manager || (apo_pid && object.can_view_content?(current_user.roles(apo_pid)))
```
darrenleeweber: Notice here that `is_manager` is included.  This is part of the clarification of how `is_admin` differs from `is_manager`.

app/views/items/_embargo_form.html.erb
```diff
@@ -1,17 +1,14 @@
-<% if current_user && current_user.is_admin %>
```
darrenleeweber: Hmm, this is removed and not replaced.  Assuming this now depends on the controller for authz.

app/controllers/items_controller.rb
LynnMcRae: The original design broke out the parts of the object — desc MD, sys MD, contents, rights, WF — in order to give maximum flexibility in crafting roles with various restrictions.  As you can see from the resulting matrix (https://consul.stanford.edu/display/chimera/Repository+Roles+and+Permissions), the critical differences that resulted and that haven’t changed were meta roles (governing items) vs APO roles (governed items) and the distinction between manage (write) and view (read).  But otherwise it’s all-or-nothing access to the items you see, making the difference between manage_item vs manage_content moot since the first includes the second for the roles we’ve implemented.

We still think there might be a place someday for the -metadata role as the audience expands to folks from the metadata unit, a much lower chance that we’d find a “you can manage the DOR object but not its content files”  ever useful. So it’s probably not worth making such distinctions in the code until we actually need them. I can collapse the table on the consul page to better indicate the parts of the object covered without implying we support that granularity of access if that helps.

Lynn

app/helpers/argo_helper.rb
LynnMcRae: The link to FoXML is a convenience to give access to any SDR admin without necessarily setting up direct Fedora admin access. So that criteria makes sense. I don’t know what or who the 'unless can? :read, document’  test adds. 

Lynn

app/helpers/argo_helper.rb
LynnMcRae: Originally only APOs had roles and therefore roleMetadata, and they are still the only roles Argo cares about.  But Hydrus implemented item level roles, so that datastream can appear in items now.  It should just be covered now by any edit/write-level permission for any object. 

Lynn

darrenleeweber:
Travis fails due to some form of recursive problem, i.e.
```
1) ApoController create should hit the registration service to register both an APO and a collection

     Failure/Error: result = find_or_create_by(:sunetid => webauth.login)
     SystemStackError:
       stack level too deep
     # ./app/models/user.rb:39:in `find_or_create_by_webauth'
     # ./app/controllers/application_controller.rb:27:in `current_user'
     # ./app/controllers/application_controller.rb:32:in `current_user'
     # ./app/controllers/application_controller.rb:32:in `current_user'
```

app/helpers/argo_helper.rb
cbeer: @LynnMcRae cancan provides the ability to separate the "who" from the "what". The view asks "who can read (view) this document", and the ability class says "admins can read (view) documents).

app/models/ability.rb
cbeer: Is the order of those checks important? It seems likely that it isn't, if we're falling through. If it is, we can simply reverse the order of the `Dor::AdminPolicyObject` and `ActiveFedora::Base` checks.

cc @LynnMcRae ?

app/models/ability.rb
cbeer: Yes, permissions cascade as expected, and short-circuit when it gets an affirmative response.

Agreed, the difference between the application `admin` and `manager` roles is very unclear. Maybe @LynnMcRae can shed some light.

app/views/catalog/_ds.html.erb
cbeer: Sure, but is that a problem?

app/views/items/_embargo_form.html.erb
cbeer: Indeed. Authorizing the view is bad practice, and actually makes things much harder to test.

cbeer:
Yes, there is very inconsistent and poorly done user stubbing that needs to be addressed, but I hope it doesn't interfere with looking at this for understanding.

app/helpers/argo_helper.rb
LynnMcRae: If the can-read-doc is itself based on is_admin criteria, then no problem there! Just wanted to make sure it wasn’t trying to bring in more generic privileges to the item. Thanks.

darrenleeweber: Under the `can?` rules, who else could view it too?  The current code indicates that only `is_admin` can view it.

app/models/ability.rb
darrenleeweber: See https://github.com/sul-dlss/argo/issues/586#issuecomment-188893192

app/helpers/argo_helper.rb
cbeer: It's exactly the same. One of the cancan rules is "admins can do anything":

> can :manage, :all if user.is_admin?

And there are no permissions for other users.

app/views/catalog/_ds.html.erb
darrenleeweber: No, it's cheap, so that's :ok_hand:

darrenleeweber:
@cbeer notes, "very inconsistent and poorly done user stubbing".  Absolutely.  One attempt to resolve this can be seen in the `spec/spec_helper.rb` in https://github.com/sul-dlss/argo/commit/ff0546b489b99bef4f219dab830b7dddf341b0aa#diff-93830fa29d616f7c87903d08b5b1b29a.  Better options may be available.  @mejackreed and I have briefly noted interests in using FactoryGirl.

app/models/ability.rb
LynnMcRae: The apo manager roles limits the user to only those items they have access to by virtue of item governance (the purpose of roles), while sdr-admin grants all possible permissions for all objects (the sdr-* roles are all meant to be repository-wide, though we only have sdr-admin to date). I can’t speak to the actual implementation, but at a practical level recognizing super powers gained considerable efficiency by bypassing the user-to-roles mapping via workgroups for all APOs (hundreds and hundreds of them) and the cost of filtering all facets and searches.  I don’t know if there are similar efficiencies at the granular level of checking permissions for individual actions/button/etc, but I imagine there could be in some cases.

Does this help, or are there other fuzzy aspects?

.rubocop_todo.yml
```diff
@@ -367,7 +367,7 @@ Style/SignalException:
 
 # Offense count: 137
 # Cop supports --auto-correct.
-Style/SingleSpaceBeforeFirstArg:
+Style/SpaceBeforeFirstArg:
```
mejackreed: Can we pop this out to a separate commit?

cbeer: Sure thing. I guess this crept in to silence some obnoxious warnings in my editor.

app/views/auth/groups.html.erb
```diff
@@ -14,15 +14,25 @@
...
-    <div>
-      <%= form_tag(auth_remember_impersonated_groups_path) do %>
-        <div class='form-group'>
-          <%= label_tag :groups, 'Enter a group or a comma separated list of groups to impersonate' %>
-          <%= text_field_tag :groups, nil, class: 'form-control' %>
-        </div>
-        <%= submit_tag 'Impersonate', class: 'btn btn-primary btn-lg btn-block' %>
-        <%= link_to 'Cancel', :back, class:'btn btn btn-default btn-lg btn-block' %>
-      <% end %>
+    <div class='panel-body'>
```
mejackreed: I think LN 17-24 here is duplicated cruft, no?

cbeer: Thanks, yes.

darrenleeweber:
I'm not well enough to evaluate and keep up with this approach.  It appears to be a potential solution to general auth issues that encapsulates more than my concerns in solving #76.  At present, it is not clear that it solves #76 without revisions to existing specs (the last time I looked at this, it seemed to inching towards compatibility with existing specs and may have arrived at that point now).  The changes for specs to solve #76 could be pulled out from somewhere in the `issue76-apo-better` branch.

LynnMcRae:
@mejackreed summarized the goal as "As a non-admin, non-manager user, I want to view an APO that I have access to in roleMetadata, so that I can... " ... it should be

"As a non-admin in any APO-based role (manager or viewer) I want access to an APO by virtue of being a role-holder in roleMetadata, so that I can do to the APO whatever I can do with items governed by that APO."

app/controllers/items_controller.rb
darrenleeweber: That was not my point.  I believe we discussed it and it should be `manage_item`.  My point was that there is no equivalent `view_item`.

app/views/catalog/_show_sections/_default_contents.html.erb
```diff
@@ -43,7 +35,7 @@
                         <% child.xpath('file').each do |file| %>
                           <li class="file">
                             <span class="label">File</span>
-                            <% if can_manage %>
+                            <% if can? :manage_content, object %>
```
darrenleeweber: If my previous solution were adopted, this might have become `manage_item` instead of `manage_content`.  So, in general, the default should be `manage_item` unless it's clear that `manage_content` is required.  In general, this distinction between the two is important and needs to be resolved.

cbeer: I'm not so sure. The original code had:

```
can_manage = current_user.is_admin? || current_user.is_manager? || (apo_pid && object.can_manage_content?(current_user.roles(apo_pid)))
```

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  context 'for items without an APO' do
+    it { should_not be_able_to(:manage_item, item) }
+    it { should_not be_able_to(:manage_content, item) }
+    it { should_not be_able_to(:manage_desc_metadata, item) }
+    it { should_not be_able_to(:view_item, item) }
+    it { should_not be_able_to(:view_content, item) }
+    it { should_not be_able_to(:view_metadata, item) }
+  end
```
darrenleeweber: What kind of `items`?  If the "item" is an APO without a governing APO, then the "item" could grant access.

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    let(:item) { Dor::AdminPolicyObject.new(pid: 'apo') }
+
+    context 'for a managed item' do
+      before { allow(item).to receive(:can_manage_item?).and_return(true) }
+      it { should be_able_to(:manage_item, item) }
+    end
+
+    context 'for a viewable item' do
+      before { allow(item).to receive(:can_view_item?).and_return(true) }
+      it { should be_able_to(:view_item, item) }
```
darrenleeweber: Does `Dor::Governable` have `#can_view_item?` or is this a fiction?

spec/models/ability_spec.rb
cbeer: I'm not sure I'm following. Could you expand on that statement? (or, I was under the impression I captured the same nuance in #595 here)

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+  end
+
+  context 'with roles assigned by an APO' do
+    let(:apo) { double(Dor::AdminPolicyObject, pid: 'apo')}
+    let(:random_item) { Dor::Item.new(pid: 'x') }
```
darrenleeweber: Why is one a `double` and the other an instance?

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    before do
+      allow(item).to receive(:admin_policy_object).and_return(apo)
```
darrenleeweber: Maybe this can go into the let for the item?

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+
+    context 'as a user with a management role for an item' do
+      before do
+        allow(item).to receive(:can_manage_item?).and_return(true)
+      end
+
+      it { should_not be_able_to(:manage_item, random_item) }
+      it { should be_able_to(:manage_item, item) }
+    end
```
darrenleeweber: How does this spec actually setup the management role in the APO?  It seems to do nothing to mock the `User#is_admin?` and similar methods to ensure that roles are required for permissions.

darrenleeweber:
It's not clear how `spec/models/ability_spec.rb` tests for role based permissions based on APO roles vs. Argo overrides in `User#is_admin?` etc.  It doesn't seem to do anything to mock the workgroups and roles.

spec/models/ability_spec.rb
cbeer: Thanks, you're right. There's no parallel to `manage_item`. I'll remove that path.

spec/spec_helper.rb
```diff
@@ -124,6 +124,7 @@ def mock_user(attributes = {})
     groups: [],
     can_view_something?: false,
     is_admin?: false,
+    is_webauth_admin?: attributes[:is_admin?],
```
darrenleeweber: Is this attribute actually in User?

spec/models/ability_spec.rb
cbeer: It could, sure. Would you like me to?

I had to use an actual instance above because of the way inheritance works (the `double` object doesn't actually inherit from `Dor::AdminPolicyObject`, so it doesn't trigger the APO code path).

spec/models/ability_spec.rb
cbeer: I was trying to isolate this from the rest of the ActiveFedora stack, but I'm happy to give it a try.

spec/models/ability_spec.rb
cbeer: This isn't testing `#is_admin?`, and is assuming `can_manage_item?` is appropriately tested upstream (and can therefore stub it out). It does seem like I might add expectations about the groups getting passed to `:can_manage_X?` though.

darrenleeweber:
The modifications to specs for APO governance occur in a few places.  The clearest example can be found in https://github.com/sul-dlss/argo/pull/597/commits and I don't yet see similar specs in this PR.

spec/spec_helper.rb
cbeer: Yes, it was added in https://github.com/sul-dlss/argo/commit/0c2196439fbae8a942c3289dc039fda1fe8816f8

spec/models/ability_spec.rb
darrenleeweber: It would help to add copious commentary on the contexts to explain what they intend to do.  In this particular section, does `random_item` have any APO?  Should it?  Why?  Does the instance get a default APO?  The code alone cannot explain these things and we shouldn't have to be an expert in DOR to understand this spec.

spec/models/ability_spec.rb
darrenleeweber: I get the impression that can...can works on object class, so it seems important to be clear about using a double vs. an instance.  If a choice is made to use one or another on purpose, please comment to explain why that choice is important.

spec/models/ability_spec.rb
darrenleeweber: https://github.com/sul-dlss/dor-services/issues/190

spec/models/ability_spec.rb
cbeer: Sure, that makes sense. As it makes no difference, let's go with real instances of those classes.

spec/models/ability_spec.rb
darrenleeweber: Yes, basically I'm looking for equivalent user specs to those in https://github.com/sul-dlss/argo/pull/595 where it considers cases for an APO with or without a governing APO.  So, the thing about issue #76 is that when the object (or 'item') is an APO, it should be able to grant permission to act on itself.  I'm not clear about what the 'item' is exactly in this particular context.

spec/models/ability_spec.rb
cbeer: Ok. Let me extract the 3 scenarios into separate tests.

cbeer: Ok, the code now explicitly covers 3 different scenarios:

- [x] an item with an APO where the user is in an appropriate role
- [x] an item with an APO where the user is not in appropriate role
- [x] an item doesn't have an APO

app/views/catalog/_show_sections/_default_contents.html.erb
darrenleeweber: There was slack discussion around defaults, but perhaps it doesn't apply here.  The partial name indicates it is about 'content' so that's good enough for me.

app/controllers/items_controller.rb
cbeer: I think I've misunderstood something. In https://github.com/sul-dlss/argo/pull/596#discussion_r54289914, you pointed out an instance where I was using `manage_item` instead of `manage_content`. After switching it to use `manage_content`, it is now parallel to `view_content`.

app/controllers/items_controller.rb
cbeer: Thanks @LynnMcRae. This distinction is already present in the code, but it would be great if we could collapse reduce the various permutations.

app/views/catalog/_show_sections/_default_contents.html.erb
cbeer: Is there anything relevant from that discussion we should add here (for posterity or context for future developers)?

spec/models/ability_spec.rb
cbeer: Is an APO without a governing APO a different use case than a normal item with a governing APO?

app/controllers/items_controller.rb
darrenleeweber: At the level of generality for the `ItemsController`, this is more likely to be `manage_item` than `manage_content`.

cbeer: That wasn't the previous behavior, and I'd feel far more comfortable making functional changes like that after discussion with @LynnMcRae and other stakeholders.

darrenleeweber: Agreed.  There was some related slack discussion in Jan 27th, see
https://sul-dlss.slack.com/archives/argo/p1453938430000152

app/models/ability.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+    end
+
+    can :view_metadata, ActiveFedora::Base do |dor_item|
+      if dor_item.admin_policy_object
+        dor_item.can_view_metadata? user.roles(dor_item.admin_policy_object.pid)
+      end
+    end
+  end
+end
```
darrenleeweber: This should probably cover all the permission methods available in `Dor::Governable`, i.e.
https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/governable.rb#L105-L128

app/models/ability.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+    end
+
+    can :view_content, Dor::AdminPolicyObject do |dor_item|
+      dor_item.can_view_content? user.roles(dor_item.pid)
+    end
+
+    can :view_metadata, Dor::AdminPolicyObject do |dor_item|
+      dor_item.can_view_metadata? user.roles(dor_item.pid)
+    end
```
darrenleeweber: For 45-51, it's not clear that an APO with a governing APO will first query for permission from the governing APO.  I would @LynnMcRae to clarify the governance logic, but I believe it should be similar to 
https://github.com/sul-dlss/argo/blob/user-permissions/app/models/user.rb#L219-L228

To summarize, any DOR object with a "governing-APO" must always query the "governing-APO" for a permission.  Only when that fails and the object is an APO itself, should it be able to query itself directly for permission.  If so, it looks like these blocks should include something like:
```
if dor_item.admin_policy_object
  dor_item.can_manage_something? user.roles(dor_item.admin_policy_object.pid)
end
```

cbeer: That should be the behavior, yes. cancan permissions both cascade and short-circuit.

app/views/catalog/_show_sections/_default_contents.html.erb
darrenleeweber: Captured it in https://github.com/sul-dlss/argo/issues/623

app/models/ability.rb
cbeer: I'll have a follow-up conversation with @LynnMcRae about what permutations we want to expose and maintain. My take-away from https://github.com/sul-dlss/argo/pull/596#discussion_r54294973:

> So it’s probably not worth making such distinctions in the code until we actually need them

Was we oughtn't expose permissions we're not using.

app/controllers/items_controller.rb
cbeer: Thanks, much appreciated.

app/models/ability.rb
darrenleeweber: I need to understand this better to get clear on what's going on.  I'll have to take a timeout to learn more about cancan patterns.

spec/spec_helper.rb
darrenleeweber: :ok_hand: -- that's a good addition.

app/models/ability.rb
darrenleeweber: It could be helpful to add commentary about such decisions in the code.

spec/models/ability_spec.rb
darrenleeweber: Yes, perhaps it amounts to more judicious use of `expect` rather than `allow`.  The granularity of these specs is murky in this regard; i.e. the specific code paths the specs execute are not clear.  I don't understand how it could be hitting all the permutations on authorization, starting with Argo global roles (is_admin etc.) and ending with object specific permissions based on APO roles.

cbeer: Hm. I'll defer to others on the team (@mejackreed, @jmartin-sul, @atz?). I'd argue we don't /actually/ care whether that precise method is called (or how many times it's called), just that if we're in a world where it responds with `true`, the user should have a particular kind of access.

app/models/ability.rb
cbeer: There's some pretty good information in the cancancan wiki (and a lot of irrelevant information about all the helpful things it can do if you're operating with a supported framework), including:

- https://github.com/CanCanCommunity/cancancan/wiki/Non-RESTful-Controllers
- https://github.com/CanCanCommunity/cancancan/wiki/Ability-Precedence
- https://github.com/CanCanCommunity/cancancan/wiki/Checking-Abilities

spec/models/ability_spec.rb
darrenleeweber: The should progressively disable access paths with higher precedence to test those with lower precedence.  If the specs never explicitly disable `is_admin` and it has the highest precedence, then nothing else will get tested no matter what else the specs try to do.  That's a sure what to get shameless green.

cbeer: I'm afraid I don't understand your comment. `is_admin` and all the other user attributes are `false` by default, so we're progressively enabling the various permutations.

darrenleeweber:
Curious as to whether can..can roles could replace the user methods for `is_admin` etc., see
https://github.com/CanCanCommunity/cancancan/wiki/Role-Based-Authorization#alternative-role-inheritance

cbeer:
Perhaps, I'm not sure off-hand, especially because our roles are not as cut-and-dried as their example.

darrenleeweber:
Can we remove any dependency on the persistence layer details by avoiding `ActiveFedora::Base` and using `Dor::Base` instead?  (This might require changes to dor-services.  Following suggestions from the Sandi Metz training, it seems fair to have dor-services isolate Argo from peristence layer details.  Argo should not need to know that Dor objects are persisted in Fedora using ActiveFedora.)

cbeer:
`Dor::Base`? I don't think such a thing exists right now.

cbeer:
Sorry, and I should be clear, I'm all for refactors like that. I suspect there's a lot of other things we'll want to clean up, but we should be clear about what MUST happen before this could ship. I'd hesitate making large changes if we can avoid them.

darrenleeweber:
Can we avoid `ActiveFedora::Base` to be more specific about the existing actual DOR classes?  There are only a few.  If a catch-all is required, maybe use a constant that holds them in an array, something like: https://github.com/sul-dlss/dor-services/blob/master/lib/dor-services.rb#L64-L66

darrenleeweber:
Trying to digest the meaning and significance of `can :manage :all` for admins and wondering whether there are any actions that even admins cannot do, like purge the Uber APO or something.

cbeer:
I'm personally ambivalent about using `ActiveFedora::Base` vs explicitly listing models. I worry about possibly missing some, or being less open to extension. Happy to defer to the team with whatever we feel is best.

cbeer:
I think it's important to distinguish between user abilities (managed by cancancan) and policies based on object state (like being able to delete the Uber APO). Cancan is good at the former, and probably less good at the latter.

darrenleeweber:
It seems we cannot use object attributes in a conventional sense described in
https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities#hash-of-conditions

darrenleeweber:
@LynnMcRae confirmed that `is_admin` can and should be able to do anything exposed by Argo, so there is no qualification on `:manage :all` for the admin role.

darrenleeweber:
I guess the issue with `ActiveFedora::Base` is that not everything includes `Dor::Governable` and that is required for the permission methods used.  So, whatever the object is, it must return true for
`dor_object.kind_of? Dor::Governable`

blalbrit:
@darrenleeweber - Just want to make clear that work is on hold on this and related tickets as we go through the deploy & release process. We will meet as a team after new Argo is safely in the field and then take the time to revisit this issue and fully plan the approach.

darrenleeweber:
I can't put everything into this PR so I started coding instead in a separate branch off this branch to try to express the ideas that way.  So, for comparison, see the `cancan_playground` branch, currently only modifies https://github.com/sul-dlss/argo/blob/cancan_playground/app/models/ability.rb

cbeer:
Thanks @blalbrit. I'm happy to set this aside until we can get the team around this. There are significant authorization gaps in our code (partially addressed here, and deserving of a holistic approach), and it would behoove us to address them in the very near future.

jmartin-sul:
overall, i'm :+1: on this.  it seems like a fine pattern to adopt, it appears to implement what's desired from issue #76, and it looks to me like it refactors existing behavior cleanly and with relatively limited scope.  i also like that the changes to the existing tests are kept to a minimum.

i had some suggestions on naming and comments that i think would make things clearer, and i had a couple expectations that i thought could be added.  it wasn't really much extra work to just implement them since they were all pretty small, so i went ahead and created a PR against this one instead of making all the comments inline.  just in case any of the suggestions were controversial, i tried to break them into separate commits, which can be squashed or cherry picked as desired.  while i would really prefer to see all of those suggestions implemented or shot down, i'd still be :ok_hand: merging this as-is.

tingulfsen:
I'm also :+1: on this, but would really like @jkeck to review.

jkeck:
:+1: from my standpoint.  I'm not knowledgable enough about the functionality in argo, but this looks like a good use of CanCan to me.

PR #597, opened 2016-02-26
User permissions applied only to catalog controller

darrenleeweber:
This applies the enhanced user only to the catalog controller and specs for it are passing, but the app as a whole does not pass specs. This is an application of #595 

This branch is based on https://github.com/sul-dlss/argo/compare/user-permissions

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -342,26 +342,18 @@ class CatalogController < ApplicationController
     sorted_info.reverse!
   end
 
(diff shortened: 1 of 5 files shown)


darrenleeweber:
In the travis results, some of the `spec/features/*` fail because the new user model has not been applied to those specs. The purpose of this PR is not to fix those, only to identify the additional need to provide a user model for these specs.

jmartin-sul:
closing this one too, since it builds on PR #595, which was just closed under the assumption that cancan work superseded it.

PR #599, opened 2016-02-26
[WIP] Revise 400/404 responses for status controller specs

darrenleeweber:
On my system, a mal-formed PID raises `Rubydora::FedoraInvalidRequest` and that is more like a 400 rather than a 404.

The full rspec error that motivates this change is:
```
  1) StatusController log with test_obj should 404 instead of 500 on bad IDs
     Failure/Error: test_item = Dor::Item.find(params[:test_obj])
     
       See logger for details
     # ./app/controllers/status_controller.rb:14:in `log'
     # ./spec/controllers/status_controller_spec.rb:52:in `block (3 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # RestClient::InternalServerError:
     #   500 Internal Server Error
     #   ./app/controllers/status_controller.rb:14:in `log'
```

### DIFF
--- a/spec/controllers/status_controller_spec.rb
+++ b/spec/controllers/status_controller_spec.rb
@@ -46,10 +46,18 @@ describe StatusController, :type => :controller do
     before :each do
       @item = instantiate_fixture(@druid)
     end
-    it 'should 404 instead of 500 on bad IDs' do
+    it 'should 400 instead of 500 on bad IDs' do
+      bad_pid = 'junk'
       expect(subject).to receive(:check_recently_indexed).and_return(false)
-      expect(Dor::Item).to receive(:find).with('junk').and_call_original
-      get 'log', :test_obj => 'junk'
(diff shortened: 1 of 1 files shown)


cbeer:
Is this commit missing the functional change?

mejackreed:
Also was thinking at some point since this controller is setup specifically for Nagios checks, could we move this functionality underneath `is_it_working` somehow?

darrenleeweber:
@cbeer -- yes, sorry, this was ripped from elsewhere and should be a [WIP] but lost that when it was converted to a PR on master instead of develop.  The functional change would be to add an exception handler to application controller to catch the exception noted.  However, this only occurs on my laptop, so this is in limbo.

atz:
Considering the pid is part of the path/route, 404 is reasonable.  The response is "that path does not exist".

darrenleeweber:
Closing this to move the problem to https://github.com/sul-dlss/argo/issues/615

PR #603, opened 2016-02-27
Update tests to use a consistent mock user object

cbeer:


spec/views/items/_embargo_form.html.erb_spec.rb
```diff
@@ -1,13 +1,16 @@
 require 'spec_helper'
 
 RSpec.describe 'items/_embargo_form.html.erb' do
-  let(:current_user) do
-    double('current_user', is_admin: true)
+  let(:current_user) { mock_user(is_admin?: true) }
+
+  before do
+    allow(view).to receive(:current_user).and_return(current_user)
+    allow(controller).to receive(:current_user).and_return(current_user)
```
mejackreed: Why the controller `allow` here?

mejackreed:
In several places in the view specs, there is an 

```ruby
allow(controller).to_receive ... 
```

Why do we need to allow for the controller to receive `current_user` from the view here?

spec/views/items/_embargo_form.html.erb_spec.rb
cbeer: cancan is injecting controller-level methods (`current_ability`, `can?`, and `cannot?`, among others) that are shared with the view. One future refactor could be to separate our view specs from users + abilities by stubbing `can?` directly, eliminating the need for this stubbing.

cbeer: Sorry, and having said that, it's obvious that change is in the wrong PR. I'll try to move it to #596.

cbeer:
Thanks for catching that. I've removed that from this PR.

mejackreed:
:+1:

PR #607, opened 2016-02-29
Use newer version of hydra-jetty, to silence solr response warnings

cbeer:


cbeer:
Maybe this fixes the original intent of #100, although our local infrastructure is now using Solr 5.

PR #611, opened 2016-02-29
Allow a non-admin to forget an impersonation group

mejackreed:
Fixes a regression in the `AuthController` where a non-admin could not forget an impersonation.


cbeer:
:+1: sorry about that. Clearly impersonation is difficult to reason about. Maybe #605 will help a little bit too.

mejackreed:
No worries our spec explanation didn't exactly meet up with what was being tested.

PR #616, opened 2016-03-01
lower default delayed_job worker count

jmartin-sul:
* use same number of workers as CPUs on prod (12/12)
* use half the number of workers as CPUs on stage (4/8)

### DIFF
--- a/config/deploy/prod-a.rb
+++ b/config/deploy/prod-a.rb
@@ -5,4 +5,4 @@ set :rails_env, 'production'
 
 set :deploy_to, '/opt/app/lyberadmin/argo'
 
-set :delayed_job_workers, 16
+set :delayed_job_workers, 12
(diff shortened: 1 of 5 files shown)


atz:
Rebased and force pushed.

mejackreed:
I'm :ok: with this. Deferring all maters of workers to @jmartin-sul and @drh-stanford

jmartin-sul:
well, i'm :+1: since i pushed it :smiley:  and values were suggested by @drh-stanford and @eefahy on chat yesterday.

FWIW:  the swap issues that caused us to lower the counts haven't recurred with these lower values.  indexing is also a bit slower, of course.  but this seems stable.  reading the effect of that change is confounded a bit, because i also tuned the eye config a bit at the same time (eye enforces the threshold for auto-restarting workers that've consumed too much memory).  specifically, i gave it more realistic values for the time it'd take to start/stop/restart delayed_job workers (compared to the far too low 10 second default it was originally using):
https://github.com/sul-dlss/argo/compare/eye-job-ctrl#diff-cdc48ac81e74d78f37efac30fa55c94eR14

PR #618, opened 2016-03-01
Don't use exceptions in place of flow control

cbeer:


cbeer:
(Some pretty annoying Liskov violations there too, perhaps for another day..)

PR #619, opened 2016-03-01
Tidying controller-level code by removing unused methods and extracting models in place of helpers

cbeer:


cbeer:
I'm happy to send any commit in a separate pull request, although I believe these are essentially mechanical extractions and hopefully very little to say about them.

cbeer:
Note, too, the hound violations are in the original extracted code.

mejackreed:
Seems :ok: and I'm ok with the straight copy hound violations. Gosh, we really have no way of testing the track sheet functionality do we?

cbeer:
Doesn't look like it -- another reason I'm hesitant to do anything besides mechanical transformation.

jmartin-sul:
tracking sheet generation is both a fragile mess and a frequently used bit of functionality.  we should def make sure to do some manual testing with this if it makes it into tomorrow's release.

i like the idea here a lot, but am slightly hesitant to ship it now, though i haven't looked over this change thoroughly at all (looks fine based on my glancing at it, though).

blalbrit:
+1 @jmartin-sul - this is used on a daily basis by production staff and is critical to their workflow. Please let me and @LynnMcRae know if there's anything ready for testing.

cbeer:
Ok, I've sliced off #621 to throw away unused code, although I'd reiterate that this extraction was purely mechanical.

atz:
Rebased and pushed.

PR #624, opened 2016-03-01
Add embargo_update to authorized before_action group.

### LINKED ISSUE #622: embargo_update must be an authorized action
Issue to track and ensure that `ItemsController#embargo_update` is included in a `before_action` for authorized access so that it short-circuits all the `after_action` activity when it's not authorized (otherwise it will save and reindex unnecessarily).

darrenleeweber:
When the action is not authorized, the after_action groups will
not do anything.  In it's current form, the action requires
authorization, but when it fails it still triggers after_actions
that run save_and_reindex needlessly.

This should fix #622

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -9,6 +9,7 @@ class ItemsController < ApplicationController
   before_action :forbid_modify, :only => [
     :add_collection, :set_collection, :remove_collection,
     :datastream_update,
+    :embargo_update,
     :mods,
     :open_version, :close_version,
     :update_resource,
@@ -226,10 +227,6 @@ class ItemsController < ApplicationController
   end
 
   def embargo_update
-    unless current_user.is_admin?
-      render :status => :forbidden, :text => 'forbidden'
-      return
-    end
     fail ArgumentError, 'Missing embargo_date parameter' unless params[:embargo_date].present?
     @object.update_embargo(DateTime.parse(params[:embargo_date]).utc)
     @object.datastreams['events'].add_event('Embargo', current_user.to_s , 'Embargo date modified')
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -71,8 +71,11 @@ describe ItemsController, :type => :controller do
   end
   describe 'embargo_update' do
     it 'should 403 if you are not an admin' do
-      allow(@current_user).to receive(:is_admin?).and_return(false)
-      post 'embargo_update', :id => @pid, :date => '12/19/2013'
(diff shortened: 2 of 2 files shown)


spec/controllers/items_controller_spec.rb
```diff
@@ -71,8 +71,10 @@
   end
   describe 'embargo_update' do
     it 'should 403 if you are not an admin' do
-      allow(@current_user).to receive(:is_admin?).and_return(false)
-      post 'embargo_update', :id => @pid, :date => '12/19/2013'
+      expect(controller).not_to receive(:save_and_reindex)
+      expect(controller).not_to receive(:flush_index)
+      expect(controller).to receive(:forbid_modify).and_return(false)
```
mejackreed: Should we submit the date format as it is created in the form?

screen shot 2016-03-01 at 6 38 35 pm

darrenleeweber: That's probably irrelevant for this spec.  (What might be more important is that this PR fixes the param name from :date to :embargo_date.)  But, if it must be perfect, name the date and you shall have it ;-)

app/controllers/items_controller.rb
```diff
@@ -9,6 +9,7 @@ class ItemsController < ApplicationController
   before_action :forbid_modify, :only => [
     :add_collection, :set_collection, :remove_collection,
     :datastream_update,
+    :embargo_update,
```
cbeer: Is this a functional change? Before, it looked like we restricted updating embargos to only `current_user.is_admin?`

spec/controllers/items_controller_spec.rb
mejackreed: Just something I noticed. Definitely an improvement passing the correct param name! (I wonder what we were testing before)

app/controllers/items_controller.rb
darrenleeweber: It's a catch-22.  If the authz is done within the action and fails, all the after actions run anyway.  If the authz is done as a before action, it can skip the action and all the after actions.  A consequence of using a before action is using the authz it uses.  So, it may be a question for @LynnMcRae as to whether the current forbid_modify action is reasonable for embargo updates.  It first checks for is_admin and then checks the object/APO for a role to `can_manage_content`.  So, yes, it's a functional change.  Without it, unauthorized requests on embargo update can trigger reindexing (unless bug).

darrenleeweber:
Looking into the travis failure, seems a bit odd that changes in this PR should raise this error.

spec/controllers/items_controller_spec.rb
darrenleeweber: Modified the date in the last push.

PR #628, opened 2016-03-02
Remove unused workflow archive count

cbeer:


app/views/report/workflow_grid.json.erb
```diff
@@ -18,11 +18,6 @@
       end
       workflows[wf_name][:processes][process_name] = p
     end
-
-    completed_count = render_workflow_archive_count('dor',wf_name)
```
cbeer: I'm not sure this JSON output is used at all. Maybe someone else could confirm?

mejackreed: Nothing that I know of is using it.

PR #636, opened 2016-03-02
Fixes an issue where prepare bulk update can not be done by adding ca…

mejackreed:
…n_open_version? inside a new class DorObjectWorkflowStatus

This will need to be merged with logic in:

 - argo/app/controllers/workflow_service_controller.rb

Error originally came in when the `DorObjectHelper` mixin was removed in 670c010797a0b23ef24b9a7f8676b6d82021718c. This logic should never be in a helper in the first place.

### DIFF
--- a/app/helpers/dor_object_helper.rb
+++ b/app/helpers/dor_object_helper.rb
@@ -96,16 +96,6 @@ module DorObjectHelper
     Sdr::Client.current_version(object.pid)
   end
 
-  ##
-  # @deprecated Please use non-blocking requests rather than blocking helpers.
-  # See WorkflowServiceController#openable for JSON API to this logic
-  def can_open_version?(pid)
-    return false unless Dor::Config.workflow.client.get_lifecycle('dor', pid, 'accessioned')
-    return false if Dor::Config.workflow.client.get_active_lifecycle('dor', pid, 'submitted')
-    return false if Dor::Config.workflow.client.get_active_lifecycle('dor', pid, 'opened')
-    true
-  end
-
   def render_qfacet_value(facet_solr_field, item, options = {})
     params = add_facet_params(facet_solr_field, item.qvalue)
     Rails.cache.fetch('route_for' + params.to_s, :expires_in => 1.hour) do
--- a/app/models/dor_object_workflow_status.rb
+++ b/app/models/dor_object_workflow_status.rb
@@ -1,0 +1,24 @@
+class DorObjectWorkflowStatus
+  attr_reader :pid
+
+  ##
+  # @param [String] pid in format "druid:abc123def4567"
+  def initialize(pid)
+    @pid = pid
+  end
+
+  ##
+  # @return [Boolean]
+  def can_open_version?
+    return false unless workflow.get_lifecycle('dor', pid, 'accessioned')
+    return false if workflow.get_active_lifecycle('dor', pid, 'submitted')
+    return false if workflow.get_active_lifecycle('dor', pid, 'opened')
+    true
+  end
+
+  private
+
+  def workflow
+    Dor::Config.workflow.client
+  end
+end
--- a/spec/models/dor_object_workflow_status_spec.rb
+++ b/spec/models/dor_object_workflow_status_spec.rb
@@ -1,0 +1,55 @@
+require 'spec_helper'
+
+RSpec.describe DorObjectWorkflowStatus do
+  subject { described_class.new(pid) }
+  let(:pid) { 'druid:abc123def4567' }
+  describe '#can_open_version?' do
+    context 'when not accessioned' do
+      before do
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_lifecycle).with('dor', pid, 'accessioned')
+          .and_return(false)
+      end
+      it { expect(subject.can_open_version?).to eq false }
+    end
+    context 'when accessioned and submitted' do
+      before do
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_lifecycle).with('dor', pid, 'accessioned')
+          .and_return(true)
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_active_lifecycle).with('dor', pid, 'submitted')
+          .and_return(true)
+      end
+      it { expect(subject.can_open_version?).to eq false }
+    end
+    context 'when accessioned, not submitted, and opened' do
+      before do
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_lifecycle).with('dor', pid, 'accessioned')
+          .and_return(true)
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_active_lifecycle).with('dor', pid, 'submitted')
+          .and_return(false)
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_active_lifecycle).with('dor', pid, 'opened')
+          .and_return(true)
+      end
+      it { expect(subject.can_open_version?).to eq false }
+    end
+    context 'when accessioned, not submitted, and not opened' do
+      before do
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_lifecycle).with('dor', pid, 'accessioned')
+          .and_return(true)
+        expect(Dor::Config.workflow.client)
+          .to receive(:get_active_lifecycle).with('dor', pid, 'submitted')
+          .and_return(false)
+        expect(Dor::Config.workflow.client)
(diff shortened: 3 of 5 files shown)


app/helpers/dor_object_helper.rb
```diff
@@ -96,16 +96,6 @@ def last_accessioned_version(object)
     Sdr::Client.current_version(object.pid)
   end
 
-  ##
-  # @deprecated Please use non-blocking requests rather than blocking helpers.
```
cbeer: Do you want to copy this comment? Or maybe it's less relevant.

mejackreed: Nope ready to kill it.

cbeer: :+1: It's definitely less confusing for a blocking request to happen in a model.

spec/models/dor_object_workflow_status_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  let(:pid) { 'druid:abc123' }
+  describe '#can_open_version?' do
+    context 'when not accessioned' do
+      before do
+        expect(Dor::WorkflowService)
+          .to receive(:get_lifecycle).with('dor', pid, 'accessioned')
+          .and_return(false)
+      end
+      it { expect(subject.can_open_version?).to eq false }
+    end
```
darrenleeweber: Prefer all these `context` with one `it` are all converted to `it`s, like:
```
it 'is not open when not accessioned' do
  expect(Dor::WorkflowService).to receive(:get_lifecycle).with('dor', pid, 'accessioned').and_return(false)
  expect(subject.can_open_version?).to be false
end
```
(Long lines up to 120 chars should be OK.)

spec/models/dor_object_workflow_status_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+
+RSpec.describe DorObjectWorkflowStatus do
+  subject { described_class.new(pid) }
+  let(:pid) { 'druid:abc123' }
```
darrenleeweber: Prefer to see a valid DRUID, like `druid:abc123def4567`.

mejackreed: Updated with that.

app/models/dor_object_workflow_status.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  ##
+  # @return [Boolean]
+  def can_open_version?
+    return false unless Dor::WorkflowService.get_lifecycle('dor', pid, 'accessioned')
+    return false if Dor::WorkflowService.get_active_lifecycle('dor', pid, 'submitted')
+    return false if Dor::WorkflowService.get_active_lifecycle('dor', pid, 'opened')
+    true
+  end
```
darrenleeweber: Maybe this could be simplified?  Like:
```ruby
Dor::WorkflowService.get_lifecycle('dor', pid, 'accessioned')  &&
!( Dor::WorkflowService.get_active_lifecycle('dor', pid, 'submitted') ||
   Dor::WorkflowService.get_active_lifecycle('dor', pid, 'opened'))
```

spec/models/dor_object_workflow_status_spec.rb
mejackreed: I guess I disagree here. I think contexts are useful in this case. If the spec fails you see the output as:

"DorObjectWorkflowStatus#can_open_version? when accessioned and submitted should eq true"

If it is switched to an `it` block, you see:

"DorObjectWorkflowStatus#can_open_version? when accessioned and submitted"

http://betterspecs.org/#contexts

app/models/dor_object_workflow_status.rb
```diff
@@ -0,0 +1,18 @@
+class DorObjectWorkflowStatus
+  attr_reader :pid
+
+  ##
+  # @param [String] pid in format "druid:abc123def4567"
+  def initialize(pid)
+    @pid = pid
+  end
```
darrenleeweber: You might to init something to hang onto the `Dor::WorkflowService`, something like:
`@workflow = Dor::WorkflowService`
If it's in one place and it needs to change later, it will be easy.  (Change could be coming.)

tingulfsen:
Looks good to me - these are pretty minimal changes to get things working again. Will merge once the Travis CI is complete.

app/models/dor_object_workflow_status.rb
mejackreed: I realize that could be done. What change would necessitate that? If we have some examples of things coming, I'm happy to support that.

darrenleeweber:
I'm on the fence about the class name.  It's sorta OK, but what about `DorWorkflow` to keep it a bit shorter?  Why is `Object` important?  If it's important, OK.

darrenleeweber:
Note that the `Dor::Workflow` might have to change, see
https://github.com/sul-dlss/argo/pull/639

app/models/dor_object_workflow_status.rb
mejackreed: It could be. This is a straight copy over method so I'm less inclined to change things here. Also to me (and I realize I might be an outlier) that next conditional is harder to read. Realistically I'll be porting over other logic and this will change to be similar to `WorkflowServiceController` private methods.

spec/models/dor_object_workflow_status_spec.rb
darrenleeweber: I vote for more readable spec code than spec output.

app/models/dor_object_workflow_status.rb
darrenleeweber: I don't like the double negative to start the whole thing off: `return false unless`  (yuk).

app/models/dor_object_workflow_status.rb
darrenleeweber: See #639

mejackreed: I've made `Dor::WorkflowService` configurable from a private method. If it needs to change we can now do that in one place.

mejackreed:
I guess "object" isn't super important, but the methods that are being called are in reference to an object and that object's abilities as determined by the `WorkflowService`. `DorWorkflow#can_open_version?`  feels more like its talking about the workflow itself and not the object. no?

darrenleeweber:
"DorWorkflow#can_open_version? feels more like its talking about the workflow itself and not the object. no?", only if you leave off the arguments like that.  Once you add the arguments, it reads fine.

app/models/dor_object_workflow_status.rb
```diff
@@ -0,0 +1,24 @@
...
+    return false unless workflow.get_lifecycle('dor', pid, 'accessioned')
+    return false if workflow.get_active_lifecycle('dor', pid, 'submitted')
+    return false if workflow.get_active_lifecycle('dor', pid, 'opened')
+    true
+  end
+
+  private
+
+  def workflow
+    Dor::WorkflowService
```
jmartin-sul: one other suggestion:  can we switch to `Dor::Config.workflow.client` instead of `Dor::WorkflowService`?  since that appears to be the preferred pattern now.

mejackreed: Yep pushing up now

jmartin-sul:
:+1: other than the comment about `Dor::Config.workflow.client`

mejackreed:
@darrenleeweber Part of me thinks that the class should be able to stand alone without the instantiation argument to know what it is about. That is why I left it out.

darrenleeweber:
I hear you singing that Sinatra tune - "and I did it myyyyyyyyyyyyy way" - so you go for it.  Thrown in my 2c for what it's worth.  Push on.

app/models/dor_object_workflow_status.rb
```diff
@@ -0,0 +1,24 @@
...
+    return false if workflow.get_active_lifecycle('dor', pid, 'opened')
+    true
+  end
+
+  private
+
+  def workflow
+    Dor::Config.workflow.client
+  end
+end
```
darrenleeweber: I'd consider additional private methods, esp. with a view to migrating more of this stuff from item controller where some of those methods are useful helpers.  Consider whether private methods could enable `can_open_version?` to become: `lifecycle('accessioned') && !( active_lifecycle('submitted') || active_lifecycle('opened') )`

mejackreed: Yes that's definitely coming when I integrate this. This PR solves the problem at hand, but leaves things ready for the integration.

PR #641, opened 2016-03-03
eye for delayed_job monitoring

jmartin-sul:
* install the eye gem
* add a configuration to restart delayed_job workers that consume too much memory:
  * check each worker every 5 min, restart if memory consumption exceeds 1.5 GB
* load the configuration upon restart after argo deployment

further implementation details in commit message and comments.


jmartin-sul:
this configuration of eye worked well for me during the recent full reindex of dor-prod to the new argo3 solr collection.  with the values as they are in this PR, it didn't seem to have any trouble restarting workers that crossed the defined memory consumption threshold (1.5 GB).  before, when the start/stop/restart timeouts were left to the defaults, eye would mostly restart workers successfully, but would occasionally allow some to keep leaking memory indefinitely, without ever being able to restart them.

open questions:
* is this something that should always be running?  should it be incorporated into the deployment scripts?
* is there a way to easily get the worker count as configured in argo?  eye doesn't load argo or rails (thankfully, other than in regards to this config duplication).  i guess i could load the config file that argo loads, or they could each get the value from a common source?  not that the duplication is the end of the world.
* is there a better way to approach this in general?

deployment opinions from @eefahy?  or general job monitoring opinions from @drh-stanford (who'd originally suggested looking into bluepill and eye)?  opinions welcome from anyone else, but thought those two in particular might want to chime in.

based on the bit of looking around i did online, the general consensus seemed to be that eye is more reliable than bluepill.  i didn't actually try bluepill out.

jmartin-sul:
i'll also definitely add a bit of documentation before removing the [WIP] tag.

Gemfile.lock
```diff
@@ -341,16 +370,16 @@ GEM
...
       rack (>= 1.0)
-    rails (4.2.5.2)
-      actionmailer (= 4.2.5.2)
-      actionpack (= 4.2.5.2)
-      actionview (= 4.2.5.2)
-      activejob (= 4.2.5.2)
-      activemodel (= 4.2.5.2)
-      activerecord (= 4.2.5.2)
-      activesupport (= 4.2.5.2)
+    rails (4.2.6)
```
jkeck: @jmartin-sul did you intend to do a rails 4.2.6 upgrade in this PR?  I didn't notice anything in the commit message about it.

jmartin-sul: that was unintentional.  though i think it was the result of `bundle install` when i added eye.  didn't investigate further, but can play around with it a bit before submitting for final review.

tingulfsen:
How does Eye log events such as restarts etc? Is that done automatically by Eye? I don't see any explicit logging here, which I'd expect.

jmartin-sul:
@tingulfsen:  log destinations are specified ([here](https://github.com/sul-dlss/argo/pull/641/files#diff-d9dc7ee63d9cccfb45591e6a25158a32R) and [here](https://github.com/sul-dlss/argo/pull/641/files#diff-d9dc7ee63d9cccfb45591e6a25158a32R24), the former for eye overall, the latter for specific per-worker stuff).  and then eye writes out what i'd consider the significant events (starts, stops, restarts, the result of each threshold check).  not sure how you'd get it to log other specific things (or whether that's possible), but it seems to get what i think we'd need to debug it if problems occurred.

tingulfsen:
:+1:  from me, let me know when you consider this branch ready to merge (it's still WIP).

jmartin-sul:
deployed to stage, worked.  re-deployed after making final touchups for the last force-push, config changes showed up on stage.  appears to work as expected.

Issue #627, opened 2016-03-02
Report Download Function - omits Admin Policy ID column (others?) on download


blalbrit:
https://argo-lb.stanford.edu/report?f%5BobjectType_ssim%5D%5B%5D=workflow and click 'download' . Did work in prod.

Needs further testing to determine which other columns are being dropped.

List of columns
- Admin Policy ID 
- Catkey

mejackreed:
When you say "Did work in prod" do you mean legacy argo 2.x ?

Currently investigating and the same result happens in v3.0.0-rc.1 and v3.0.0-rc.2

blalbrit:
@mejackreed Yes - works in legacy argo 2.x. Not a blocker to release.

mejackreed:
A few things happening here:

`@field` assigning here doesn't compare symbols and strings correctly. Some field names are strings some are symbols. https://github.com/sul-dlss/argo/blob/master/app/models/report.rb#L169

PR #659, opened 2016-03-07
add the Registration datetime to the report columns and fix download

### LINKED ISSUE #648: Some required columns missing from report view
Current list includes:

Registration Date (missing from report)
Admin Policy (present online but dropped in downloaded report)

Will add to the list as I find more. Related to #627 ?

drh-stanford:
This PR fixes #648 and #627 by adding a registration datetime to the available reporting columns, and fixing the report field `:field` parameters such that they are all symbols and downstream expects them as such. The latter fixes the download functionality.

### DIFF
--- a/spec/controllers/report_controller_spec.rb
+++ b/spec/controllers/report_controller_spec.rb
@@ -61,10 +61,15 @@ describe ReportController, :type => :controller do
   end
   describe 'download' do
     it 'should download valid CSV data' do
-      get :download
-      expect(response.status).to eq(200)
+      get :download, fields: ' '
+      expect(response).to have_http_status(:ok)
       expect(response.header['Content-Disposition']).to eq('attachment; filename=report.csv')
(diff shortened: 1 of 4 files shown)


spec/controllers/report_controller_spec.rb
```diff
@@ -61,11 +61,16 @@
   end
   describe 'download' do
     it 'should download valid CSV data' do
-      get :download
-      expect(response.status).to eq(200)
+      get :download, fields: ' '
+      expect(response).to have_http_status(:ok)
```
mejackreed: Nice :heart: !

mejackreed:
It works great and looks good! Is there a simple way to add a test to see if the `Admin Policy ID` column is not omitted if future changes happen?

drh-stanford:
Ok, I updated the spec to include the mandatory fields and whether they're on by default

mejackreed:
Awesome!

PR #662, opened 2016-03-09
remove deprecated sdr.url setting

jmartin-sul:
allow the dor-services sdr client to autoconfigure (to proxy to sdr via dor-services-app) by removing the (deprecated) explicit setting of sdr.url.

remove corresponding entry from settings.yml, since it isn't used anywhere else.


jmartin-sul:
coverage on the last commit merged to master was 81.46%, so this is a 0.03% improvement (should coveralls stay yellow).

i deployed this to argo-dev, along with the shared_configs change that removes `Settings.SDR_URL`, and i was able to call `Sdr::Client.current_version 'druid:bc849vk0855'` from rails console (which requires talking to SDR via the now autoconfigured dor-services-app proxy).  all worked as expected (a 404 was returned in this case, since that object doesn't exist on dev).

cbeer:
:+1: :clap:

PR #665, opened 2016-03-09
only activate Purge button if object is in registered state

### LINKED ISSUE #634: Purge is enabled for an object in accessioning
In stage, object druid:cz273df8849is in accessioning but the Purge button is enabled.  It should oinly be available for "Registered" objects.

screen shot 2016-03-02 at 2 49 11 pm

drh-stanford:
This PR fixes #634 by changing the disabled/enabled check on the Purge button to use the processing status facet value.

### DIFF
--- a/app/helpers/argo_helper.rb
+++ b/app/helpers/argo_helper.rb
@@ -118,7 +118,7 @@ module ArgoHelper
         label: 'Purge',
         new_page: true,
         confirm: 'This object will be permanently purged from DOR. This action cannot be undone. Are you sure?',
-        check_url: workflow_service_submitted_path(pid)
+        check_url: !registered_only?(doc)
       }
 
       buttons << {:url => '/items/' + pid + '/source_id_ui', :label => 'Change source id'}
@@ -191,4 +191,9 @@ module ArgoHelper
     link_to 'MODS bulk loads', bulk_jobs_index_path(@document), :id => 'bulk-button', :class => 'button btn btn-primary'
   end
 
+  protected
+
+  def registered_only?(document)
+    ['Registered', 'Unknown Status'].include?(document['processing_status_text_ssi'])
+  end
 end
--- a/spec/features/enable_buttons_spec.rb
+++ b/spec/features/enable_buttons_spec.rb
@@ -18,7 +18,6 @@ feature 'Enable buttons' do
     expect(page).to have_css 'a.disabled', text: 'Close Version'
     expect(page).to have_css 'a.disabled', text: 'Open for modification'
     expect(page).to have_css 'a.disabled', text: 'Republish'
-    expect(page).to have_css 'a.disabled', text: 'Purge'
   end
   scenario 'buttons are enabled if their services return true', js: true do
     allow_any_instance_of(WorkflowServiceController).to receive(:check_if_can_close_version).and_return(true)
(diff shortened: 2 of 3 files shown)


mejackreed:
Is the only way to do this using the `Dor::Item` call?

drh-stanford:
The dor-services code has a bunch of application logic for determining the active milestones and thus the "Status" that you see on the show page. It accounts for versioning and whatnot. See https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/processable.rb#L107-L135

drh-stanford:
marking this [WIP] until I get some functional requirements clarification...

drh-stanford:
I was able to implement this using a Solr query rather than a Fedora fetch. Please review.

jmartin-sul:
well... my initial reaction is that i think i would rather base checks like this on fedora data instead of solr data?  for one object, it seems better to me to take the performance hit and know that we have an accurate idea of the object's state, since we know indexing can lag or otherwise be flaky.  unless we're really sure it's going to always be right (compared to fedora) in this situation.

well, it does appear there's at least a final state check before the object is purged, so there's some safeguarding there:
https://github.com/sul-dlss/argo/blob/69b7e97f5973c2323a314c1442d0375ce99d1534/app/controllers/items_controller.rb#L498

still slightly uncomfortable w/ using the index, but less so in light of that.  so... :ok_hand: by me.

mejackreed:
I'd be fine for the approach of checking the document in the helper. It breaks the pattern of the async queries, but not necessary if we already have the information.

mejackreed:
The original reason to move to the async queries was to remove external blocking requests from the request / response cycle. The ones moved out were wfs ones that seemed self contained.

drh-stanford:
One of the issues I ran into is where to put the check -- I decided on DorController, but if the render_buttons needs to check for the Purge active/inactive then I don't need that route.

There isn't an Argo example I see already of toggling a button active/inactive other than the `check_url`. Is this possible without going through check_url? It looks to me like I'd need to modify this line in button_link_list:

```
    <%= link_to button[:label], button[:url], class: "btn button btn-primary #{'disabled' if button[:check_url]}", data: button_data %>
```

mejackreed:
Ah I was thinking about doing the check within the `ArgoHelper` LN109. 

For reference the 2.x way of checking to show this was:

https://github.com/sul-dlss/argo/blob/2.x-stable/app/helpers/argo_helper.rb#L208-L210

Which checked here:

https://github.com/sul-dlss/argo/blob/2.x-stable/app/helpers/dor_object_helper.rb#L142-L144

The current behavior in Argo is a straight port over of that.

mejackreed:
So if there is no need for an external request.. just put the conditional in the `render_buttons` method. This method is rough going already, but at least its not making blocking external requests.

drh-stanford:
I'm going to try to code it up so that it disables the button rather than omit it. I think I can do that with passing in true/false to `check_url` even though it's not a external URL.

LynnMcRae:
@drh-stanford said: I'm going to try to code it up so that it disables the button rather than omit it.

I thought the behavior was already correct, that it was disabled/enabled as appropriate and that only In Accessioning was incorrect.  Poking around now I see that Purge is always on!  So the pattern of other buttons -- enabled/disabled based on object status as well as user privs -- would indeed be ideal.

drh-stanford:
Ok, I've changed the implementation and this PR is ready for review.

PR #674, opened 2016-03-12
fix bulk operations on search results

jmartin-sul:
`Report#pids` should expect search result fields to be keyed on symbols, not strings.

looks like a minor regression as a result of:  https://github.com/sul-dlss/argo/commit/4b158da1293edccfee8d3ddb8578c1ac55f0c5f5  (a quick test where i reverted that commit fixed the issue, but i doubt reverting the commit is actually the right fix)

i assume report field configs were changed from using strings to symbols for some reason, so i'm guessing the right thing to do here is to fix the `pids` method?

this is deployed to stage, and seems to fix the problem with getting a list of druids from search results for bulk update (but it replaces the branch with the dor-services upgrade that fixes a couple indexing things, so indexing on argo-stage is back to being like argo-prod).

leaving this marked as WIP, because it'd be nice to write a test, but i'm not really sure what to do there.  i'm inclined to have `search_results` return fake data in the expected format, but i'm not sure how to mock the call to it.

### DIFF
--- a/app/controllers/report_controller.rb
+++ b/app/controllers/report_controller.rb
@@ -15,11 +15,7 @@ class ReportController < CatalogController
   end
 
   def data
-    # if !params[:sidx] || params[:sidx] == 'druid'
-    #  params[:sidx] = 'id'
-    # end
     params[:sord] ||= 'asc'
-    # params[:sort] = "#{params.delete(:sidx)} #{params.delete(:sord)}" if params[:sidx].present?
     rows_per_page = params[:rows] ? params.delete(:rows).to_i : 10
     params[:per_page] = rows_per_page * [params.delete(:npage).to_i, 1].max
     @report = Report.new(params, current_user: current_user)
@@ -41,13 +37,13 @@ class ReportController < CatalogController
   end
 
   def pids
-    # params[:per_page]=100
-    # params[:rows]=100
-    ids = Report.new(params, ['druid'], current_user: current_user).pids params
     respond_to do |format|
       format.json do
         render :json => {
-          :druids => ids
+          :druids => Report.new(params, current_user: current_user).pids(
+            source_id: params[:source_id].present?,
+            tags: params[:tags].present?
+          )
         }
       end
     end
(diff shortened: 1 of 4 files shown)


drh-stanford:
Re: tests. I think you can return some fake data for the document since there's only 3 cases in that `if` statement in `pids`. We should have a test that would have at least caught the regression.

drh-stanford:
I've added a test for `Report.pids` (which just querys for "report") but I had to change the method so that the `params` variable wasn't shadowed. So, I'm not sure I fully understand that method's intent and why `params` and `@params` are different. ???

drh-stanford:
the controller sends in the same params value
```
ids = Report.new(params, ['druid'], current_user: current_user).pids params
```

also I don’t think this method in the controller actually works…the `fields` parameter is busted imho (i.e., there's no match for `'druids'`)
```
  def pids_from_report(params)
    Report.new(params, ['druids'], current_user: current_user).pids params
  end
```

mejackreed:
Yeah the way that `ReportController` and `Report` interact is strange indeed. I would be for removing the params passed in value if possible in favor of removing the passed in `params` to the `pids` method if possible in favor of just using the already available instance variable. Does that reduce some complexity?

drh-stanford:
I've updated the `Report.pids` signature to make it more clear.

But the Report and ReportController tests are thin at best. They need to check the return data. For example, if I go to http://localhost:3000/report/pids.json it works fine, but in the controller spec the `get :pids` method doesn't return any pids...there's something wrong with the mocking there???

jmartin-sul:
:+1: @drh-stanford's further implementation and tests.  in this case, i like the integration-ish test approach more than mocking `search_results` i'd initially thought about.  seems more likely to catch the type of regression that occurred here.

and i prefer the new `params`/`opts` naming to the old style anyway, and it's tested, so happy to keep it, even if the breakage with the old version was somewhat unexpected.  the old style was a messy way to name that stuff, happy to see it cleaned up.

took out of [WIP], but seemed appropriate to defer the merge to someone who didn't commit on the branch?

drh-stanford:
I've fixed the Report Controller tests since they were using a bogus mock user. Now it uses an admin user and also tests some basic properties of the data returned by the queries.

drh-stanford:
So, I'm convinced that `pids_from_report` didn't work as intended and it wasn't being tested anyways, so I've removed it and put the needed code inside `reset`. I've refactored the `reset` tests so that they query for pids rather than mock the pids.

Issue #696, opened 2016-03-23
In stage, unable to Open an accessioned item for modification


LynnMcRae:
The "Open for modification" blue button should be enabled for any item with a Processing Status of "accessioned" (or "Accessioned" as show in the facet).  This is the only status where this has meaning and should be allowed. It works ok in prod.
![screen shot 2016-03-23 at 1 49 00 pm](https://cloud.githubusercontent.com/assets/1824732/14000339/10d92af6-f0fe-11e5-8100-30acc5cb464a.png)

LynnMcRae:
The blue button seems to be coming on in stage today, but leads to a system error ("Sorry, something went wrong") -- https://sul-squash.stanford.edu/projects/argo/environments/staging/bugs/38

LynnMcRae:
This latest may be an artificial testing situation ... an object that claims to be accessioned without a version ingested into SDR, a situation that itself calls for a more explicit consistency error perhaps, but not a flaw in the current Open implementation.

Leaving ticket for the original bug, where the blue button logic didn't seem to be enabling properly. Will attempt to reproduce.

ndushay:
This may be related to #698.  Investigating.

LynnMcRae:
Here are some current examples in stage where the "Open for modification" button does not come on.  The solr doc and status information look fine, so I can't tell yet what distinguishes these from seemingly similar cases where the blue button is enabled. 

Bad:
https://argo-stage.stanford.edu/catalog/druid:bb027sf6965
https://argo-stage.stanford.edu/catalog/druid:bb050dj7711

I'd love it if this is a test data issue. I can't yet find any failing cases in prod but need to determine this is not a logic issue that would migrate forward with a new release. Still looking.

ndushay:
I am looking at this.  So far found a test with a typo.

##### bb050dj7711
- https://sul-solr-d.stanford.edu/solr/argo_stage_shard1_replica4/select?q=druid:bb050dj7711
```xml
<doc>
<str name="id">druid:bb050dj7711</str>
<str name="status_ssi">v4 Accessioned</str>
<int name="current_version_isi">4</int>
<date name="modified_latest_dttsi">2015-04-09T20:19:32Z</date>
<str name="processing_status_text_ssi">Accessioned</str>
<int name="processing_status_code_isi">6</int>

LynnMcRae:
[...]

PR #701, opened 2016-03-24
enable_buttons_spec feature had a typo

ndushay:
Given that this test works with a typo, it's not clear to me if the test is doing its job.  Is it supposed to have a line  

```ruby
visit catalog_path 'druid:hj185vb7593'
```
after the allow statements?


ndushay:
making a diff branch for this;  it's not related to issue #696  anymore.

PR #702, opened 2016-03-24
fix test for enabled buttons when services return true

ndushay:
previously, the test gave false positive.  This was confirmed when fixing the typo made no change.   My changes actually do test something -- changing each "allow" result to return false instead of true in fact makes the test fail as expected.

@jmartin-sul @tingulfsen    this should be a nobrainer merge ...


jmartin-sul:
tests failed before, but error looked like a transient issue getting an .xsd file from loc.gov.  re-running the travis tests.  :+1: if they pass.

mejackreed:
ty! My fault here. Also an issue with `allow_any_instance_of` :+1:

PR #703, opened 2016-03-25
Install whenever gem and configure it to run argo:htaccess…

jkeck:
… every 5 minutes.

This needs to be to deployed to stage to validate if it works as expected.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -27,6 +27,7 @@ require 'capistrano/passenger'
 require 'dlss/capistrano'
 require 'squash/rails/capistrano3'
 require 'capistrano/delayed-job'
+require 'whenever/capistrano'
 
 # Loads custom tasks from `lib/capistrano/tasks' if you have any defined.
 Dir.glob('lib/capistrano/tasks/*.rake').each { |r| import r }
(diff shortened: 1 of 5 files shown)


jmartin-sul:
:+1: 

@tingulfsen had a branch deployed to stage, but i think it's possible that the things it was deployed to test have been tested?  if that's the case, let's deploy this branch, and if it looks good, let's merge it to master.

jkeck:
Just deployed this to -dev and verified that the cron-job was written.

```bash

# Begin Whenever generated tasks for: argo_dev
0,5,10,15,20,25,30,35,40,45,50,55 * * * * /bin/bash -l -c 'cd /opt/app... RAILS_ENV=development bundle exec rake argo:htaccess --silent'

# End Whenever generated tasks for: argo_dev
```

I think this should be ready to :ship:.

jmartin-sul:
looks like the job is running:
```
[lyberadmin@argo-dev ~]$ cd argo/current/
[lyberadmin@argo-dev current]$ ls -lah public/.htaccess 
-rw-rw-r-- 1 lyberadmin lyberadmin 3.0K Mar 25 15:15 public/.htaccess
[lyberadmin@argo-dev current]$ ls -lah public/.htaccess 
-rw-rw-r-- 1 lyberadmin lyberadmin 3.0K Mar 25 15:20 public/.htaccess
```

viewed the contents of the file and it looks good.  merging...

PR #704, opened 2016-03-28
Fix versioning of new objects to include correct arguments.

### LINKED ISSUE #671: Open Version via Bulk Upload produces invalid versionMD
Possibly related to #670 - it appears that bulk upload, when opening a new version, produces an incomplete versionMD stanza that does not supply semantic version tag or description. @LynnMcRae - should we try the same exercise you did with the bulk MODS upload with a spreadsheet upload to verify?

screenshot 2016-03-11 07 08 25

screenshot 2016-03-11 07 08 35

### LINKED ISSUE #693: Spreadsheet/MODS upload in stage fails to update objects, no errors logged
See https://argo-stage.stanford.edu/catalog/druid:ws024gz0298/bulk_jobs_index

The first (earlier) attempt labeled "Testing" was a spreadsheet convert&load.  The conversion was fine (see XML) but no objects got loaded.  No errors indicated in the log.

The second run used the output from the first to do just a MODS load.  Same result ... zero successes, no indication of error in the log.

tingulfsen:
Fixes #693
Fixes #671

### DIFF
--- a/app/jobs/modsulator_job.rb
+++ b/app/jobs/modsulator_job.rb
@@ -45,7 +45,7 @@ class ModsulatorJob < ActiveJob::Base
         log.puts('argo.bulk_metadata.bulk_log_xml_only true')
       elsif filetype != 'xml' # If the submitted file is XML, we never want to load anything into DOR
         log.puts('argo.bulk_metadata.bulk_log_xml_only false')
-        update_metadata(apo_id, response_xml, log) # Load into DOR
+        update_metadata(apo_id, response_xml, original_filename, user_login, log) # Load into DOR
       end
 
       log.puts("argo.bulk_metadata.bulk_log_job_complete #{Time.now.strftime(TIME_FORMAT)}")
@@ -56,11 +56,13 @@ class ModsulatorJob < ActiveJob::Base
 
   # Upload metadata into DOR.
   #
-  # @param  [String] druid         The governing APO's druid.
-  # @param  [String] xml_string    A MODS XML string.
-  # @param  [File]   log           Log file handle.
+  # @param  [String] druid               The governing APO's druid.
+  # @param  [String] xml_string          A MODS XML string.
+  # @param  [File]   log                 Log file handle.
+  # @param  [String] user_login          The login name of the current user_login
+  # @param  [String] original_filename   The name of the uploaded file
   # @return [Void]
-  def update_metadata(druid, xml_string, log)
+  def update_metadata(druid, xml_string, original_filename, user_login, log)
     return if xml_string.nil?
 
     root = Nokogiri::XML(xml_string).root
@@ -94,15 +96,7 @@ class ModsulatorJob < ActiveJob::Base
           next
         end
 
-        # If the object is currently in the opened version state, then go ahead, otherwise (unless the status is "Registered")
-        # open a new version first, but do not close it
-        if dor_object.status_info[:status_code] != 9 && dor_object.status_info[:status_code] != 1
-          unless DorObjectWorkflowStatus.new(dor_object.pid).can_open_version?
-            log.puts("argo.bulk_metadata.bulk_log_unable_to_version #{current_druid}")  # totally unexpected
-            next
-          end
-          dor_object.open_new_version()
-        end
+        version_object(dor_object, original_filename, user_login, log)
 
         dor_object.descMetadata.content = mods_node.to_s
         dor_object.save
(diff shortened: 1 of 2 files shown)


app/jobs/modsulator_job.rb
```diff
@@ -60,7 +60,7 @@ def perform(apo_id, uploaded_filename, output_directory, user_login, filetype =
   # @param  [String] xml_string    A MODS XML string.
   # @param  [File]   log           Log file handle.
   # @return [Void]
```
jkeck: Does this comment require updating to match the new method signature?

app/jobs/modsulator_job.rb
```diff
@@ -121,6 +113,32 @@ def update_metadata(druid, xml_string, log)
...
+      :description => "Descriptive metadata upload from #{original_filename}",
+      :opening_user_name => user_login
+    }
+    dor_object.open_new_version({:vers_md_upd_info => vers_md_upd_info})
+  end
+
+  def accessioned(dor_object)
+    return (dor_object.status_info[:status_code] == 6 ||
+            dor_object.status_info[:status_code] == 7 ||
+            dor_object.status_info[:status_code] == 8)
```
jkeck: This could be more succinctly written as 

```ruby
(6..8).include?(dor_object.status_info[:status_code])
```

jkeck: Also, it appears that we have a bit of primitive obsession going on here.  If the owner of this object has not modeled these statuses other than their internal numeric representation, it may be worth considering extracting a class to model them in a future refactor.

app/jobs/modsulator_job.rb
```diff
@@ -121,6 +113,32 @@ def update_metadata(druid, xml_string, log)
     end
   end
 
+  # Open a new version for the given object if it is in the accessioned state.
+  def version_object(dor_object, original_filename, user_login, log)
```
jkeck: I tend to agree with avoiding unused arguments.  It doesn't appear that this method requires a `user_login`.  Is it possible to remove that from the method signature?

app/jobs/modsulator_job.rb
jmartin-sul: the dor-services code that results in this issue is actually the dor-services code i mentioned on chat recently as an example of primitive obsession that could stand to be fixed (and the primitive obsession is at least a result of refactoring it to be more modular and reusable by different callers).

app/jobs/modsulator_job.rb
```diff
@@ -121,6 +115,40 @@ def update_metadata(druid, xml_string, log)
     end
   end
 
+  # Open a new version for the given object if it is in the accessioned state.
+  # @param   [Dor::Item]  dor_object          The object to version
+  # @param   [String]     original_filename   The name of the uploaded file
+  # @param   [String]     user_login          The current user_login
+  # @param   [File]       log                 Log file handle
+  def version_object(dor_object, original_filename, user_login, log)
+    if(accessioned(dor_object))
```
ndushay: I agree with hound on this

app/jobs/modsulator_job.rb
```diff
@@ -121,6 +115,40 @@ def update_metadata(druid, xml_string, log)
...
+      if !DorObjectWorkflowStatus.new(dor_object.pid).can_open_version?
```
ndushay: don't care about unless vs if, personally, but line 124 is missing the space after the if ...

spec/jobs/modsulator_job_spec.rb
```diff
@@ -106,12 +106,81 @@
...
+
+    it 'does not update the version if the object is in the registered state' do
+      expect(@dor_object).to receive(:status_info).at_least(:once).and_return({status_code: 1})
+      expect(@mj).not_to receive(:commit_new_version)
+
+      @mj.version_object(@dor_object, 'any_filename', 'any_user', @log)
+    end
+
+    it 'does not update the version if the object is in the opened state' do
+      expect(@dor_object).to receive(:status_info).at_least(:once).and_return({status_code: 1})
```
ndushay: is status_code 1  "opened" as well as "registered" ?   Or is this a typo?   Are there class constants for these status_code values with more human values, like  REGISTERED, OPENED, etc.  ?    Having to know the number codes is not the greatest thing ...

jmartin-sul: status_code 1 is indeed both opened and registered.  i think the idea is that registered is a special case of opened.  i agree that the way it's implemented is confusing, but IIRC, there is logic that currently depends on the two being the same, and it would take a bit of effort to understand how to untangle things.  i think i tried giving opened and registered separate status codes at one point, and the tests started failing, and i declined to investigate further at the time.

ndushay:
My argo-n00b eyes found this:
- a probably typo on status_code value in a test
- a dislike of using hardcoded integer status codes instead of class constants or symbolic values
- this PR does indeed do what it says:  it passes the arguments to versioning ... but I can't tell if that will fix the problem.  Perhaps that's in a different code base.
- wondering if we know that filename and username will always be reasonable values, of it it's up to ??dor-services?? to test for that ...

app/jobs/modsulator_job.rb
jmartin-sul: i would do space or parens, but not both.  i.e., i'd vote for leaving it as-is, or switching to `if accessioned(dor_object)`.  i guess the latter would be preferable just for consistency's sake.

app/jobs/modsulator_job.rb
```diff
@@ -121,6 +115,40 @@ def update_metadata(druid, xml_string, log)
...
+      :description => "Descriptive metadata upload from #{original_filename}",
+      :opening_user_name => user_login
+    }
+    dor_object.open_new_version({:vers_md_upd_info => vers_md_upd_info})
+  end
+
+  # Returns true if the given object is accessioned, false otherwise.
+  # @param   [Dor::Item]  dor_object  A DOR object
+  def accessioned(dor_object)
+    (6..8).include?(dor_object.status_info[:status_code])
```
jkeck: I always forget that Ranges have the `#cover?` method.  We should use that instead as it is more efficient.

tingulfsen:
I've updated this pull request to address most of the comments. I don't like the way Rubocop wants me to align the the rescue statements with the end block, because that would indent the rescue statements on the same level as the function definition.

I agree that this code is ugly, but given our current situation I wonder if we should leave it as is or take on a major refactoring. I'm leaning towards creating a refactoring ticket for the modsulator_job to be tackled when porting this to the new bulk actions framework.

spec/jobs/modsulator_job_spec.rb
ndushay: I think my concern is that these 2 tests have exactly the same code ... so do we need both of them?

ndushay:
I'm okay with ignoring rubocop for the alignment;  My only remaining concern is about the two tests that have exactly the same code but different description text.

ndushay:
also: are we waiting for verification that this code fixes the problem before we merge?

tingulfsen:
OK, @ndushay thanks for spotting that spurious test. I've cleaned up the commit history and fixed the test.

ndushay:
:+1: to merge.

Issue #652, opened 2016-03-04
Bulk Download Not Functioning on Large Queries


blalbrit:
See https://argo.stanford.edu/catalog?f%5Bnonhydrus_collection_title_ssim%5D%5B%5D=Stanford+Historical+Photograph+Collection with 14K objects. This is one of the key use-cases for bulk download support so need to determine if it's a size issue or something else happening.

tingulfsen:
This is a size issue. When starting the bulk download job, we create a list of all the druids to download. This list is stored within MySQL, which barfs on the long list. We may be able to alter the code to pass the druids in a more sane way to the job. The crash happens on this line:

https://github.com/sul-dlss/argo/blob/master/app/models/bulk_action.rb#L54

The job_params hash has too many elements.

@blalbrit  if you can find a way to reduce the number of objects by manually chunking the download job, that could be a quick workaround for now.

blalbrit:
@tingulfsen - I suppose our stakeholders can be forced to chunk the download job, but there is no clear mechanism for that in large, undifferentiated collections like the one in question where there *are* 25K objects that need to be downloaded as a group and are not already sub-divided in any way. Maybe they could generate the list of druids, and then paste in 1000 or max-number-downloadable at a time?

Do you have any way of knowing what max-number-downloadable actually is?

@LynnMcRae - do you see any other work-arounds than forcing a stakeholder to create arbitrary groupings of content in order to be able to download large collections?

LynnMcRae:
Is the druid list simply saved as a column in the table? Given the 64k row size limit in MySql, we should have known that would impose a druid list limit around 5000. Can it be saved as an artifact in the file system instead? It's not really a good use of the database in the first place if the facility is meant to scale arbitrarily, which was a requirement for the same reasons we run it in the background instead of in the user's session.  @tingulfsen Did you have other ideas for getting past this limit?

drh-stanford:
In MySQL, the `:text` column can be changed to `:longtext` so that it turns into a 4G blob rather than a 64k blob.

tingulfsen:
Thanks, @drh-stanford - that's a good idea. I'm a little reluctant to alter the MySQL schema because it is not really within Argo's domain. It's actually defined within DelayedJob/ActiveJob. So while fixing the schema is a possibility, I'm more inclined to change the ways arguments are passed into perform_later().

jkeck:
[...]

PR #707, opened 2016-03-28
Increase Delayed Job column size to accommodate many more druids when…

tingulfsen:
… creating a new job.

### DIFF
--- a/db/migrate/20160318203331_increase_delayed_job_column.rb
+++ b/db/migrate/20160318203331_increase_delayed_job_column.rb
@@ -1,0 +1,8 @@
+class IncreaseDelayedJobColumn < ActiveRecord::Migration
+  def change
+    # Ensure that the handler column in the delayed_job table is MySQL longtext
+    # so that we can pass in lots of druids in the params when a new Delayed
+    # Job is created.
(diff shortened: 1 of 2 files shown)


tingulfsen:
Fixes #652

ndushay:
:+1:  from me.  Again, @jkeck may have some insights as he is more knowledgeable re: databases.

jmartin-sul:
:+1: from me too.

can we test it on stage?

LynnMcRae:
We can test this on stage but need to be told if/when it is on stage please.

jkeck:
:+1:

ndushay:
also: are we waiting for verification that this code fixes the problem before we merge?

tingulfsen:
Yep, it would be ideal to verify in staging before merging, although I don't think merging at this time would do anything bad.

LynnMcRae:
From a large run this afternoon:
screen shot 2016-03-29 at 6 05 34 pm

- From the download file, all but one object's descMetadata was downloaded
- From the Log, there was one error midway with detail diagnostics, then the run seemed to resume just fine (Tommy has the log)
- So the numbers reported are off but the utility definitely want way past the older ~500 item limit and processed them correctly.

PR #708, opened 2016-03-29
Remove rubocop rake task

jkeck:
`¯\_(ツ)_/¯`

### DIFF
--- a/lib/tasks/rubocop.rake
+++ b/lib/tasks/rubocop.rake
@@ -1,9 +1,0 @@
-begin
-  require 'rubocop/rake_task'
-
-  RuboCop::RakeTask.new do |task|
-    task.options = %w(--format simple)
-  end
-rescue LoadError
-  puts 'Unable to load RuboCop.'
-end


ndushay:
:+1:

lib/tasks/rubocop.rake
```diff
@@ -5,5 +5,5 @@ begin
     task.options = %w(--format simple)
   end
 rescue LoadError
-  puts 'Unable to load RuboCop.'
+  puts 'Unable to load RuboCop.' if Rails.env.development? || Rails.env.test?
```
ndushay: argo-dev  runs in development mode, so if we deploy to argo-dev, will we also get bombarded with these messages?

jkeck: Classic @tingulfsen. :wink: 

This was deployed to -dev for a while.  Since rubocop is a development dependency, it is installed and loads fine.

jmartin-sul:
i was left slightly confused by the code change, but the PR description cleared that misunderstanding right up.

PR #709, opened 2016-03-30
Consolidate wfs logic

mejackreed:
Adding this as a PR as work that @darrenleeweber and I paired on to consolidate some of the places where Argo calls the WorkflowService. Will solely leave it up to the Margo team to evaluate. Feel free to delete the branch if it isn't useful.


tingulfsen:
The team has decided to leave this pull request for inclusion in Argo during a later sprint, in order to minimize the changes for the releases that we do this time around (the work looks good and useful but is not considered to solve any blockers for our current sprint).

PR #716, opened 2016-04-01
Retry DOR connections when downloading descriptive metadata in bulk.

### LINKED ISSUE #714: Descriptive metadata download timeout
When using Bulk Actions to download descriptive metadata for large numbers of objects (more than 1000) it seems fairly common that some druids fail to load due to network problems. Investigate Fedora logs, and maybe retry the connection?

tingulfsen:
Closes #714 (hopefully).

This pull request adds a retry to Dor.find in case of timeout. I hope that this can mitigate some of the errors we are seeing with the descriptive metadata download function. Given the current time constraints our team is facing, I am implementing a pretty simple fix. If this does not help, some more time consuming debugging will be required.

### DIFF
--- a/app/jobs/descmetadata_download_job.rb
+++ b/app/jobs/descmetadata_download_job.rb
@@ -1,6 +1,9 @@
 class DescmetadataDownloadJob < GenericJob
   queue_as :default
 
+  MAX_TRIES = 3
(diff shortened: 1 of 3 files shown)


tingulfsen:
This branch was also tested on stage, and appeared to work.

PR #720, opened 2016-04-04
Extend maximum runtime for Delayed Job to a much larger value.

tingulfsen:
30 minutes is not enough time for e.g. descriptive metadata download jobs to complete when thousands of objects are selected. Let's bump the max runtime.


ndushay:
:+1:

PR #727, opened 2016-04-13
apo indexing fix

### LINKED ISSUE #725: APOs indexed incorrectly after an online update - roles disappear
We've had reports two of people losing role-based privileges suddenly, and in both cases it was fixed by reindexing the APO.

I have found that an online update of the APO does a reindex in a way that indexes the object as a Dor::Item instead of a Dor::AdminPolicyObject.  It get indexed correctly via the Blue Button.

Privileges are lost when the object is indexed as an Item because the roleMetadata datastream is only indexed for AdminPolicy object types.

After online update (a test APO in prod):
screen shot 2016-04-08 at 10 19 50 am

After manual reindex:
screen shot 2016-04-08 at 10 21 29 am

jmartin-sul:
fixed #725 by loading the object using `Dor.find` instead of `Dor::Item.find`

Bug was when you edited a datastream directly in argo, upon saving, the object would always be indexed in Solr as an item.  Now, it's actually indexed as its object type.

paired w/ @ndushay, who got us to add some `puts` statements and find an items_controller indexing bug much more quickly than if i'd kept being perplexed about why dor-indexing-app didn't seem to be getting indexing messages (but i'm still perplexed about that, and will file an investigation ticket for it).

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -670,7 +670,11 @@ class ItemsController < ApplicationController
   # Filters
   def create_obj
     raise 'missing druid' unless params[:id]
-    @object = Dor::Item.find params[:id]
+    create_obj_and_apo params[:id]
+  end
+
+  def create_obj_and_apo(obj_pid)
+    @object = Dor.find obj_pid
     @apo = @object.admin_policy_object
     @apo = ( @apo ? @apo.pid : '' )
   end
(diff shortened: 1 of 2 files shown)


jmartin-sul:
while i was able to reliably reproduce this on my laptop, i was not able to reproduce this via web UI on stage.  however, if i manually did the old broken thing in rails console (index an obj that was loaded using `Dor::Item.find` instead of `Dor.find`), i was able to reproduce on stage.  somewhat contrived, but a good indication that the old behavior was, in fact, incorrect, since it generated the same sort of bad solr doc we saw in prod and on my laptop.

@ndushay thinks that the delay lynn saw yesterday could be due to the solr update not being committed immediately.  i.e., the object is reindexed synchronously by argo, but the new solr document isn't returned on requests for it till the commit happens a few minutes later.

PR #742, opened 2016-05-04
prevent 500 error when a bulk job log file is missing

### LINKED ISSUE #734: Add check for null in the controller for bulk upload
screenshot 2016-04-25 10 17 12

https://argo.stanford.edu/catalog/druid:yf767bj4831/bulk_jobs_index

jmartin-sul:
* CatalogController#load_bulk_jobs: coerce sort field to string to prevent ArgumentError
* _bulk_index_table.html.erb: prevent nil String references when job['dir'] is nil

resolve #734 
resolve sul-dlss/media#45

### DIFF
--- a/app/views/catalog/_bulk_index_table.html.erb
+++ b/app/views/catalog/_bulk_index_table.html.erb
@@ -1,62 +1,62 @@
 <div id="bulk-upload-table">
-    <table class="table table-hover">
-	<tr>
-	    <th>When</th><th>Who</th><th>File Name</th><th>Note</th><th>Status <%= link_to bulk_status_help_path, :data => { ajax_modal: 'trigger', ajax_modal_title: 'Status Information' } do %><span class="glyphicon glyphicon-question-sign" aria-hidden="true"></span><% end %></th><th>Druids</th><th>Success</th><th/><th/>
-	</tr>
-	<% @bulk_jobs.each do |job| %>
-	<tr>
-	    <td><%= job['argo.bulk_metadata.bulk_log_job_start'] %></td>
-	    <td><%= job['argo.bulk_metadata.bulk_log_user'] %></td>
-	    <td><%= job['argo.bulk_metadata.bulk_log_input_file'] %></td>
-	    <td><%= job['argo.bulk_metadata.bulk_log_note'] %></td>
-	    <td>
-		<% if(job.has_key?('error')) %>
-		  system errors
-		<% elsif(job['argo.bulk_metadata.bulk_log_job_complete']) %>
-		  completed
-		<% else %>
-		  in process
-		<% end %>
-            </td>
-	    <td><%= job['argo.bulk_metadata.bulk_log_record_count'] %></td>
-	    <td><%= job['argo.bulk_metadata.bulk_log_druids_loaded'] %></td>
+  <table class="table table-hover">
+    <tr>
+      <th>When</th><th>Who</th><th>File Name</th><th>Note</th><th>Status <%= link_to bulk_status_help_path, :data => { ajax_modal: 'trigger', ajax_modal_title: 'Status Information' } do %><span class="glyphicon glyphicon-question-sign" aria-hidden="true"></span><% end %></th><th>Druids</th><th>Success</th><th/><th/>
+    </tr>
+    <% @bulk_jobs.each do |job| %>
+      <tr>
+        <td><%= job['argo.bulk_metadata.bulk_log_job_start'] %></td>
+        <td><%= job['argo.bulk_metadata.bulk_log_user'] %></td>
+        <td><%= job['argo.bulk_metadata.bulk_log_input_file'] %></td>
+        <td><%= job['argo.bulk_metadata.bulk_log_note'] %></td>
+        <td>
+          <% if(job.has_key?('error')) %>
+            system errors
+          <% elsif(job['argo.bulk_metadata.bulk_log_job_complete']) %>
+            completed
+          <% else %>
+            in process
+          <% end %>
+        </td>
+        <td><%= job['argo.bulk_metadata.bulk_log_record_count'] %></td>
+        <td><%= job['argo.bulk_metadata.bulk_log_druids_loaded'] %></td>
 
-      <% if(job.has_key?('dir')) %>
-        <% druid_and_time = job['dir'].split(/\//) %>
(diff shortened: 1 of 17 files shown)


jmartin-sul:
this does appear to resolve the 500 error described in issue #734, but i'm leaving it marked as [WIP] because:
* i'm not sure this is the right solution in the application code
* i'd like to add relevant unit tests

jmartin-sul:
alright, so there's one minor known problem/regression with this fix:  when running unit tests, some job related processing directories get created, and don't get cleaned up.  e.g.:
```
spec/fixtures/bulk_upload/workspace/DescmetadataDownloadJob_1/
spec/fixtures/bulk_upload/workspace/DescmetadataDownloadJob_2/
spec/fixtures/bulk_upload/workspace/GenericJob_1/
spec/fixtures/bulk_upload/workspace/GenericJob_2/
spec/fixtures/bulk_upload/workspace/GenericJob_3/
```

this wasn't happening previously, because the entire bulk upload workspace was getting deleted by one of the tests (which pointed to the bulk upload workspace as its output dir, then removed the whole thing after that set of tests finished, see [descmetadata_download_job_spec.rb](https://github.com/sul-dlss/argo/blob/e178c766d562a5cbf14384921a5b064710224ae0/spec/jobs/descmetadata_download_job_spec.rb#L24)).  however, the tests added in this PR need fixtures in the bulk upload workspace, and so the aggressive cleanup of the existing test would prevent the user from easily running the test suite multiple times on any environment that didn't get re-created for each test run (e.g., a dev laptop).

after discussing w/ @tingulfsen, it seemed like the best option was to do what's been done in this PR:
* have the test that was hijacking `Settings.BULK_METADATA.DIRECTORY` for its own output point to its own subdirectory in the argo `tmp` dir.
* have test.yml specify the bulk upload workspace in the fixtures directory for `Settings.BULK_METADATA.DIRECTORY`.
* file a bug for dealing with the strays that are now generated by tests.

we agreed that the regression here is just a minor inconvenience for laptop developers.  shouldn't affect travis.  and it keeps us from perpetuating an overly broad teardown, and makes the new and existing tests a bit more realistic for having to work inside of a more persistent bulk upload workspace.

we tried to suss out what exactly is creating the files that don't get cleaned up, so as to fix those tests to clean up those files specifically (via `BulkAction#destroy`).  but we didn't have any luck doing that (we weren't able to get all the strays, and still weren't sure what was creating them or where further cleanup should get added).

happy to revisit if others disagree with that decision, but after a couple hours of testing, it didn't seem worth further pursuit for now.

jmartin-sul:
otherwise, this is ready for review.  @tingulfsen (and @ndushay and @jkeck)

jmartin-sul:
i think i added test coverage for the little bit of new code that was added, and added coverage for some uncovered existing code, so i'm not sure how overall coverage dropped.

spec/views/catalog/_bulk_index_table.html.erb_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  it 'has links to log info and XML for each row with log info' do
+    render
+    bulk_jobs[1..2].each do |job|
+      druid_and_time = job['dir'].split %r(\/)
+      expect(rendered).to have_link 'Log', href: bulk_jobs_log_path(druid_and_time[0], druid_and_time[1])
+      expect(rendered).to have_link 'XML', href: bulk_jobs_xml_path(druid_and_time[0], druid_and_time[1])
+    end
+  end
+
+  it 'prints error placeholders for each row with no long info' do
```
tingulfsen: You meant "no log info" ?

spec/views/catalog/_bulk_index_table.html.erb_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    render
+    expect(rendered).to have_css 'div#bulk-upload-table table tr td', text: 'error:  job log dir not found', count: 3
+    expect(rendered).to have_css 'div#bulk-upload-table table tr', text: 'error:  job log dir not found', count: 1
```
tingulfsen: I think I can see where this comes from, but would you be able to attach a representative screenshot?

tingulfsen:
Other than my comments above, I'm 👍 on this.

jmartin-sul:
test description fixed, deployed to stage, screenshot attached (see lower right-hand corner):
screen shot 2016-05-12 at 3 20 07 am

app/views/catalog/_bulk_index_table.html.erb
```diff
@@ -21,14 +21,23 @@
 	    <td><%= job['argo.bulk_metadata.bulk_log_record_count'] %></td>
 	    <td><%= job['argo.bulk_metadata.bulk_log_druids_loaded'] %></td>
 
-	    <% druid_and_time = job['dir'].split(/\//) %>
-	    <td><%= link_to('Log', bulk_jobs_log_path(druid_and_time[0], druid_and_time[1]), data: { ajax_modal: 'trigger', ajax_modal_title: 'MODS Bulk Load Log' }) %></td>
-	    <td><%= link_to('XML', bulk_jobs_xml_path(druid_and_time[0], druid_and_time[1])) %></td>
+      <% if(job.has_key?('dir')) %>
```
ndushay: indentation woes.  whoa whoa whoa

ndushay:
I made one line comment on indentation in the view file.   o.w. nothing jumped out at me.  So 👌 w me.

app/views/catalog/_bulk_index_table.html.erb
jmartin-sul: ugh, yeah, this looked ok in my text editor, but it's a hodgepodge of tabs and spaces, as it seems is often the case in view templates.  and actually, the second newly added if block should be indented another level regardless, now that i look at it.

will just add a commit fixing spacing.

ndushay: 👌

jmartin-sul:
indentation fixed.  i did the whole file, so i left it as a separate commit.  deploying update to stage...

jmartin-sul:
maybe @LynnMcRae or @blalbrit can confirm on stage, and then we can merge?  this is the URL from the screenshot demoing the fix:  https://argo-stage.stanford.edu/catalog/druid:bc682xk5613/bulk_jobs_index

jkeck:
👍

LynnMcRae:
Not an ideal situation that the history and detail of the aborted upload is orphaned and dangling off the end, but correcting the cause of the missing log file is another ticket. This restores access to the page as a whole which was the critical stopper.

jmartin-sul:
merging, as this has passed code review and has been verified on stage

PR #744, opened 2016-05-12
fix purge command to delete solr doc

### LINKED ISSUE #728: Purge no longer deletes the solr doc
Test object druid:np266dm3062 in prod Argo can be found via a search and shows on a result page based on solr, but if you try to show the detail page you get "Object not found" because it was purged.

It's possible the part of purge that deletes a doc from solr is pointing to the wrong solr code since the new index?

jmartin-sul:
fix ItemsController#purge_object so that it deletes the object from solr reliably and so that it handles object creation and user auth in a way that's more consistent with other methods in the controller/application.

items_controller_spec.rb: fix #purge_object tests.  note that i added a failing test in the first commit, and that if you check out 68dff40 and run `rspec`, you'll get one failure:
```
  1) ItemsController#purge_object deletes the object from fedora and solr
     Failure/Error: expect(Dor::SearchService.solr).to receive(:delete_by_id).with(@pid)
     
       (#<RSolr::Client:0x007fc46bf31308>).delete_by_id("druid:oo201oo0001")
           expected: 1 time with arguments: ("druid:oo201oo0001")
           received: 0 times
     # ./spec/controllers/items_controller_spec.rb:79:in `block (3 levels) in <top (required)>'
```
(i actually wrote the test after the fix, but `git rebase -i` was able to make me look a bit more thoughtful than i actually was in the moment)

was able to reproduce the problem locally on laptop, also tested fix manually on laptop.  can deploy to stage for further testing.

on a side-note, looking at the commit history, i'm not sure how this ever worked.  the solr doc delete used to be in a begin/rescue that wrapped inline object creation and auth check, and even when that rescue wasn't limited to `ActiveFedora::ObjectNotFoundError`, deleting the solr doc still would've required one of those things to fail.

resolve #728 
resolve sul-dlss/media#43

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -2,7 +2,6 @@ class ItemsController < ApplicationController
   include ModsDisplay::ControllerExtension
   before_action :create_obj, :except => [
     :open_bulk,
-    :purge_object,
     :register
   ]
   before_action :authorize_manage_obj_content!, :only => [
@@ -10,6 +9,7 @@ class ItemsController < ApplicationController
     :datastream_update,
     :mods,
     :open_version, :close_version,
+    :purge_object,
     :update_resource,
     :set_content_type,
     :source_id,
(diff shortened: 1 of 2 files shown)


ndushay:
FWIW, nothing jumps out at me, so 👌 by me ...

jmartin-sul:
once PR #742 is verified on stage, i can deploy this branch for testing.  or, i could create a branch that incorporated the branches from both PRs, and deploy that, if that'd be significantly easier for the testers.  @LynnMcRae @blalbrit

jkeck:
👍

jmartin-sul:
merging, as this has passed code review and has been verified on stage

Issue #746, opened 2016-05-13
Bulk upload can sometimes fail to make a log file 


LynnMcRae:
This bug led to page access errors which have been fixed by #734, but the underlying cause, the job creating a directory but not a log file, has not been fixed.  The user might perceive this as a "lost" submission because the only evidence of it is an unlabeled, undated entry dangling off the the bottom of the upload history display. So far it has been difficult to reproduce.

Issue #731, opened 2016-04-20
Metadata Upload Fails on Large Numbers of Objects


blalbrit:
Example, with a spreadsheet of 252 items, system errors are reported (Net::ReadTimeout on the Modsulator server). Running smaller batches of the spreadsheet work fine. History available here: https://argo.stanford.edu/catalog/druid:sb559ym0369/bulk_jobs_index (see Sri Lanka spreadsheet attached).

![screenshot 2016-04-20 10 17 54](https://cloud.githubusercontent.com/assets/796990/14683787/6dee8d0a-06e1-11e6-83d8-41dd715a59fa.png)

[SriLanka_BulkUpload.xlsx](https://github.com/sul-dlss/argo/files/228446/SriLanka_BulkUpload.xlsx)

tingulfsen:
I've done a little investigation, but the nature of this problem has proven to be quite difficult to track down. I think I'll need some input from @blalbrit and @LynnMcRae to figure out how to proceed, with respect to how much time we should devote to this.

LynnMcRae:
Here is a MODS/XML file of 500 records that fails (from Robert Rohrbacher, April 28):

> Hello,
> 	I'm trying to load a batch of 600 MODS records for the Rumsey Map Collection. I got system errors (https://argo.stanford.edu/catalog/druid:sq161jk2248/bulk_jobs_index)
> 
> Robert
> 

Note I have changed the file extension from .xml to .txt so I could upload it here.
[Batch4_1st.txt](https://github.com/sul-dlss/argo/files/241575/Batch4_1st.txt)

blalbrit:
Just tested this on production files using the Batch4_1st file listed above from Robert.

Results are:

<img width="903" alt="screenshot 2016-05-26 20 51 12" src="https://cloud.githubusercontent.com/assets/796990/15597311/c794390e-2383-11e6-9d60-9a22f42d9842.png">

blalbrit:
re: above comment, this can be found at https://argo.stanford.edu/catalog/druid:sq161jk2248/bulk_jobs_index

PR #747, opened 2016-05-13
Switch from Faraday to Rest Client, and wait longer for request timeo…

tingulfsen:
…uts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem 'config'
 gem 'delayed_job'
 gem 'delayed_job_active_record'
 gem 'equivalent-xml', '>= 0.6.0'   # For ignoring_attr_values() with arguments
-gem 'faraday'
+gem 'rest-client'
 gem 'eye'  # NOTE: if eye is upgraded, see the note in the 'bin/eye' script about checking to see whether that script needs upgrading (which won't happen automatically).
 gem 'jqgrid-jquery-rails'
 gem 'jquery-rails'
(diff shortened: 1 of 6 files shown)


tingulfsen:
This pull request partially addresses issues #746  and #731 by extending the timeout on POST requests to the MODSulator server. It also switches over from using Faraday to Rest Client, because I was unable to figure out how to set timeouts with POST requests using Faraday (sad, I know...).

This is only a partial fix, because there are performance issues when large files are uploaded (particularly large XML files). The current workaround is to break the files up into smaller pieces (e.g. under 10M.

During a sprint later this year, we plan to change the processing of XML files such that you can upload a zip file of many small XML files (each corresponding to a single MODS record). This will enable a "round trip" with descriptive metadata download, which currently provides just such a zip file as its output.

tingulfsen:
Note that the changes in app/jobs/modsulator_job.rb tie in with the corresponding changes in modsulator-app sul-dlss/modsulator-app#5

That is the API that is called.

jkeck:
@tingulfsen so the more that I think about it, I wonder if we're taking the wrong approach here.  If setting timeouts in Faraday does not work, we should report that sort of thing upstream.

I just setup the following to test Faraday timeouts for multipart POST requests: https://gist.github.com/jkeck/a973a12d2ea8b2bc4f85e067c9fb4724  This seems to work exactly as I would expect, so I wonder if there isn't something else that is the problem.

I am kind-of uncomfortable with reverting the use of Faraday back to rest-client.  It seems like a step in the wrong direction as we have consciously tried to remove rest-client in our production applications.

tingulfsen:
Superseded by #751

ndushay:
can this be closed for the time being to avoid confusion?

Issue #741, opened 2016-05-04
Bulk Update of tags does not copy forward existing tags for pasted druid list


LynnMcRae:
When doing a bulk tag update (the Tags button), the current set of tags should seed the box with existing values so they can be augmented, edited or removed. This works when "Get druids from search" is used, but the update box is empty for a pasted druid list.

oceanofsound:
2019 update: this is still a problem. As a user who occasionally works on content registered by other people, the ability to bulk add tags makes retrieval much easier. However, with this functionality broken, it can be very difficult to corral a set of objects without manually going to each one and adding a tag.

PR #751, opened 2016-05-21
Set modsulator-app request timeouts, and prevent files from being del…

tingulfsen:
…eted upon the first modsulator job failure.

### DIFF
--- a/app/jobs/modsulator_job.rb
+++ b/app/jobs/modsulator_job.rb
@@ -10,6 +10,9 @@ class ModsulatorJob < ActiveJob::Base
   # A somewhat easy to understand and informative time stamp format
   TIME_FORMAT = '%Y-%m-%d %H:%M%P'
 
+  # Wait 30 minutes for remote requests to complete
+  TIMEOUT = 1800
+
   # This method is called by the caller running perform_later(), so we're using ActiveJob with Delayed Job as a backend.
   # The method does all the work of converting any input spreadsheets to XML, writing a log file as it goes along.
   # Later, this log file will be used to generate a nicer looking log for the user to view and to generate the list of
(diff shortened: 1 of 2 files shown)


tingulfsen:
Thanks to @jkeck for showing how to use the not exactly intuitive Faraday library to POST requests with timeouts.

This pull request supersedes #747, which will be deleted once this pull request is approved.

tingulfsen:
Note that the changes in app/jobs/modsulator_job.rb tie in with the corresponding changes in sul-dlss/modsulator-app#5

tingulfsen:
This pull request partially addresses issues #741 and #731 by extending the timeout on POST requests to the MODSulator server. It also switches over from using Faraday to Rest Client, because I was unable to figure out how to set timeouts with POST requests using Faraday (sad, I know...).

This is only a partial fix, because there are performance issues when large files are uploaded (particularly large XML files). The current workaround is to break the files up into smaller pieces (e.g. under 10M.

During a sprint later this year, we plan to change the processing of XML files such that you can upload a zip file of many small XML files (each corresponding to a single MODS record). This will enable a "round trip" with descriptive metadata download, which currently provides just such a zip file as its output.

ndushay:
In your comment above, it says " It also switches over from using Faraday to Rest Client, because I was unable to figure out how to set timeouts with POST requests using Faraday (sad, I know...)."  - perhaps the cut and pasted text needs a small edit?

tingulfsen:
Sorry I edited that comment.

ndushay:
Not spending a ton of brainpower on this;  it looks okay to me in that "nothing-is-screaming-at-me" sort of way.  Shameless green!

jmartin-sul:
👍 from me.  would be nice to have someone more familiar with faraday give it a look, but it looks fine to me.

the only thing i can think of is something to test that faraday respects timeout settings, but how to go about writing that test isn't immediately apparent to me.  wouldn't hold up merging over it.

app/jobs/modsulator_job.rb
```diff
@@ -10,6 +10,9 @@ class ModsulatorJob < ActiveJob::Base
   # A somewhat easy to understand and informative time stamp format
   TIME_FORMAT = '%Y-%m-%d %H:%M%P'
 
+  # Wait 30 minutes for remote requests to complete
+  TIMEOUT = 1800
```
ndushay: in a perfect world, this would be in a Settings.yml file

PR #760, opened 2016-07-05
PermittedQueries should use the lucene query type; fixes #759

### LINKED ISSUE #759: Role-based access stopped working with solr 6.1 upgrade
at least we believe the two must be connected.  Try impersonating any workgroup that should give a users lots of stuff:

   workgroup:dlss:pmag-staff
   workgroup:sdr:dpg-staff
   workgroup:dlss:pmag-staff

If we can figure out why impersonate doesn't work, we willprobably know why user's role-based access doesn't work either.

cbeer:


jmartin-sul:
rebased on master and deployed to stage.  @LynnMcRae, can you do some testing on stage for this one?

app/models/concerns/permitted_queries.rb
```diff
@@ -41,12 +48,20 @@ def permitted_apos
   # FIXME: seems to include display logic
   # @return [Array<Array<String>>] Sorted array of pairs of strings, each pair like: ["Title (PID)", "PID"]
   def permitted_collections
-    q = 'objectType_ssim:collection AND !project_tag_ssim:"Hydrus" '
-    q += permitted_apos.map { |pid| "#{SolrDocument::FIELD_APO_ID}:\"info:fedora/#{pid}\"" }.join(' OR ') unless is_admin
+    q = if is_admin
+          '*:*'
+        elsif permitted_apos.empty?
+          '-id:[* TO *]'
```
ndushay: could this line be  '-id:*'  instead of the range query?

jmartin-sul:
a general 👍 from me after further explanation of the query mechanics by @ndushay (no opinion on her suggestion about the `-id` query since that's out of my solr depth).

app/models/concerns/permitted_queries.rb
```diff
@@ -21,16 +21,23 @@ def initialize(groups, known_roles, is_admin)
   # @return [Array[String]] list of DRUIDs from APOs that this User can view
   def permitted_apos
     query = groups.map { |g| g.gsub(':', '\:') }.join(' OR ')
-    q = 'apo_role_group_manager_ssim:(' + query + ') OR apo_role_person_manager_ssim:(' + query + ')'
-    known_roles.each do |role|
-      q += ' OR apo_role_' + role + '_ssim:(' + query + ')'
+
+    q = ''
+    if is_admin
+      q += '*:*'
```
ndushay: I'm assuming you checked to make sure the request handler is okay with *:* ?    I think it's okay for e/dismax but can't remember if it's ok for lucene??

ndushay:
LGTM except for my two questions.

app/models/concerns/permitted_queries.rb
cbeer: Yep, it's a lucene-ism in the first place.

app/models/concerns/permitted_queries.rb
cbeer: I'm not sure, but we've used this construct elsewhere. Honestly, the whole thing could use a refactor (why send the query if you're going to throw away the results?), but I'm nervous about doing too much with this.. lightly.. tested code.

LynnMcRae:
Impersonate on stage looks good, as well as subsequent queries.

app/models/concerns/permitted_queries.rb
```diff
@@ -21,16 +21,23 @@ def initialize(groups, known_roles, is_admin)
...
     end
-    q = 'objectType_ssim:adminPolicy' if is_admin
+
     resp = repository.search(
       q: q,
+      defType: 'lucene',
       rows: 1000,
       fl: 'id',
-      fq: '!project_tag_ssim:"Hydrus"'
+      fq: ['objectType_ssim:adminPolicy', '!project_tag_ssim:"Hydrus"']
```
ndushay: Hydrus shouldn't need to be in quotes??

app/models/concerns/permitted_queries.rb
ndushay: indeed.  silly me.

app/models/concerns/permitted_queries.rb
cbeer: Need to be? Probably not. Harmful? Not really. Anyway, that's how it was before.

app/models/concerns/permitted_queries.rb
cbeer: (I also wouldn't put too much thought into it... this is an extreme edge case, hopefully even prevented by restrictions elsewhere in the application)

ndushay: I just tested in web browser direct to Solr and get same result for id:*

app/models/concerns/permitted_queries.rb
```diff
@@ -41,12 +48,20 @@ def permitted_apos
...
+          permitted_apos.map { |pid| "#{SolrDocument::FIELD_APO_ID}:\"info:fedora/#{pid}\"" }.join(' OR ')
+        end
+
     result = repository.search(
       q: q,
+      defType: 'lucene',
       rows: 1000,
-      fl: 'id,tag_ssim,dc_title_tesim'
+      fl: 'id,tag_ssim,dc_title_tesim',
+      fq: ['objectType_ssim:collection', '!project_tag_ssim:"Hydrus"']
```
ndushay: again, no quotes needed for single term Hydrus

ndushay:
I believe this is a functioning fix;  I also believe there could be a few chars cleanup with little effort.  Whatever spirit moves y'all.

jmartin-sul:
@atz

jmartin-sul:
we should test with an actual non-super-user, since impersonate doesn't result in actual querying.

LynnMcRae:
Having little recourse, I made myself a mere mortal, subject to the heartache and the thousand natural shocks that flesh is heir to.  I confirmed that I had access to nothing in production, while in stage I had access to about 2/3 of the items and a subset of APOs and collections as when fully charged, a number consistent with the access based mainly on being in the dlss:developers group (among others) as verified afterwards via impersonate.

PR #769, opened 2016-07-08
better rights facets

jmartin-sul:
* upgrade to dor-services 5.9.0 to get rights indexing improvements
* catalog_controller.rb: replace rights_primary_ssim with rights_descriptions_ssim, add rights_errors_ssim below empties section

will require a full reindex to take advantage of the new facet.

new facet shows some data on stage, for a handful of objects with location restrictions that have been reindexed.

we should figure out how we want to deploy this and when we want to reindex before this gets merged.

i haven't added tests, since i don't think we're in the habit of adding tests for each facet.  but we can figure out testing if people think that'd be a good idea.

connect to sul-dlss/dor-rights-auth#17


jmartin-sul:
i believe this is now ready for review, since the dor-services changes should be done.  dor_indexing_app will also need to be upgraded.  then argo (the `argo3_prod` solr collection) will need to be reindexed to take advantage of the new facet (new rights info is in a new multivalued field, and the old single value field is no longer being presented in argo).

in the intervening few days, while the indexing is running, argo will show incomplete info in the rights facet, since the new field will not have been populated for all objects.  but reindexing an individual object manually should allow that object to properly contribute to rights facet counts.

ndushay:
Tests failures are due to a temporarily unavailable schema, http://www.loc.gov/mods/xml.xsd 

Merging.

PR #770, opened 2016-07-15
dump_fedora_pids.rb: update top level documentation comments

jmartin-sul:
i tried to retrieve the full pid list on dor-test yesterday (in the course of discussing indexing w/ @eefahy).  i ran into permissions issues, and eventually figured out that file writing privileges had to be granted to the fedora mysql user (which i'd previously discovered when first writing the script, but had failed to include in the explanation).

added comments that would've been helpful in sorting out the problem yesterday.


jmartin-sul:
@tingulfsen?  @atz?  should hopefully be uncontroversial.

PR #771, opened 2016-07-18
cleanup items helper spec

jmartin-sul:
we went in to fix a failing test, but realized that the failure is likely to be due to loc.gov being temporarily unreachable.  but we saw some easy cleanup while we were in the code, so we went ahead an implemented @ndushay's suggestions.

branch name is a misnomer now, since this was just a refactoring, and no test behavior was changed.


ndushay:
tests failing now due to http://www.loc.gov/mods/xml.xsd not being available at the present time.

Issue #774, opened 2016-07-21
items_helper_spec could use cleanup


jmartin-sul:
the test XML is both declared and laid out in a way that's messy, and it's unclear if all of it is necessary.  https://github.com/sul-dlss/argo/commit/717e6ffe8525550c9931ed6a1b57e4642960658d made an attempt to clean it up, but inadvertently broke the test (masked at the time of the commit by the failure of loc.gov, from which that test depends on a response).

PR #775, opened 2016-07-21
items_helper_spec.rb: revert 717e6ff

jmartin-sul:
since it appears to break the test (masked at the time by the failure of an external service on which the test depends).

i filed #774 to clean up that code, and we can prioritize that as usual.


jmartin-sul:
sorry for putting the mess back, @houndci-bot, i know you're OCD about that sort of thing.

PR #776, opened 2016-07-21
items_helper_spec: fix failing schema validation spec

### LINKED ISSUE #774: items_helper_spec could use cleanup

ndushay:
@jmartin-sul   This is a better fix for the problem than #775

Closes #774


ndushay:
p.s.  I don't care about hound violations

PR #784, opened 2016-08-15
Update dependencies

cbeer:


Gemfile
```diff
@@ -54,7 +54,7 @@ gem 'pry-byebug' # Adds step-by-step debugging and stack navigation capabilities
 gem 'pry-rails' # use pry as the rails console shell instead of IRB
 
 group :test, :development do
-  gem 'capybara' # used by wfs_rails in development
+  gem 'capybara', '~> 2.6.0' # used by wfs_rails in development
```
jmartin-sul: a comment explaining the reason for pinning to 2.6 would be nice

jmartin-sul:
i'm surprised by the capybara errors, though they appeared to be consistent when i re-ran the tests.  will have to investigate further.

PR #811, opened 2016-09-21
Adds Release object bulk action

### LINKED ISSUE #806: Bulk release
Requires SDR admin role or APO manager role
I believe the use case is for white-listing or black-listing items only and should not apply to collection druids.  But since the point is the ability to drop in a druid list, have the processing skip (with a skipped status or message) any druid that is not object type "item".
The Item dialog above should suffice minus the line about current release state and with these choices:
( * ) Release this item regardless of release status of any collection it is in
(  ) Do not release this item regardless of release status of any collection it is in

mejackreed:
Closes #806

### DIFF
--- a/app/jobs/release_object_job.rb
+++ b/app/jobs/release_object_job.rb
@@ -25,8 +25,11 @@ class ReleaseObjectJob < GenericJob
       update_druid_count
 
       pids.each do |current_druid|
-        next unless can_manage?(current_druid)
         log.puts("#{Time.current} Beginning ReleaseObjectJob for #{current_druid}")
+        unless can_manage?(current_druid)
+          log.puts("#{Time.current} Not authorized for #{current_druid}")
+          next
+        end
         log.puts("#{Time.current} Adding release tag for #{manage_release['to']}")
         begin
           response = post_to_release(current_druid, params_to_body(manage_release))
--- a/app/jobs/release_object_job.rb
+++ b/app/jobs/release_object_job.rb
@@ -1,0 +1,114 @@
+##
+# Job to add release and then release a Dor Object
+class ReleaseObjectJob < GenericJob
+  queue_as :release_object
+
+  attr_reader :manage_release, :pids, :webauth
+  ##
+  # This is a shameless green approach to a job that calls release from dor
+  # services app and then kicks off release WF.
+  # @param [Integer] bulk_action_id GlobalID for a BulkAction object
+  # @param [Hash] params additional parameters that an Argo job may need
+  # @option params [Array] :pids required list of pids
+  # @option params [Hash] :manage_release required Hash of release options
+  # @option manage_release [String] :to required release to target
+  # @option manage_release [String] :who required username of releaser
+  # @option manage_release [String] :what required type of release (self, collection)
+  # @option manage_release [String] :tag required (true, false)
+  # @option webauth [Hash] required for permissions check
+  def perform(bulk_action_id, params)
+    @manage_release = params[:manage_release]
+    @webauth = OpenStruct.new params[:webauth]
+    @pids = params[:pids]
+    File.open(bulk_action.log_name, 'w') do |log|
+      log.puts("#{Time.current} Starting ReleaseObjectJob for BulkAction #{bulk_action_id}")
+      update_druid_count
+
+      pids.each do |current_druid|
+        next unless can_manage?(current_druid)
+        log.puts("#{Time.current} Beginning ReleaseObjectJob for #{current_druid}")
+        log.puts("#{Time.current} Adding release tag for #{manage_release['to']}")
+        begin
+          response = post_to_release(current_druid, params_to_body(manage_release))
+          if response.status == 201
+            log.puts("#{Time.current} Release tag added successfully")
+          else
+            log.puts("#{Time.current} Release tag failed POST #{response.env.url}, status: #{response.status}")
+            bulk_action.increment(:druid_count_fail).save
+            next
+          end
+        rescue Faraday::TimeoutError, Faraday::ConnectionFailed => e
+          log.puts("#{Time.current} Release tag failed POST #{e.class} #{e.message}")
+          bulk_action.increment(:druid_count_fail).save
+          next
+        end
+        log.puts("#{Time.current} Trying to start release workflow")
+        begin
+          response = post_to_release_wf(current_druid)
+          if response.status == 201
+            log.puts("#{Time.current} Workflow creation successful")
+            bulk_action.increment(:druid_count_success).save
+          else
+            log.puts("#{Time.current} Workflow creation failed POST #{response.env.url}, status: #{response.status}")
+            bulk_action.increment(:druid_count_fail).save
+          end
+        rescue Faraday::TimeoutError, Faraday::ConnectionFailed => e
+          log.puts("#{Time.current} Workflow creation failed POST #{e.class} #{e.message}")
+          bulk_action.increment(:druid_count_fail).save
+        end
+      end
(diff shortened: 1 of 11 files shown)


spec/jobs/release_object_job_spec.rb
```diff
@@ -0,0 +1,182 @@
+require 'spec_helper'
+
+describe ReleaseObjectJob do
+  describe '#perform' do
+    let(:bulk_action_no_proccess_callback) do
```
tingulfsen: I guess you meant ...process...

spec/jobs/release_object_job_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+      it 'logs information to the logfile' do
+        # Stub out the file, and send it to a string buffer instead
+        buffer = StringIO.new
+        expect(File).to receive(:open).with(bulk_action_no_proccess_callback.log_name, 'w').and_yield(buffer)
+        subject.perform(bulk_action_no_proccess_callback.id, params)
+        expect(buffer.string).to include 'Starting ReleaseObjectJob for BulkAction'
+        pids.each do |pid|
+          expect(buffer.string).to include "Beginning ReleaseObjectJob for #{pid}"
+        end
+        expect(buffer.string).to include 'Adding release tag for SEARCHWORKS'
```
tingulfsen: Is it worth to change the test slightly to verify that the buffer contains one of these messages per pid?

app/jobs/release_object_job.rb
```diff
@@ -0,0 +1,112 @@
...
+  def perform(bulk_action_id, params)
+    @manage_release = params[:manage_release]
+    @webauth = OpenStruct.new params[:webauth]
+    @pids = params[:pids]
+    File.open(bulk_action.log_name, 'w') do |log|
+      log.puts("#{Time.current} Starting ReleaseObjectJob for BulkAction #{bulk_action_id}")
+      update_druid_count
+
+      pids.each do |current_druid|
+        next unless can_manage?(current_druid)
```
tingulfsen: What about adding a log message here, so we don't "silently" pass on to the next one in case the user by mistake tried to operate on a druid they can't manage?

mejackreed: I thought about that, but didn't want to leak druid id's to a user if they don't have permissions to manage it. An information leakage reason only.

Thoughts?

spec/controllers/bulk_actions_controller_spec.rb
```diff
@@ -52,28 +55,36 @@
...
+      it 'has a 302 status code' do
+        post :create, bulk_action: {action_type: 'GenericJob', pids: ''}
+        expect(response.status).to eq 302
+      end
+      it 'when not saveable render new' do
+        fake_bulk_action = double('fake', save: false, 'user=' => nil)
+        expect(BulkAction).to receive(:new).and_return fake_bulk_action
+        post :create, bulk_action: {action_type: 'GenericJob'}
+        expect(response).to render_template('new')
+      end
```
mejackreed: An action isn't savable if required params aren't passed through.  That message will be rendered from the form: https://github.com/sul-dlss/argo/blob/master/app/views/bulk_actions/_form.html.erb#L4

This is a previously created spec where an indent was added. Happy to address in a separate PR if an issue is created.

spec/jobs/release_object_job_spec.rb
mejackreed: Totally! Thanks for catching that.. pushed up the change.

spec/jobs/release_object_job_spec.rb
mejackreed: I think it does? See LN47. Is that what you mean?

app/jobs/release_object_job.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+      log.puts("#{Time.current} Finished ReleaseObjectJob for BulkAction #{bulk_action_id}")
+    end
+  end
+
+  def can_manage?(pid)
+    Ability.new(
+      User.find_or_create_by_webauth(webauth)
+    ).can?(:manage_item, Dor.find(pid))
+  end
```
mejackreed: Thanks for that suggestion. Added and pushed up.

app/jobs/release_object_job.rb
tingulfsen: I just had a chat with @LynnMcRae and @blalbrit about this, and they said that a log message would be useful. In fact, they'd like the log to account for every druid processed, just so you know. I think a really simple log message would solve this nicely.

spec/controllers/bulk_actions_controller_spec.rb
tingulfsen: 👍

spec/jobs/release_object_job_spec.rb
tingulfsen: OK, not a big deal, I'd rather keep the tests short and to the point.

app/jobs/release_object_job.rb
mejackreed: Added in c7615820bad0106405cdade45f422d7ff7da498c

spec/jobs/release_object_job_spec.rb
```diff
@@ -126,15 +126,15 @@
         subject.perform(bulk_action_no_process_callback.id, params)
         expect(bulk_action_no_process_callback.druid_count_total).to eq pids.length
       end
-      # no information leakage to an unauthorized user
       it 'logs no druid info to logfile' do
```
tingulfsen: Does it make more sense for this to say "logs druid info to logfile" with the most recent change?

mejackreed: yep.. good catch.. pushed with the change.

tingulfsen:
I am now 👍  on this.

app/jobs/release_object_job.rb
```diff
@@ -0,0 +1,117 @@
...
+            log.puts("#{Time.current} Workflow creation successful")
+            bulk_action.increment(:druid_count_success).save
+          else
+            log.puts("#{Time.current} Workflow creation failed POST #{response.env.url}, status: #{response.status}")
+            bulk_action.increment(:druid_count_fail).save
+          end
+        rescue Faraday::TimeoutError, Faraday::ConnectionFailed => e
+          log.puts("#{Time.current} Workflow creation failed POST #{e.class} #{e.message}")
+          bulk_action.increment(:druid_count_fail).save
+        end
```
mejackreed: So I have in a branch breaking this out even further (release management client class) but wanted to offer a simpler approach in this PR. Just wanted to reduce overall complexity here and leave this less complicated for now. Thoughts?

app/views/bulk_actions/_form.html.erb
```diff
@@ -11,12 +11,61 @@
...
+                  Do not release this object*
+                </label>
+              </div>
+            </div>
+            <div class='form-group'>
+              <label for='bulk_action_manage_release_to'>to</label>
+              <%= manage_release_fields.select :to, Constants::RELEASE_TARGETS, {}, { class: 'form-control' } %>
+            </div>
+            <%= manage_release_fields.hidden_field('what', value: 'self') %>
+            <%= manage_release_fields.hidden_field('who', value: current_user.sunetid) %>
```
mejackreed: It won't always be `self` just in this context of the form. Multiple forms will be used to offer this type of functionality in different ways and I wanted to leave the input as clean as possible (using form to generate values).

mejackreed: ^^ rather than munging in controller. We could do this differently since the action is already tied to the user, but having the form element seemed to make since if we needed to change down the road.

app/jobs/release_object_job.rb
jmartin-sul: sounds good to me...

PR #821, opened 2016-09-28
Update release history

### LINKED ISSUE #809: Release tag history doesn't show all releases
The history uses DorServices `release_tags` method, which interestingly enough doesn't return all release tags.

mejackreed:
Closes #809 

Some of this could also be used in simplifying #811

### DIFF
--- a/spec/models/release_tag_spec.rb
+++ b/spec/models/release_tag_spec.rb
@@ -1,0 +1,39 @@
+require 'spec_helper'
+
+describe ReleaseTag do
+  describe '#initialize' do
+    subject do
+      described_class.new(
+        to: 'SearchWorks',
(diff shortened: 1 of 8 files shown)


spec/models/release_tag_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        to: 'SearchWorks',
+        what: 'self',
+        when_time: '2016-09-13T20:00:00.000Z',
+        who: 'esnowden',
+        release: 'true'
+      )
+    end
+    it 'casts a Boolean string to a Boolean' do
+      expect(subject.release).to eq true
+    end
```
mejackreed: good suggestion.. added.

PR #824, opened 2016-09-30
Add rack mini profiler for diagnosing page performance issues.

mejackreed:
I think this will be useful to Argo devs to show the consequences of many of the dor-services calls and how async can help.

### 20 blocking requests on a show page
screen shot 2016-09-30 at 3 31 23 pm


jkeck:
This may make the dev. env a bit slow, no?  Perhaps that's not a huge issue, but maybe there would be a way to initialize it when starting rails in development?

Just a thought, I really like the idea of being able to easily get this info, I've typically just made the modifications to add the dependency locally and then removed it after my profiling.

mejackreed:
Do you have suggestions? I've added this in locally about 10 different times so I figure there should be a way to get it in for devs. Really highlights argo performance issues.

config/initializers/rack_profiler.rb
```diff
@@ -0,0 +1,6 @@
+if Rails.env == 'development'
```
jkeck: What about something like

```ruby
if Rails.env == 'development' && ENV['PROFILE'] == 'true'
```

And then if you want to do your profiling your can start rails w/  `PROFILE=true rails s`.

Not sure if rails-config settings can be made available in an initializer (I think I have run into complications with that before), but that might be another option (where a development.local.yml could be used).

It's just a thought, if you have success with a particular pattern I would love to add it to some other projects.

mejackreed: 👍 I like it thanks. I'll try and get that  change up

cbeer: I'd check if the profiler makes that big a difference; I'd guess compared to everything else argo does, probably not.

jmartin-sul: there's a `PROFILER` section in the settings.yml.  if we were to add a flag, i'd propose adding one there, maybe `ENABLE_RACK_PROFILING`.  default to `false`, enable as needed in local settings yml files, maybe enable in the laptop and dev (and stage?) shared configs.

i'd also be totally fine with just doing it for `Rails.env == 'development'`, but the flexibility of the flag has proved beneficial for the new relic profiler as well as the ruby prof stuff that can be enabled on bulk indexing.  it's nice to be able to turn the profiling on and off at a whim for whatever instance you're debugging.

jmartin-sul: also: profiling did noticeably slow indexing throughput, FWIW (both RubyProf and New Relic).  i don't have the numbers handy, i think there are actually some recorded in some of the github issues, but i think we saw up to a 5-10x slowdown on indexing workers when profiling was enabled.  we typically had to run it for limited windows to collect enough info to debug the memory issues we were running into.  if we left profiling on full time, indexing would've taken something like a month instead of 3 days.

i have neither anecdotal nor scientific evidence for noticeable regular application slow down when new relic was enabled (which is the only profiling we've left on in the app for any extended period).  argo's slow enough anyway that i didn't really notice slowdown from new relic.  and we weren't collecting performance data before new relic, so i have nothing to judge its numbers against.

jmartin-sul: ah, sorry, i totally missed @jkeck's point about setting availability at config/initializer time.  totally fine w/ env var if settings are problematic on that front.

mejackreed:
Thanks for all of the comments, updated the PR to enable localized configurability per comments. Adding as default on in development (as I hope this helps future Argo developers).

PR #826, opened 2016-10-07
Blue button release

### LINKED ISSUE #805: Add a blue button for "Manage release"
Requires SDR admin role or APO manager role
If a collection:
```
Manage release to discovery applications for collection sj775xm6965
       Douglas Menuez Photography, circa 1986-2003
Currently released to Searchworks[, Target2 ...]
-------
Release to : [Searchworks] <-- pulldown list for available targets
( * ) Release this collection and all it's members *
(  ) Release only this collection description but not any of its members (members previously released will be withdrawn) <-- needed?
(  ) Withdraw this collection from release * <-- only offered if there is a current "true" for the selected target
* Any members that have their own overriding release information will not be affected
```
If an item:
```
Manage release to discovery applications for item bb423jr2180
       The complaint, and the consolation: or, Night thoughts
Currently released to Searchworks[,target 2...] as part of collection "Douglas Menuez Photography, circa 1986-2003"
Release to : [Searchworks] <-- pulldown list for available targets
( * ) Release this item
(  ) Withdraw this item from release  <-- only offered if there is a current "true" for the selected target either locally or in its collection
```
Consider the release of an item that is already released by a collection a no-op?  Or allow if to create a local release tag in case it's an effort to white-list items before withdrawing the collection? To start, assume the creation of a local tag regardless of collection status.
Confirmation information, an opportunity to say things might take time for large collections as things work through workflow.

```
New release settings saved for Searchworks 
Please leave time for affected items to be republished in PURL and processed by the Searchworks indexer. 
```

mejackreed:
Closes #805 

Adds show page blue button functionality for items and collections.

screen shot 2016-10-07 at 9 07 55 am

## Item release
screen shot 2016-10-07 at 9 09 35 am

## Collection release
screen shot 2016-10-07 at 9 10 00 am


tingulfsen:
👍 

Over to you @jmartin-sul

jmartin-sul:
👍 from me too, merging...

PR #844, opened 2016-10-17
Update dependencies

cbeer:
Argo also needed changes to fixture object loading; whereas in AF 6.0, the `ActiveFedora::Base` class provided access to all datastreams, in later versions these needed to be made explicit. The easy thing to do was to simply cast objects to their expected classes.


spec/controllers/catalog_controller_spec.rb
```diff
@@ -8,7 +8,7 @@
     @item = instantiate_fixture(@druid, Dor::Item)
     @user = User.find_or_create_by_webauth double('WebAuth', :login => 'sunetid', :logged_in? => true, :attributes => {'DISPLAYNAME' => 'Example User'}, :privgroup => '')
     allow(Dor).to receive(:find).with("druid:#{@druid}").and_return(@item)
-    allow(Dor::Item).to receive(:find).with("druid:#{@druid}").and_return(@item)
+    allow(Dor).to receive(:find).with("druid:#{@druid}").and_return(@item)
```
jmartin-sul: since it's an allow, and the two lines are now identical, could we just remove this?

lib/tasks/argo_testing.rake
```diff
@@ -39,6 +39,7 @@ end
 
 if ['test', 'development'].include? Rails.env
   require 'jettywrapper'
+  Jettywrapper.hydra_jetty_version = 'v7.3.0'
```
jmartin-sul: could you add a comment explaining the version specification?

PR #857, opened 2016-10-27
Update blacklight to latest version 6.7.2

mejackreed:


cbeer:
👏

Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'equivalent-xml', '>= 0.6.0'   # For ignoring_attr_values() with arguments
 gem 'faraday'
 gem 'eye'  # NOTE: if eye is upgraded, see the note in the 'bin/eye' script about checking to see whether that script needs upgrading (which won't happen automatically).
 gem 'honeybadger', '~> 2.0'
+gem 'jbuilder', '~> 2.0'
```
atz: Just to inform my blacklight migrations in other apps, what is the need for `jbuilder` here?

jmartin-sul:
@LynnMcRae and @blalbrit say they noticed no difference with this on stage, which i think is a 👍 from them.  merging so this doesn't hang out forever and go stale...

LynnMcRae:
Confirming I see no visible difference or unexpected interactions

PR #867, opened 2016-11-08
Update dependencies

cbeer:


drh-stanford:
This PR failed the dependency checks -- some gems don't have licenses. Do we consider that a broken build?

PR #872, opened 2016-11-08
okcomputer app monitoring at /status

cbeer:


cbeer:
This is @ndushay's code in the first place, with a small tweak to get the rubydora check working and a shim route to keep `/is_it_working`... working.

PR #873, opened 2016-11-08
Okcomputer tweaks

cbeer:


config/initializers/okcomputer.rb
```diff
@@ -61,9 +62,7 @@ def profile
 OkComputer::Registry.register 'suri_url', OkComputer::HttpCheck.new(Settings.SURI.URL)
 
 # Stacks
-OkComputer::Registry.register 'stacks_doc_cache_storage_root', OkComputer::DirectoryCheck.new(Settings.STACKS.DOCUMENT_CACHE_STORAGE_ROOT)
```
ndushay: so are we saying that these aren't on -dev environment, so they're not worth checking?

cbeer: They aren't on -dev, -stage, or -prod.

ndushay: I wonder if it's for running locally?

ndushay: If these settings are not used, make 'em disappear altogether??

cbeer: Ticketed: https://github.com/sul-dlss/argo/issues/874

jmartin-sul: IIRC (and i may be misremembering, it's been a while):  file content retrieval is now proxied through dor-services-app, and argo no longer does the SFTP thing.

PR #875, opened 2016-11-09
Update to Rails 5

cbeer:


cbeer:
The hound comments are on Rails-generated config files. I'm not inclined to change them, since we seem to only rely on hound-ci, not rubocop, in this project.

spec/controllers/bulk_actions_controller_spec.rb
```diff
@@ -103,7 +103,7 @@
     end
   end
   describe 'GET file' do
-    it 'sends through a BulkActions file' do
+    xit 'sends through a BulkActions file' do
```
mejackreed: Maybe something like:

```ruby
it 'sends through a BulkActions file' do
      b_action = create(:bulk_action, user: current_user)
      File.new(b_action.file('test.log'), 'w')
      get :file, id: b_action.id, filename: 'test.log', mime_type: 'text/plain'
      expect(response.status).to eq 200
    end
```

cbeer: Ok, I'll give that a try. Ought to work for now, at least.

cbeer:
I've put 3a5ebbae0b15b9dcc21a0832b221ee00fe06259a in a separate commit. In Rails 5, directly feeding `params` into a `*_url` or `*_path` helper raises an exception (for generally very good security reasons). What I've done is the equivalent to what we were doing in Rails 4, and we should feel really bad about it, but I don't think the analysis and remediation should prevent us from updating either.

PR #876, opened 2016-11-09
Add index queue mechanism

cbeer:
This reverts commit 7db26871c2a45159268d62c5a1cd157ad571fbce.

screen shot 2016-11-09 at 12 59 40 pm

Fixes https://github.com/sul-dlss/dor_indexing_app/issues/59?


drh-stanford:
NOTE that the shared configs PRs are also merged, so we need to update them on the next deploy.

PR #883, opened 2016-11-14
Prefer MODS-derived fields instead of public-dc derived fields

cbeer:


cbeer:
I reviewed these changes with @blalbrit, who gave it a 👍.

PR #891, opened 2016-11-17
Use proper MODS for descMetadata for workflow objects

drh-stanford:
This fixes fixture data that causes `to_solr` being called on Workflow objects in dor-services 5.17 to fail.

```
Undefined namespace prefix: /oai_dc:dc/*[count(text()) = 0]
```

This PR also updates to dor-services 5.17.x where these fixtures break.

TODO: 
- [x] Fix this line too https://github.com/sul-dlss/argo/blob/master/lib/tasks/argo.rake#L12

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,7 +82,7 @@ gem 'about_page'
 gem 'active-fedora', '~> 8.2'
 gem 'blacklight', '~> 6.0'
 gem 'blacklight-hierarchy'
-gem 'dor-services', '>= 5.14.0', '< 6'
+gem 'dor-services', '>= 5.17.0', '< 6'
 gem 'moab-versioning'
 gem 'mods_display'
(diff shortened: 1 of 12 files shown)


LynnMcRae:
All production workflow objects now have proper MODS descMetadata

drh-stanford:
@cbeer I got it down to 3 failures. The first is a Blacklight search for all the collections associated with an APO, and the other 2 are Blacklight search result matches. 

The latter works fine in my browser, but not on Travis. 

The former doesn't work either place -- I'm not sure why `collection_list` is going to both Solr and DOR for the titles -- but it gets a 404 when trying to go to Solr. My blacklight.yml looks ok but it's likely a configuration issue:

  http://localhost:3000/registration/collection_list?apo_id=druid:fg464dn8891

https://travis-ci.org/sul-dlss/argo/builds/176884919

PR #897, opened 2016-11-18
/dor/reindex/:pid route will now index remotely, and delete_from_index route removed

drh-stanford:
This PR is connected to https://github.com/sul-dlss/dor_indexing_app/issues/105. It changes the behavior of the `/dor/reindex/:pid` route to use the dor-indexing-app service rather than calling `to_solr` and `RSolr` itself.

This PR also removes the `/dor/delete_from_index/:pid` route as it is no longer needed (use dor-indexing-app for that).

**NOTE** reindexing via the UI on laptop deployments will be broken unless you setup and configure a dor-indexing-app also on your laptop. The default configuration goes to the -dev instance, like the dor-services-app laptop configuration.


LynnMcRae:
A question about the delete_from_index route.  The only case I can think of where Argo needs  to nuke a solr doc is for Purge.  Just want to make sure Purge is not depending on this route.

cbeer:
Purge is entirely separate logic (and encapsulates both removing it from DOR and the index).

drh-stanford:
tested in dev -- ready for review.

drh-stanford:
changed the error status code from 404 to 500

PR #899, opened 2016-11-18
Use standard Rails and Blacklight routing approach

cbeer:


config/routes.rb
```diff
@@ -9,31 +9,27 @@
...
+    concerns :searchable
   end
 
-  # Catalog stuff.
-  match 'view/opensearch', :to => 'catalog#opensearch', :via => [:get, :post]
-  match 'view/facet/:id',  :to => 'catalog#facet', :via => [:get, :post]
-  resources :catalog, :path => '/view', :only => [:index, :show, :update]
-  match 'view/:id/dc',                :to => 'catalog#dc', :via => [:get, :post], :as => 'dc_aspect_view_catalog'
-  match 'view/:id/ds/:dsid',          :to => 'catalog#ds', :via => [:get, :post], :as => 'ds_aspect_view_catalog'
-  match 'view/:id/datastreams/:dsid', :to => 'catalog#datastream_view', :via => [:get, :post], :as => 'datastream_view_catalog'
```
drh-stanford: did this one get removed?

cbeer: Yep, unused as far as I can tell.

cbeer:
Sounds reasonable. In 47e763f, I've added specs for the ones I've changed in this commit.

cbeer:
I was confused too. My understanding is @LynnMcRae prefers `/view`, and `/catalog` is just the Blacklight default that no one changed before.

LynnMcRae:
It goes back to when we first adopted vuFind then Blacklight, both I think use /catalog for their primary display route. The searchworks design team, Jennifer, Tom, Naomi, myself, a bunch of other, wanted to downplay the concept of a discovery system that simply equated to a library holdings catalog. Blacklight could have /catalog but searchworks got /view and still uses it rigorously in the UI today though /catalog also works.

We built Argo in the same spirit (from the same codebase basically) since it also is not a "catalog", and /view is just a better choice for the UI.  Argo's behavior has eroded over time with new blacklight injections -- you now see /catalog in the location bar unless you explicitly type /view. I would love for that to be fixed but no one is losing sleep over it. I don't think /view can be dropped at this point since that's been a widely shared form of URL.

cbeer:
> Blacklight could have /catalog but searchworks got /view and still uses it rigorously in the UI today though /catalog also works.

The goal of that commit in this PR was to try to actually get argo to use `/view/*` in routes for records. We had routes that were attempting to do that already, but weren't quite right.

config/routes.rb
```diff
@@ -44,8 +40,8 @@
   get 'catalog/:id/:time/bulk_jobs_csv', :to => 'catalog#bulk_jobs_csv',     :as => 'bulk_jobs_csv'
   post 'catalog/:id/upload',             :to => 'catalog#upload',            :as => 'upload'
   delete 'catalog/:id/bulk_jobs_delete', :to => 'catalog#bulk_jobs_delete',  :as => 'bulk_jobs_delete'
-  match 'catalog',      :via => [:get, :post], :to => redirect { |params, req| req.fullpath.sub(%r{^/catalog}, '/view') }
-  match 'catalog/*all', :via => [:get, :post], :to => redirect { |params, req| req.fullpath.sub(%r{^/catalog}, '/view') }
+  match 'catalog',      :via => [:get, :post], :to => redirect { |params, req| req.fullpath.sub(%r{^/catalog}, '/view') }, as: 'search_catalog_redirect'
+  match 'catalog/*all', :via => [:get, :post], :to => redirect { |params, req| req.fullpath.sub(%r{^/catalog}, '/view') }, as: 'catalog_redirect'
```
drh-stanford: I'm confused by this configuration here for the redirects. Are we trying to provide backward compatibility here for old URLs, or is this functional somehow?

cbeer: Yes, that's the idea, as far as I can tell. It seems like that's how it behaves, at least.

drh-stanford:
Ok, I'm 👍 for this PR. Let's be sure to document the route changes in the release notes.

PR #900, opened 2016-11-28
deprecate some of the indexing documentation in favor of dor_indexing_app

drh-stanford:
This PR deprecates some of the indexing documentation in favor of the Wiki on dor_indexing_app. Argo makes these reindexing calls to dor_indexing_app now. The PidGatherer functionality, however, has not be migrated. It's only needed for a from-scratch reindex, but the documentation/code is valuable to retain here.

### DIFF
--- a/INDEXING.md
+++ b/INDEXING.md
@@ -4,17 +4,18 @@
 
 Argo uses Solr as an index for faceting and searching data in the corresponding Fedora repository.
 
-Objects can be reindexed by Argo as a result of a couple different things:
-* An object is edited in Fedora (by Argo or another application), and a message is sent to an Argo endpoint (`/dor/reindx/PID`), triggering a reindex of the object.
-* An object is part of a batch of objects queued for bulk indexing.
+Objects can be reindexed as a result of a couple different things:
+* An object is edited in Fedora (by Argo or another application), and a message is sent to an endpoint in the dor_indexing_app service (`/dor/reindex/PID`), triggering a reindex of the object.
+* An object is part of a batch of objects queued for bulk indexing within Argo (deprecated)
 
-In both cases, Argo uses dor-services to:
+To support indexing, dor-services will:
 * load the object from Fedora.
 * create a Solr document representing the object (using its `to_solr` method).
 * save the document to the Solr instance specified in the configuration, possibly committing immediately.
 
-
 ## Argo::Indexer
+
+This is deprecated for `Dor::IndexingService` in `dor-services`.
 
 ### Reusable methods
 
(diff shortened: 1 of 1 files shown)


INDEXING.md
```diff
@@ -4,48 +4,29 @@
...
-    render :status => 404, :plain => 'Object does not exist in Fedora.'
-    return
-  end
-```
-
-
 ## Bulk reindexing
 
+This method is deprecated. See the [`dor_indexing_app`](https://github.com/sul-dlss/dor_indexing_app) Wiki for the new methods. The `Argo::PidGatherer`, however, is not migrated so if you need to gather pids directly from Fedora, refer to that code.
+
```
jmartin-sul: is there a robust/stable way to do a full reindex of DOR without this mechanism?  the web UI that uses the old javascript bulk operations approach for bulk reindexing is pretty flakey, since the browser is essentially managing the queuing.  should we assume that we can go back to the old full-reindex approach of throwing tons of PID lists at karaf?  that was also flakey in the past, but maybe that's improved now that dor-indexing-app (and the underlying dor-services code) has been improved?  but karaf seemed to get wedged in weird ways whenever argo got wedged (dor-indexing-app taking on the work that argo used to do for karaf's reindexing messages).

totally fine to be moving away from this approach, just hesitant to call it deprecated if there's no immediate replacement, since it's what i'd still be inclined to use for a full reindex if told to do one today.

drh-stanford: We were successful doing a full reindex in ~3 days using sulmq only. It was not a from-scratch index (where we needed to do ordering by object type), just a reindex of all the pids in the Solr index already. We weren't asked to do a from-scratch index, but I think your PidGatherer approach is still valid for that.

I'm not sure about the flakeyness with the message broker approach. dor_indexing_app is much faster and stable now, so it may be more reliable.

The Argo bulk update for reindexing is still supported (the javascript stuff). See https://github.com/sul-dlss/dor_indexing_app/issues/105 for that work.

jmartin-sul: alright, so that's a really big performance improvement, and might totally alleviate the issues with doing a full reindex using sulmq.  i think the last time i used sulmq to do a full reindex was in late 2014, and it took something like 3.5 days to do 400k objects.  i'd be happy to try the new improved sulmq and dor_indexing_app setup instead of the argo reindexing workers if/when the time comes for another full reindex (or a reindex of a significant chunk of the objects in DOR).

and actually, while we'd love to use the `PidGatherer` mechanism, that doesn't provide reliable results, because, for whatever reason, the Fedora resource index (used via `risearch`) doesn't return complete results.  see #380.  as such, we still have to dump a list of pids from fedora's mysql db if we want to index from scratch and we don't have/trust a current solr index.  the script for doing that isn't called out anywhere in the top level documentation, but it lives here, and has decent usage comments at the top of it (lives in the argo codebase git repo, but is run on the DOR machine):
https://github.com/sul-dlss/argo/blob/master/bin/dump_fedora_pids.rb

anyway, discussed for a bit yesterday afternoon w/ @cbeer and @eefahy.  as i understood it, we all agreed that
* we'd like `PidGatherer`/`risearch` to work, but probably won't have the time to actually investigate why the resource index doesn't return complete results.  if we can't get it to work, maybe we should shut it off?
* full index rebuilds (or rebuilds of a significant chunk of the index) tend to be exceptional situations, and the most helpful thing is to have complete documentation of what works and what doesn't, and how things fit together, since it's hard to have reliable readymade solutions for unexpected circumstances.

so, i propose something like the following text:

> This approach is deprecated, but we haven't yet ironed out a process for doing a full reindex from scratch.  One alternate option to this approach would be to queue the IDs of the objects that need to be reindexed using sulmq, for processing by dor_indexing_app.  See the [`dor_indexing_app`](https://github.com/sul-dlss/dor_indexing_app) Wiki for the new methods. The `Argo::PidGatherer`, is not migrated, but we've had issues with our Fedora instance's `risearch` results (see #380).  As such, the following script may be useful in obtaining a full PID list from DOR, in the event that there's no trustworthy Solr index from which to get a list of IDs:
> https://github.com/sul-dlss/argo/blob/master/bin/dump_fedora_pids.rb

@eefahy had suggested improving the `dump_fedora_pids.rb` approach by having it automatically SCP the results to the box where the PIDs will be fed to sulmq.

anyway, that's my $0.02.  no strong objection to merging as-is, if you'd still prefer to do that.

drh-stanford:
thanks for the great feedback. i've changed this paragraph and pushed the branch.

jmartin-sul:
@drh-stanford: no problem, glad it was helpful!

PR #906, opened 2016-12-06
Actually generate the public descriptive metadata instead of doing so…

cbeer:
…me of the same things


atz:
It seemed out of place, but what was the intent of the `@object.add_collection_reference @object.descMetadata.ng_xml` line there?  Some _ad hoc_ insurance for a confused consumer?

cbeer:
That line adds a collection reference to the MODS. I believe the intent was to simply generate the public descriptive metadata for the preview, but instead we chose to do something similar but not quite that.

PR #908, opened 2016-12-06
Use sw_display_title_tesim as the document title

cbeer:


LynnMcRae:
Why is it sw_display_title_tesim?  why the "sw" that is? We want the DOR object to have a display title. It should reflects a canonical way of assembling a title from MODS <title> parts, or receive any title substitute (like object label) in the absence of descMetadata. title_ssi/title_tesim seems to fit the bill (it exists even without descMetadata). The sw_ set of fields are intended to support Searchworks view of data where differences or distinctions make them worth reflecting in facets or reports. Is the sw_display_title anything more than what the MODS Display gem produces for a title? Does it differ from title_ssi, like maybe single valued? If I wanted a DOR title or titles I would not look to a sw_ field. 

If we are eliminating the dc_ versions of fields that come from the DC datastream, they should just go away. Do any come from the Full Dublin Core derived from MODS?

cbeer:
`sw_display_title_tesim` is the display title derived from MODS using the rules in stanford-mods. `title_ssi` and `title_tesim` are derived from the `DC` datastream. From conversation with you in the past (and dealing with a variety of other tickets), I was under the impression we wanted to avoid using the values from `DC`. Is that not the case?

LynnMcRae:
Yes, we don't care about DC (except marginally in weaning people off anything to do with it). But we do want a field that reflects the object's title that has a name that suggests that. title_ssi does and  sw_display_title doesn't.  It also sounds like sw_display_title only exists in the presence of MODS, which doesn't make it a reliable title for all objects per subject line (and makes it a mods_title_). A separate MODS only title might be useful in the way the separate obj_label field is (it adds to the search vocabulary) but limited and mainly in its absence -- an Empty Fields entry that says "No MODS" or "No descMetadata datastream" would be more direct.

Were you dropping title_ssi because of it's DC origins?  then what is dc_title_ssi? I thought only the dc_ field were disappearing.  I'd rather repurpose title_ssi to be the the DOR title. Or add a dor_title or dor_object_title if necessary, but would that leave title_ssi out there as a confusing choice?

LynnMcRae:
we can chat more directly if this is getting too swirly. The DC cleanup is a good thing and eliminating redundant and overlapping choices of fields whose differences are obscure is great as well. Unfortunately I am out one more day without a car and would love to pin it down in front of a whiteboard if it can wait until tomorrow.

LynnMcRae:
There might be two simpler questions here.  The sw_ display_title comes from the use of Stanford MODS and its sw_display_title method, but that specific view of the title, if different at all from a more generic one, is not of interest as such in the DOR index. The MODS Display gem offers a generic title that makes more sense. I'd like to know how they differ if we never reconciled them, but ideally they would be the same. If there's a reason for a descMetadata/MODS title field tied to the presence of MODS, it's to know one exists and how it looks in DOR.

The second question is whether, in eliminating the dc_ flavored fields, we have a reliable always-present DOR object title
- one whose value is set from descMetadata, else DC/dc:title (still?), else identityMetadata/objectLabel, else Fedora Label
- an object title good for Collection or APO title lookups (what are we using today?)
- for labeling items in a search result, populating a title column in a report, etc.

Some people were using dc_title for this. I thought DOR Services had a method to return such a guaranteed title/label for indexing, but I can't find it (is it obj_title?).

I'm trying to answer some of these questions myself from the code but I seem to lack the savvy to navigate the labyrinth.

cbeer:
I'm not sure I'm following, or we're confusing several different issues.

One thing I hear is that you're concerned about the name of the Solr field. I agree that `title_*` is a much better name for this field, and we can certainly get there eventually, and switching from the `DC` datastream derived field name to this existing field derived from the `descMetadata` is a stepping stone that allows us to free up that field for other purposes like this.

Another thing I hear is you're concerned that there are objects without descriptive metadata, or with descriptive metadata that lacks a title. Several weeks ago, I thought we agreed that all objects should have at least minimal `descMetadata` and we remediated several objects for which this was not the case (particularly, a handful of workflow objects). I believe we also identified a number of objects (mostly from Revs?) that lacked MODS titles and agreed those should be remediated.

I also hear you describing a future state, where we try harder to extract an appropriate title field from a variety of sources. To my knowledge, no such logic exists now, and we're inconsistent in our titles currently.

Finally, note that the changes in this pull request are in a handful of administrative areas, including titles for APOs, agreements, and workflow objects, which appear to all have MODS titles (of which some are incomplete, but no worse than the DC-derived ones).

LynnMcRae:
It is true that we have a commitment to making sure well-formed objects have descMetadata with a MODS title (which would make collection and APO lookups simpler), but some of the concern is when this doesn't happen -- newly registered objects that get orphaned by an error for instance, or wayward logic that results in an empty MODS doc. There are also 10k+ ETD "part" objects that lack descMetadata and identityMetadata but do have a dc:title and Fedora label.  Argo uses title_ssim I think as the reliable title for display, so fascinating to know it is no different than dc_title (this explains some things).

Sorry that lifting this rock has surfaced so many wiggly worms. It has always been a confusing set of title fields that I suspect got added to through generations rather than solved. So we are excited that this will really clarify and clean things up. I could see just two fields perhaps ... a title based on presence of descMetadata/MODS bibliographic title, and perhaps obj_title based on best-available string, of interest primarily to Argo and admin reporting. If this got rid of dc titles and info title and main titles, things would be so clean. title_sort might have a place if it actually drops nonsort characters (and if we we were to actually enable a sort by title in Argo).

cbeer:
Again, note that this change will have no impact on ETD "part" objects.

blalbrit:
thumbsup from me and @LynnMcRae

atz:
👍

LynnMcRae:
👍  My summary of background, for review: sw_display_title_tesim has been identified as the best representation of the MODS title in the index as it stands, supplanting the redundant public_dc_title we intend to remove. But we agreed that sw_ as a prefix is misleading and should apply to SW specific transformations of MODS output, not generic MODS values. So we plan to introduce a more neutrally named field not tied to Searchworks, one that provides a title for the DOR object even if it lacks MODS or descMetadata for some reason.

PR #915, opened 2017-01-18
Use upstream logic for validating source ids

cbeer:
Upstream does:

```ruby
   parts = value.split(':').map(&:strip)
    # parts = value.split(/:/, 2).map(&:strip) # if we needed to allow colons in values or bunched colon separators
    raise ArgumentError, "Source ID must follow the format 'namespace:value', not '#{value}'" unless
      parts.length == 2 && parts[0].present? && parts[1].present?
```

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -412,22 +412,9 @@ class ItemsController < ApplicationController
     end
   end
 
-  # TODO: this should be a method in dor-services, invoked within set_source_id
-  def self.normalize_source_id(new_source_id)
-    src_id_arr = new_source_id.split(':').map(&:strip)
-    raise 'new source_id must be of the form "source:value"' if src_id_arr.length != 2
-    src_id_arr.join(':')
-  end
-
   def source_id
-    new_id = ItemsController.normalize_source_id(params[:new_id])
-    @object.source_id = new_id
-    # TODO: the content= and content_will_change! calls belong in dor-services,
-    # for this method and other similar methods. can then clean up
-    # allow(idmd).to receive(:"ng_xml") (and "content_will_change!") in tests.
-    # getting rid of "Metadata\.content\s*=\s*"
-    @object.identityMetadata.content = @object.identityMetadata.ng_xml.to_xml
-    @object.identityMetadata.content_will_change!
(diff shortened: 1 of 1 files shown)


ndushay:
upstream (dor-services gem) WILL do:  

```
    parts = value.split(':', 2).map(&:strip)
    raise ArgumentError, "Source ID must follow the format 'namespace:value', not '#{value}'" unless
      parts.length == 2 && parts[0].present? && parts[1].present?
```

ndushay:
Okay, I was confused earlier.  Now I understand that validation is happening in dor-services gem, as the commit comment explains.  :-P   So 👍  from me.

app/controllers/items_controller.rb
```diff
@@ -412,22 +412,9 @@ def close_version
     end
   end
 
-  # TODO: this should be a method in dor-services, invoked within set_source_id
-  def self.normalize_source_id(new_source_id)
-    src_id_arr = new_source_id.split(':').map(&:strip)
-    raise 'new source_id must be of the form "source:value"' if src_id_arr.length != 2
```
ndushay: ignore above comment

app/controllers/items_controller.rb
```diff
@@ -412,22 +412,9 @@ def close_version
...
-    src_id_arr.join(':')
-  end
-
   def source_id
-    new_id = ItemsController.normalize_source_id(params[:new_id])
-    @object.source_id = new_id
```
ndushay: ignore above comment

app/controllers/items_controller.rb
```diff
@@ -412,22 +412,9 @@ def close_version
...
-    # TODO: the content= and content_will_change! calls belong in dor-services,
-    # for this method and other similar methods. can then clean up
-    # allow(idmd).to receive(:"ng_xml") (and "content_will_change!") in tests.
-    # getting rid of "Metadata\.content\s*=\s*"
-    @object.identityMetadata.content = @object.identityMetadata.ng_xml.to_xml
-    @object.identityMetadata.content_will_change!
```
atz: I agree this TODO and `@object` lines are obsolete.  The lines it references in test appear to already be gone and the setter from Dor::Services now calls `ng_xml_will_change!`.

PR #938, opened 2017-02-14
Add goobi workflow selector drop-down

### LINKED ISSUE #937: Add choice for external Goobi workflow to item registration
The DPG lab uses a commercial workflow package called Goobi to implement the in-house digitization workflow.  All DOR-side processing that passes work to DPG is handled by a single workflow, formerly dpgImageWF, now goobiWF.  Goobi requires that any item coming from DOR specify a specific workflow within Goobi, based on the specific kinds of material being digitized. 

Add a new drop-down menu to the ARGO item registration form called “External Workflow” which has a fixed list of values.  The selected value creates a tag called “DPG : Workflow : [VALUE]”.  

Ideally only show this drop-down menu when the goobiWF is selected in the “Initial workflow” drop down menu, else we might have to have the list show more specific values that identify the choices as DPG and Goobi related (simply show the whole tag as choices?).

The Item Registration form should pass this down to dor-services registration as simply another admin tag, so no changes to dor-services should be required.

peetucket:
fixes #937 

add a new drop-down menu to select the goobi (external) workflow name if the goobiWF is selected for the initial workflow

### DIFF
--- a/app/assets/javascripts/registration/grid.js.erb
+++ b/app/assets/javascripts/registration/grid.js.erb
@@ -10,7 +10,7 @@ var gridContext = function() {
 
   var statusFormatter = function(val, opts, rowObj) {
     if (val in $t.statusIcons) {
-      var result = '<span class="' + $t.statusIcons[val] + '" title="' + 
+      var result = '<span class="' + $t.statusIcons[val] + '" title="' +
         (rowObj.error||val)+'" aria-hidden="true"></span>';
       if (rowObj.druid) {
         var href = dor_path + "objects/druid:" + rowObj.druid;
@@ -428,7 +428,7 @@ var gridContext = function() {
       $('#object_type').change(function(evt) {
         var sender = evt.target
         var valid_controls = {
-          'item'       : ["object_type", "apo_id", "rights", "id_source", "workflow_id", "content_type", "project", "registered_by", "tags_0", "tags_1", "tags_2", "tags_3", "tag_4", "collection"],
+          'item'       : ["object_type", "apo_id", "rights", "id_source", "workflow_id", "external_workflow_id", "content_type", "project", "registered_by", "tags_0", "tags_1", "tags_2", "tags_3", "tag_4", "collection"],
           'set'        : ["object_type", "apo_id", "rights", "id_source", "project", "registered_by", "tags_0", "tags_1", "tags_2", "tags_3"],
           'collection' : ["object_type", "apo_id", "rights", "id_source", "project", "registered_by", "tags_0", "tags_1", "tags_2", "tags_3"],
           'adminPolicy': ["object_type", "registered_by"],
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/registration/grid.js.erb
```diff
@@ -470,6 +470,17 @@ var gridContext = function() {
           }
         })
 
+        // Show/hide external workflow drop-down when workflow selector changes
```
cbeer: Does this deserve an integration test?

peetucket: am i missing other integration tests for the reg form? (other than it shows without an error)

is there an existing pattern for adding what will be a pretty simple test like this in?

mejackreed: I think the test could go here: https://github.com/sul-dlss/argo/blob/ac6483915e12ca0cfe4735ca8bab907d962e97c7/spec/features/item_registration_spec.rb

As far as examples for testing that form specifically.. no. In the Argo work last year we didn't touch this stuff, thus it didn't get tests added.

An example of testing javascript forms with Capybara/Poltergeist: https://github.com/sul-cidr/histonets/blob/master/spec/features/image_clean_spec.rb#L24-L40

not the cleanest example there, but the most recent.

jmartin-sul: @cbeer and @peetucket, i think i might have something that addresses basic UI/integration testing.  it's not particularly comprehensive considering that the page is a big dynamic JS edit form, but it at least tests the basic use cases for what peter added and for item registration in general.  thanks to @mejackreed for the help in debugging.

see https://github.com/sul-dlss/argo/pull/946

app/assets/javascripts/registration/register.js.erb
```diff
@@ -1,4 +1,5 @@
 $.ajaxQ('register', { maxRequests: 10 });
+$('#external_workflow_dropdown').hide(); // external (e.g. goobiWF) workflow selector hidden by default
```
cbeer: Should we hide this by default using a css class or something in the HTML to prevent possibly displaying the field briefly?

jmartin-sul: https://github.com/sul-dlss/argo/pull/952

jmartin-sul:
travis was failing because the item reg test expected a tag based on the content types as they were prior to #944.  i'm actually not sure why only one of the travis runs failed, as both should have failed, since that shouldn't be an intermittent error.  anyway, pushed a branch that should hopefully fix the tests, while also cleaning up the commit history a bit.  https://github.com/sul-dlss/argo/tree/add-goobi-workflows-to-reg-jm-fixup

will switch this PR to use that if it passes travis for the push.

jmartin-sul:
well, apparently you can switch the base branch, but not the branch being pulled.  just opened: https://github.com/sul-dlss/argo/pull/953

jmartin-sul:
same work slightly cleaned up and merged to master in PR #953

PR #944, opened 2017-02-16
update content types

### LINKED ISSUE #939: Add/update content types shown in reg drop down menu
List should be:

Book (ltr)
Book (rtl)
File
Image
Manuscript (ltr)
Manuscript (rtl)
Map
Media
Software
3D

peetucket:
fixes #939

### DIFF
--- a/app/helpers/registration_helper.rb
+++ b/app/helpers/registration_helper.rb
@@ -32,17 +32,16 @@ module RegistrationHelper
 
   def valid_content_types
     [
-      'Book (flipbook, ltr)',
-      'Book (flipbook, rtl)',
-      'Book (image-only)',
+      'Book (ltr)',
+      'Book (rtl)',
       'File',
       'Image',
-      'Manuscript (flipbook, ltr)',
-      'Manuscript (flipbook, rtl)',
-      'Manuscript (image-only)',
+      'Manuscript (ltr)',
+      'Manuscript (rtl)',
       'Map',
       'Media',
-      'Software'
+      'Software',
+      '3D'
     ]
   end
 


cbeer:
I don't know how the data flows, but we've determined that the new values will work with downstream environments and that no changes will be needed in e.g. purl, stacks, embed?

LynnMcRae:
The pulldown in question populates a tag in a new DOR object that is used by assembly to set the right content type when it creates contentMetadata. The use of any new content type of course relies on both the processing and delivery chains to support it, which is coming, so we'd be careful about releasing any object that didn't have support, even if that support initially was a fallback to file-delivery (I think that is built into embed).

peetucket:
actually I don't believe the assembly robots or pre-assembly use this tag to decide on how to generate content metadata.... up until very recently the assembly robots never generated contentMetadata, and the recent addition to the robot suite either generates from a "stub" (currently only produced by goobi) or vanilla CM of type "file"

pre-assembly uses info given to it in a .yaml config file for the specific job being kicked off

LynnMcRae:
Sorry, I didn't actually mean the pre-assembly code. It was used by the labs in creating the objects, guiding some aspects of production (like ltr vs rtl)) and perhaps the source of what goes into the yaml file.  I will check with DPG and SMPL to verify this is still a useful part of their workflow.

peetucket:
ahh cool, so DPG is the consumer of this knowledge...was worried i was forgetting something important here

cbeer:
👍 I'd just want to sure we update any consumers of this information before producing data that may disrupt downstream activities.

jmartin-sul:
is this still WIP?  still uncertainty about affected consumers?  or good to go?  @peetucket @cbeer

peetucket:
uncertainty only about the content types being listed as correct -- may need to confirm with @blalbrit

LynnMcRae:
I checked and they match the list on the board, so looks good to me. @blalbrit?

PR #945, opened 2017-02-16
add "refresh descMetadata" blue button to detail pages for objects w/ entry in symphony (i.e. objects w/ catkey)

### LINKED ISSUE #941: Add "Refresh metadata" blue button to item display
This is a capability to refresh the MODS in descMetadata from Symphony when there is a catkey in the record.  It is identical in functionality to the parallel action in Bulk Updates so there should already be a DOR service to do this. The button should only be enabled when there is a catkey in the record (\<identityMetadata>\<otherId name="catkey">)

Put the new button between "Set Rights" and "Manage Release".

jmartin-sul:
closes #941

### DIFF
--- a/app/helpers/argo_helper.rb
+++ b/app/helpers/argo_helper.rb
@@ -104,6 +104,9 @@ module ArgoHelper
       if object.datastreams.include? 'rightsMetadata'
         buttons << {:url => url_for(:controller => :items, :action => :rights, :id => pid), :label => 'Set rights'}
       end
+      if object.datastreams.include?('identityMetadata') && object.identityMetadata.otherId('catkey').present? # indicates there's a symphony record
+        buttons << {url: refresh_metadata_item_path(id: pid), label: 'Refresh descMetadata', new_page: true}
+      end
       buttons << { url: manage_release_solr_document_path(pid), label: 'Manage release' }
     end
     if doc.key?('embargo_status_ssim')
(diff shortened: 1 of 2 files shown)


spec/helpers/argo_helper_spec.rb
```diff
@@ -14,12 +14,13 @@
       allow(@object).to receive(:can_manage_item?).and_return(true)
       allow(@object).to receive(:pid).and_return(@item_id)
       desc_md = double(Dor::DescMetadataDS)
-      id_md   = double(Dor::DescMetadataDS)
+      @id_md  = double(Dor::IdentityMetadataDS)
```
jmartin-sul: arguably, this switch to `Dor::IdentityMetadataDS` could've been it's own commit, since it's fixing an incorrect mocking that wasn't specific to the test i added, but it seemed fine to bundle this in since it's so small.  happy to break out if that's preferred.  making `id_md` an instance var was specific to the new test, which needed to `allow(@id_md).to receive(:otherId).with('catkey').and_return(['1234567'])`.

app/helpers/argo_helper.rb
```diff
@@ -104,6 +104,9 @@ def render_buttons(doc, object = nil)
       if object.datastreams.include? 'rightsMetadata'
         buttons << {:url => url_for(:controller => :items, :action => :rights, :id => pid), :label => 'Set rights'}
       end
+      if object.datastreams.include?('identityMetadata') && object.identityMetadata.otherId('catkey').present? # indicates there's a symphony record
+        buttons << {url: url_for(controller: :items, action: :refresh_metadata, id: pid), label: 'Refresh descMetadata', new_page: true}
```
cbeer: I think it's preferable to use (or create) a named route (allowing us to refactor things like controller and action names without breaking compatibility). Is that possible here?

jmartin-sul: @cbeer: done using `refresh_metadata_item_path` (which was already there)

blalbrit:
In -stage, refresh descMD blue button is failing (with both opened and unopened versions):

screenshot 2017-03-03 13 14 44

Issue #940, opened 2017-02-15
Improve integration testing of registration form


peetucket:
Currently, the registration form is not well tested in Argo.  We should have some integration tests to confirm form submissions produces the expected parameters in various scenarios.

jcoyne:
Integration testing now exists for the registration form.

PR #946, opened 2017-02-16
feature tests for Add goobi workflows to registration & basic item registration in general

jmartin-sul:
taking a shot at addressing https://github.com/sul-dlss/argo/pull/938/files#r101326897

addresses #940 (not sure if it's comprehensive enough to close it)

### DIFF
--- a/app/controllers/registration_controller.rb
+++ b/app/controllers/registration_controller.rb
@@ -12,17 +12,20 @@ class RegistrationController < ApplicationController
     render :plain => pdf.render, :content_type => :pdf
   end
 
-  def workflow_list
-    docs = Dor::SearchService.query(%(id:"#{params[:apo_id]}"))['response']['docs']
+  def workflows_for_apo(apo_id)
+    docs = Dor::SearchService.query(%(id:"#{apo_id}"))['response']['docs']
     result = docs.collect { |doc| doc['registration_workflow_id_ssim'] }.compact
-    apo_object = Dor.find(params[:apo_id], :lightweight => true)
+    apo_object = Dor.find(apo_id, :lightweight => true)
     adm_xml = apo_object.administrativeMetadata.ng_xml
     adm_xml.search('//registration/workflow').each do |wf|
       result << wf['id']
     end
+    result.flatten.uniq.sort
+  end
 
+  def workflow_list
     respond_to do |format|
-      format.any(:json, :xml) { render request.format.to_sym => result.flatten.uniq.sort }
+      format.any(:json, :xml) { render request.format.to_sym => workflows_for_apo(params[:apo_id]) }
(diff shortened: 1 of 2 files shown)


app/controllers/registration_controller.rb
```diff
@@ -12,17 +12,20 @@ def tracksheet
     render :plain => pdf.render, :content_type => :pdf
   end
 
-  def workflow_list
-    docs = Dor::SearchService.query(%(id:"#{params[:apo_id]}"))['response']['docs']
+  def workflows_for_apo(apo_id)
+    docs = Dor::SearchService.query(%(id:"#{apo_id}"))['response']['docs']
     result = docs.collect { |doc| doc['registration_workflow_id_ssim'] }.compact
-    apo_object = Dor.find(params[:apo_id], :lightweight => true)
+    apo_object = Dor.find(apo_id, :lightweight => true)
```
cbeer: I don't think the `lightweight` option does anything any more.

app/controllers/registration_controller.rb
```diff
@@ -12,17 +12,20 @@ def tracksheet
     render :plain => pdf.render, :content_type => :pdf
   end
 
-  def workflow_list
-    docs = Dor::SearchService.query(%(id:"#{params[:apo_id]}"))['response']['docs']
+  def workflows_for_apo(apo_id)
+    docs = Dor::SearchService.query(%(id:"#{apo_id}"))['response']['docs']
```
cbeer: I'm not sure I understand why we do this with weird solr indirection, especially when we load the object below anyway?

cbeer:
Two comments, not directly related to your new test, though.

app/controllers/registration_controller.rb
jmartin-sul: filed https://github.com/sul-dlss/argo/issues/949

app/controllers/registration_controller.rb
jmartin-sul: me neither.  filed a new issue to investigate:  https://github.com/sul-dlss/argo/issues/950

jmartin-sul:
@cbeer:  yeah, agreed, we should probably clean up the two things you mentioned.  i'm happy to get rid of lightweight usage pretty much immediately, since it only appears in a couple places, and should be easy to test.  i'm also not sure why we do the other thing you mentioned, but it wouldn't surprise me if we were loading things unnecessarily.

but i'd sort of prefer to fix both under new tickets, since neither thing was affected by this PR or the PR it's augmenting, and since these PRs should make the app better without making the specific things you mention worse.

jmartin-sul:
merging based on @peetucket's general approval, and @cbeer's sign-off on my last comment...

PR #951, opened 2017-02-23
use mods 3.6

### LINKED ISSUE #942: Update code references for MODS from 3.4 to 3.6
This ticket refers only to those places where we generate the MODS header that currently looks like this

<mods:mods xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:mods="http://www.loc.gov/mods/v3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-4.xsd" version="3.4">

The two references to "3.4" just need to change to 3.6.  The places we know need looking at are:
- When an object is first registered, an initial descMetadata datastream is created with stub MODS just for the title. 
- Modsulator creates MODS from a spreadsheet

Both ideally get the needed document header from one place, perhaps 
https://github.com/sul-dlss/mods/blob/master/lib/mods/constants.rb#L3

Note that there is code to call a versioned XSLT for the MARC to MODS conversion. This should be left at 3.4 until a 3.6 XSLT is installed.

jmartin-sul:
* upgrade to a version of dor-services that uses MODS 3.6 when creating MODS header in descMetadata
* switch to using MODS 3.6 XSD for descMetadata MODS validation (in the old bulk operations)

things that seem worth testing based on the dor-services and argo changes:
* Create minimal descriptive metadata from the title (in bulk ops in argo, ultimately calls set_desc_metadata_using_label via create_minimal_mods in argo)
* item registration
* Run mods schema validation with fields that'd pass 3.6 but not 3.4 (also in old bulk ops)

@LynnMcRae or @blalbrit, would one of you mind doing a bit of testing ~once~ now that this branch is deployed to stage?

closes #942


jmartin-sul:
@LynnMcRae says this is good to go.  rebased and pushed up again (which evaporated the dor-services upgrade out, since this was rebased on work that upgraded to a later version of dor-services).

PR #956, opened 2017-03-03
do not show manage catkey button for apos

peetucket:


app/helpers/argo_helper.rb
```diff
@@ -94,9 +94,9 @@ def render_buttons(doc, object = nil)
       }
 
       buttons << {:url => source_id_ui_item_path(id: pid), :label => 'Change source id'}
-      buttons << {:url => catkey_ui_item_path(id: pid), :label => 'Manage catkey'}
       buttons << {:url => tags_ui_item_path(id: pid), :label => 'Edit tags'}
-      unless object.datastreams.include? 'administrativeMetadata' # apos cant be members of collections
+      if [Dor::Item, Dor::Set, Dor::Collection].include? object.class # these only apply for items, sets and collections
```
jmartin-sul: how about:
```ruby
if [Dor::Item, Dor::Set].any? { |clazz| object.is_a? clazz }
```

instead?  then you can do the type checking a bit more polymorphically.  and you wouldn't have to check for `Collection`, since `Dor::Collection` subclasses `Dor::Set`.

peetucket: good idea...didn't realize Dor::Collection subclasses ... made the change and squashed commits..

PR #958, opened 2017-03-16
Set APO (single object and bulk)

### LINKED ISSUE #957: add UI for moving objects to a new APO, individually, or in bulk
currently, the only way to move an object to a new governing APO is to manually edit the XML (in the `RELS-EXT` datastream).

we'd like to add a blue button on the object detail page for moving an object to a new APO, as well as a bulk action (in the newer bulk action framework) for updating multiple items.

further info:
* the user must have management privileges on both the current governing APO and the target APO (and the target governing object should be an APO, not sure if we have to check that manually since we'll be setting via an ActiveFedora accessor that should only allow references to APOs).
* if there's an `adminPolicy` reference in `identityMetadata` datastream, blank it out (we used to track APO here too, but don't anymore, and this is a bit of remediation, since we don't want to perpetuate writing APO here, nor do we want two different APO references on some objects that have been updated by this UI).

when i was initially starting in on this, @mejackreed suggested these PRs as references for the bulk update part of things:
https://github.com/sul-dlss/argo/pull/811/files
https://github.com/sul-dlss/argo/pull/804/files

jmartin-sul:
this is getting pretty close to done, so here's a WIP PR, since it's easier to comment on that than on a branch by itself.

here's what's left to do, AFAIK at the moment:
* [x] feature spec for the individual object update case (partially done on my laptop, but not yet pushed up)
* [x] items_controller_spec.rb additions
* [x] ability_spec additions.rb
* [x] manual testing on stage

here are some questions i have for what's done so far:
* the APO update (not including saving or reindexing) consists of the following two lines:
```ruby
@object.admin_policy_object = Dor.find(params[:new_apo_id])
@object.identityMetadata.adminPolicy = nil if @object.identityMetadata.adminPolicy # no longer supported, erase if present as a bit of remediation
```
seems like that should be broken out into a common helper method.  or maybe placed in dor-services as a helper method?  it's only two repeated lines, but it seems like it'd be good to centralize them.  then, if other things wanted to update APO, they'd be doing it correctly via common code.
* the permissions checking is done three different ways in three different places:  items_controller does it in its usual style, using `before_action` filters.  the job does the check in a single conditional statement called from the method that does the update.  the button rendering (necessarily) does a slightly more relaxed check in a nested conditional where the button is added to the list of buttons to render (it can't check that the user has permission to move the object to the target APO, because there's no way to know what APO the user might want to move the object to at button render time).
* there's some repetition between the edit forms used for the bulk job and the individual obj edit, but this is actually the repetition i'm least concerned about.  the form is pretty simple, and i'm not sure it'd be less complex or more maintainable to abstract it to something shared.

closes #957

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -22,7 +22,7 @@ class Ability
     can :manage, :all if user.is_admin?
     cannot :impersonate, User unless user.is_webauth_admin?
 
-    can [:manage_item, :manage_content, :manage_desc_metadata, :view_content, :view_metadata], ActiveFedora::Base if user.is_manager?
+    can [:manage_item, :manage_content, :manage_desc_metadata, :manage_governing_apo, :view_content, :view_metadata], ActiveFedora::Base if user.is_manager?
     can :create, Dor::AdminPolicyObject if user.is_manager?
 
     can [:view_metadata, :view_content], ActiveFedora::Base if user.is_viewer?
@@ -55,6 +55,11 @@ class Ability
       if dor_item.admin_policy_object
         dor_item.can_manage_desc_metadata? user.roles(dor_item.admin_policy_object.pid)
       end
+    end
+
+    can :manage_governing_apo, ActiveFedora::Base do |dor_item, new_apo_id|
+      # user must have management privileges on both the target APO and the APO currently governing the item
+      dor_item.can_manage_item?(user.roles(new_apo_id)) && can?(:manage_item, dor_item)
     end
 
     can :view_content, Dor::AdminPolicyObject do |dor_item|
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -13,6 +13,11 @@
 #   check to the APO's own ID for Dor::AdminPolicyObject types in addition to the checks for ActiveFedora::Base types (e.g. the
 #   "can :manage_item, Dor::AdminPolicyObject" and "can :manage_item, ActiveFedora::Base" definitions, with the latter being checked
 #   first).  see https://github.com/sul-dlss/argo/issues/76
+#
+# note about confusing dor-services method declarations:
+#  * the can_manage_*? and can_view_*? method calls (e.g. dor_item.can_manage_item? or dor_item.can_view_metadata?) don't
+#  actually do anything with the state of the object that receives the message.  they could all be static methods in Dor::Governable,
+#  since  they just check the intersection of the given roles with the appropriate static list of known roles.
 class Ability
   include CanCan::Ability
 
--- a/app/controllers/bulk_actions_controller.rb
+++ b/app/controllers/bulk_actions_controller.rb
@@ -50,7 +50,8 @@ class BulkActionsController < ApplicationController
       :action_type,
       :description,
       :pids,
-      manage_release: [:tag, :what, :who, :to]
+      manage_release: [:tag, :what, :who, :to],
+      set_governing_apo: [:new_apo_id]
     ).merge(webauth: { privgroup: webauth.privgroup, login: webauth.login })
   end
 end
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -29,6 +29,9 @@ class ItemsController < ApplicationController
   before_action :authorize_manage_desc_metadata!, :only => [
     :refresh_metadata
   ]
+  before_action :authorize_set_governing_apo!, :only => [
+    :set_governing_apo
+  ]
   before_action :enforce_versioning, :only => [
     :add_collection, :set_collection, :remove_collection,
     :source_id, :set_source_id,
@@ -36,6 +39,7 @@ class ItemsController < ApplicationController
     :refresh_metadata,
     :set_content_type,
     :set_rights,
+    :set_governing_apo,
     :tags,
     :update_rights
   ]
@@ -49,7 +53,8 @@ class ItemsController < ApplicationController
     :catkey,
     :refresh_metadata,
     :set_rights,
-    :set_content_type
+    :set_content_type,
+    :set_governing_apo
   ]
   # must run after save_and_reindex
   prepend_after_action :flush_index, :only => [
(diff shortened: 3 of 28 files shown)


app/models/ability.rb
```diff
@@ -57,6 +57,14 @@ def initialize(user)
       end
     end
 
+    can :manage_governing_apo, ActiveFedora::Base do |dor_item, new_apo_id|
+      next false unless dor_item.can_manage_item?(user.roles(new_apo_id))
```
tingulfsen: Can you add a comment here to explain this gotcha?

app/models/ability.rb
```diff
@@ -57,6 +57,14 @@ def initialize(user)
...
+
+      if dor_item.admin_policy_object
```
tingulfsen: This if block is repeated from line 48. Can you try to see if we can "call" the cancan block on L48 so we don't have to  copy the code?

app/models/ability.rb
```diff
@@ -22,7 +22,7 @@ def initialize(user)
     can :manage, :all if user.is_admin?
```
tingulfsen: At the top of this file, can you add a comment about the confusing nature of the DOR can_manage_item? method and how it should be made static in the future?

app/models/ability.rb
jmartin-sul: that worked

app/models/ability.rb
jmartin-sul: no longer needed, because the whole thing was compacted into a single `&&` conditional once the `if dor_item.admin_policy_object ...` part got turned into `can?(:manage_item, dor_item)`

jmartin-sul:
@tingulfsen: i think i've addressed all of your suggestions.

@LynnMcRae: you'd noted some objects that failed to get moved to a new APO when you tested them with the bulk update in stage (druid:bb251ny2915, druid:hf427fj3231, druid:hx034wg5194).  when i looked at these objects the other day, it appeared that none of them were open for modification.  so, when you get a chance, could you try: 1) moving those same objects again, and 2) moving some objects that are open.  (1) should still not move the objects, but it should give a more specific error message (it'll now differentiate between permission denied based on the APO and permission denied based on openness).  (2) should work (hopefully).

also, the "set governing APO" blue button should now pop up a modal with the expected close button and no app navbar, and should now also default to the current governing APO (i noticed that it had not been doing that previously).

latest code is deployed on stage.  took off the [WIP] since things appear to be working based on the testing i've done, and since i think i've addressed all the feedback so far.

LynnMcRae:
BulkAction job 4 with these druids in stage

druid:bb040vm6679
druid:bb059bk1259
druid:by882sx9664
druid:bb097pj4363

All 4 succeeded
- the 3rd is in accessioning so should not have been succeeded. The Blue Button for Set APO is also enabled on the detail page and shiould not be. 
- the other three are accessioned. Should they have been rejected or should they have been opened for modification as part of the bulk update (I think the latter is how other bulk actions/updates work but would have to confirm. @blalbrit did I make this up?).

blalbrit:
@LynnMcRae - I think the pattern for something like this is to require opening a version first, making the change, and then closing.

LynnMcRae:
hmmm, these three items In Accessioning all failed as expected (BulkAction 41):

druid:bb251ny2915, druid:hf427fj3231, druid:hx034wg5194

jmartin-sul:
weird... i'll have to look into the exact logic of the `allows_modification?` method.  FWIW, that's the same method that all of the other `ItemsController` methods use to check whether an object can be modified.

also maybe worth seeing if there's something obviously wrong or inconsistent with the datastreams on the objects that behaved unexpectedly.

as ben said, this functionality shouldn't open objects for modification.  the user should open the objects first, as usual with other commands that require things to be open for modification.

LynnMcRae:
I'm looking for oddities in the objects in stage but don't see any so far. There are objects that claim to be "Accessioned" and their WF is complete but they have an sdrIngestWF error. I am voiding these in tests now (and didn't report any of those as issues.

LynnMcRae:
We still have the special exemption for items "In Accessioning" but that have sdr-ingest-transfer on "hold".  I think that explains what I thought were false successes.  Plus the display status and indexed status didn't necessarily match.   The tests all look good now. Ben to confirm from his trials.

blalbrit:
Looks good from my tests too @LynnMcRae . Is this ready to ship?

LynnMcRae:
Single button dialog looks good too ... where's the github thumbs up emoji?

blalbrit:
👍

LynnMcRae:
:+1:

jmartin-sul:
nice!  well, let me know if i should do any further investigation into this morning's test oddities.

absent that, i defer to @tingulfsen to confirm that his concerns have been addressed.  would also be nice to get feedback from @mejackreed since he was the other implementer of the bulk actions framework, but seems like tommy did a pretty thorough review, so i'd be fine merging without jack's input if he doesn't have time to review a large PR in between wrangling conference stuff.

LynnMcRae:
We're good for a release. Test anomalies have been explained.

jmartin-sul:
met with @mejackreed this afternoon.  we agreed that `ItemsController` should reject requests that have `params[:bulk]` for the new `set_governing_apo` method, under the assumption that we shouldn't be adding usage of the old bulk update mechanism, esp for actions that are implemented in the new bulk action system.

we also agreed that it'd be nice if the new bulk action framework had a straight-forward way to perform an action immediately/synchronously, so that we could invoke that for single-object updates that also have bulk action counterparts.  but that would take a bit of work/investigation, and seems a bit outside the scope of this PR.  so we'll defer that as a future tech debt task, and ship this in its current form, since the code repeated between the controller method and the job is minimal, and since this is working and pretty much ready to go.

if @mejackreed agrees that the fixup i just pushed addresses his concern, i think this is ready for merge.

blalbrit:
Awesome @jmartin-sul . I agree completely with the move away from the old bulk update mechanism. If you haven't already, could you add the single/bulk task to a backlog ticket so that we can keep it in the prioritization process for upcoming work?

jmartin-sul:
@blalbrit: yup, will do.

PR #960, opened 2017-03-23
fix refresh descMetadata UI

jmartin-sul:
previously, the controller just showed a plaintext success method, as it had only been invokable from the old bulk actions page.

this updates it to keep the same behavior for bulk actions, but for single object update, it now redirects back to the object detail page with a flash message indicating success.  it also guards against invocation on objects without catkeys.

also, tests for all the aforementioned behavior.

and a bit of refactoring to have the standard `after_action` handle saving the updated object, instead of doing it explicitly in the updating controller method.

screen shot 2017-03-22 at 10 09 15 pm

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -26,10 +26,14 @@ class ItemsController < ApplicationController
     :embargo_update,
     :embargo_form
   ]
+  before_action :authorize_manage_desc_metadata!, :only => [
+    :refresh_metadata
+  ]
   before_action :enforce_versioning, :only => [
     :add_collection, :set_collection, :remove_collection,
     :source_id, :set_source_id,
     :catkey,
+    :refresh_metadata,
     :set_content_type,
     :set_rights,
     :tags,
@@ -709,6 +713,10 @@ class ItemsController < ApplicationController
 
   def authorize_manage_item!
     authorize! :manage_item, @object
+  end
+
+  def authorize_manage_desc_metadata!
+    authorize! :manage_desc_metadata, @object
   end
 
   def enforce_versioning
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -537,9 +537,21 @@ class ItemsController < ApplicationController
   end
 
   def refresh_metadata
+    if @object.catkey.blank?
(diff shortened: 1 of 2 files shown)


jmartin-sul:
two questions for @LynnMcRae and/or @blalbrit:
1) should this only be allowed for items that allow modification?
2) should this only be allowed for users who can manage object content?

currently it doesn't look like either of those conditions is enforced for this operation.

blalbrit:
@jmartin-sul 

1. It seems like we would want to be able to support this on objects in a registered state, which along with 'accessioned' or 'opened' state, would limit it to modifiable objects. We would not this to be available on 'in accessioning' state.

2. Yes - this should be limited to users who can manage object content.

app/controllers/items_controller.rb
```diff
@@ -536,10 +537,21 @@ def schema_validation
   end
 
   def refresh_metadata
+    if @object.catkey.blank?
+      render status: :forbidden, plain: 'object must have catkey to refresh descMetadata'
```
peetucket: does the user's browser just return a standard error at this point?

app/controllers/items_controller.rb
```diff
@@ -536,10 +537,21 @@ def schema_validation
...
+      return
+    end
+
     @object.build_datastream('descMetadata', true)
     @object.descMetadata.content = @object.descMetadata.ng_xml.to_s
-    @object.descMetadata.save
-    render :status => :ok, :plain => 'Refreshed.'
```
peetucket: just double-checking, so save needed here on the object?

jmartin-sul: most of the operations in this controller use an `after_action` to handle the saving on the object as a whole.  so i moved responsibility out of the controller method and to that `after_action` (the first commit of this PR).  that `after_action` saves the object as a whole, as opposed to saving on the datastream specifically, but testing indicates that things still work fine with saving on the object as a whole.

app/controllers/items_controller.rb
jmartin-sul: yeah.  this is a case that the user really shouldn't run into if they're not manually constructing URLs, because we won't present the button for things without a catkey.  if the user did run into this, they'd get a plaintext page with the error message a 403 forbidden as the HTTP status code.

PR #961, opened 2017-03-23
update dependencies

jmartin-sul:
also, update stubbing in test that failed after blacklight upgrade.

impetus for this was bundle audit preventing deploy due to nokogiri vulnerabilities.  figured i'd try to upgrade all the dependencies, since argo was overdue for that.  attempted to address the test failure we deferred investigating in #914.


spec/views/catalog/show.html.erb_spec.rb
```diff
@@ -4,9 +4,13 @@
   let(:document) { SolrDocument.new :id => 'xyz', :format => 'a' }
   let(:title) { 'Long title that should be truncated at 50 characters' }
   let(:blacklight_config) { Blacklight::Configuration.new }
+  let(:query_params) { { controller: 'catalog', action: 'show' } }
+  let(:search_state) { Blacklight::SearchState.new(query_params, blacklight_config) }
   before(:each) do
     assign :document, document
     allow(view).to receive(:blacklight_config).and_return(blacklight_config)
+    allow(view).to receive(:search_state).and_return search_state
```
jmartin-sul: the intent of this line is to fix the failure seen here:
https://travis-ci.org/sul-dlss/argo/builds/193079542#L787
```
  1) catalog/show.html.erb assigns page title, truncating it
     Failure/Error: <% content_for(:head) { show_presenter(@document).link_rel_alternates } -%>
     
     ActionView::Template::Error:
       undefined method `search_state' for #<#<Class:0x0000000b3aa790>:0x00000012560d30>
     # ./app/views/catalog/show.html.erb:2:in `block in _app_views_catalog_show_html_erb___3659222435883736110_181830440'
     # ./app/views/catalog/show.html.erb:2:in `_app_views_catalog_show_html_erb___3659222435883736110_181830440'
     # ./spec/views/catalog/show.html.erb_spec.rb:18:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method `search_state' for #<#<Class:0x0000000b3aa790>:0x00000012560d30>
     #   ./app/views/catalog/show.html.erb:2:in `block in _app_views_catalog_show_html_erb___3659222435883736110_181830440'
```

i'm slightly uncomfortable with it, because i don't fully understand why i had to add it.  but it seems to work, as long as the below line is also added. because i haven't traced all the way through, i'm not sure if the unexpected `search_state` call is coming from code internal to blacklight, or from one of the `search_state` invocations we make in workflow_helper.rb methods.  i adapted this stubbing from https://github.com/sul-dlss/argo/blob/master/spec/helpers/workflow_helper_spec.rb#L9-L13

a couple months ago, @cbeer tried to update the argo dependencies, and ran into the same failure (see #914).  that update saw the same test failure (which is where the linked travis error comes from).  at the time, we deferred doing an upgrade of all dependencies, and just did a narrow upgrade that didn't run into this error.  in a slack discussion @cbeer said "more investigation is needed to figure out a) what’s going on with all the crazy stubbing in that spec and, b) was that blacklight change a compatibility issue or just bad stubbing in argo."  hoping he or @jcoyne or another blacklight expert can weigh in and suggest whether this extra stubbing is a terrible idea (and hopefully suggest a not-too-arduous alternative if it is indeed a terrible idea).

spec/views/catalog/show.html.erb_spec.rb
```diff
@@ -4,9 +4,13 @@
...
+    allow(view).to receive(:application_name).and_return 'Argo'
```
jmartin-sul: when i added the `search_state` stubbing, but not this line, i got this error:
```
  1) catalog/show.html.erb assigns page title, truncating it
     Failure/Error:
       expect(view.instance_variable_get(:@page_title))
         .to eq 'Long title that should be truncated at 50 chara... - Argo'
     
       expected: "Long title that should be truncated at 50 chara... - Argo"
            got: "Long title that should be truncated at 50 chara... - Blacklight"
     
       (compared using ==)
     # ./spec/views/catalog/show.html.erb_spec.rb:22:in `block (2 levels) in <top (required)>'
```

again, slightly uncomfortable with this change due to lack of understanding, because i was reusing the same `blacklight_config` var as was already working with the test.  but, as this is a view spec, it seemed that the point was not to test the configuration done in the rest of the app, and it seemed defensible to stub the `application_name` result.

jmartin-sul:
also, FWIW, my limited (laptop instance) manual testing shows that the object detail page works just fine with the upgrade, with no changes required.  it seems like only the test broke.

spec/views/catalog/show.html.erb_spec.rb
jcoyne: The search state is provided by a blacklight controller (https://github.com/projectblacklight/blacklight/blob/4e5af828ded57bc426b640bac35a7f94556eee30/app/controllers/concerns/blacklight/controller.rb#L27)  If you're in a view test, it's using a TestController, which doesn't provide that method. So you've done the right thing here by stubbing it out. 👍

spec/views/catalog/show.html.erb_spec.rb
jcoyne: This is coming from here: https://github.com/sul-dlss/argo/blob/ac6483915e12ca0cfe4735ca8bab907d962e97c7/config/locales/blacklight.en.yml#L3  I think you should fix that there and not stub out this instance.

jmartin-sul:
great, thanks for the feedback @jcoyne.  fixed per your suggestion and pushed the updated version, looks like tests pass.

PR #973, opened 2017-04-07
argo helper polish

### LINKED ISSUE #962: refresh descMD error bug - plain text
On trying to refresh descMD when object is not in an opened/etc. state, get plain text error message.

screenshot 2017-03-23 17 42 10

Preferred behavior would be to disable button if object is not in an editable state.

### LINKED ISSUE #963: Blue Buttons: Disable Create Collection and Edit APO when not on an APO object
screenshot 2017-03-24 10 51 45

Clicking either, on an item, throws a "we're sorry, something went wrong" error.

### LINKED ISSUE #967: Blue button tests have non-item item fixtures

### LINKED ISSUE #968: Blue buttons that are disabled based on available object info end up hitting a URL that 404s
in some places, `check_url` (as defined in some of the button hashes returned by `ArgoHelper#render_buttons` for use by `_button_link_list.html.erb` and `button_checker.js`) isn't a URL, but a boolean.  the result is that the `check_url` field can also act as a disabled flag for occasions where the decision to enable the button or not can be made using info already available from the object, as opposed to needing to reach out to e.g. workflow service.

https://github.com/sul-dlss/argo/blob/master/app/helpers/argo_helper.rb#L87
```ruby
buttons << {
  url: set_governing_apo_ui_item_path(id: pid),
  label: 'Set governing APO',
  check_url: !object.allows_modification?
}
```

with the boolean usage, if `check_url` is `false`, the button is enabled, because whether the button is disabled on page load is determined by (https://github.com/sul-dlss/argo/blob/master/app/views/items/_button_link_list.html.erb#L15):
```ruby
link_to button[:label], button[:url], class: "btn button btn-primary #{'disabled' if button[:check_url]}", data: button_data
```

and no request is made, because `check_url` is only added to `button_data` if it's truthy (https://github.com/sul-dlss/argo/blob/master/app/views/items/_button_link_list.html.erb#L11-L13):
```ruby
if button[:check_url]
  button_data[:check_url] = button[:check_url]
end
```

if, `check_url` is `true`, the button is disabled, the value for `check_url` is added to `button_data`, and the javascript that makes the remote call to enable the button will try the URL `true` (and of course get a 404; https://github.com/sul-dlss/argo/blob/master/app/assets/javascripts/modules/button_checker.js#L25):
```javascript
$('a.disabled[data-check-url]').buttonChecker();
```

since the response is a 404, the button never gets enabled, and nothing misleading happens in the UI.  but there's an unnecessary network request and 404, and the code's a little confusing.  not a big deal at all, and not a bug functionally, just a bit of a rough edge that i stumbled across.  feels worth sanding down if time allows.  i was thinking something like the addition of a simple `disabled` flag field at the same level as the `check_url` field in the button hash.  then change the `link_to` in the template to do `#{'disabled' if (button[:check_url] || button[:disabled])}`.  then swap the non-URL uses of the `check_url` field to use the `disabled` flag instead, and add a test for the new flag (in _button_link_list.html.erb_spec.rb, i think).

FWIW, i totally perpetuated this usage with the "Set governing APO" button (PR #958, didn't get to fixing this thing as part of that).

### LINKED ISSUE #969: Disable the "Set governing APO" button correctly

jmartin-sul:
* hide the "Open" and "Close" buttons when the user doesn't have permission to perform those operations (using the same cancan perm check that `ItemsController` uses to determine whether the user can actually perform the operation).
* add a `disabled` flag to the possible button configuration fields, for use instead of `check_url` when it's possible to disable a button without reaching out to another service with a network request (e.g. perm info based on the already loaded dor-services `object`).  #968
 * convert instances where `check_url` is used with readily locally available info to use the `disabled` flag instead.  #969
  * employ the `disabled` flag for "Refresh descMetadata" button rendering (disabled if obj doesn't allow modification).  #962
* fix check to show/hide "Edit APO" and "Create Collection" buttons to be based on whether the object is a `Dor::AdminPolicyObject` (as opposed to using presence of `datastreams['roleMetadata']` as a marker for object type).  #963
* take the two things that do a `can?(:manage_item, object)` check of their own, and put them inside the main `can?(:manage_item, object)` conditional to cut down on repetition.
* add tests for the "Close", "Open", and "Purge" buttons.
* clearer var names, more accurate test descriptions, more precise expectations/description for `#registered_only?` tests.
* more precise expectations on which buttons appear with different object states and privilege levels.  more accurate setup of test for no mgmt privs and no governing APO on item.
* more explicit testing for the set of buttons available on APOs.
* more accurate fixture instantiation and mocking, remove unnecessary allow statements.  #967
* misc formatting cleanup.

closes #969 
closes #968 
closes #967 
closes #963 
closes #962

### DIFF
--- a/app/helpers/argo_helper.rb
+++ b/app/helpers/argo_helper.rb
@@ -131,16 +131,18 @@ module ArgoHelper
         buttons << {url: refresh_metadata_item_path(id: pid), label: 'Refresh descMetadata', new_page: true, disabled: !object.allows_modification?}
       end
       buttons << { url: manage_release_solr_document_path(pid), label: 'Manage release' }
-    end
-    if doc.key?('embargo_status_ssim')
-      embargo_data = doc['embargo_status_ssim']
-      text = embargo_data.split.first
-      # date=embargo_data.split.last
-      if text != 'released'
-        # TODO: add a date picker and button to change the embargo date for those who should be able to.
-        buttons << {:label => 'Update embargo', :url => embargo_form_item_path(pid)} if can?(:manage_item, object)
+
+      if doc.key?('embargo_status_ssim')
+        embargo_data = doc['embargo_status_ssim']
+        text = embargo_data.split.first
+        # date=embargo_data.split.last
+        if text != 'released'
+          # TODO: add a date picker and button to change the embargo date for those who should be able to.
+          buttons << {:label => 'Update embargo', :url => embargo_form_item_path(pid)}
+        end
       end
     end
+
     buttons
   end
 
--- a/app/helpers/argo_helper.rb
+++ b/app/helpers/argo_helper.rb
@@ -65,12 +65,12 @@ module ArgoHelper
       }
     end
 
-    # if this is an apo and the user has permission for the apo, let them edit it.
-    if object.datastreams.include?('roleMetadata') && can?(:manage_item, object)
-      buttons << {:url => register_apo_index_path(id: pid), :label => 'Edit APO', :new_page => true}
-      buttons << {:url => register_collection_apo_path(id: pid), :label => 'Create Collection'}
-    end
     if can?(:manage_item, object)
+      if object.is_a? Dor::AdminPolicyObject
+        buttons << {:url => register_apo_index_path(id: pid), :label => 'Edit APO', :new_page => true}
+        buttons << {:url => register_collection_apo_path(id: pid), :label => 'Create Collection'}
+      end
+
       buttons << {
         url: dor_reindex_path(pid: pid),
         label: 'Reindex',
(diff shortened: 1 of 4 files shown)


jmartin-sul:
i tried to break this into logical commits that built on one another, but am open to some squashing or other restructuring if people feel this breakdown is too granular.

jmartin-sul:
ah, yeah, sorry, that bullet point is incorrect.  the removal happened on the subsequent commit, which also has a note to that effect.  will edit the commit message and re-push...

jmartin-sul:
commit comment should be fixed.

this is the `allow` that was removed: https://github.com/sul-dlss/argo/pull/973/commits/b0da03034dc04645469af4ea049424c899143b92#diff-72df251f7fd40057ac1f8ec7c6c066b5L14

tingulfsen:
👍  from me.

PR #978, opened 2017-04-15
logging and status count improvements for failed job situations

### LINKED ISSUE #972: Background Release job too successful
A Release job for 6900 objects already has 10697 successes suggesting some replay or resubmission? a counting issue?  It may still be out there running. 

pastedimage

jmartin-sul:
* append to the log file when opening it instead of truncating it, so that log info from an unsuccessful job attempt isn't lost.
* zero out the success/failure/total druid counts before performing the job, in case this isn't the first time the job is being attempted for a bulk action, so that the job is incrementing fresh counts.

closes #972

### DIFF
--- a/app/jobs/descmetadata_download_job.rb
+++ b/app/jobs/descmetadata_download_job.rb
@@ -11,7 +11,6 @@ class DescmetadataDownloadJob < GenericJob
   # requires `:pids` (an Array of pids) and output_directory
   def perform(bulk_action_id, params)
     zip_filename = generate_zip_filename(params[:output_directory])
-    initialize_counters(bulk_action)
     with_bulk_action_log do |log|
       #  Fail with an error message if the calling BulkAction doesn't exist
       if bulk_action.nil?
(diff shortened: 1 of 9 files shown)


app/jobs/release_object_job.rb
```diff
@@ -20,7 +20,7 @@ def perform(bulk_action_id, params)
     @manage_release = params[:manage_release]
     @webauth = OpenStruct.new params[:webauth]
     @pids = params[:pids]
-    File.open(bulk_action.log_name, 'w') do |log|
+    with_bulk_action_log do |log|
```
jmartin-sul: inclined to leave this, since i didn't change the line count, and i think i cleaned it up slightly overall.

app/jobs/descmetadata_download_job.rb
```diff
@@ -11,8 +11,7 @@ class DescmetadataDownloadJob < GenericJob
   # requires `:pids` (an Array of pids) and output_directory
   def perform(bulk_action_id, params)
     zip_filename = generate_zip_filename(params[:output_directory])
-    initialize_counters(bulk_action)
-    File.open(bulk_action.log_name, 'w') do |log|
+    with_bulk_action_log do |log|
```
jmartin-sul: inclined to leave this as-is for the same reason

tingulfsen: Agreed.

Issue #970, opened 2017-03-31
Port old bulk jobs to the new bulk actions framework


tingulfsen:

tingulfsen:
Specifically, move the functionality behind the various buttons on the bulk update page (reachable by selecting a subset of records then clicking on the "Bulk Update" button).

PR #980, opened 2017-04-19
move bulk reindex from old bulk updates page to new bulk actions framework

### LINKED ISSUE #743: Make bulk reindexing a background job in bulk framework
"In a future sprint it'd be awesome if we could work on how long it takes to reindex a list of druids within Argo. I'm breaking up the druid lists into batches of about 200 which it can usually handle, but sometimes times out of. I'm doing this via Argo for approx 27k items so it's a lot of granular work.
Thanks for considering!
Laura"

jmartin-sul:
there should now be a "Reindex" option when creating a new bulk action.  the "Reindex" command on the old bulk updates page should now be gone.

deployed to stage, did a quick test by reindexing all the workflows using the new bulk action, seemed to work fine.

closes #743
partially addresses #970

### DIFF
--- a/INDEXING.md
+++ b/INDEXING.md
@@ -47,15 +47,15 @@ The simplest usage is to instantiate `PidGatherer` and just call either `pid_lis
 
 However, there are setters for each list that the query methods memoize to.  As such, you can provide values of your own for each of those lists (or some subset of them).  Then you can take that `PidGatherer` instance and call `pid_lists_for_full_reindex` or `pid_lists_for_unindexed`.  It will use whatever overridden values you've already provided, and for any that weren't manually specified, it'll run the query methods for the first time (which will retrieve and memoize results).
 
-#### Argo::IndexingJob
-`Argo::IndexingJob` is a subclass of `ActiveJob::Base`.  Its `perform` method just calls `Argo::Indexer.reindex_pid_list` (or its profiled counterpart, based on `Settings.INDEXING_JOB.SHOULD_PROFILE`).  Batches are committed immediately if `Settings.INDEXING_JOB.SHOULD_COMMIT_BATCHES`, and not otherwise.  So, you can create a job to reindex a batch of PIDs by doing something like `IndexingJob.delay(priority: priority).perform_later(pid_list)`.
+#### Argo::LocalIndexingJob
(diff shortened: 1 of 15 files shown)


INDEXING.md
```diff
@@ -47,15 +47,15 @@ The simplest usage is to instantiate `PidGatherer` and just call either `pid_lis
...
+ * for each object type sub-list, it will break the list into chunks of `Settings.BULK_REINDEXER.BATCH_SIZE`, and create an `LocalIndexingJob` for each chunk of PIDs.  It will also set the priority on each job, such that each batch has a priority corresponding to the object type contained in that batch (since the lists are already in priority order, it just uses the index of the parent list from `pid_lists`).  The delayed_job workers will then pick up the queued jobs in order of priority, and execute them.  Everything will get logged properly since the jobs are ultimately executing calls to `reindex_pid` (via a few intermediate layers).  Additionally, the delayed_job workers will log higher level info about their success/failure to execute.
```
jmartin-sul: Still exists.  The only change is that what was `IndexingJob` has been renamed to `LocalIndexingJob`.  But `LocalIndexingJob` still uses `Argo::Indexer`, which extends `Dor::IndexingService`, which implements `reindex_pid`:
https://github.com/sul-dlss/dor-services/blob/master/lib/dor/services/indexing_service.rb#L68

A larger question is whether the bulk reindexer that's documented in `INDEXING.md` (which was written for the argo 3 upgrade's reindexing) should live on as-is, get removed entirely, or be modified somehow.  I punted on that question for this PR, as I couldn't come up with an immediately obvious answer.  It still seems like that mechanism might be useful for a full reindex of DOR, though I'd be inclined to modify it so that it too calls out to dor_indexing_app to do the actual indexing work.  I'd be hesitant to do a full reindex of 1.5 million objects using this new bulk reindexing UI, since it does all the objects in one large batch (which should be fine for 10k or so objects, but maybe not for 100k or 500k).  Since the `BulkReindexer` mechanism documented here has batching, it seems like it'd be more sensible to just swap in the remote reindexing call, or have that `BulkReindexer` mechanism use the `RemoteIndexingJob` and then just remove the `LocalIndexingJob` part.

app/jobs/remote_indexing_job.rb
```diff
@@ -0,0 +1,39 @@
+##
+# job to reindex a DOR object to Solr using the dor_indexing_app endpoint
+class RemoteIndexingJob < GenericJob
+  queue_as :indexing_remote
```
jmartin-sul: that was my intent.

i was under the impression that the queue name just gets used at the time the job is queued, and that the queue doesn't have to be declared ahead of time.  if that's wrong, where else would i have to add the new queue name?  and, i assume i'd also have to change the old `:indexing` queue name to `:indexing_local` in that case?

PR #983, opened 2017-05-03
Update travis build matrix

cbeer:


jmartin-sul:
good to know that the error in the failed travis run only happens under ruby 2.4.  i've been running into that on my laptop (which defaults to ruby 2.4).  but it didn't seem to happen elsewhere, so i edited my dor_config.rb to get around the error, and deferred investigation till later.

would be nice to figure out what the issue is, though i do remember @cbeer saying he's also run into bugginess running some of our apps under 2.4.

PR #986, opened 2017-07-05
Add content_file_mimetypes_ssim as a facet field

cbeer:
We should wait to merge / deploy until we've fully reindexed Argo with this new field.


blalbrit:
Exciting! Can't wait to see it in -stage!

PR #989, opened 2017-07-13
Dependency update fixes

jmartin-sul:
* allow deprecation warnings to be silenced by checking an env var (intended for use only by cron jobs that invoke rake tasks)
* fix the gem inclusion that caused invocations of the eye script to fail


jmartin-sul:
i believe this is ready for review (and hopefully merge).  tests pass, deployment happens cleanly, eye successfully restarts on deployment (and operates correctly when you log into the box and ask it for info on the processes it's watching).  also, the .htaccess file is successfully being updated by the rake task (invoked by cron via the whenever gem).

PR #996, opened 2017-07-26
prevent RegistrationController#collection_list from returning entries for collections that don't exist in fedora

jmartin-sul:
also, fix tests accordingly.

previously, if an APO referenced a collection druid in `administrativeMetadata`'s `registration` section, but the collection didn't exist, the collection druid would still be available when registering an item under the APO, listed as "Unknown Collection".  @blalbrit agrees that this is not actually a desirable feature, so this PR gets rid of it.

turns out we have some bad data like this in stage, and in the fixture data for local development work.  but it appears we don't have any data like this in production, thankfully.  i used the following code to check for registration collection druids listed on APOs, where the collection druid doesn't actually exist in the fedora repository:

```ruby
# rows=2000 because we only have 955 APOs on prod at the moment, adjust accordingly if running this in the future
solr_resp = Dor::SearchService.query "objectType_ssim:\"adminPolicy\"", { fl: 'id', rows: 2000 }
apo_id_list = solr_resp['response']['docs'].map{ |doc| doc['id'] }

def registration_collection_ids_for_apo(apo_id)
  apo_object = Dor.find(apo_id)
  apo_object.administrativeMetadata.ng_xml.search('//registration/collection/@id').map(&:to_s)
rescue ActiveFedora::ObjectNotFoundError
  []
end

collection_to_apo_hash = { }
apo_id_list.each do |apo_id|
  registration_collection_ids_for_apo(apo_id).each do |collection_id|
    if collection_to_apo_hash[collection_id]
      collection_to_apo_hash[collection_id].push(apo_id)
    else
      collection_to_apo_hash[collection_id] = [apo_id]
    end
  end
end
# all_registration_collections = apo_id_list.map { |apo_id| registration_collection_ids_for_apo(apo_id) }.flatten.uniq
all_registration_collections = collection_to_apo_hash.keys

missing_registration_collections = all_registration_collections.reject do |col_id|
  begin
    Dor.find(col_id) 
  rescue ActiveFedora::ObjectNotFoundError
    false
  end
end

# use collection_to_apo_hash to look up any collection IDs in missing_registration_collections
```

on argo-stage, APO `druid:ww057vk7675` references collection `druid:nq879cq5811`, but `druid:nq879cq5811` does not appear to exist in fedora on stage.

### DIFF
--- a/app/controllers/registration_controller.rb
+++ b/app/controllers/registration_controller.rb
@@ -33,12 +33,11 @@ class RegistrationController < ApplicationController
   # Data route to return all the registration collections listed for the given APO
   # @option params [String] `:apo_id` the druid for the APO
   # @option params [String] `:truncate` word boundary truncation limit
+  # @return [Hash<String, String>] key represents collection druid, value represents collection title. entries sorted by title, except leading "None" option.
   def collection_list
     truncate_limit = (params[:truncate] || 60).to_i
     collections = { }
-    apo_object = Dor.find(params[:apo_id])
-    apo_object.administrativeMetadata.ng_xml.search('//registration/collection/@id').each do |node|
-      col_id = node.to_s
+    registration_collection_ids_for_apo(params[:apo_id]).each do |col_id|
       col_druid = col_id.gsub(/^druid:/, '')
       col_title_field = SolrDocument::FIELD_TITLE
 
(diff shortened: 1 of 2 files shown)


app/controllers/registration_controller.rb
```diff
@@ -33,12 +33,11 @@ def workflow_list
   # Data route to return all the registration collections listed for the given APO
   # @option params [String] `:apo_id` the druid for the APO
   # @option params [String] `:truncate` word boundary truncation limit
+  # @return [Hash] keys represent collection druids, values represent collection titles, alpha-sorted except for the leading "None" option
```
tingulfsen: I think I understand what you mean, but what about saying what the keys actually are?

jmartin-sul: ah, yeah, good point.  strings, but that's not obvious since they could just as easily be, e.g., symbols.

PR #1000, opened 2017-08-05
When adding a new catkey, strip whitespace from beginning and end.

### LINKED ISSUE #981: Manage Catkey is not trimming leading/trailing spaces from input
Recording them as part of the value has bad consequences for downstream processes that use the value to link records, e.g., SW understanding digital collection membership in released collections. 

Example in stage:
screen shot 2017-04-21 at 4 16 25 pm

tingulfsen:
Fixes #981

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -439,7 +439,7 @@ class ItemsController < ApplicationController
   end
 
   def catkey
-    @object.catkey = params[:new_catkey]
+    @object.catkey = params[:new_catkey].strip
 
     respond_to do |format|
       if params[:bulk]
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -163,10 +163,10 @@ describe ItemsController, :type => :controller do
       post 'catkey', params: { :id => @pid, :new_catkey => '12345' }
       expect(response.code).to eq('403')
     end
-    it 'should update the catkey' do
+    it 'should update the catkey, trimming whitespace' do
       expect(@item).to receive(:catkey=).with('12345')
       expect(Dor::SearchService.solr).to receive(:add)
-      post 'catkey', params: { :id => @pid, :new_catkey => '12345' }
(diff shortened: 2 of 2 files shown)


jmartin-sul:
this prevents new bad data from being created, and is an improvement over the way things were, so merged it.

@LynnMcRae or @blalbrit, should there be a ticket for remediating existing bad data?

blalbrit:
ugh. probably. Fairly low priority at this point, but yes, we should ticket it.

blalbrit:
@jmartin-sul - is it just whitespace, or is there also a hidden line-break or special characters involved?

jmartin-sul:
@blalbrit: off-hand, i can't think of a super quick way to do it.

first thing to do would be to identify the affected records, which i can't think of a quick way to do for all the affected objects.  we index the field, but i don't know how to search for an arbitrary run of whitespace in a field.  someone with better solr-fu (@ndushay? @cbeer?) might know of a way to query for documents that have fields with extraneous whitespace.

either way, we'd have to iterate over the relevant objects and open/edit/close them, right?  which sounds like either a bunch of manual work, or a bit of custom scripting.

re: "is it just whitespace":  i'm guessing it's spaces, tabs, and line-breaks (all of which i'd count as whitespace generally, though not all text processing tools treat line breaks the same as tabs/spaces).  not sure about other special characters.  whatever the remediation approach is, probably best to just trim all leading/trailing whitespace, including line-breaks (but i wouldn't bother looking for other special chars unless we know of a specific reason to).

ndushay:
here is a way to find druids for records of ckeys with a single trailing space:  https://sul-solr.stanford.edu/solr/argo3_prod/select?fq=catkey_id_ssim:*\%20*&fl=id,catkey_id_ssim&facet=false&spellcheck=false

ndushay:
I think this is a way to find ckeys starting with whitespace, but it only found one and it appears to be blank:

https://sul-solr.stanford.edu/solr/argo3_prod/select?q=catkey_id_ssim:/\%20*/&fl=id,catkey_id_ssim&facet=false&spellcheck=false

PR #1004, opened 2017-08-11
Manage versions automatically for APO bulk update

### LINKED ISSUE #987: Bulk "Update governing APO" should open objects not opened instead of failing

tingulfsen:
Fixes #987

### DIFF
--- a/app/jobs/set_governing_apo_job.rb
+++ b/app/jobs/set_governing_apo_job.rb
@@ -8,6 +8,7 @@ class SetGoverningApoJob < GenericJob
   def perform(bulk_action_id, params)
     @new_apo_id = params[:set_governing_apo]['new_apo_id']
     @webauth = OpenStruct.new params[:webauth]
+
     @pids = params[:pids]
 
     with_bulk_action_log do |log|
@@ -36,7 +37,7 @@ class SetGoverningApoJob < GenericJob
       return
     end
 
-    open_new_version(current_obj, log) unless current_obj.allows_modification?
+    open_new_version(current_obj, log, @webauth) unless current_obj.allows_modification?
 
     current_obj.admin_policy_object = Dor.find(new_apo_id)
     current_obj.identityMetadata.adminPolicy = nil if current_obj.identityMetadata.adminPolicy # no longer supported, erase if present as a bit of remediation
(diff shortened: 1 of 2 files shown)


app/jobs/set_governing_apo_job.rb
```diff
@@ -57,4 +62,22 @@ def update_druid_count
...
+      begin
+        vers_md_upd_info = {
+          :significance => 'minor',
+          :description => 'Set new governing APO',
+          :opening_user_name => current_user.to_s
+        }
+        object.open_new_version({:vers_md_upd_info => vers_md_upd_info})
+      rescue Dor::Exception => e
+        log.puts("#{Time.current} Failed to open new version for #{object.pid} (bulk_action.id=#{bulk_action.id}): #{e}")
+        return
```
jmartin-sul: should this increment `:druid_count_fail`?  if so, i'd advocate for just letting the exception propagate up and letting the `rescue` that catches all errors for `set_governing_apo_and_index_safely` deal with it, and then i'd just get rid of the wrapping `rescue` here.  (and i'd prob use that to bolster my case for returning to the `check_can_set_governing_apo!` and just dropping its first check about `obj.allows_modification?`, since that would be handled by the newly added `open_new_version(current_obj, log) unless current_obj.allows_modification?` just below the check.

app/jobs/set_governing_apo_job.rb
```diff
@@ -57,4 +62,22 @@ def update_druid_count
...
+      end
+    else
+      log.puts("#{Time.current} Unable to open new version for #{object.pid} (bulk_action.id=#{bulk_action.id})")
```
jmartin-sul: along the lines of the other two comments about failure counts and logging, i'd just throw an exception here, with the message being `Unable to open new version for #{object.pid}`, and then the enclosing `rescue` for `set_governing_apo_and_index_safely` can handle logging and failure count.

app/jobs/set_governing_apo_job.rb
jmartin-sul: plus, wouldn't this swallow the exception and let it proceed to try setting the APO?  (running the code in my head and not in the app, so very possible i'm guessing wrong 😄 )

tingulfsen:
Superseded by #1172 .

PR #1005, opened 2017-08-14
Bumpbed version to 3.15.3

SaravShah:
Updated version from 3.15.2 to 3.15.3


jmartin-sul:
will merge once it goes green.  we're just being extra-formal here since this is @SaravShah's first deployment, but normally we'd just bump the version in `VERSION` directly on master, and then tag a release (or at least, that's what i usually do).

PR #1007, opened 2017-08-16
Update registration_helper.rb

blalbrit:
add one Goobi tag, remove old extraneous Goobi tags

### DIFF
--- a/app/helpers/registration_helper.rb
+++ b/app/helpers/registration_helper.rb
@@ -48,7 +48,7 @@ module RegistrationHelper
   # the names of the workflows defined in the external system (e.g Goobi)
   # the selected one will create a tag in the object with the value to be passed (to goobi), the tag prefix is defined in constants.rb
   def external_workflow_names
-    %w(Muybridge_Workflow)
+    %w(Test_Workflow_QA Muybridge_Workflow)
   end
 
   def metadata_sources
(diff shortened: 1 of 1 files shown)


jmartin-sul:
looks like the item registration feature tests look for a couple of the removed workflow names:
https://github.com/sul-dlss/argo/blob/master/spec/features/item_registration_spec.rb#L37
https://github.com/sul-dlss/argo/blob/master/spec/features/item_registration_spec.rb#L47-L48
https://github.com/sul-dlss/argo/blob/master/spec/features/item_registration_spec.rb#L95

the current travis failures only point to lines 37 and 47 (https://travis-ci.org/sul-dlss/argo/jobs/265228255#L753-L766), but i believe line 48 and 95 will cause failures once line 47 is fixed, since 47-48 and 95 are all part of one long test.

also, during standup, @blalbrit pointed out that these tags will change somewhat regularly.  as such, it seems like the better long term solution would be to:
* parameterize the tag list and have it come from settings.
* have a known dummy tag list in the `test.yml` config file.
* update the real tag list in shared_configs as needed (which should be easier to review and redeploy than argo code changes, since fixing tests and tagging an argo release would not be required for config changes).

@blalbrit would you mind filing a ticket for the "right" way?

once this PR is fixed and merged, we should tag an argo release and deploy it to prod so that DPG can use this new tag.  should be easy enough to do that this afternoon.

jmartin-sul:
ha, yes, this newly pushed solution also works, and is lower effort than my suggestion, so 👍 

will merge this once the tests pass (which i think will happen, since the one thing that looks for `Test_Workflow` doesn't appear to need that to be the exact value in the field).

then we can do the more robust thing in a day or two, hopefully.

PR #1009, opened 2017-08-17
[WIP] bulk action sub batches

### LINKED ISSUE #990: Bulk "Update governing APO" seems to fail after 600-700 objects
Per recent conversation on SLACK, Josh Schneider (@joshschneider) from UArchives pointed out a problem in the "Update governing APO" functionality which is that for large jobs (13K+), the update job fails after ~600 objects, and then reruns those ~600 objects several more times before the process dies.

I have tested on batches of 600, which complete with no error, and on batches of 1000, which fail with the same problem as above. Historically, we have submitted batches over 1000 which have succeeded - it is unclear if the problem we're seeing is caused by certain content, or by some change in the system.

@jmartin-sul and @jcoyne have tracked the conversation on SLACK with some suggestions and troubleshooting and are aware of the issue.

This is primarily an analysis ticket, which might spawn a technical ticket(s) for a fix. @jmartin-sul suggested a small update to batch large jobs up front, which would be transparent to the user.

Any input from anyone is welcome here, since this is a bit of a mysterious failure. @LynnMcRae has agreed that this could be worked in the upcoming functional debt mini-sprint.

Anyone with better knowledge is welcome to comment or edit to better characterize this problem.

jmartin-sul:
don't merge this yet!

still needs unit tests for the new functionality, and some manual testing to make sure it works as advertised in practice.

fixes #990

### DIFF
--- a/spec/models/bulk_action_spec.rb
+++ b/spec/models/bulk_action_spec.rb
@@ -54,13 +54,36 @@ RSpec.describe BulkAction do
     @bulk_action = BulkAction.create(
       action_type: 'GenericJob', pids: 'a b c', webauth: {}, manage_release: {}
     )
+    # TODO: offset is for the bulk action's spawned by the parent bulk action explicitly created above.  but why's the offset start at 4 instead of 1?
     expect(GenericJob).to receive(:perform_later)
       .with(
-        @bulk_action.id,
+        @bulk_action.id + 4,
         hash_including(
-          pids: %w(a b c),
+          pids: %w(a),
           output_directory: Settings.BULK_METADATA.DIRECTORY +
-            "#{@bulk_action.action_type}_#{@bulk_action.id}",
+            "#{@bulk_action.action_type}_#{@bulk_action.id + 4}",
+          manage_release: {},
+          webauth: {}
+        )
+      )
+    expect(GenericJob).to receive(:perform_later)
+      .with(
+        @bulk_action.id + 5,
+        hash_including(
+          pids: %w(b),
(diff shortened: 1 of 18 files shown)


spec/models/bulk_action_spec.rb
```diff
@@ -54,13 +54,36 @@
...
+          pids: %w(a),
           output_directory: Settings.BULK_METADATA.DIRECTORY +
-            "#{@bulk_action.action_type}_#{@bulk_action.id}",
+            "#{@bulk_action.action_type}_#{@bulk_action.id + 4}",
+          manage_release: {},
+          webauth: {}
+        )
+      )
+    expect(GenericJob).to receive(:perform_later)
+      .with(
```
jmartin-sul: this was the existing style (`.with` instead of `.` on the preceding line), and rubocop has the opposite opinion on this style decision, and i find this more readable, so i'm going to ignore hound here unless others would prefer not to.

mjgiarlo:
@jmartin-sul worth keeping this PR around?

jmartin-sul:
since we no longer seem to be experiencing the problem this PR was attempting to address, and since it'd need some re-basing/re-constituting to be viable for merge 2.5 years later, _and_ since i never really got it working as i'd hoped before the preservation catalog work cycle started, i'm going to go ahead and close this.

we can revisit/reconsider if timeouts on this job type become a problem again, and if we think breaking large batches into smaller batches in the background is a solution worth exploring.

mjgiarlo:
@jmartin-sul should we keep the branch around or dump it?

PR #1022, opened 2017-11-20
Move to testing on solr 6 and use solr_wrapper for Solr

jcoyne:
Do we need to add directions for how to create a solr collection for development, or is the single collection (how we do it in Blacklight) a paradigm that's going to work?

### DIFF
--- a/.solr_wrapper.yml
+++ b/.solr_wrapper.yml
@@ -1,0 +1,6 @@
+port: 8984
+version: 6.6.0
+mirror_url: http://lib-solr-mirror.princeton.edu/dist/
+collection:
+  dir: solr_conf/conf/
+  name: argo-test
--- a/Gemfile
+++ b/Gemfile
@@ -106,6 +106,7 @@ group :test, :development do
   gem 'rails-controller-testing'
   gem 'rspec-rails', '~> 3.5'
   gem 'rubocop', '~> 0.58.1', require: false
+  gem 'solr_wrapper'
   gem 'sqlite3'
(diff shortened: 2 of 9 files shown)


ndushay:
@jmartin-sul you might want to look at this PR, even if it gets merged, just to make sure you don't want to tweak anything.

mejackreed:
👏

atz:
I like getting Solr 6.  Some reservations though.

Obviously Argo is concerned more with indexing logic and index writes than a read-only app like Blacklight.  Single core makes perfect sense for an app that only ever cares about solr read operations.  But if you want to test two different index configurations/schemas, our dev practices should support that.  

It seems like a major concession to default to nuking a dev's entire index every time they want to run tests, since it basically precludes working on the schema and being able to run the same code/tests against both solr configurations.  Samvera stack as a whole has kept itself painfully detached from using or configuring solr expertly.  This kinda cements that in place in a way I consider unproductive.  

Maybe Argo isn't where we would decide to refine our solr schema (though it was previously), but it seems as likely an app as any.  It is more practical to do it here than in a higher level of abstraction (e.g. an engine/gem).  Argo's index is **the** main index of SDR.  When we test something like a major updates, we have performed costly operations like reindexing hundreds of thousands of objects under a new schema.  The precursor to that is having revisions to the schema.  Making that process dumber or more cumbersome increases the risk of incurring that cost multiple times.  

TL;DR: Not going to block this, but not going to approve it either.

PR #1027, opened 2018-02-05
Avoid the extra memory footprint of pry in production

jcoyne:


jmartin-sul:
does this actually consume a lot of memory?  i've found the rails console to be useful for prod debugging, and pry is so much more comfortable than IRB.

not totally opposed to this change, but would prefer to leave it unless it'll result in an actual performance boost (happy to merge it if it does though, since it's not like we need to pull up a console *that* often on prod).

jcoyne:
@jmartin-sul It's not a good idea to be using console on production.  It's too easy to accidentally make an irreversible error with live data (I've done it, don't be like me). If you did find you absolutely couldn't do the debugging you needed to on the dev server, then 

1. We need a better logging
2. We need a mechanism to move a copy of live data to dev/test
3. As a last resort we go to console on prod (ideally with a buddy). In that case we could do a deploy with all the tools we need.

atz:
It's easy to make any kind of error from a command line, interactive or not.  We could also nuke the whole index from curl.  But that conversation isn't really relevant.  The question is about how `pry` is used.

For this app, I don't see a way around using console in production, but I would always use rails console instead of IRB.  I specifically would want to avoid a scenario where production runtime files were being edited to add `binding.pry` statements or anything similar.  @jmartin-sul Is that how you are saying it is being used?

jmartin-sul:
@atz 

no, my usage would typically be something like using dor-services methods to instantiate ActiveFedora objects and see what datastreams and such are there.  i'd be pretty unlikely to do anything that was a write operation, and i wouldn't modify production deployed code to add a `binding.pry` unless it was a pretty exceptional circumstance and i'd run it by someone to make sure i wasn't totally crazy.

@jcoyne, i'd agree that using a rails console on prod is to be avoided if possible, and i'd be happy to have a rule in place where it was standard procedure to pair when doing so, but i've definitely had to use the rails console to diagnose unexpected `NoMethodError: undefined method 'method_i_expect_to_be_there' for nil:NilClass` because some field on some datastream was unexpectedly `nil` in some production data.

> If you did find you absolutely couldn't do the debugging you needed to on the dev server, then
>
>  #. We need a better logging
>  #. We need a mechanism to move a copy of live data to dev/test
>  #. As a last resort we go to console on prod (ideally with a buddy). In that case we could do a deploy with all the tools we need.

amen to all of that, but that feels outside the scope of this PR or patch party, to me at least.

i've also accidentally hosed prod data in prior jobs, and cleaned up after prod data was hosed at this job, so prod consoles are something i try to be very careful about using.

jmartin-sul:
@jcoyne & @atz:  i wouldn't fight tooth and nail to keep pry in the production gem set, so if there's consensus among other infrastructure people that it should go, i'm happy to play along.  i'm just saying that i'm still personally in favor of keeping it.

i'll also note that getting rid of pry doesn't get rid of the ability to use a rails console on prod, it just gets rid of your ability to use one that has syntax highlighting and other niceties.

jcoyne:
I don't care about this enough to keep this open

PR #1029, opened 2018-02-05
Use the Capistrano Git plugin

jcoyne:


atz:
Fixes warning:
```
[Deprecation Notice] Future versions of Capistrano will not load the Git SCM
plugin by default. To silence this deprecation warning, add the following to
your Capfile after `require "capistrano/deploy"`:

    require "capistrano/scm/git"
    install_plugin Capistrano::SCM::Git
```

PR #1036, opened 2018-02-14
Stop using implicit actions with the wrong mime type for the rendered file

### LINKED ISSUE #1035: Weird UI bug for open/close version modal
screenshot 2018-02-12 17 51 03

Within the last week, the open/close version modal has gotten out of whack. Note that the modal now includes a search bar and full html for Argo (rather than being limited to the open/close information. In addition, the Type and Version description boxes are now larger than the modal itself.

Not sure how this happened - but @cbeer it seems to have begun about the time of last week's Argo redeploy.

cbeer:
Fixes #1035


app/controllers/catalog_controller.rb
```diff
@@ -182,9 +182,15 @@ def show
   end
 
   def dc
+    respond_to do |format|
+      format.html { render layout: request.xhr? }
```
mejackreed: In some methods added we have the logic as `!request.xhr?` but here its `request.xhr?` why would we want different logic in these similar stanzas?

cbeer: Nope, that's a mistake.

jcoyne:
@mejackreed Would you please take another look?

PR #1053, opened 2018-03-12
URL-escape file names in stacks urls

cbeer:


mejackreed:
For context, this was spawned by https://github.com/sul-dlss/sul-embed/issues/616

PR #1059, opened 2018-04-05
Resolve dependency error for rubyzip in prod

atz:
See: https://app.honeybadger.io/projects/49894/faults/37153380#notice-trace

`zip` was previously available by chance as a sub-dependency of `roo`, but `roo` recently was removed, thankfully. Tests continued to pass because `jettywrapper` also included `rubyzip`, but `jettywrapper` is intentionally not included in prod.

### DIFF
--- a/app/jobs/descmetadata_download_job.rb
+++ b/app/jobs/descmetadata_download_job.rb
@@ -1,13 +1,13 @@
+require 'zip'
+
 class DescmetadataDownloadJob < GenericJob
   queue_as :default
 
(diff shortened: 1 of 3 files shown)


app/jobs/descmetadata_download_job.rb
```diff
@@ -1,13 +1,13 @@
+require 'zip'
```
jcoyne: This is not necessary if you've declared it in your Gemfile.  🤷‍♂️

PR #1072, opened 2018-05-02
Update user test

jcoyne:
Reverts part of #563

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -54,8 +54,6 @@ describe User, :type => :model do
     it 'should be false for a blank group' do
       allow(subject).to receive(:groups).and_return([])
       expect(subject.is_admin?).to be false
-      allow(subject).to receive(:groups).and_return(nil)
-      expect(subject.is_admin?).to be false
     end
     it 'should be false with an inadequate group membership' do
       allow(subject).to receive(:groups).and_return(['workgroup:dlss:not-admin'])
@@ -101,8 +99,6 @@ describe User, :type => :model do
     it 'should be false for a blank group' do
       allow(subject).to receive(:groups).and_return([])
       expect(subject.is_manager?).to be false
-      allow(subject).to receive(:groups).and_return(nil)
-      expect(subject.is_manager?).to be false
     end
     it 'should be false with an inadequate group membership' do
       allow(subject).to receive(:groups).and_return(['workgroup:dlss:not-manager'])
@@ -133,8 +129,6 @@ describe User, :type => :model do
     end
     it 'should be false for a blank group' do
       allow(subject).to receive(:groups).and_return([])
-      expect(subject.is_viewer?).to be false
-      allow(subject).to receive(:groups).and_return(nil)
       expect(subject.is_viewer?).to be false
     end
     it 'should be false with an inadequate group membership' do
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -42,116 +42,130 @@
...
+      it { is_expected.to be false }
     end
   end
 
-  describe 'is_webauth_admin?' do
-    it 'should be true for ADMIN_GROUPS' do
-      allow(subject).to receive(:webauth_groups).and_return(User::ADMIN_GROUPS)
-      expect(subject.is_webauth_admin?).to be true
+  describe '#is_webauth_admin?' do
+    subject { user.is_admin? }
```
jmartin-sul: this doesn't actually seem to be testing `#is_webauth_admin?` anymore?

spec/models/user_spec.rb
```diff
@@ -42,110 +42,130 @@
...
+    let(:user) { described_class.new }
+    before do
+      allow(user).to receive(:webauth_groups).and_return(groups)
     end
 
-    it 'should be true when impersonating other group membership' do
-      allow(subject).to receive(:webauth_groups).and_return(User::ADMIN_GROUPS)
-      allow(subject).to receive(:groups).and_return(['workgroup:dlss:not-admin'])
```
jmartin-sul: this change seems to lose the point of this test:  that the user is only in a non-admin group according to `#groups` (presumably due to impersonation of another group list).  but they are actually an admin according to `#webauth_groups` (this is what allows them to keep the link to re-set the groups to impersonate, even when they are impersonating a non-super-user).

spec/models/user_spec.rb
```diff
@@ -243,33 +263,46 @@
...
+    end
+
+    context 'when impersonating' do
+      before do
+        user.set_groups_to_impersonate(%w(workgroup:dlss:impersonatedgroup1 workgroup:dlss:impersonatedgroup2))
       end
-      it 'should return the groups by impersonation' do
-        impersonated_groups = %w(workgroup:dlss:impersonatedgroup1 workgroup:dlss:impersonatedgroup2)
-        @user.set_groups_to_impersonate(impersonated_groups)
-        expect(@user.groups).to eq(impersonated_groups)
```
jmartin-sul: i like the more thorough tests for what shouldn't come through from the actual webauth groups, but why lose the simple test for whether `#groups` returns the impersonated group list?  e.g., why not `let(:groups_to_impersonate) { %w(workgroup:dlss:impersonatedgroup1 workgroup:dlss:impersonatedgroup2) }`, use that in the `user.set_groups_to_impersonate` call, and then have something like
```ruby
it 'returns the impersonated groups' do
  expect(subject).to eq groups_to_impersonate
end
```

spec/models/user_spec.rb
jcoyne: https://github.com/sul-dlss/argo/pull/1075

spec/models/user_spec.rb
jcoyne: I'm not sure I follow. Here I'm only setting `webauth_groups`, and we see that when the `webauth_groups` have the admin group, then `is_webauth_admin?` returns `true`. Maybe your concern is addressed by #1075

PR #1073, opened 2018-05-02
Create a HomeTextPresenter

jcoyne:
This is the only place we use the view_something? logic, so extract it
from the model. It is not a model concern.


spec/models/user_spec.rb
```diff
@@ -327,48 +327,4 @@ def groups_to_impersonate
       expect(groups_to_impersonate).to eq ['groupA']
     end
   end
-
-  describe '#can_view_something?' do
```
mjgiarlo: Would you consider adapting this block of tests in a new `spec/presenters/home_presenter_spec.rb`?

jcoyne: ack! I think i just forgot to `git add` that file. Thanks for the good 👁

mjgiarlo: np. Thanks! (I see some conflicts below. Happy to review again once this PR is green and ready to go, if you need another review.)

PR #1076, opened 2018-05-02
Remove webapp context from the jobs

jcoyne:
Jobs should not need to be aware of webauth, and since the `BulkAction` already has a user, there is no need for this.


jcoyne:
replaced by #1078

PR #1103, opened 2018-05-09
now that fedora and solrizer URLs are in the default development.yml, put the dor indexing app URL in there too

jmartin-sul:


config/settings/development.yml
```diff
@@ -2,7 +2,8 @@ PROFILER:
   RACKMINI_ENABLED: true
 
 # URLs
-FEDORA_URL:   'http://fedoraAdmin:fedoraAdmin@localhost:8983/fedora'
+DOR_INDEXING_URL: 'http://localhost:4000/dor'
```
aaron-collier: Is this the correct port? Or am I looking at the wrong spot here:

https://github.com/sul-dlss/argo/blob/master/.travis.yml#L18

jcoyne: Yes, those are different things. 3001 is the workflow service.

jmartin-sul: different service dependency 🙂 

https://github.com/sul-dlss/argo/#run-the-servers

PR #1114, opened 2018-05-16
Use docker-compose to start the services

jcoyne:


jmartin-sul:
weird, the build hung, according to travis.

but i really like this idea, and actually, it'd be really nice if we could fold dor_indexing_app into this too, and have both of those dependencies start up with the one docker command.  but i'd be happy to ticket that and do it as follow-on work.

jcoyne:
I guess I don't understand why we'd want the indexing-app to be in the docker compose.  It doesn't seem to be a dependency of the tests as written now, does it?

jmartin-sul:
i guess it's a thing that i always spin up when doing argo development, because it's necessary for any indexing to happen, and indexing is often necessary to do testing while developing, even when not developing schema specific things.  but it is true, the indexing app isn't necessary to populating fedora or running the tests (though actually i'm not sure why it isn't required to populate fedora and initially index those fedora objects to solr).

but yeah, not strictly necessary.  something i'd find convenient, but like i said, certainly wouldn't hold up the PR over it.  what's currently in this PR seems like the more important thing.

jcoyne:
Closed because I'm probably not going to have time to figure out how to get Travis to start docker-compose.

PR #1117, opened 2018-05-17
Refactor APO register into three distinct actions: new, edit and create

jcoyne:
Merged two tests in spec/integration/apo_spec.rb because the second spec
depended on the first one.

Added some bootstrap styles to the form.


app/controllers/apo_controller.rb
```diff
@@ -17,6 +18,9 @@ class ApoController < ApplicationController
   before_action :authorize, except: [
     :is_valid_role_list_endpoint,
     :register,
```
jmartin-sul: this can get dropped now that the method's gone, right?

PR #1133, opened 2018-05-31
Update dor-services to 5.30.0

aaron-collier:
While this change should have minimal impact, I'm testing on stage before removing [WIP].


jcoyne:
@aaron-collier can you provide an update on this?

PR #1150, opened 2018-07-17
Update dependencies

jcoyne:


Gemfile.lock
```diff
@@ -84,7 +84,7 @@ GEM
       rails (>= 4.2, < 6)
       rsolr (>= 1.0.6, < 3)
       twitter-typeahead-rails (= 0.11.1.pre.corejavascript)
-    blacklight-hierarchy (1.1.0)
+    blacklight-hierarchy (2.0.0)
```
jmartin-sul: this worries me slightly, just because BL hierarchy has been squirrely in the past.  but the change set from 1.1.0 to 2.0.0 looks small, so if there has been a bit of manual testing and the hierarchical facets still appear to work, i'm happy to merge.  i'm not super-well versed in BL, so the BLH 1.1.0 to 2.0.0 changes aren't very intelligible to me at a glance.

PR #1170, opened 2018-10-30
Automatically manage versions for APO bulk update

### LINKED ISSUE #987: Bulk "Update governing APO" should open objects not opened instead of failing

tingulfsen:
Fixes #987


tingulfsen:
Superseded by #1172

PR #1172, opened 2018-10-31
APO bulk update: manage versions

### LINKED ISSUE #987: Bulk "Update governing APO" should open objects not opened instead of failing

tingulfsen:
Fixes #987


jcoyne:
👍 LGTM.  @jmartin-sul do you want to add anything?

tingulfsen:
Given all the testing, the conversations I've had with @jmartin-sul and the review above, I'm merging this and cutting a new release of Argo that I'm deploying to stage and production.

PR #1179, opened 2018-11-15
Remove the diagram generation.

jcoyne:
It was discussed at the patch party Nov 2018 that this feature is little used and requires a dependency on GraphViz.  By removing this feature, we can remove the graphviz dependency from dor-services.


blalbrit:
Cool - I thought we got rid of this a long time ago!

mejackreed:
👏 🍾

PR #1181, opened 2018-11-28
Update dor-services to 6.0

jcoyne:


jmartin-sul:
looks like the build is failing due to a dor-services method that didn't get delegated/moved with the calling indexer code that was moved in sul-dlss/dor-services#381:
https://github.com/sul-dlss/dor-services/blob/9e99a703383aa50abe4cf3f2ee61406d6ccf0439/lib/dor/indexers/identifiable_indexer.rb#L65

jcoyne:
@jmartin-sul Yep, the fix is in https://github.com/sul-dlss/dor-services/pull/384

PR #1200, opened 2018-11-30
Fix to suri url env in docker-compose.yml.

justinlittman:


docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
       SOLR_URL: http://solr:8983/solr/argo
       SETTINGS__SOLRIZER_URL: http://solr:8983/solr/argo
       SETTINGS__FEDORA_URL: http://fedoraAdmin:fedoraAdmin@fcrepo:8080/fedora
-      SETTINGS__SURI__URL: http://suri:3000
+      SETTINGS__SURI_URL: http://suri:3000
```
jcoyne: No, the double underscore translates to a separate level of YAML: https://github.com/sul-dlss/argo/blob/master/config/settings.yml#L95

PR #1205, opened 2018-12-03
Only set the Solr url in one place

jcoyne:


jcoyne:
@mjgiarlo we have defaluts set here: https://github.com/sul-dlss/argo/blob/master/config/settings/test.yml#L6

PR #1215, opened 2018-12-05
Added filtering of more facets.

### LINKED ISSUE #201: Improved interactions for Collection and APO facet navigation
These facets have grown to be several hundreds of entries long (over 500), yet is often the first or primary means of limiting queries to the entries you want. Finding the name you want is a challenge that an alphabetized sort would not address, because you many not know/remember that Red-Dennis is "The Reid W. Dennis Collection ...", that it's "Glen McLaughlin Maps Collection -- Maps of Malta" but "The Glen McLaughlin Map Collection of California as an Island", that several collections end "....collection of the Revs Institute". Because of this, we have let the facet display grow arbitrarily large in order use the browser find (cmd-F) to do string matching. This is actually not a bad solution, but we assume could be improved. 

The "more" option built into blacklight is not suitable .. it will only support navigation to leading letters in a sub-divided alphabetical entry. One idea is a simple text entry box with some form of auto-complete behavior, dynamically offering matches based on string match anywhere in the value. Whether this supplements or replaces the full facet listing is a separate question, as the browse of a full list of available collections and APO can be useful sometimes.

justinlittman:
closes #201

### DIFF
--- a/app/views/catalog/facet.html.erb
+++ b/app/views/catalog/facet.html.erb
@@ -1,0 +1,26 @@
+<div class="facet_pagination top">
+  <%= render :partial=>'facet_pagination' %>
+</div>
+
+<div class="modal-header">
+  <button type="button" class="ajax-modal-close close" data-dismiss="modal" aria-hidden="true">×</button>
+
+  <h3 class="modal-title"><%= facet_field_label(@facet.key) %></h3>
+  <%= render partial: 'facet_index_navigation' if @facet.index_range && @display_facet.index? %>
+
+</div>
+<div class="modal-body">
+  <div class="facet_extended_list">
+    <%= render_facet_limit(@display_facet, layout: false) %>
+  </div>
+</div>
+
+<div class="modal-footer">
+
+  <div class="facet_pagination bottom">
+    <div class="col-sm-4">
+      <input type="text" id="filterInput" placeholder="Filter..." class="form-control form-control-sm" />
+    </div>
+    <%= render 'facet_pagination' %>
+  </div>
+</div>
--- a/app/assets/javascripts/argo.js
+++ b/app/assets/javascripts/argo.js
@@ -69,3 +69,24 @@ function validate_spreadsheet_filetype()
     if((filename.lastIndexOf(".xlsx") == -1) && (filename.lastIndexOf(".xls") == -1) &&  (filename.lastIndexOf(".xml") == -1) && (filename.lastIndexOf(".csv") == -1))
         $('span#bulk-spreadsheet-warning').text("Note: Only spreadsheets or XML files are allowed. Please check your selected file.");
 }
+
+// Allows filtering a list of facets.
+function filterList() {
+    const input = document.getElementById('filterInput');
+    const filter = input.value.toUpperCase();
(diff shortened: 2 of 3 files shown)


app/views/catalog/facet.html.erb
```diff
@@ -0,0 +1,46 @@
+<div class="facet_pagination top">
+  <%= render :partial=>'facet_pagination' %>
```
jcoyne: `render 'facet_pagination'` will suffice here.

app/views/catalog/facet.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  </div>
+</div>
+
+<div class="modal-footer">
+
+  <div class="facet_pagination bottom">
+    <div class="col-sm-4">
+      <input type="text" id="filterInput" onkeyup="filterList()" placeholder="Filter..." class="form-control form-control-sm" />
+    </div>
+    <%= render :partial=>'facet_pagination' %>
```
jcoyne: `render 'facet_pagination'` will suffice here.

app/views/catalog/facet.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  </div>
+</div>
+
+<script>
+    function filterList() {
```
jcoyne: put this in `app/assets/javascript` somewhere.

app/views/catalog/facet.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  <div class="facet_extended_list">
+    <%= render_facet_limit(@display_facet, layout: false) %>
+  </div>
+</div>
+
+<div class="modal-footer">
+
+  <div class="facet_pagination bottom">
+    <div class="col-sm-4">
+      <input type="text" id="filterInput" onkeyup="filterList()" placeholder="Filter..." class="form-control form-control-sm" />
```
jcoyne: Can you bind this in the javascript (rather than the html)?

justinlittman:
@jcoyne Yes, this only works on the current page. However, in Argo the `more_limit` is set to a very high value, such that all facet values always appear on the current page.

justinlittman:
Look at jasmine gem.

app/views/catalog/facet.html.erb
```diff
@@ -0,0 +1,35 @@
...
+  <div class="facet_pagination bottom">
+    <div class="col-sm-4">
+      <input type="text" id="filterInput" placeholder="Filter..." class="form-control form-control-sm" />
+    </div>
+    <%= render 'facet_pagination' %>
+  </div>
+</div>
+
+<script>
+$(document).ready(function() {
```
jcoyne: Can we move this code into the asset pipeline (e.g. `app/assets/javascript/....`)? 

Can you use `Blacklight.onLoad` rather than `$(document).ready` (accounts for turbolinks events)

justinlittman: Yes.

jcoyne: would it work if you did `$(document).on('keyup', '#filterInput', function() {} )`?

justinlittman: Yes, that worked. Fixed.

PR #1216, opened 2018-12-05
Use metadata object labels instead of fedora object label in edit collection dialog

### LINKED ISSUE #1006: "Edit Collection" collection title drawn from wrong source data
Reported by a user, noticed when using the "edit collection" functionality, the collection title seems to be drawing from the Fedora label rather than descMetadata or the other defaults (identityMD or DC).

See below:

Object to be modified:
screenshot 2017-08-15 08 56 26

Incorrect collection title in modal:
screenshot 2017-08-15 08 56 36

Collection item view showing correct title (also verified in identityMD and DC):
screenshot 2017-08-15 08 56 03

View of Fedora label:
screenshot 2017-08-15 08 56 12

(see for instance, any object associated with the collection druid:jt898xc8096 - https://argo.stanford.edu/catalog?page=1&per_page=100&q=Chinese+Non-Profit+Organizations&utf8=%E2%9C%93 ).

peetucket:
add a helper for pulling the title out of metadata instead of just using the fedora label; use in edit collection modal

fixes #1006

### DIFF
--- a/app/helpers/value_helper.rb
+++ b/app/helpers/value_helper.rb
@@ -2,6 +2,15 @@
 
 module ValueHelper
   # Renderers
+
+  # Given a fedora object, fetch the preferred object title from MODs, if not found, notify Honeybadger and return Fedora label
+  # @return [String] object title, as pulled from descMetadata
+  def object_title(obj)
+    return obj.descMetadata.title_info.first.strip if obj.descMetadata.title_info.first.present?
+    Honeybadger.notify("No title found in descMetadata for collection #{obj.pid}")
+    obj.label
+  end
+
   def label_for_druid(druid)
     druid = druid.to_s.split(/\//).last # strip "info:fedora/"
     Rails.cache.fetch("label_for_#{druid}", expires_in: 1.hour) do
--- a/app/views/items/_collection_ui.html.erb
+++ b/app/views/items/_collection_ui.html.erb
@@ -6,7 +6,7 @@
     <ul class='list-group'>
       <% @object.collections.each do |col| %>
         <li class='list-group-item'>
-          <%= col.label %>
+          <%= object_title(col) %>
           <%= link_to url_for(:controller => :items, :action => :remove_collection, :id => @object.pid, :collection => col.pid) do %>
             <span class='glyphicon glyphicon-remove text-danger'></span>
           <% end %>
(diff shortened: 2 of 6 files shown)


app/helpers/value_helper.rb
```diff
@@ -2,6 +2,18 @@
 
 module ValueHelper
   # Renderers
+
+  # Given a fedora object, fetch the preferred object title
+  # @return [String] object title, as pulled from metadata, given the following order of preference:
+  #  1. MODS title, 2. identityMetadata label, 3. DC title, 4. Fedora label
+  def object_title(obj)
+    obj_title = obj.descMetadata.title_info.first.strip if obj.descMetadata.title_info.first.present?
```
jcoyne: Can you just return here, rather than doing the `||=` below?

peetucket: yup, something like this?

```    
return obj.descMetadata.title_info.first.strip if obj.descMetadata.title_info.first.present?
obj_title = obj.identityMetadata.objectLabel.first.strip if obj.identityMetadata.objectLabel.first.present?
obj_title ||= obj.DC.title.first.strip if obj.DC.title.first.present?
obj_title ||= obj.label
obj_title```

Comment (jcoyne) [reply app/helpers/value_helper.rb]: yes, but then also on the next 3 lines too.

Comment (peetucket) [reply app/helpers/value_helper.rb]: ahh right, fixed

File: app/helpers/value_helper.rb [hunk 2]
```diff
@@ -2,6 +2,17 @@
 
 module ValueHelper
   # Renderers
+
+  # Given a fedora object, fetch the preferred object title
+  # @return [String] object title, as pulled from metadata, given the following order of preference:
+  #  1. MODS title, 2. identityMetadata label, 3. DC title, 4. Fedora label
+  def object_title(obj)
+    return obj.descMetadata.title_info.first.strip if obj.descMetadata.title_info.first.present?
```
jcoyne: is this different than `obj.mods_title`?

jcoyne: I'm sort of against adding all these conditions here as Ben is saying:

> Collection items should always have a MODS title, the other datastreams are there as backup in case something goes awry with creation of the descMetadata datastream.

This seems to indicate the other situations are exceptional.  Should we raise an error or log this (perhaps in Honeybadger.notify)?

peetucket: 1. Didn't realize there was a method for that.  I looked into it and discovered that `obj.mods_title` is part of the `Editable` module in dor-services, which is only included in APO type objects.  Not sure why that is, but that method is not available to collections (or items for that matter).

2. Let me check with Ben on your idea for just throwing an exception when there is no title in MODs instead of all the conditionals.

peetucket: Ben liked this idea - implemented now and rebased/pushed.  Note that I decided to return the fedora label after notifying honeybadger so at least the app can continue for the user (with a potentially non-useful fedora label).  Asked Ben to confirm this strategy as well.

PR #1221, opened 2018-12-11
 Remove active workflows when object is purged

### LINKED ISSUE #1020: "Purge" action should also clean up workflow table
Currently, "purge" removes a registered (not accessioned) object from DOR. However, it leaves the register workflow for that object in the Oracle workflow table.

Expected behavior for "purge" would be to also clean up the workflow table while removing the object from DOR

mjgiarlo:
Fixes #1020

In so doing:

* Stop stubbing the item pid with a bogus pid—this appears to add nothing other than a needless stub
* Use argo core for testing—this allows developers run the test suite with the given docker-compose configuration, simplifying setup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ email: false
 before_install:
   - docker pull suldlss/workflow-server:latest
   - docker run -d -p 127.0.0.1:3001:3000 suldlss/workflow-server:latest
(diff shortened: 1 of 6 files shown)


mjgiarlo:
@peetucket @jcoyne Would you mind taking another look at this PR? I made some additional changes to 1) get the build passing, and 2) clarify local development documentation.

PR #1222, opened 2018-12-12
Use dor-services-app to register new objects

### LINKED ISSUE #1212: Use dor-services-app for registration
Rather than RegistrationService.  We should check to see if dor-services-app can support registration of collections too.

mjgiarlo:
Do this instead of using `Dor::RegistrationService` within the `dor-services` gem, which creates a service seam and reduces the reach and impact of `dor-services`.

Note that this makes a small change to an integration spec that used to test two side-effects of `Dor::RegistrationService` that are no longer as easy to test with the new service boundary. We may want to revisit this integration spec if we want the additional assurance this test affords us, say, if dor-services-app is dockerized and added as a dependency of the dev/test environments for Argo.

Fixes #1212

### DIFF
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -16,7 +16,7 @@ class CollectionForm < BaseForm
 
   # Copies the values to the model and saves and indexes
   def save
-    register_model
+    @model ||= register_model
     sync
     model.save
     model.update_index
@@ -37,9 +37,9 @@ class CollectionForm < BaseForm
 
   # @return [Dor::Collection] registers the Collection
   def register_model
-    response = Dor::RegistrationService.create_from_request(register_params)
+    response = DorServices::Client.register(params: register_params)
     # Once it's been created we populate it with its metadata
-    @model = Dor.find(response[:pid])
+    Dor.find(response[:pid])
   end
 
   # @return [Hash] the parameters used to register an apo. Must be called after `validate`
--- a/spec/integration/apo_spec.rb
+++ b/spec/integration/apo_spec.rb
@@ -6,6 +6,8 @@ RSpec.describe 'apo', type: :request, js: true do
   let(:user) { create(:user) }
   let(:new_apo_druid) { 'druid:zy987wv6543' }
   let(:new_collection_druid) { 'druid:zy333wv6543' }
+  let(:new_apo) { Dor::AdminPolicyObject.new(pid: new_apo_druid) }
+  let(:new_collection) { Dor::Collection.new(pid: new_collection_druid) }
 
   after do
     Dor::AdminPolicyObject.find(new_apo_druid).destroy # clean up after ourselves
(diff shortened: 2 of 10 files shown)


jcoyne:
@mjgiarlo can you stub out `Dor.find(response[:pid])` on the integration test?  Would that be enough?

mjgiarlo:
> @mjgiarlo can you stub out `Dor.find(response[:pid])` on the integration test? Would that be enough?

This is ready for another :eyes:, @jcoyne.

app/forms/collection_form.rb
```diff
@@ -37,9 +37,9 @@ def sync
 
   # @return [Dor::Collection] registers the Collection
   def register_model
-    response = Dor::RegistrationService.create_from_request(register_params)
+    response = DorServices::Client.register(params: register_params)
     # Once it's been created we populate it with its metadata
-    @model = Dor.find(response[:pid])
+    Dor.find(response[:pid])
```
jcoyne: Why do we need to find this?

spec/integration/apo_spec.rb
```diff
@@ -48,7 +58,12 @@
     expect(page).to have_selector('.permissionName', text: 'someone')
     expect(find_field('desc_md').value).to eq('MODS')
     expect(find_field('use_license').value).to eq('by-sa')
-    expect(page).to have_link('New Testing Collection')
+    # TODO: This was removed when we switched from using
```
jcoyne: If we split this test into a "create" test and an "edit" test, could we restore this assertion?

spec/integration/apo_spec.rb
```diff
@@ -6,20 +6,29 @@
...
-    allow(Dor::SuriService).to receive(:mint_id).and_return(new_apo_druid, new_collection_druid)
+    allow_any_instance_of(ApoForm).to receive(:register_model).and_return(new_apo)
+    allow_any_instance_of(CollectionForm).to receive(:register_model).and_return(new_collection)
     allow(ApoController).to receive(:update_index).with(any_args)
     sign_in user, groups: ['sdr:administrator-role']
   end
 
   it 'creates and edits an apo' do
-    expect(Dor::Config.workflow.client).to receive(:create_workflow).twice
+    # TODO: This was removed when we switched from using
```
jcoyne: I don't think this is going to improve the tests. Can we just remove this?

app/forms/collection_form.rb
jcoyne: Why are we doing this find?

spec/integration/apo_spec.rb
mjgiarlo: Plausible, for sure. I'll give it a shot.

app/forms/collection_form.rb
mjgiarlo: So that it is assigned to `@model` in the calling method (which the form uses throughout). I moved the assignment out of this method to make it easier to test.

spec/integration/apo_spec.rb
mjgiarlo: I would be happy to ditch it. Coming soon.

app/forms/collection_form.rb
jcoyne: Ah, I see now! 👍

app/forms/collection_form.rb
```diff
@@ -16,7 +16,7 @@ def validate(params)
 
   # Copies the values to the model and saves and indexes
   def save
```
jcoyne: Why are we doing the `||=`. Is there a case where `@model` is already set?

mjgiarlo: Yes, it can be set on form initialization per `BaseForm`.

spec/integration/apo_spec.rb
mjgiarlo: Not happy with the attempt I made. Still too tangly. Inclined to pass on it for now, and revisit when dor-services-app is dockerized.

mjgiarlo:
@jcoyne ready for another look.

PR #1229, opened 2018-12-13
Use correct route helper

jcoyne:
Also handle format extension


jcoyne:
I added a couple of feature tests, so we shouldn't hit this error again.

PR #1235, opened 2018-12-14
Call the dor-services-app for content data

jcoyne:


peetucket:
looks like a rebase is needed here

jcoyne:
@peetucket rebased

PR #1238, opened 2018-12-14
Fix the route to FilesController#preserved

jcoyne:


mjgiarlo:
Approved, but not merged bc of one failing build (which I just restarted).

PR #1241, opened 2018-12-17
Use dor-services-app container & docker-compose for testing 

mjgiarlo:
And remove JDK from Travis since it is no longer needed (was for spinning up Fedora/Solr before switching to Docker)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,17 +8,17 @@ addons:
 rvm:
   - 2.5.3 # deployed
 
-jdk:
-  - oraclejdk8
-
 email: false
 
 before_install:
-  - docker pull suldlss/workflow-server:latest
-  - docker run -d -p 127.0.0.1:3001:3000 suldlss/workflow-server:latest
-  - docker run -d -p 8984:8983 -v $PWD/solr_conf/conf/:/myconfig solr:7 solr-create -c argo -d /myconfig
-  - docker run -d -p 8983:8080 suldlss/fcrepo:no-messaging-latest
-  - docker ps -a
+  # Travis's docker-compose is too old to handle v3.6 docker-compose templates.
+  # This code is from Travis's documentation: https://docs.travis-ci.com/user/docker/#using-docker-compose
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
+  - docker-compose up -d solr fcrepo dor-services-app suri workflow
+  - docker-compose ps
   - google-chrome-stable --headless --disable-gpu --no-sandbox --remote-debugging-port=9222 http://localhost &
 
 before_script:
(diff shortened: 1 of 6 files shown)


mjgiarlo:
Rebasing.

.travis.yml
```diff
@@ -8,17 +8,15 @@ addons:
 rvm:
   - 2.5.3 # deployed
 
-jdk:
-  - oraclejdk8
```
jcoyne: 👏

.travis.yml
```diff
@@ -8,17 +8,15 @@ addons:
...
-
 email: false
 
 before_install:
-  - docker pull suldlss/workflow-server:latest
-  - docker run -d -p 127.0.0.1:3001:3000 suldlss/workflow-server:latest
-  - docker run -d -p 8984:8983 -v $PWD/solr_conf/conf/:/myconfig solr:7 solr-create -c argo -d /myconfig
-  - docker run -d -p 8983:8080 suldlss/fcrepo:no-messaging-latest
-  - docker ps -a
+  - sudo rm /usr/local/bin/docker-compose
```
jcoyne: Why do we need to remove the system version?  Can you add a comment to explain the reason?

docker-compose.yml
```diff
@@ -17,6 +17,23 @@ services:
...
+      SETTINGS__CONTENT__SDR_SERVER_URL: http://dor-services-app:3000/v1/sdr
+      SETTINGS__METADATA__CATALOG_URL: http://dor-services-app:3000/v1/catalog/mods
+
+  dor-services-app:
+    image: suldlss/dor-services-app:latest
+    ports:
+      - 3003:3000
+    environment:
+      SOLR_URL: http://solr:8983/solr/argo
+      SETTINGS__SSL__CERT_FILE: /app/spec/support/certs/spec.crt
```
jcoyne: I wonder if we can set these SSL variables as defaults within the Dockerfile in dor-services-app, so we don't have to copy these settings every time we want to run it.

spec/integration/apo_spec.rb
```diff
@@ -6,18 +6,20 @@
...
   after do
     Dor::AdminPolicyObject.find(new_apo_druid).destroy # clean up after ourselves
     Dor::Collection.find(new_collection_druid).destroy # clean up after ourselves
   end
 
   before do
-    allow_any_instance_of(ApoForm).to receive(:register_model).and_return(new_apo)
-    allow_any_instance_of(CollectionForm).to receive(:register_model).and_return(new_collection)
-    allow(ApoController).to receive(:update_index).with(any_args)
+    # Inject pid into dor-services client's params so the destroys above work
```
jcoyne: This sentence is missing a noun.

jcoyne: I wonder if it's now possible to run this test without hardcoding pids.  Shouldn't dor-services-app and suri just give us one?

.travis.yml
mjgiarlo: Comment added.

mjgiarlo:
@jcoyne Good idea re: defaulting the SSL settings in DSA's docker config. Can you take a look at https://github.com/sul-dlss/dor-services-app/pull/183 before we merge this PR?

docker-compose.yml
mjgiarlo: Done.

PR #1242, opened 2018-12-17
Rename argo to es6 because the const keyword was added in es6

jcoyne:


jcoyne:
Superceded by https://github.com/sul-dlss/argo/pull/1244

PR #1243, opened 2018-12-17
Changes fields used to look for associated items.

### LINKED ISSUE #449: Navigation from one collection member to all collection members
When a user goes to an individual object, it would be nice to have two options:
1. the user can link to the collection object (this currently exists)
2. the user can link to all objects in the collection to which the object belongs (new feature) – this is currently a two-step process and would improve speed moving between individual objects and all of its fellow children.

transferred from https://jirasul.stanford.edu/jira/browse/ARGO-488

justinlittman:
closes #449


justinlittman:
@jcoyne Test added.

PR #1248, opened 2018-12-17
bulk remove all collections if none is selected

### LINKED ISSUE #926: Reconcile bulk "Add Collection" and "Set Collection" with item "Edit Collection"
These are confusing options that behave differently for one object vs many. Would like to simply provide the proper collection info and have the system add it if missing else replace the collection reference(s) that's already there, or offer add/replace choice uniformly across both facilities.

peetucket:
fixes #926 

see latest comments, where the desire is that if the user selects "None" for the collection when bulk setting the collection, it should be removed from items without setting a new one (this actually already happens, but the user gets an error message when a collection with no druid is attempted to be added each time)

### DIFF
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -419,6 +419,10 @@ RSpec.describe ItemsController, type: :controller do
         expect(@item).to receive(:add_collection).with('druid:1234')
         post 'add_collection', params: { id: @pid, collection: 'druid:1234' }
       end
+      it 'does not add a collection if no param supplied' do
+        expect(@item).not_to receive(:add_collection).with('druid:1234')
+        post 'add_collection', params: { id: @pid, collection: '' }
+      end
     end
 
     context "when they don't have manage_content access" do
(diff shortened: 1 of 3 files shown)


spec/controllers/items_controller_spec.rb
```diff
@@ -445,6 +449,15 @@
         expect(response.code).to eq('200')
       end
 
+      it 'remove a collection if a new one is not selected' do
```
jcoyne: missing an 's' on "removes"

jcoyne: Can you put this in a context block?

e.g. `context 'when a new collection is not selected' do ...`

spec/controllers/items_controller_spec.rb
```diff
@@ -419,6 +419,10 @@
         expect(@item).to receive(:add_collection).with('druid:1234')
         post 'add_collection', params: { id: @pid, collection: 'druid:1234' }
       end
+      it 'does not add a collection if no param supplied' do
```
jcoyne: can you put this in a context block?
e.g. :
```
context 'when no parameter is supplied' do
  it 'does not add a collection' do
    ...
  end
end
```

peetucket:
updates pushed

PR #1252, opened 2018-12-18
always put default workflow first in list regardless of APO

### LINKED ISSUE #1251: Always show registrationWF (default workflow) in reg drop down list regardless of what's in the APO
@blalbrit  indicates that regardless of what is in the APO, we should show the default workflow first in the drop-down list on the registration form.

peetucket:
fixes #1251

### DIFF
--- a/spec/controllers/registration_controller_spec.rb
+++ b/spec/controllers/registration_controller_spec.rb
@@ -284,13 +284,6 @@ RSpec.describe RegistrationController, type: :controller do
   end
 
   describe '#workflow_list' do
-    it 'should handle an APO with a single default workflow, putting the default workflow first always' do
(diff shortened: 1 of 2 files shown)


spec/controllers/registration_controller_spec.rb
```diff
@@ -284,21 +284,18 @@
   end
 
   describe '#workflow_list' do
-    it 'should handle an APO with a single default workflow' do
+    it 'should handle an APO with a single default workflow, putting the default workflow first always' do
```
jcoyne: This isn't really a "single default workflow" anymore is it.  Do we still need this test?

peetucket: yeah, i guess not, basically the same as the second test now, i'll remove

PR #1256, opened 2018-12-19
create collection modal should not render layout from an ajax call

peetucket:
* creating a collection makes an ajax call which should not render the layout (else you get JS classes when all of the JS tries to load again)
* also hide the cancel button from an ajax call to be consistent with other modals, since you already have a close modal button and the cancel would have forced a full page reload

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -5,6 +5,9 @@ class CollectionsController < ApplicationController
   def new
     @apo = Dor.find params[:apo_id]
     authorize! :manage_item, @apo
+    respond_to do |format|
+      format.html { render layout: !request.xhr? }
+    end
   end
 
   def create
--- a/app/views/collections/new.html.erb
+++ b/app/views/collections/new.html.erb
@@ -59,8 +59,10 @@
       <%= button_tag 'register', id: 'register', class: 'btn btn-default' do 'Register Collection' end %>
     </div>
     <% end %>
-    <div class="form-group">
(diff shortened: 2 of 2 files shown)


jcoyne:
So this is sometimes modal content and sometimes not? Can you add a comment about where each case originates?

peetucket:
I couldn't find any other reference to this page other than this specific location (at least by searching for the named route of `new_apo_collection`), but if you disable javascript, this allows it to degrade more gracefully.

jcoyne:
Why do we try to support JavaScript being off? Clearly that wouldn’t work with the registration grid amongst other features. How do we decide that some features should have a non-JavaScript fallback? Which users  population does that represent?

justinlittman:
I don't think the non-JS fallback is the point. This makes it consistent with how other modals on the page are served and get rids of the JS error that required "Create collection" to be pressed twice.

jcoyne:
@justinlittman I'm not nit picking the change. I'm just wondering why we have the conditional for `!request.xhr?` Can't we just do `layout: false`?

peetucket:
i decided to just matched the same style that was used in all other the similar "blue button" modals in the item controller, e.g. https://github.com/sul-dlss/argo/blob/master/app/controllers/items_controller.rb#L692-L696

I guess I'd vote for consistency for now (those other modals work with no JS on, even if that is not a stated requirement).  Though I know turning of JS breaks other stuff (like item registration) so its not all totally consistent either.

PR #1257, opened 2018-12-19
Hide Metadata Source drop-down menu in item registration; default to "Auto"

peetucket:
addresses #1258

### DIFF
--- a/app/assets/javascripts/registration/grid.js
+++ b/app/assets/javascripts/registration/grid.js
@@ -274,7 +274,7 @@ var gridContext = function() {
         cellsubmit: 'clientArray',
         colModel: [
           {label:' ',name:'status',index:'status',width:18,sortable:false,formatter: statusFormatter },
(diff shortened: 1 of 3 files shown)


peetucket:
Note this PR changes a label on the Registration page grid from "Metadata ID" to the more appropriate "Catkey" (which is all it can be):

BEFORE
 
Screen Shot 2019-09-04 at 4 12 10 PM


AFTER

Screen Shot 2019-09-04 at 4 13 06 PM

peetucket:
@andrewjbtw : small change to the registration for simplification... as there is no longer a choice in "Metadata source", we propose to eliminate that drop-down menu and relabel the "Metadata ID" column to "Catkey" (which is the only identifier you can put in there anyway).

andrewjbtw:
That sounds good to me. Go ahead and make the change. 

I don't know if there will ever be a reason to have multiple metadata sources again, but we can address that if it happens later. By that point, the registration form may already have had a bigger redesign anyway.

PR #1261, opened 2018-12-20
Warn if collection with duplicate title or catkey is being created.

### LINKED ISSUE #927: Prevent (or warn) if duplicate collection object is being created
Problem: We don't currently check for uniqueness when registering a collection object from a catkey
Desired solution: Warning (or prevention) that a duplicate collection is being created
Description (from Doris):

Also, could there be an alert if one person tries to create a collection that already exists? Astrid just mentioned to me (she's reading over my shoulder) that she and Geoff have done this before, it's come up for me as well.

justinlittman:
closes #927

### DIFF
--- a/app/assets/javascripts/argo.js
+++ b/app/assets/javascripts/argo.js
@@ -76,3 +76,26 @@ function filterList() {
 }
 
 $(document).on('keyup', '#filterInput', function(e) { filterList() });
+
+// Provide warnings when creating a collection.
+function collectionExistsWarning(warningElem, field, value) {
+    var client = new XMLHttpRequest();
+    client.onreadystatechange = function() {
+        if (this.readyState == 4 && this.status == 200) {
+            if (this.responseText == 'true') {
+                warningElem.style.display = "block";
+            } else {
+                warningElem.style.display = "none";
+            }
+        }
+    };
+    client.open("GET", '/collections/exists?' + field + '=' + value, true);
+    client.send();
+}
+$(document).on('keyup', '#collection_title', function(e) {
+    collectionExistsWarning(document.getElementById('collection_title_warning'), 'title', e.target.value);
+});
+
+$(document).on('keyup', '#collection_catkey', function(e) {
+    collectionExistsWarning(document.getElementById('collection_catkey_warning'), 'catkey', e.target.value);
+});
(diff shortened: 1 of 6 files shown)


justinlittman:
Before merging, I'd like @blalbrit to approve:

app/assets/javascripts/argo.js
```diff
@@ -76,3 +76,26 @@ function filterList() {
 }
 
 $(document).on('keyup', '#filterInput', function(e) { filterList() });
+
+// Provide warnings when creating a collection.
+function collectionExistsWarning(warningElem, field, value) {
+    const client = new XMLHttpRequest();
```
jcoyne: Const is introduced in es6, so you can only use it in an es6 file. Otherwise Uglifier will raise an error when you try to deploy.

app/assets/javascripts/argo.js
```diff
@@ -76,3 +76,26 @@ function filterList() {
...
+    client.onreadystatechange = function() {
+        if (this.readyState == 4 && this.status == 200) {
+            if (this.responseText == 'true') {
+                warningElem.style.display = "block";
+            } else {
+                warningElem.style.display = "none";
+            }
+        }
+    };
+    client.open("GET", `/collections/exists?${field}=${value}`, true);
```
jcoyne: string interpolation is also an es6 thing

app/controllers/collections_controller.rb
```diff
@@ -23,4 +23,15 @@ def create
     redirect_to solr_document_path(params[:apo_id]), notice: "Created collection #{collection_pid}"
     @apo.save # indexing happens automatically
   end
+
+  def exists
+    resp = if params[:title].present?
```
jcoyne: Do we need to consider the case where title is present and unique and catkey is present and not unique?

peetucket: good question for Ben, but I'd guess a non-unique catkey should trigger a validation problem regardless of title uniqueness

jcoyne:
@justinlittman I wonder if this should be two different endpoints in the controller?  One is for "is title unique" and one is for "is catkey unique" as you ought to never call `#exists?` with both parameters

app/controllers/collections_controller.rb
justinlittman: Because of how the form is constructed, you can only add one or the other.

app/assets/javascripts/argo.js
justinlittman: For my education, what indicates an ES6 file?

justinlittman:
@jcoyne Rather than add additional endpoints, I made the `exists` endpoint able to handle both parameters, in addition to one parameter at a time.

PR #1285, opened 2019-01-11
when creating a bulk action job, prefix druid: to each pid if it does not have one yet

### LINKED ISSUE #676: Bulk Actions "Paste a druid list" Requires Druid Prefix - Need to make this consistent with other "Paste a druid list" which allow both prefixed and non-prefixed druids
Situation:

User pastes in list of druids for download in Bulk Actions. Download fails without druid: prefix. Succeeds if druid: prefix included.

Desired behavior:

Exactly like every other "paste a druid list" screen in the app - allow both prefixed and non-prefixed druid lists.

peetucket:
fixes #676 

- [x] needs a test

### DIFF
--- a/app/controllers/bulk_actions_controller.rb
+++ b/app/controllers/bulk_actions_controller.rb
@@ -20,7 +20,7 @@ class BulkActionsController < ApplicationController
   def create
     # Since the groups aren't persisted, we need to pass them here.
     @bulk_action = BulkAction.new(
-      bulk_action_params.merge(user: current_user, groups: current_user.groups)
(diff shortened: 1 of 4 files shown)


jcoyne:
Rather than having this conditional in the model, is it possible to prevent this bad data (pids without namespaces) from getting to the model level in the first place?  I would like to see something like this in a Form object (think Reform, see ApoForm) or baring that, in the controller.

peetucket:
OK, moved logic to controller and updated test to reflect its new location (a form object for one method seemed potentially overkill).  Can squash commits if this approach makes more sense.

PR #1288, opened 2019-01-16
add new facet for earliest accessioned date

peetucket:


app/models/concerns/document_date_concern.rb
```diff
@@ -3,13 +3,14 @@
...
-  FIELD_REGISTERED_DATE       = :registered_dttsim
-  FIELD_LAST_ACCESSIONED_DATE = :accessioned_latest_dttsi
-  FIELD_LAST_PUBLISHED_DATE   = :published_latest_dttsi
-  FIELD_LAST_SUBMITTED_DATE   = :submitted_latest_dttsi
-  FIELD_LAST_DEPOSITED_DATE   = :deposited_latest_dttsi
-  FIELD_LAST_MODIFIED_DATE    = :modified_latest_dttsi
-  FIELD_LAST_OPENED_DATE      = :opened_latest_dttsi
+  FIELD_REGISTERED_DATE           = :registered_dttsim
+  FIELD_LAST_ACCESSIONED_DATE     = :accessioned_latest_dttsi
+  FIELD_EARLIEST_ACCESSIONED_DATE = :accessioned_earliest_dttsi
```
peetucket: note...this is the only line added, everything else is just re-aligning the spacing

jcoyne:
Argo is pretty slow to draw pages, and I've heard that a lot of this is because the faceting is expensive for solr.  Do we know what the impact of adding this will be?  How much performance degradation are we willing to accept in order to add this feature?

peetucket:
I am not sure what the impact will be of adding yet more more facet -- we could do some smoke testing on argo-stage and have @blalbrit provide his opinion.

blalbrit:
Happy to smoke test before we go all out on this. We would definitely want to add this to the facets that are buried behind the "more facets" button so that it doesn't slow initial load.

peetucket:
Ben has smoked tested this branch on argo-stage and approved.  My best guess on performance (no hard data though) is that give its just another solr query to run, its probably going to decrease performance in only a marginally linear way - i.e. if 20 facets make the page take 30 seconds to load, then going to 21 facets will make it take an extra second or so.

PR #1289, opened 2019-01-17
New manage catkey bulk action

### LINKED ISSUE #984: Add bulk action for "Manage Catkey"
Quite often, objects need to be associated with a catkey after the fact. This enhancement would allow a user to add catkeys to a list of objects produced either by a search or a pasted druid list.

The desired action would be to:

1. add catkey to identityMD if missing
2. change or delete existing catkey as needed
3. do NOT automatically refresh descMD from source (the association with catkey could be for purely organizational needs, not descMD needs), be offer an option to do so?

Normal prepare/close version process should handle the republish process. 

This would be an action that requires versioning.

As @LynnMcRae has pointed out, this would be great to consider from a pattern point of view - there will be many examples of desired added/altered tuples across datastreams beyond just identityMetadata and otherId="catkey".

peetucket:
fixes #984 

- add new Manage Catkey Job class
- update bulk actions UI to allow new job to be submitted
- write new tests for manage catkey job class
- move methods that we being used by multiple jobs (such as open_new_version, can_manage?, etc) into the generic job class so they aren't duplicated in jobs that need them
- move tests related to these shared methods into the generic_job_spec
- update rubocop_todo to reflect new location of these tests (did not add to rubucop violations, but did not cleanup either when moving tests)


spec/jobs/generic_job_spec.rb
```diff
@@ -52,4 +52,83 @@ def perform(_bulk_action_id, _params)
...
   end
+
+  describe '#open_new_version' do
+    let(:current_user) do
+      instance_double(User,
+                      is_admin?: true)
+    end
+
+    before do
+      @dor_object = double(pid: 'druid:123abc')
```
jcoyne: Can you switch these to the `let()` syntax rather than instance variables? http://www.betterspecs.org/#let

app/views/bulk_actions/_form.html.erb
```diff
@@ -60,6 +65,12 @@
           Reindexes the DOR object to Solr
         </span>
       </div>
+      <div role='tabpanel' class= 'tab-pane' id='manage_catkey_options'>
+        <span class='help-block'>
+          Adds or updates catkeys associated with objects.  You need two lists: druids and catkeys, and they need to correspond 1:1 in the correct order.
+        </span>
+        <%= render 'bulk_actions/forms/manage_catkey_form', f: f %>
```
jcoyne: Why are you passing `f` and then not using it in the partial?

.rubocop_todo.yml
```diff
@@ -167,6 +168,7 @@ RSpec/BeforeAfterAll:
     - 'spec/support/**/*.rb'
     - 'spec/jobs/descmetadata_download_job_spec.rb'
     - 'spec/jobs/modsulator_job_spec.rb'
+    - 'spec/jobs/generic_job_spec.rb'
```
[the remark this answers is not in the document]
peetucket: nope, will remove

app/views/bulk_actions/_form.html.erb
peetucket: oops

spec/jobs/generic_job_spec.rb
peetucket: yes, will do

peetucket:
What kind of test were you thinking of for BulkActionsController?  Only new thing is permitting the catkey text box params through

PR #1297, opened 2019-01-28
Update dependencies

sul-devops-team:


Gemfile.lock
```diff
@@ -198,7 +198,7 @@ GEM
       unf (>= 0.0.5, < 1.0.0)
     dor-rights-auth (1.3.0)
       nokogiri
-    dor-services (6.1.14)
+    dor-services (6.2.0)
```
justinlittman: Any reason to be nervous about this change?

jcoyne: nope, we follow semver.

PR #1307, opened 2019-01-29
Only display the workflow status from the most recent version

jcoyne:
Also:
* Extract a presenter for the workflow processes

### DIFF
--- a/app/presenters/workflow_presenter.rb
+++ b/app/presenters/workflow_presenter.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class WorkflowPresenter
+  def initialize(object:, workflow_name:, xml:)
+    @object = object
+    @workflow_name = workflow_name
+    @xml = xml
+  end
+
+  delegate :pid, to: :object
+  attr_reader :workflow_name
+
+  # This is going to find the workflow definition:
+  #   Dor::WorkflowObject.find_by_name(workflowId.first) and get the steps from there.
+  #   then it will fill in process status with that found in the xml.
+  # @return [Array<Dor::Workflow::Process>]
+  def processes
+    return [] if workflow_document.nil?
+
+    workflow_document.processes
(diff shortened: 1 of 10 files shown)


app/presenters/workflow_presenter.rb
```diff
@@ -0,0 +1,42 @@
...
+
+class WorkflowPresenter
+  def initialize(object:, workflow_name:, xml:)
+    @object = object
+    @workflow_name = workflow_name
+    @xml = xml
+  end
+
+  delegate :pid, to: :object
+  delegate :processes, to: :workflow_document
```
mjgiarlo: It looks like you're both delegating and defining `#processes`. Won't one make the other dead code?

jcoyne: Ah, good call. I thought I removed this. I suspect it's just noop, because the other one overwrites this.

mjgiarlo:
@jcoyne 	restarted the failing build. Feel free to merge when all :green_heart:

app/presenters/workflow_presenter.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def pretty_xml
+    # rubocop:disable Rails/OutputSafety
+    CodeRay::Duo[:xml, :div].highlight(Nokogiri::XML(xml).prettify).html_safe
+    # rubocop:enable Rails/OutputSafety
+  end
+
+  private
+
+  attr_reader :object, :xml, :workflow_steps
```
mjgiarlo: Are the private attrs for `:object` and the `:workflow_steps` used, or are they cruft?

jcoyne: used on lines 15 & 24 respectively

mjgiarlo: :eyes: 	:roll_eyes:  :zzz:

PR #1316, opened 2019-02-01
Avoid deprecated WorkflowDs#[]

jcoyne:
Also changed the response for trying to activate an already activated workflow
to a 403 instead of a 500

### DIFF
--- a/app/controllers/workflows_controller.rb
+++ b/app/controllers/workflows_controller.rb
@@ -60,12 +60,13 @@ class WorkflowsController < ApplicationController
       end
     end
     wf_name = params[:wf]
-    wf = @object.workflows[wf_name]
+
     # check the workflow is present and active (not archived)
-    if wf&.active?
-      render status: 500, plain: "#{wf_name} already exists!"
+    if workflow_active?(wf_name)
+      render status: 403, plain: "#{wf_name} already exists!"
       return
     end
+
     Dor::CreateWorkflowService.create_workflow(@object, name: wf_name)
 
     # We need to sync up the workflows datastream with workflow service (using #find)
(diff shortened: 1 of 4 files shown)


app/controllers/workflows_controller.rb
```diff
@@ -91,6 +92,14 @@ def history
...
   private
 
+  # Fetches the workflow from the workflow service and checks to see if it's active
+  def workflow_active?(wf_name)
+    xml = Dor::Config.workflow.client.get_workflow_xml('dor', @object.pid, wf_name)
+    ng_xml = Nokogiri::XML(xml)
+    result = ng_xml.at_xpath("/workflow/process[@version=#{@object.current_version}]")
+    result ? true : false
+  end
+
```
peetucket: could you do this to avoid having to directly parse workflow XML outside of the workflow service? ```Dor::Config.workflow.client.get_active_workflows('dor', @object.pid).include? wf_name```

jcoyne: I don't think so, because that depends on `archived` behavior, which is something we removed from the new workflow service.

peetucket: ok

peetucket: how about moving this into a new workflow service method ?  don't feel strongly enough to block progress on this PR, but seems cleaner and removes specific XML knowledge from a consumer

mjgiarlo: Ideally the workflow client could answer this question, yeah. I too don't care too strongly.

jcoyne: I think that sounds good.  Earlier I was hoping we could move all the workflow models there, but I'm waiting on https://github.com/sul-dlss/workflow-server-rails/issues/102, first. However, I think it wouldn't be difficult to model the response.

PR #1341, opened 2019-03-13
[Draft] Avoid looking in solr to draw the milestones

jcoyne:
This attempts to reduce the amount of things that we need to index which
ought to greatly reduce the load on the workflow service.

Before we proceed down this path, we need to devise a way of showing the workflow grid just for certain pids. For example: https://argo.stanford.edu/report/workflow_grid?f%5Bnonhydrus_collection_title_ssim%5D%5B%5D=Papyrus+fragments+%28papyrology+collection%29%2C+circa+200+BC+-+200+AD


jcoyne:
Superseded by #2098

PR #1351, opened 2019-04-03
Upgrade to dor-workflow-service 2.11

jcoyne:


justinlittman:
Give me a shout when green.

PR #1355, opened 2019-04-09
Remove unused bulk indexing code

jcoyne:
None of this is useful as it was all based on risearch,
which required the triple store in fedora to be enabled.
See https://github.com/sul-dlss/argo/issues/380


jmartin-sul:
historical note, just to lessen confusion for future readers of this PR...

it is true that risearch didn't work (good tip that it's the triplestore, definitely didn't realize that, but i've never been super familiar with fedora configuration or internals).  but we did use this machinery heavily for reindexing to the new solr schema in the argo 2 to 3 transition in early 2016.  by way of the hack snarkily described in a now excised portion of the docs:  https://github.com/sul-dlss/argo/pull/1355/files#diff-8c4163152cc2d2dd95f4bb46512506dcL60 (and the dump scripts also referenced in that doc)

anyway, totally agree this code has outlived its usefulness, and that it makes sense to get rid of it and use sul-mq or the (not so new anymore) bulk actions UI in the web app.  just wanted to clarify a bit.

PR #1364, opened 2019-04-16
Move the ontology services from dor-services

jcoyne:
Argo is the only consumer of these services
Ref https://github.com/sul-dlss/dor-services/issues/563


jcoyne:
These services are needed within the DefaultObjectRightsDS for APOs

PR #1366, opened 2019-04-17
Remove unused method fix_missing_provenance

jcoyne:


jcoyne:
Yes, I discussed with him on slack this morning:

> I don’t believe we have used this from the Argo side - perhaps ever. There is a semi-persistent problem of objects making it to accessioning without a required datastream (usu. versioning or descMetadata) - in which case being able to add that datastream from Argo would be useful.
> I don’t think provenance metadata has been one of those datastreams - at least not in > 5 years.

mjgiarlo:
Thanks, @jcoyne

PR #1370, opened 2019-04-23
Remove unused create_minimal_mods route

jcoyne:
This was introduced https://github.com/sul-dlss/argo/commit/8e68e66c775761085496a60e6f72d7b3175f7df7
but that button has since been removed


justinlittman:
Congratulations. You've eliminated HELL.

PR #1385, opened 2019-05-02
Remove unused routes

jcoyne:
Using the `traceroute` gem to audit the routes


jcoyne:
@justinlittman no routes have been added.

PR #1395, opened 2019-05-07
Update the modsulator_job to use the Dor::StatusService as Dor::Item.…

aaron-collier:
…status_info has been removed

Fixes https://github.com/sul-dlss/modsulator-app-rails/issues/65

### DIFF
--- a/app/jobs/modsulator_job.rb
+++ b/app/jobs/modsulator_job.rb
@@ -289,7 +289,7 @@ class ModsulatorJob < ActiveJob::Base
   # @param  [Dor::Item]   dor_object    DOR object to check
   # @return [Boolean]     true if the object is currently being accessioned, false otherwise
   def in_accessioning(dor_object)
-    status = dor_object.status_info[:status_code]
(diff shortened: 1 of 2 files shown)


app/jobs/modsulator_job.rb
```diff
@@ -299,7 +299,15 @@ def in_accessioning(dor_object)
...
-    status = dor_object.status_info[:status_code]
+    status = get_status(dor_object) # dor_object.status_info[:status_code]
     [1, 6, 7, 8, 9].include?(status)
   end
+
+  # Returns the status_info for a DOR object from the StatusService
+  #
+  # @param  [Dor::Item]   dor_object    DOR object to check
+  # @return [Integer]     value cooresponding to the status info list
+  def get_status(dor_object)
```
jcoyne: can we call this `status` or `status_info`?  the current name violates https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/AccessorMethodName

aaron-collier: Renamed to `status`

PR #1401, opened 2019-05-09
Use token auth for connecting to dor-services-app

jcoyne:


config/settings.yml
```diff
@@ -71,6 +71,8 @@ DOR_SERVICES:
   URL: 'http://localhost:3003'
   USER: dor-service-user
   PASS: dor-service-password
+  TOKEN: 'secret-token'
+  TOKEN_HEADER: 'X-Auth'
```
justinlittman: Will this always be X-Auth or is that a dummy value?

jcoyne: It just has to match whatever is in dor-services-app: https://github.com/sul-dlss/dor-services-app/blob/master/app/controllers/application_controller.rb#L12-L14

justinlittman: Can you add a comment to that effect?

PR #1404, opened 2019-05-10
Database cleaner is not necessary since Rails 5.1 implemented system …

jcoyne:
…tests


spec/spec_helper.rb
```diff
@@ -60,20 +60,20 @@ def instantiate_fixture(druid, klass = ActiveFedora::Base)
...
-    end
-    DatabaseCleaner.start
-  end
-
-  config.after do
-    DatabaseCleaner.clean
-  end
+  config.use_transactional_fixtures = true
+
+  # config.before do
```
justinlittman: Cruft?

PR #1410, opened 2019-05-14
Use the workflow server to start workflow

jcoyne:
The dor-services-app endpoint is deprecated


jcoyne:
@justinlittman it's not green because the dor-services-app docker image was updated and was incompatible with the assumptions we had earlier.

PR #1411, opened 2019-05-15
Fix to docker-compose for latest images.

justinlittman:


jcoyne:
@justinlittman do you want to take out the first commit?

justinlittman:
@jcoyne I merged #1410 and then rebased.

PR #1423, opened 2019-06-04
Add a notification when we're off the nominal path

jcoyne:


jcoyne:
I'm not sure what you want that's not already in the code.  It says here this is a fallback: https://github.com/sul-dlss/argo/blob/5907f957267e4d851ae4113b58fdfe9a356dbcd8/app/controllers/registration_controller.rb#L47

I don't want to "fallback" without knowing about it.

PR #1448, opened 2019-08-05
Avoid an extra Fedora lookup

jcoyne:
Pass the looked up object to render_buttons().  Previously it was passing an undefined ivar

### DIFF
--- a/app/views/catalog/_show_external_links.html.erb
+++ b/app/views/catalog/_show_external_links.html.erb
@@ -1,5 +1,3 @@
-<% document ||= @document %>
-
 <div class="show_sidebar">
   <h3 class="start-open">View in new window</h3>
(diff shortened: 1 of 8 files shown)


app/views/catalog/_show_external_links.html.erb
```diff
@@ -15,5 +13,5 @@
     <li><small><%= render_index_info document %></small></li>
   </ul>
   <hr>
-  <%= render(:partial => 'items/button_link_list', :locals => {:buttons => (render_buttons document, @obj)})%>
+  <%= render(:partial => 'items/button_link_list', :locals => {:buttons => (render_buttons document, object)})%>
```
jcoyne: This is the most important change in this PR.

PR #1449, opened 2019-08-05
Always call render-buttons with two arguments

jcoyne:


app/views/catalog/_show_sidebar.html.erb
```diff
@@ -1,2 +1,2 @@
-<%= render :partial => 'catalog/show_external_links', :locals => { :document => @document, :object => @object } %>
+<%= render :partial => 'catalog/show_external_links', :locals => { :document => @document, :object => @obj } %>
```
jcoyne: This wasn't using the right ivar previously.

app/views/catalog/_show_external_links.html.erb
```diff
@@ -15,5 +13,5 @@
     <li><small><%= render_index_info document %></small></li>
   </ul>
   <hr>
-  <%= render(:partial => 'items/button_link_list', :locals => {:buttons => (render_buttons document, @obj)})%>
```
jcoyne: Now that correct ivar is being passed in we can use the local variable.

PR #1453, opened 2019-08-06
Blacklight 7

jcoyne:


astridu:
image
1. Search button has a slightly darker background behind the magnifying glass (purely cosmetic to make it look nicer)
2. Separate each button so it looks more like a button
3. 2 pt radius on all the button corners to make them look more like buttons
4. Bold, white font on active buttons
5. Regular, black font in inactive buttons (so it's accessible)
6. Colors: The blue is from the Stanford palette.
Darkest blue (behind magnifying glass, you can also use this for the hover): 003961
Dark blue (buttons): 00548F
Gray inactive button: C5C5C5

jcoyne:
@astridu is this acceptable?
Screen Shot 2020-04-24 at 4 19 13 PM

astridu:
Nice! the inactive buttons should be gray: C5C5C5 
Thank you.

jcoyne:
@astridu Let's add that improvement as a separate ticket. Currently this is just about the same as we have in production currently. I don't want to add more complexity to this ticket as there's already 1200 lines of code changed here.

Screen Shot 2020-04-24 at 4 51 01 PM

app/controllers/catalog_controller.rb
```diff
@@ -161,9 +160,12 @@ class CatalogController < ApplicationController
       }
     }
 
-    config.add_results_collection_tool(:report_view_toggle)
     config.add_results_collection_tool(:bulk_update_view_button)
     config.add_results_collection_tool(:bulk_action_button)
+    config.add_results_collection_tool(:sort_widget)
+    config.add_results_collection_tool(:per_page_widget)
+    # config.add_results_collection_tool(:view_type_group)
```
justinlittman, line 167: Cruft?

jcoyne: This is a default that blacklight 7 generates into the config, so I thought we should leave it in case we want to add that tool back.

PR #1470, opened 2019-08-14
Bump to dor-services-client 2.0.0 and dor-services 7.2.4

mjgiarlo:
And correct `openeable` typo now that DSC uses `openable` instead as of 2.0.0.

Refs sul-dlss/dor-services-app#322


mjgiarlo:
@jcoyne yep! Just pushed a commit that should fix the test suite.

Issue #1346, opened 2019-03-27
Improve workflow step status management


blalbrit:
Summary: Currently, a user has limited ability to manage workflow step statuses in the Argo workflow grid ("set to waiting" in some instances, "set to completed" in some instances). However, there are instances where greater control is desired and/or necessary: re-instantiating and entire workflow; set a series of steps to completed; re-enqueuing a single step; etc.

This is a story ticket (no implementation suggestions - would defer to @astridu about where and how to provide hooks) requesting Argo-based self-service (possibly limited to certain users?) to, for individual items and in bulk:
*  set workflow to completed (single step or multiple steps)
* re-instantiate entire workflow
* set status on workflow step to waiting/skipped/completed individually and in bulk.

There are existing API calls in the dor-services-app to accomplish all of these tasks.

blalbrit:
@vivnwong  ^^

ndushay:
@astridu can you also provide a design for an individual object?

Easiest would be to add the same pulldown to this view, I suppose:

![image](https://user-images.githubusercontent.com/96775/62584516-cfddb080-b869-11e9-8390-3b185a996f7e.png)

ndushay:
I'm pretty sure we need to enable different messages to be sent to the workflow service, and we have to change the UI to trigger the correct action when pulldown is selected.  We might want to have user confirm when "skip" or "complete" is selected.  Maybe start with a search on "reset" in the argo code.

https://github.com/sul-dlss/argo/blob/2b1df709450fcbc0f80611e512c41e99f2d10e52/app/controllers/report_controller.rb#L63

@jcoyne is this the latest, greatest way to update a workflow status?

jcoyne:
@blalbrit can you confirm that re-instantiate entire workflow or set an entire workflow to completed are important parts of this ticket? I'm not seeing how the designs proposed address those parts.

blalbrit:
@jcoyne - the use-cases to-date have been:
* need to completely re-instantiate an entire workflow
* granularly control status on a workflow step (set to waiting, completed, skipped)
* granularly control status on multiple workflow steps in succession

I'm not sure we've had a case of needing to set an entire workflow to completed, though I can imagine scenarios where that would be a nice-to-have.

ndushay:
@blalbrit so we do, or don't want to do this in workflow_grid view as well as single object specific workflow view?

blalbrit:
[...]

PR #1473, opened 2019-08-15
[Needs PO approval] Add dropdown to set a status for an errored workflow item (design needed - blocker)

jcoyne:
***DO NOT MERGE*** without approval from Product Owner. 

Ref #1346

Ref #1516

### DIFF
--- a/app/presenters/workflow_process_presenter.rb
+++ b/app/presenters/workflow_process_presenter.rb
@@ -30,15 +30,27 @@ class WorkflowProcessPresenter
 
   attr_reader :view, :process_status
 
-  delegate :form_tag, :item_workflow_path, :hidden_field_tag, :button_tag, :options_for_select, :select_tag, to: :view
+  delegate :form_tag, :item_workflow_path, :hidden_field_tag, :button_tag, :content_tag, :options_for_select, :select_tag, to: :view
   delegate :pid, :workflow_name, to: :process_status
+
+  CONFIRM_MESSAGE = 'You have selected to manually change the status. ' \
+    'This could result in processing errors. Are you sure you want to proceed?'
 
   def error_choices
     # workflow update requires id, workflow, process, and status parameters
     form_tag item_workflow_path(pid, workflow_name), method: 'put' do
       hidden_field_tag('process', name) +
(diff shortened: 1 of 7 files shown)


ndushay:
This looks good but I think it needs to be deployed to stage and taken for a spin by @andrewjbtw or someone before it gets merged to master.  Also, before and after pix would be great here ...

(I'm psyched to see this functionality get added, btw!)

app/presenters/workflow_process_presenter.rb
```diff
@@ -23,25 +17,40 @@ def elapsed
...
-    form_tag item_workflow_path(pid, workflow_name), method: 'put' do
-      hidden_field_tag('process', name) +
-        hidden_field_tag('status', new_status) +
-        button_tag('Set to ' + new_status, type: 'submit')
+  def status_action
+    case status
+    when 'error'
+      error_choices
+    when 'waiting'
+      completed_button('completed')
```
ndushay: could this also be 'skipped'?  I'm not clear on the difference ...

jcoyne: I'm not sure I understand.  This is the existing functionality for "waiting".  In https://github.com/sul-dlss/argo/issues/1346#issuecomment-521484228 we agreed the ticket was only affecting rows with the "error" status.

ndushay:
Houston, I think we have a problem:  argo-stage shows this:

image

what is the "assign" button??

Also -- I suspect this will also need to be in a feature branch.

jcoyne:
Assign is the submit button. As in assign this status

ndushay:
Up to @andrewjbtw to sign off on UI, then.  Thanks for explanation.

andrewjbtw:
The functionality looks good, but I would like to revisit the UI for this. We definitely want a mechanism to confirm the choice, but I was initially confused by the "Assign" label.

Also, depending on the layout of the workflow view, the button is sometimes on the right of the dropdown and sometimes below it. This seems to be related to how word wrapping applies to the process names on the left. Example from druid:hx447jk4615

Screen Shot 2019-08-16 at 9 22 37 AM

ndushay:
I would like one more set of eyes on this.

PR #1475, opened 2019-08-19
Update dependencies

sul-devops-team:


jmartin-sul:
approved pending green PR build.  if it fails again w/ the same error (i.e. doesn't look like something transient), will remove approval and investigate.

PR #1482, opened 2019-08-19
Remove unnecessary activejob spec helper

jcoyne:


mjgiarlo:
@jcoyne Not entirely clear how your changes broke the build in such a way.

PR #1483, opened 2019-08-19
Switch to sassc as the pure ruby sass gem is deprecated

jcoyne:


mjgiarlo:
Approving despite flappy spec failure. Restarting build as well.

PR #1485, opened 2019-08-20
Put the identifier in the route for close version

jcoyne:


config/routes.rb
```diff
@@ -144,8 +145,6 @@
   end
 
   namespace :items do
-    post 'version/close',  action: :close_version, as: 'close_version'
-    post 'version/open',   action: :open_version,  as: 'open_version'
```
ndushay: Just making sure -- is this going to remove argo routes?  If so, we many need to ensure no one uses them ...

jcoyne: I don't believe these are used now that we have the member routes above.

mjgiarlo: I searched through Argo and sul-dlss and could find no code using these endpoints.

PR #1486, opened 2019-08-21
Convert close to a async background job

jcoyne:
Ref #1466


andrewjbtw:
This looks good to me. I've run a few tests - including opening, modifying, and closing 1000 objects in a batch - and it's working as expected.

PR #1489, opened 2019-08-23
Extract VersionsController from ItemsController

jcoyne:


justinlittman:
@jcoyne It would be helpful to have a screenshot indicating what part of the UI this affects. (I understand that it won't change the UI.)

Issue #1491, opened 2019-08-26
add missing permissions checks 


ndushay:
from @jmartin-sul 
>I think these jobs probably should run the GenericJob#can_manage?(ActiveFedora::Base) check (inherited from the parent class, some jobs already do this, see the writeup in argo 1478):
- [x] close_version_job.rb
- [x] prepare_job.rb
- [x] modsulator_job.rb
- [x] virtual_merge_job.rb

>Also maybe:
- [ ] descmetadata_download_job.rb should check can_view?(ActiveFedora::Base) (but prob doesn't need to authorize management on the objects in the job).

ndushay:
Leaving open per last bullet point in description.

PR #1520, opened 2019-08-28
[Needs PO review] Add missing permissions checks to background jobs

jcoyne:
Ref #1491


mjgiarlo:
@jcoyne 	There are some conflicts to resolve on this PR now.

ndushay:
I think it's also worth doing a smoke test in stage -- impersonate a group that isn't enabled, make sure it correctly doesn't allow stuff;  impersonate a group that is enabled, make sure it does allow stuff.

Also, since this may affect argo users that aren't in the right groups, I think we should  do a PO review @andrewjbtw

app/jobs/virtual_merge_job.rb
```diff
@@ -1,20 +1,30 @@
 # frozen_string_literal: true
 
 ##
-# job to merge objects
+# job to merge objects and then close the version
 class VirtualMergeJob < GenericJob
   queue_as :default
 
   def perform(_bulk_action_id, parent_druid:, child_druids:)
+    return unless authorized_to_manage?(parent_druid)
```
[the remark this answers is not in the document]
andrewjbtw: Now that I understand better what's happening behind the scenes with virtual objects, I think if the user has permission to manage the parent, then it shouldn't be necessary to check the child objects. So my answer is no.

andrewjbtw:
I believe this is working correctly, from what I've been able to determine on stage. I think we should go ahead and watch carefully to see if users experience a problem running background jobs.

app/jobs/close_version_job.rb
```diff
@@ -10,11 +10,10 @@ class CloseVersionJob < GenericJob
   # @param [Integer] bulk_action_id GlobalID for a BulkAction object
   # @param [Hash] params additional parameters that an Argo job may need
   # @option params [Array] :pids required list of pids
-  # @option params [Array] :groups the groups the user belonged to when the started the job. Required for permissions check (can_manage?)
+  # @option params [Array] :groups the groups the user belonged to when the started the job. Required for because groups are not persisted with the user.
```
mjgiarlo: `s/for because/because/`

app/jobs/modsulator_job.rb
```diff
@@ -16,18 +16,21 @@ class ModsulatorJob < ActiveJob::Base
   # @param  [String]  apo_id             DRUID of the DOR APO that governs all of the objects we're trying to upload metadata for.
   # @param  [String]  uploaded_filename  Full path to the temporary uploaded file. Deleted upon completion.
   # @param  [String]  output_directory   Where to store output (log, generated XML etc.).
-  # @param  [String]  user_login         Acting user's username.
+  # @param  [User]    user               Acting user
   # @param  [String]  filetype           One of 'xml, 'spreadsheet', or 'xml_only'. If not 'xml', the input is to be loaded into the datastream.
   # @param  [String]  note               An optional note that the user entered to go with the job.
   # @return [Void]
-  def perform(apo_id, uploaded_filename, output_directory, user_login, filetype = 'spreadsheet', note = '')
+  def perform(apo_id, uploaded_filename, output_directory, user, groups, filetype = 'spreadsheet', note = '')
```
mjgiarlo: Do we also need to add `groups` as a `@param` above?

app/jobs/modsulator_job.rb
```diff
@@ -124,13 +127,13 @@ def generate_original_filename(uploaded_filename)
 
   # Write initial job information to the log file.
   #
-  # @param [File]    log_file  The log file to write to.
-  # @param [String]  username  The login name of the current user.
-  # @param [String]  filename  The name of this job's input file.
-  # @param [String]  note      An optional comment that describes this job.
-  def start_log(log_file, username, filename, note = '')
+  # @param [File] log_file The log file to write to.
+  # @param [user] username The login name of the current user.
```
mjgiarlo: I think this should be `@param [User] user` instead of `@param [user] username`?

mjgiarlo:
Up to @jcoyne 	whether he wants to act on my code review. Nothing big there. Merge when ready.

ndushay:
I'm gonna merge this to get it out, but leave @mjgiarlo and @jcoyne to do a followon PR for stuff Mike mentioned.

Issue #1428, opened 2019-06-07
(uber ticket) Newly created collections not showing up in "Set Collection" or "Edit Collection" drop-downs


blalbrit:
#1490 investigation is needed to determine the root cause 

**Describe the bug**
Some (all?) newly created collections are not showing up in the list of collections for the Bulk Update "Set Collection" or Item View "Edit Collection" tasks.

**User Impact**
A user (in this case @hannahfrost ) is not able to add a newly created collection to items that were not formerly associated with a collection (see below for details).

**To Reproduce**
Steps to reproduce the behavior (for Bulk Update):
1. Go to https://argo.stanford.edu/catalog?f%5Bis_governed_by_ssim%5D%5B%5D=info%3Afedora%2Fdruid%3Apw482xr1298
2. Click on 'Bulk Update'
3. Click on 'Get druids from search'
4. Scroll down to 'Set Collection'
5. Scroll down to the drop-down list of collections and look for 'Stanford University, Department of Physics, senior honors theses' (https://argo.stanford.edu/view/druid:ds182hd6984)

(for Item View):
1. Go to https://argo.stanford.edu/view/druid:bm962wh8682
2. Click on 'Edit Collections'
3. From the 'Add a Collection' drop-down, look for 'Stanford University, Department of Physics, senior honors theses' (https://argo.stanford.edu/view/druid:ds182hd6984)

**Expected behavior**
The collection 'Stanford University, Department of Physics, senior honors theses' (https://argo.stanford.edu/view/druid:ds182hd6984) should appear in either/both drop-down lists.

**Screenshots**
Let us know if you need screenshots for any of the above.

cc. @hannahfrost

blalbrit:
See https://app.honeybadger.io/projects/49894/faults/50335590

mgolson-stanford:
**Describe the bug:**  We recently created a new collection object for the Shikamura collection https://argo.stanford.edu/view/druid:my748dz4281.  This newly created collection object is NOT showing up in the collection drop down and we are unable to associate digitized objects with this collection or release them to SearchWorks.

**User Impact:** A user (@mgolson) is unable to add a newly created collection to items that were not formerly associated with a collection.  This is preventing us from associating digital objects with the new Shikamura collection object and releasing this content to SearchWorks.

**To Reproduce:** (for bulk update)

1.  Go to https://argo.stanford.edu/catalog?f%5Bexploded_tag_ssim%5D%5B%5D=DPG+%3A+Shikamura
[...]

PR #1528, opened 2019-08-29
Increased the number of collections returned in query

justinlittman:
As there are now > 1000. refs #1428.


ndushay:
it would be funny in a sad way if this is the problem in #1428.   

Also - didn't the uber ticket mention more than one place where coll records might be missing?  Does this PR cover all those places?

justinlittman:
@ndushay To answer your question: based on @peetucket's research, they both use the same code (fixed in this PR).

PR #1539, opened 2019-08-29
Upgrade jquery3

jcoyne:
The problem we saw earlier where modals were broken has been fixed.

### DIFF
--- a/app/assets/javascripts/jquery.persistentModal.js
+++ b/app/assets/javascripts/jquery.persistentModal.js
@@ -1,18 +1,18 @@
 /*global Blacklight */
 /*
-  usage:  
+  usage:
     if you want to have a link load the contents of the link target in a persistent modal, add the attribute 'data-behavior="persistent-modal"'.
     e.g.:
       <a title="DC" data-behavior="persistent-modal" href="/view/druid:dy196vh8233/ds/DC">DC</a>
     or, if you're building a link from ruby, something like:
       link_to specs[:dsid], ds_solr_document_path(doc['id'], specs[:dsid]), :title => specs[:dsid], :data => { behavior: 'persistent-modal' }
 
-  the modal is persistent in that subsequent invocations of the modal on the same overall page load will load the same retained 
+  the modal is persistent in that subsequent invocations of the modal on the same overall page load will load the same retained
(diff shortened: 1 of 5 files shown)


ndushay:
So should we smoke test this on stage before merging to master?  LGTM ... but so did the broken PR.

jcoyne:
@ndushay sure.  I smoke tested the other one too, just missed looking at datastreams.

app/assets/javascripts/jquery.persistentModal.js
```diff
@@ -50,30 +50,30 @@
...
         }
       }
 
-      function receiveAjax(data) {      
+      function receiveAjax(data) {
+
         //default to error text, replace if we got something back
         var contents = "Error retrieving content";
         if (data.readyState !== 0) {
-          contents = data.responseText;
```
ndushay: this looks ... different

app/assets/javascripts/jquery.persistentModal.js
```diff
@@ -82,8 +82,7 @@
 
       function createModal() {
         var jqxhr = $.ajax({
-          url: linkTarget,
-          dataType: 'script'
```
ndushay: this is different too ...

app/assets/javascripts/registration/register.js
```diff
@@ -77,28 +77,28 @@ function DorRegistration(initOpts) {
         beforeSend: function(xhr) {
           $t.setStatus(data, 'pending')
         },
-        dataType: 'json'
+        dataType: 'json',
+        success: function(response,status,xhr) {
```
ndushay: I see this is ordered / grouped differently.

jcoyne: No, it's refactored because xhr no longer has a `success` method.

app/assets/javascripts/jquery.persistentModal.js
jcoyne: Yes, it's asking for HTML, but it was saying "script" which would attempt to eval the results as javascript.

app/assets/javascripts/jquery.persistentModal.js
jcoyne: Yes, the jQuery API changed.

PR #1541, opened 2019-08-30
fix modsulator bulk upload jobs

### LINKED ISSUE #1540: MODs Bulk Job update failing in Argo
See https://app.honeybadger.io/projects/49894/faults/54082542

Users report stalled jobs, but that MODs appear to be updated.

Perhaps related to recent updates in versioning in argo.

peetucket:
fixes #1540

Could @jcoyne and/or @mjgiarlo please review?


peetucket:
its not clear to me why the tests here https://github.com/sul-dlss/argo/blob/master/spec/services/apply_mods_metadata_spec.rb didn't find this problem.  At least one of the specs seems to check for the logging message, which is where the bug would be.

PR #1552, opened 2019-09-05
[Needs PO Approval] Create virtual objects in the UI

### LINKED ISSUE #1463: Create one or more virtual objects as a bulk action (formerly "virtual merge")
Add GUI part for virtual-merge async bulk action. The bulk action will be named "Create virtual object(s)"

GUI should consist of a single "Import CSV" button. See #1511 (but ignore the other form fields in that design).

The CSV should be parsed on the Argo side and Argo should then use dor-services-client to push structured JSON to dor-services-app. Dor-services-app will return either a 204 if all virtual objects were created without error, or a 422 if there were any errors. The 422 will include JSON that provides information on which objects failed and why.

Output to the user will continue to use the same log file as other bulk actions (so argo will display object counts (total, success, reject) just like other bulk actions (e.g. reindex).) Note: it is known that the argo user will need to refresh to see progress on bulk actions.

mjgiarlo:
Connects to #1463

Here's how the UI looks in this PR:

Screenshot from 2019-09-06 14-41-55

When there are errors doing virtual merge, the log now looks like this:

```
2019-09-05 22:51:39 UTC Starting VirtualMergeJob for BulkAction 17
2019-09-05 22:51:40 UTC Creating virtual object druid:hj185vb7593 failed because one or more objects are not combinable: 2 objects could not be combined: Item druid:kv840rx2720 is dark and Item druid:xb482bw3979 is dark
2019-09-05 22:52:09 UTC Finished VirtualMergeJob for BulkAction 17
```


app/jobs/generic_job.rb
```diff
@@ -70,7 +70,6 @@ def string_to_boolean(string)
 
   def update_druid_count
     bulk_action.update(druid_count_total: pids.length)
-    bulk_action.save
```
mjgiarlo: `#save` is not needed because `#update` persists the object.

README.md
```diff
@@ -58,12 +57,40 @@ docker-compose build web
...
-## Load and index records
+### Load and index records
 
 ```
 docker-compose run --rm web rake argo:repo:load
 ```
 
+## Run the servers (for development)
+
+Spin up the services Argo depends upon, but *not* Argo itself; you will run Argo outside of containerland for maximum agility.
```
jcoyne: You certainly don't have to do it this way.  This seems *way* more complicated to me.  Can this be a different commit?

mjgiarlo:
@jcoyne Here you go. I guess I'm wondering how much work is worth doing to make one of eight bulk actions look **completely** different from every other one. Will those changes make creation of virtual objects as a bulk action more or less confusing? cc: @ndushay @andrewjbtw @astridu 
 
Screenshot from 2019-09-06 08-23-28

ndushay:
I think it's up to @andrewjbtw re:  using Astrid's design with the correct text.  We already need to pull out the "populate with search" button;  we also spoke about including the warning text in Astrid's design about "order of child druid matters."

We knew this was a diff animal than other bulk actions ... is this the PR to bite the bullet and take care of it?  I would say we probably don't want to merge this to master (or at least hold off deployment) without addressing the UI further.

README.md
ndushay: I'm glad for this info, but agree a separate PR would be better.

README.md
```diff
@@ -94,26 +121,6 @@ Dor.find("druid:pv820dk6668").destroy
...
-```
-
-where `pp000pp0000` is the druid for the virtual composite object and the child objects are `cc111cc1111` and `cc222cc2222`.  More info, including object requirements: <https://consul.stanford.edu/display/chimera/Virtual+Object+Combinator+I>
-
-For help:
-
-```bash
-bundle exec ./bin/virtual-merge -h
-```
-
```
ndushay: I would prefer we remove this as a last step, after all the virtual merge pieces are there and we're satisfied.  I can make a separate ticket, or you can just make a PR with "[blocked by virtual merge bulk action]" or some such.

Rakefile
```diff
@@ -6,4 +6,4 @@
 require File.expand_path('../config/application', __FILE__)
 require 'rake'
 
-Argo::Application.load_tasks
+Rails.application.load_tasks
```
ndushay: 🎉

config.ru
```diff
@@ -3,4 +3,4 @@
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path('../config/environment', __FILE__)
-run Argo::Application
+run Rails.application
```
ndushay: 🎉

README.md
mjgiarlo: @jcoyne If there's an easier way to spin up a fully functional Argo that immediately reflects changes in `app/`, I am all ears. When running the web app in containerland, I recall having to restart and/or rebuild to see changes which is less than ideal.

README.md
mjgiarlo: :+1: to making a separate ticket to tearing out the CLI after the UI is confirmed sufficient. I'll restore that stuff in this PR.

mjgiarlo:
@ndushay :speech_balloon: 
> I think it's up to @andrewjbtw re: using Astrid's design with the correct text. We already need to pull out the "populate with search" button; 

I remember we touched on this in storytime, but it was before I had any connection to this issue or how this stuff was implemented. (This is why I am finding it difficult to "connect" to the storytime discussions; that is on me!)

Now that I see how it's all wired together, I am wondering why we need to pull out the "populate with search" button, @andrewjbtw. That's not pushback -- I just want to understand the rationale, mostly because I found that button really helpful when working on this feature. (Granted, that is a very different context from a normal usage context.)

> we also spoke about including the warning text in Astrid's design about "order of child druid matters."

That's already included. Blow up the screenshot and you can see it up in the area where every other bulk action's documentation is rendered.
 
> We knew this was a diff animal than other bulk actions ... is this the PR to bite the bullet and take care of it? I would say we probably don't want to merge this to master (or at least hold off deployment) without addressing the UI further.

It is a diff animal, but I guess my point is that it doesn't necessarily need to be unless we want it to be and are OK with adding another day's (or whatever) worth of work to this issue.

ndushay:
@mjgiarlo the order of child objects matters, thus "populate with search" cannot be used ... because order matters.   Think of an atlas with 300 individual map objects.  

Up to @andrewjbtw , but the text about child objects will serve the users better if it's by the input to child objects.  I think we sorta decided that "bulk actions" are gonna be straying a bit from the original notion and maybe the UI code needs to be more flexible.  Checksum report is less of a strain, but also will only have one result for a bunch of ids.

mjgiarlo:
@ndushay :speech_balloon: 
> @mjgiarlo the order of child objects matters, thus "populate with search" cannot be used ... because order matters. Think of an atlas with 300 individual map objects.

So you're suggesting that users don't arbitrarily glom together these virtual objects? But... that's what **I** do! OK, I'm convinced. On that now. Thanks, and sorry this discussion had to drag on past storytime. :smiley:

mjgiarlo:
This is deployed to stage and awaiting testing by @andrewjbtw

mjgiarlo:
@jcoyne @ndushay 	can you both pls re-review? (Note, do not merge until @andrewjbtw has given us the go-ahead.)

mjgiarlo:
After chatting with @ndushay @andrewjbtw and @jcoyne, I am going to take another crack at virtual merge. Instead of the current form + single-line CSV input, we're going all in on CSV import (of one-liners or batches). Closing this PR.

PR #1564, opened 2019-09-06
Use the default remove icon

jcoyne:
This means less custom code in Argo that we need to support.  I think this looks nicer too.

Before:

Screen Shot 2019-09-06 at 11 01 46 AM

After:

Screen Shot 2019-09-06 at 11 03 19 AM


andrewjbtw:
Looks good to me.

PR #1566, opened 2019-09-06
[Needs PO approval] New bulk action should use a pulldown -- redo

jcoyne:
This reverts commit 48c9fee3eeaccdff0c7ee1edc693b8c021706615 which itself was a revert.


andrewjbtw:
Reviewed on stage, looks good to me.

ndushay:
So this is the original PR, but now approved?  I approve.

PR #1567, opened 2019-09-06
Remove flappy spec

jcoyne:
Added additional assurances this is working via the controller spec


ndushay:
nice - you didn't remove it -- you addressed it.

PR #1573, opened 2019-09-06
Improve refresh metadata err handling

### LINKED ISSUE #1545: Improve user experience if refresh_metadata gets an error in symphony response
(part of uber ticket sul-dlss/common-accessioning#188)

https://github.com/sul-dlss/dor-services-app/pull/358 means dor-services-app will return an HTTP 500 error if it receives the wrong number of bytes from Symphony:

after hitting "Refresh Metadata" button in left "nav" for an existing item:

### Argo UI:

image

This is probably not the best user experience.  Something super simple: keep the same error page, but change the text
- to have the specific error message:  "Dor::Services::Client::UnexpectedResponse (Internal Server Error: 500 (Incomplete response received from Symphony for 111 - expected 1960 bytes but got 1965))", where 111 is the ckey.
  - could add the druid
  - could "sweeten" the error message for an end user
  - etc.

ndushay:
Fixes #1545;  The vanilla rails error experience here in Argo, if dor-services-app has trouble retrieving metadata from Symphony, did not give enough information to help with debugging.  This does help and also gives the user a better experience.

## Before

image

## After

image

Note that this is specific to refresh_metadata;  it accommodates UnexpectedResponse errors with any message from dor-services-client, and Andrew approved this "design" here:  https://github.com/sul-dlss/argo/issues/1545#issuecomment-528909307

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ vendor/ruby/
 .ruby-version
 .ruby-gemset
 .byebug_history
+.pry_history
 
 # The docker volume for the workflow database
 postgres-data
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -55,7 +55,7 @@ Metrics/AbcSize:
 # Offense count: 6
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 412
+  Max: 417
 
 # Offense count: 7
 Metrics/CyclomaticComplexity:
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -21,6 +21,8 @@ vendor/ruby/
 .ruby-version
 .ruby-gemset
 .byebug_history
+.pry_history
+jetty/
```
jcoyne: Why is this here?

ndushay: b/c i use pry.  I dunno how jetty got there.

ndushay: haha - jetty dir on my laptop is from 2017.  I can remove from .gitignore if you like.

PR #1591, opened 2019-09-11
Remove unused code 'delete_role'

jcoyne:
dor-services hasn't had Item#delete_role for many years, this code doesn't work
and isn't called from anywhere.


mjgiarlo:
@jcoyne 	Need to remove a reference from the test suite as well.

jcoyne:
@mjgiarlo Thanks. It looks like I forgot to save those changes.  Should be good now.

PR #1593, opened 2019-09-11
Add CSV import of virtual object batches as a bulk action

mjgiarlo:
**DO NOT MERGE**. This PR requires PO approval.

Connects to #1463

# Screenshot

Screenshot from 2019-09-11 10-45-48

# Includes

* Add "Create virtual object(s)" tab to bulk actions
* Hide common fields (populate search results, PIDs list) when switching to
"Create virtual object(s)" tab
* Bump dor-services-client to version that supports new virtual object batch endpoint
* Rename virtual merge-related stuff to "create virtual object" stuff
* Rejigger virtual object-related job to a bulk action job
* Persist CSV within bulk action persister
* Add service for converting CSV to virtual object JSON (expected by dor-services-app)
* Allow bulk actions to write logs within containerland


app/jobs/create_virtual_objects_job.rb
```diff
@@ -0,0 +1,75 @@
...
+# job to create virtual objects
+class CreateVirtualObjectsJob < GenericJob
+  queue_as :default
+
+  ##
+  # A job that creates virtual objects
+  # @param [Integer] bulk_action_id GlobalID for a BulkAction object
+  # @param [Hash] params additional parameters that an Argo job may need
+  # @option params [String] :create_virtual_objects CSV string
+  def perform(bulk_action_id, params)
```
mjgiarlo: This method is indeed gnarly, even gnarlier than it was before I renamed the job. If I have time, I'd like to spend a couple hours refactoring it. I can also ticket this for later.

app/views/bulk_actions/new.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class='container'>
   <h1>New Bulk Action</h1>
 
-  <%= form_for(@bulk_action) do |f| %>
+  <%= form_for(@bulk_action, html: { 'data-controller' => 'bulk_actions' }) do |f| %>
```
mjgiarlo: Move the StimulusJS controller up in scope so its actions can hide the common fields div.

app/assets/javascripts/controllers/bulk_actions.es6
```diff
@@ -8,5 +8,18 @@ export default class extends Stimulus.Controller {
 
     const newTab = this.element.querySelector(`.tab-content > #${selected}`)
     newTab.classList.add('active')
+
+    this.toggleCommonFieldVisibility(selected)
```
ndushay: slick approach - nice.

app/jobs/create_virtual_objects_job.rb
ndushay: refactoring this as a separate ticket seems like a great approach.  Let's get 'er in front of Andrew asap.  It would be so cool to get this deployed!

app/jobs/create_virtual_objects_job.rb
```diff
@@ -0,0 +1,75 @@
...
+      parent_druids = virtual_objects[:virtual_objects].map { |hash| hash[:parent_id] }
+
+      # NOTE: We use this instead of `update_druid_count` because virtual object
+      #       creation does not use the `pids` form field.
+      bulk_action.update(druid_count_total: parent_druids.length)
+
+      # rubocop:disable Metrics/LineLength
+      problematic_druids = find_and_authorize_druids(parent_druids)
+      if problematic_druids[:unauthorized].any? || problematic_druids[:not_found].any?
+        log.puts("#{Time.current} Could not create virtual objects because user lacks ability to manage the following parent druids: #{problematic_druids[:unauthorized].to_sentence}") if problematic_druids[:unauthorized].any?
```
ndushay: I often find putting a bunch of the message into a var helps with long lines like this and makes the code more readable, IMO:

```
not_allowed_msg = "Could not create virtual objects because user lacks ability to manage the following parent druids"
log.puts("#{Time.current} #{not_allowed_msg}: #{problematic_druids[:unauthorized].to_sentence}") if problematic_druids[:unauthorized].any?
```

app/jobs/create_virtual_objects_job.rb
```diff
@@ -0,0 +1,75 @@
...
+      log.puts("#{Time.current} Starting #{self.class} for BulkAction #{bulk_action_id}")
+
+      parent_druids = virtual_objects[:virtual_objects].map { |hash| hash[:parent_id] }
+
+      # NOTE: We use this instead of `update_druid_count` because virtual object
+      #       creation does not use the `pids` form field.
+      bulk_action.update(druid_count_total: parent_druids.length)
+
+      # rubocop:disable Metrics/LineLength
+      problematic_druids = find_and_authorize_druids(parent_druids)
```
ndushay: Is it more normal for a method named `find_and_authorize_druids` to return the approved druids rather than the errors?  `problematic_druids` would be a better name, no?  and if you can memoize, you lose a line of code.

app/jobs/create_virtual_objects_job.rb
```diff
@@ -0,0 +1,75 @@
...
+      if problematic_druids[:unauthorized].any? || problematic_druids[:not_found].any?
+        log.puts("#{Time.current} Could not create virtual objects because user lacks ability to manage the following parent druids: #{problematic_druids[:unauthorized].to_sentence}") if problematic_druids[:unauthorized].any?
+        log.puts("#{Time.current} Could not create virtual objects because the following parent druids were not found: #{problematic_druids[:not_found].to_sentence}") if problematic_druids[:not_found].any?
+        bulk_action.update(druid_count_fail: parent_druids.length)
+        return
```
ndushay: so the code is basically only going to run if ALL the parent druids are ok.   Which is fine as a first pass but I'm wondering if Andrew would prefer that the ok parent druids get run (no idea - I'm ok w whatever he wants.  I can see the value in "we'll either run this csv, or we won't.  do your homework."

app/services/csv_to_json_service.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# Convert CSV to JSON for virtual objects
+class CsvToJsonService
```
jcoyne: I'd like to see this have a less generic name. add something about virtual merge in there.

app/services/bulk_action_persister.rb
```diff
@@ -55,7 +55,15 @@ def job_params
       set_governing_apo: set_governing_apo,
       manage_catkeys: manage_catkeys,
       prepare: prepare,
+      create_virtual_objects: csv_string(create_virtual_objects),
       groups: groups
     }
   end
+
+  def csv_string(params)
```
ndushay: can this be named so its unique purpose is clearer?   Maybe csv_file?  Or, if it's only for create_virtual_objects, indicate THAT in the method name?  and ... you do a File.read here?  Not a File.exists?  I'm not up on my bulk actions, so maybe it makes sense ...

app/views/bulk_actions/forms/_create_virtual_objects_form.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= f.fields_for :create_virtual_objects do |create_virtual_objects_fields| %>
+  <div class='form-group' id="import-csv">
+    <%= create_virtual_objects_fields.label :csv_file, 'Upload a CSV file' %>
+      <small><em>Please note that each row should contain the parent druid as the first field and child druids as all the other fields. Child druids <strong>must</strong> be entered in the order they should appear in the catalog.</em></small>
```
jcoyne: Do you want to say something about "no header rows"?

docker-compose.yml
```diff
@@ -14,6 +14,9 @@ services:
       REMOTE_USER: blalbrit@stanford.edu
       RAILS_LOG_TO_STDOUT: 'true'
       SOLR_URL: http://solr:8983/solr/argo
+      # Allow bulk action logs to be written (can't write to /tmp)
+      SETTINGS__BULK_METADATA__DIRECTORY: '/app/tmp'
+      SETTINGS__BULK_METADATA__TEMPORARY_DIRECTORY: '/app/tmp/tmp'
```
ndushay: okay ... uh ... what?

app/jobs/create_virtual_objects_job.rb
mjgiarlo: @ndushay roger that!

app/services/csv_to_json_service.rb
mjgiarlo: Coming.

app/views/bulk_actions/forms/_create_virtual_objects_form.html.erb
mjgiarlo: Meh. AFAIK, it is not current practice to include header rows.

app/jobs/create_virtual_objects_job.rb
mjgiarlo: Coming!

app/jobs/create_virtual_objects_job.rb
mjgiarlo: I'll make these changes.

app/jobs/create_virtual_objects_job.rb
mjgiarlo: I'm changing this behavior so it runs all non-problematic druids.

app/services/bulk_action_persister.rb
mjgiarlo: It returns a CSV in string form. I can reflect that it's used for virtual objects in the method name. In this case, the file being read is the file that was just uploaded by the user.

docker-compose.yml
mjgiarlo: Until I overrode these values, I didn't see bulk actions successfully write their log files, thus rendering bulk actions inoperable when testing locally using containers. This change only affects local dev and testing.

app/services/csv_to_json_service.rb
mjgiarlo: Done.

app/assets/javascripts/controllers/bulk_actions.es6
mjgiarlo: thx

mjgiarlo:
@ndushay @jcoyne 	I responded to all your excellent code review feedback (either with replies or with code changes).

app/jobs/create_virtual_objects_job.rb
```diff
@@ -0,0 +1,97 @@
...
+      log.puts("#{Time.current} Finished #{self.class} for BulkAction #{bulk_action_id}")
+    end
+    # rubocop:enable Metrics/BlockLength
+  end
+
+  private
+
+  attr_reader :problematic_druids
+
+  def find_problematic_druids!(druids)
```
jcoyne: I don't understand why we need to memoize and have an accessor as I only see this being invoked the one time.  How about just returning the data structure?

mjgiarlo:
This was approved by @andrewjbtw on 9/16/2019 and will be merged to `master` momentarily, and will be deployed to production by Monday at the latest. At the moment, we will "soft launch" this feature, which means it will be available in production but we're not otherwise announcing it until I have a chance to tackle sul-dlss/dor-services-app#336 (adding background job processing to dor-services-app and making the virtual object endpoint use background jobs). cc: @ndushay

PR #1594, opened 2019-09-11
Prevent stale DelayedJob workers from hanging around between deploys

mjgiarlo:


mjgiarlo:
@ndushay It is not clear to me how graceful the `stop` is, but the state now is the jobs are never refreshed (until a reboot or some manual intervention). FWIW, the other project using this configuration is SW indexing, so we'd at least be using a consistent practice, I guess?

mjgiarlo:
Would like other @sul-dlss/infrastructure-team 	folks to chime in on :point_up:

peetucket:
seems reasonable to me

PR #1609, opened 2019-09-13
Convert inline script to a Stimulus controller

jcoyne:


app/components/throbber.html.erb
```diff
@@ -0,0 +1 @@
+<div class="throbber"><div></div><div></div><div></div><div></div><div></div><div></div><div></div><div></div><div></div><div></div><div></div><div></div></div>
```
mjgiarlo: *blinks*

mjgiarlo: @jcoyne 	but srsly, what's going on here? :smiley:

mjgiarlo:
@jcoyne 	can you drop a comment here too about what ActionView::Component is doing for us?

app/components/throbber.html.erb
jcoyne: Those divs are the "dots" in the animated throbber

jcoyne:
@mjgiarlo ActionView::Component is server-side component rendering.  It's by github and it's going to be part of rails 6.1.   It's better than view helpers because it's encapsulated and testable. We're already using it in was-registrar-app.

PR #1612, opened 2019-09-13
Update base Ruby image and add a note to the README about stale images

mjgiarlo:
This is what @peetucket was running into, so we thought it worth documenting.


ndushay:
Some days I hate the pace of change in our field.  It kinda does feel like we're building an airplane in the air.

PR #1619, opened 2019-09-16
Upgrade to dor-services 8

jcoyne:


jcoyne:
Those configs were for elements of dor-services that have been retired. Only dor-services-app should care about such things.

PR #1603, opened 2019-09-12
new bulk action for checksum reporting

peetucket:
fixes #1530 

To Do:
~~Fix new ChecksumReportJob test so it doesn't fail:~~

- [x] add expectation to confirm the pres cat API call is made
- [x] add expectation to confirm the CSV file is written
- [x] remove debuging `raise` statement in ChecksumReportJob class (around L35)

~~Get PO Approval~~

PR #1630, opened 2019-09-18
[WIP - do not merge] get last_accessioned_version from preservation

ndushay:
DO NOT MERGE -- I'm working on the PR that will use the new preservation-client gem instead.

get `last_accessioned_version` from preservation rather than dor-services-app proxying for sdr-services-app.

Note that CodeClimate has our coverage at 97% but it also didn't have coverage stats for helpers.  I tweaked the SimpleCov config to hopefully fix this.

### DIFF
--- a/app/helpers/dor_object_helper.rb
+++ b/app/helpers/dor_object_helper.rb
@@ -56,7 +56,7 @@ module DorObjectHelper
   end
 
   def last_accessioned_version(pid)
-    Dor::Services::Client.object(pid).sdr.current_version
+    PreservationClient.current_version(pid)
   end
 
   def render_qfacet_value(facet_solr_field, item, options = {})
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -58,7 +58,7 @@ DOR_SERVICES:
   TOKEN: eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJGb28ifQ.-BVfLTW9Q1_ZQEsGv4tuzGLs5rESN7LgdtEwUltnKv4
 
 preservation_catalog:
-  url: 'http://PRESCATURL'
+  url: 'https://example.org/prescat'
 
 # URLs
 DOR_INDEXING_URL: 'https://dor-indexing-app.example.com/dor'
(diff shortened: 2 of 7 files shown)


ndushay:
Note to self - verify this after deployed to prod (objects in stage are not properly "preserved" per preservation catalog, so this is impossible to check in stage without heroics).

Reverting is a one liner in dor_object_helper.

app/helpers/dor_object_helper.rb
```diff
@@ -56,7 +56,7 @@ def has_been_accessioned?(pid)
   end
 
   def last_accessioned_version(pid)
-    Dor::Services::Client.object(pid).sdr.current_version
+    PreservationClient.current_version(pid)
```
jcoyne: Should PreservationClient be extracted to a Gem.  This is something dor-services-app could also use.

config/settings.yml
```diff
@@ -51,6 +51,9 @@ DOR_SERVICES:
   # To generate the token: docker-compose run dor-services-app rake generate_token
   TOKEN: eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJGb28ifQ.-BVfLTW9Q1_ZQEsGv4tuzGLs5rESN7LgdtEwUltnKv4
 
+PRESERVATION_CATALOG:
```
jcoyne: Please lowercase these. The config gem documents the correct usage as lowercase: https://github.com/railsconfig/config#embedded-ruby-erb We have lowercase variables already in this file. If we use uppercase the config gem ends up creating methods with the ruby convention for constants, which is confusing.

app/helpers/dor_object_helper.rb
ndushay: Per discussion after standup this morning:  we'll extract it to a gem after we have a second use case.

config/settings.yml
ndushay: will do.  Actually, already done in PR #1603, and will rebase this afterwards.

ndushay:
Closing this in favor of #1658

PR #1643, opened 2019-09-19
DescmetadataDownloadJob:  add authorization check for each exported object, miscellaneous cleanup

### LINKED ISSUE #1491: add missing permissions checks

jmartin-sul:
*  check whether user who started the job can view metadata for each object in the job's pid list.  if the user cannot, log the event, skip exporting the metadata for that object, and continue processing the rest.
  * corresponding new test for the permission check
* tighten up an existing expectation for the happy path use case (user has permission and all objects in job's pid list are successfully retrieved)
* refactoring/cleanup:
  * spec:  turn all the instance vars into `let` vars, turn a local example var into a `let` var (and consolidate on the longer pid list for all the tests)
  * job:  remove an unneeded setting of an instance var in the job, and instead rely on the parent class to do setup

i believe this...

closes #1491

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ If you want to use the rails console use:
 docker-compose run --rm web rails console
 ```
 
-If you want to run background jobs, which are necessary for spreadsheet bulk uploads and indexing run:
+If you want to run background jobs, which are necessary for spreadsheet bulk uploads and indexing to run:
 
 ```
 docker-compose run web bin/delayed_job start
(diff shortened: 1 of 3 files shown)


jmartin-sul:
we should probably test on stage and get @andrewjbtw's approval before merging and deploying?

jmartin-sul:
@andrewjbtw gives the thumbs up after manually testing, and @jcoyne approves.  merging...

PR #1652, opened 2019-09-20
Extract code from views by creating an ActionButton component

jcoyne:
## Why was this change made?

Keep code out of the views.

N/A

### DIFF
--- a/app/components/action_button.rb
+++ b/app/components/action_button.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# Draws a blue button in the side menu
+class ActionButton < ApplicationComponent
+  # @param [Hash] properties the button properties
+  def initialize(label:, url:, confirm: nil, new_page: nil, check_url: nil, disabled: nil)
+    @label = label
+    @url = url
+    @confirm = confirm
+    @new_page = new_page
+    @check_url = check_url
+    @disabled = disabled
+  end
+
+  def disabled?
+    check_url || disabled
+  end
+
+  def button_data
+    {}.tap do |data|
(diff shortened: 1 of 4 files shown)


jcoyne:
@mjgiarlo as you can see by the test failure, this component is adequately covered by the view spec.  If you want to add coverage, adding a component spec is a good way.

mjgiarlo:
@jcoyne :speech_balloon: 
> @mjgiarlo as you can see by the test failure, this component is adequately covered by the view spec. If you want to add coverage, adding a component spec is a good way.

I don't so much care about the overall coverage. I care more about getting a sense of whether folks would rather sling view specs or component specs (or both) going forward, as we convert views to being backed by components. Maybe that's not the right question to ask, though.

peetucket:
I don't think both kinds of tests are really providing a benefit.  Having to pick one, I'd do a component spec, but don't think I would both converting when a refactoring like this is done.

app/components/action_button.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Draws a blue button in the side menu
+class ActionButton < ApplicationComponent
+  # @param [Hash] properties the button properties
+  def initialize(label:, url:, confirm: nil, new_page: nil, check_url: nil, disabled: nil)
+    @label = label
```
peetucket: it feels like there ought to be a more elegant way to avoid so much boilerplate code like this

jcoyne: We could make a struct, but that's just moving the boilerplate, isn't it?

PR #1658, opened 2019-09-20
use preservation-client gem to get last_accessioned_version

ndushay:
## Why was this change made?

get `last_accessioned_version` from preservation rather than dor-services-app proxying for sdr-services-app.  This PR uses the new preservation-client gem to get the info from a preservation-catalog ReST endpoint.

N/A

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -69,6 +69,7 @@ gem 'dor-services-client', '~> 2.2'
 gem 'dor-workflow-client', '~> 3.7'
 gem 'mods_display'
 gem 'okcomputer' # monitors application and its dependencies
+gem 'preservation-client'
 gem 'responders', '~> 2.0'
 gem 'rsolr'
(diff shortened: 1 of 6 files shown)


ndushay:
lest you are worried about the hookup, this is from my laptop's rails console:

```
$ bx rails c
Loading development environment (Rails 5.2.3)
2.6.4 (main):0 Pry > Preservation::Client.current_version('aa000aa0000')
NoMethodError: undefined method `current_version' for Preservation::Client:Class
from (pry):1:in `<main>'
2.6.4 (main):0 Pry > Preservation::Client.objects.current_version('aa000aa0000')
Preservation::Client::UnexpectedResponseError: Preservation::Client.current_version for aa000aa0000 got Not Found (404) from Preservation Catalog at https://example.org/prescat/objects/aa000aa0000.json
```

PR #1662, opened 2019-09-21
Use the cocina model to display release tags

jcoyne:
Blocked by https://github.com/sul-dlss/cocina-models/pull/12

## Why was this change made?

So that Argo doesn't have to parse XML or be aware of Fedora 3

N/A


Gemfile
```diff
@@ -64,8 +64,9 @@ gem 'rubyzip'
 # Stanford/Hydra related gems
 gem 'blacklight', '~> 6.0'
 gem 'blacklight-hierarchy'
+gem 'cocina-models', github: 'sul-dlss/cocina-models', branch: 'deserialize'
```
ndushay: I would rather wait for the gem to have its PR merged ...

jcoyne: Agree.  I forgot that was why this was a draft and marked it "ready"

PR #1664, opened 2019-09-22
convert ChecksumReportJob to use preservation-client gem

ndushay:
## Why was this change made?

To be consistent in using the preservation-client gem to interact with the Preservation API (preservation catalog).

n/a

## Reviewers, please:

- does this notify honeybadger appropriately?


app/jobs/checksum_report_job.rb
```diff
@@ -22,23 +22,35 @@ def perform(bulk_action_id, params)
...
+        Honeybadger.notify message
+        raise message
       rescue StandardError => e
         bulk_action.update(druid_count_fail: pids.length)
-        log.puts("#{Time.current} ChecksumReportJob creation failed #{e.class} #{e.message}")
+        message = "#{Time.current} ChecksumReportJob creation failed #{e.class} #{e.message}"
+        log.puts(message)
+        # if we couldn't write out the file, we have an issue
+        Honeybadger.notify message
+        raise message
```
ndushay: Is this where we end up if user is disallowed?  If so, should we be more specific in the error we're trapping for, and NOT raise the exception?

I thought this was catching for problems opening file.

spec/jobs/checksum_report_job_spec.rb
```diff
@@ -18,45 +18,69 @@
...
       it 'does not call the presevation_catalog API, does not write a CSV file, and records failure counts' do
-        allow(my_conn).to receive(:post)
-        subject.perform(bulk_action.id,
+        expect { subject.perform(bulk_action.id,
                         output_directory: output_directory_fail,
                         pids: pids,
                         groups: groups,
-                        user: user)
-        expect(my_conn).not_to have_received(:post).with('/objects/checksums', druids: pids, format: 'csv')
+                        user: user) }.to raise_error(RuntimeError, /ChecksumReportJob not authorized to view all content/)
```
ndushay: I'm assuming this is the ability filtering, so StandardError has to do with opening the output file?

app/jobs/checksum_report_job.rb
peetucket: i'm not sure if this is where unauthorized users end up actually -- i was more thinking it was a just a general catch all ... but we could just let it go for now and see what it catches (and if it catches a bunch of authorization errors, extend to improve messaging)

spec/jobs/checksum_report_job_spec.rb
peetucket: sounds reasonable

PR #1673, opened 2019-09-24
Spend less time scanning irrelevant and potentially large directories for Ruby files that are not there

mjgiarlo:
## Why was this change made?

This helps rubocop start up more quickly, shaving at least a few seconds off the build

n/a

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,13 @@ AllCops:
     - db/schema.rb
     - vendor/bundle/**/*
     - node_modules/**/*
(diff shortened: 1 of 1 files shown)


.rubocop.yml
```diff
@@ -11,6 +11,13 @@ AllCops:
     - db/schema.rb
     - vendor/bundle/**/*
     - node_modules/**/*
+    - tmp/**/*
```
mjgiarlo: I don't really like listing all this crud here, but it does noticeably speed up rubocop.

PR #1678, opened 2019-09-27
add webpacker compile when building web container

peetucket:
## Why was this change made?

John and I had problems spinning up the web docker container unless we ran this step manually.

Note: Even though the Dockerfile currently installs Yarn, I had to re-do that before running the compile command, like this:

```yarn```

Not sure why.

Should we update documentation instead?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,5 +23,3 @@ RUN bundle install
 
 COPY package.json yarn.lock ./
 RUN yarn install
-
-RUN bundle exec rake webpacker:compile
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,3 +23,5 @@ RUN bundle install
 
 COPY package.json yarn.lock ./
 RUN yarn install
+
+RUN bundle exec rake webpacker:compile
--- a/README.md
+++ b/README.md
@@ -62,11 +62,19 @@ docker-compose run --rm web rake argo:repo:load
 
 If you run into errors related to the version of bundler when building the `web` container, that likely means you need to pull a newer copy of the base Ruby image specified in `Dockerfile`, e.g., `docker pull ruby:{MAJOR}.{MINOR}-stretch`.
 
+Also, if you run into webpacker related issues, you may need to manually install yarn and compile webpacker in your Docker container (or local laptop if you running that way):
+
+```
+docker-compose run --rm web yarn
+docker-compose run --rm web bundle exec rake webpacker:compile
+```
+
 ## Load and index records
 
 ```
 docker-compose run --rm web rake argo:repo:load
 ```
+
 
 ## Common tasks
 

### CODE REVIEW

**Comment by jmartin-sul** [comment]
it's possible that i had, in prior flailing with my local setup, done `bundle exec rake assets:precompile` from outside the web container (i.e. just natively in OS X).  i wonder whether this interfered with the web service running in the docker container?  until running the below per @peetucket's advice, my local argo instance was serving everything unstyled, and i saw the following on my browser's web console:

```
Error: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
Error: Missing binding /app/node_modules/node-sass/vendor/linux-x64-72/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 12.x

Found bindings for the following environments:
  - OS X 64-bit with Node.js 8.x

This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass` to download the binding for your current environment.
```

despite trying to clean up node modules, maybe i had some cached asset stuff lying around, but built for the wrong environment (OS X instead of the linux docker container)?

was discussing with peter on slack, and i wonder whether troubleshooting tips in the README would be an appropriate supplement/alternative to the Dockerfile changes in this PR?  no idea, as i'm a webpacker noob.

input from @jcoyne would probably be helpful.

File: Dockerfile [hunk 1]
```diff
@@ -23,3 +23,5 @@ RUN bundle install
 
 COPY package.json yarn.lock ./
 RUN yarn install
+
+RUN bundle exec rake webpacker:compile
```
jcoyne: Webpacker compile is not necessary in development mode, as it does this when you request the assets.

jmartin-sul:
@jcoyne said
> Webpacker compile is not necessary in development mode, as it does this when you request the assets.

makes sense, and that was my experience when editing JS assets while developing locally -- i'd make changes, things would automatically recompile, and changes would be reflected when visiting my local instance.

however, till i manually ran the asset compilation step (using an instance of the web container, while the `docker-compose`d stack was up and running, including the web service), i was getting the error described in my comment [here](https://github.com/sul-dlss/argo/pull/1678#issuecomment-536097915).  that was perplexing (precisely because i didn't expect to have to manually compile assets).  i suspect this was because i may at some point have done `bundle exec rake assets:precompile` directly in OS X, outside the linux web container.  probably while troubleshooting some other unrelated issue with my local instance.  and since i hadn't made any JS changes, it didn't think it had anything to rebuild.

as such, i think it'd at least be good to put this in the README as a troubleshooting note.  i don't think it has to go in Dockerfile, necessarily -- automatically having it run might've saved me some trouble, but it also might've hidden (and prevented me from learning about) env specific asset compilation, and it might unnecessarily slow the argo docker image build.  but without running this manually at some point, i think my local dev instance would still be busted.  so i think it'd be good to put this pointer somewhere.

i guess if i'd have just started editing JS, this would have happened automatically.  but i had no way of knowing that, and had no reason to suspect that my JS edits would magically fix my busted local instance.  i didn't want to waste time making and trying changes locally when it looked like my local instance wasn't even properly set up yet.

Dockerfile
peetucket: Moved to readme for troubleshooting if needed.  I ran into the same issue and had to manually compile into my web docker container.

peetucket:
> Does #1686 fix this problem?

probably won't know until John or I build a new web container, is that right?

jmartin-sul:
also, executing the steps added to the README did no harm to my currently working argo instance

jmartin-sul:
ah, i see, it's possible that i wasn't able to reproduce the original issue because of #1686, which i'd forgotten about, and didn't notice at first when revisiting this.  happy to revert this PR if @jcoyne would strongly prefer that.

if we keep the change from this PR, i have a follow on that tidies it up just a bit, though some of that would be worth keeping regardless of whether the changes in this PR stay:  https://github.com/sul-dlss/argo/pull/1702

jcoyne:
@jmartin-sul I think we should revert it if it's not relevant.  Too much documentation that isn't relevant diminishes the value of the whole thing.

Issue #1671, opened 2019-09-24
Unable to download a report


caaster:
**Describe the bug**
I navigate to `https://argo.stanford.edu/report?f%5Bnonhydrus_apo_title_ssim%5D%5B%5D=ETDs&f%5Bprocessing_status_text_ssi%5D%5B%5D=Accessioned&f%5Breleased_to_ssim%5D%5B%5D=Searchworks` -- and I select `Select View` and then I select `Report View`. Next, I select `Download` in the lower left-hand corner, and I wait, and wait, and wait. No report is downloaded.

**User Impact**
This is critical functionality. I pull several reports a month as I check on and perform analysis of content queues. I sometimes download reports to provide to stakeholders as well.

**To Reproduce**
1. Go to 'https://argo.stanford.edu/report?f%5Bnonhydrus_apo_title_ssim%5D%5B%5D=ETDs&f%5Bprocessing_status_text_ssi%5D%5B%5D=Accessioned&f%5Breleased_to_ssim%5D%5B%5D=Searchworks'
2. Click on 'Select View'
3. Click on `Report View`
4. Scroll down to and select 'Download'
5. I wait and wait and wait and no report is downloaded. This always worked in the past.

**Expected behavior**
I expect to execute the steps listed above, and after a minute, or 3, or 5 (generally, depending upon the number of items), I expect to see a .csv report in my download directory.

**Screenshots**
This is self-explanatory.

**Additional context**
n/a

jcoyne:
I was able to download a report of 1200 items, but it took 131s.  It's possible that a bigger report would time out.

peetucket:
For additional information, it would be helpful if we could identify the columns that are needed for downloaded reports (as opposed to viewing on screen).  The reason is that there may be some columns that are causing performance issues and if those columns are not actually needed when downloading reports, removing them may speed things up greatly.  Displaying is less of a performance issue because the entire result set is not fetched at once when showing in the grid.
^^
@dinahhandel @caaster and @andrewjbtw

peetucket:
We can also use the most common list of columns identified as needed in reports to set as the default set that is offered for download.

peetucket:
Note: some optimizations in #1693 should end up speeding up downloaded reports (once fully deployed to -prod).  There is also work to make this into a background job if needed (#1688)

PR #1688, opened 2019-10-01
bulk action for download report as csv

peetucket:
## Why was this change made?

Some big reports time out when you download.  Attempt at implementing the ability to download a report as a job.

Should helps address #1671, assuming #1693 doesn't make things so fast as to make this totally unncessary

### DIFF
--- a/app/controllers/bulk_actions_controller.rb
+++ b/app/controllers/bulk_actions_controller.rb
@@ -29,7 +29,8 @@ class BulkActionsController < ApplicationController
 
     # BulkActionPersister is responsible for enqueuing the job
     if BulkActionPersister.persist(@bulk_action)
-      redirect_to action: :index, notice: 'Bulk action was successfully created.'
+      flash[:notice] = 'Bulk action was successfully created.'
+      redirect_to action: :index
     else
       render :new
     end
--- a/app/controllers/bulk_actions_controller.rb
+++ b/app/controllers/bulk_actions_controller.rb
@@ -2,6 +2,8 @@
 
 class BulkActionsController < ApplicationController
   before_action :set_bulk_action, only: [:destroy, :file]
+  include Blacklight::Catalog
+  copy_blacklight_config_from CatalogController
 
   rescue_from ActiveRecord::RecordNotFound, with: -> { render plain: 'Record Not Found', status: :not_found }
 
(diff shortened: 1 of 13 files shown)


app/javascript/controllers/bulk_actions.js
```diff
@@ -17,8 +17,8 @@ export default class extends Controller {
   // Toggles visibility of common fields based on selected tab
   toggleCommonFieldVisibility(selectedTab) {
     const commonFields = this.element.querySelector('#common_fields')
-
-    if (selectedTab == 'CreateVirtualObjectsJob') {
+    var common_fields_hidden_jobs = [ 'CreateVirtualObjectsJob', 'DownloadReportJob' ]
```
mjgiarlo: Can this be defined as a `const` outside this function?

app/jobs/download_report_job.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  private
+
+  # produce the report given the search results
+  def download_report(params)
+    log.puts("#{Time.current} Running report with #{params}")
+    results = []
+    bulk_action.update(druid_count_total: results.size)
```
mjgiarlo: Given the prior line, won't this always be 0?

peetucket: yeah, there isn't any implementation here yet, still WIP

mjgiarlo: okie dokie.

jcoyne:
@peetucket rather than this radical change, what if we just removed all the facet stuff from the report. I don't think it's used on the report and that is the part that makes the report generation slow.

peetucket:
> @peetucket rather than this radical change, what if we just removed all the facet stuff from the report. I don't think it's used on the report and that is the part that makes the report generation slow.

Can you say more about this?  I'm not sure what you mean.

app/controllers/bulk_actions_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 class BulkActionsController < ApplicationController
   before_action :set_bulk_action, only: [:destroy, :file]
+  include Blacklight::Catalog
+  copy_blacklight_config_from CatalogController
```
jcoyne: I don't understand why this is necessary. Can you explain?

peetucket: This is needed so I can render the selected search with `render_constraint` in the create job view...unless there is a more clever way

peetucket:
> Does this imply that #1693 is not sufficient?

possibly, we are doing some testing now...if big reports still take a couple minutes or more, I think its worth making this available as an option

peetucket:
Closing for now as we have made current immediate downloads faster

Issue #1670, opened 2019-09-24
In report view, unable to select columns


caaster:
**Describe the bug**
In Argo, I navigate to: `https://argo.stanford.edu/report?f%5Bnonhydrus_apo_title_ssim%5D%5B%5D=ETDs&f%5Bprocessing_status_text_ssi%5D%5B%5D=Accessioned&f%5Breleased_to_ssim%5D%5B%5D=Searchworks`, and I select view > report view. Next, I select `columns` so I can specify the columns I want in my report. Nothing happens: no column labels appear in a dropdown menu so that I can select the ones I want.

**User Impact**
The columns are preset (at least they were in the past), so that all columns are not included when I download a report -- if I do nothing to specify which columns I want to see in the report. This limits my ability to perform robust analysis on a group of items. This is an important feature to maintain. I use this feature several times a month for various tasks.

**To Reproduce**
1. Go to 'https://argo.stanford.edu/report?f%5Bnonhydrus_apo_title_ssim%5D%5B%5D=ETDs&f%5Bprocessing_status_text_ssi%5D%5B%5D=Accessioned&f%5Breleased_to_ssim%5D%5B%5D=Searchworks'
2. Click on `Select View`
3. Select `Report View`
4. In bottom left-hand corner, select `Columns`
5. Nothing happens. I should see a drop down menu display with a list of all available columns that I can select from.

**Expected behavior**
Please see the above description, which explains the expected behavior.

**Screenshots**
This behavior is self explanatory.

**Additional context**
n/a

jcoyne:
I believe this broke when we upgraded to jquery 3.
We had to upgrade to Jquery 3 because of a security vulnerability.  
The column selector is built on Jquery-ui 1.12, which was last updated over 3 years ago and doesn't work with the latest version of Jquery 3.

caaster:
Is PSM being consulted when choices like this need to be made? Or even being notified?

jcoyne:
@caaster Andrew was aware and tested the changes.

jcoyne:
The UI worked in this commit (but was very ugly):
https://github.com/sul-dlss/argo/commit/adf8e86b848ede1d838e51641f1a5a92a5cf0620

There have been 88 changes to master since: https://github.com/sul-dlss/argo/compare/adf8e86b848ede1d838e51641f1a5a92a5cf0620...master

Highlights:
* switched to free-jqgrid (the old library was taken private and they required us to have an enterprise license)
* webpacker
[...]

PR #1691, opened 2019-10-03
actually show notices correctly in bulk actions

peetucket:
## Why was this change made?

because notices were not actually being shown to users when bulk actions were created, and should be

n/a


jmartin-sul:
nice!  i'd noticed this when testing for #1670, and had meant to file a bug about it (and/or just try to fix it).

feel free to merge if/when this goes green.

screenshot might be nice, but also prob not super necessary (this does seem like the right approach AFAICT from cursory web search, and this seems like a very low risk change anyway).

jmartin-sul:
just went green, merged

PR #1692, opened 2019-10-03
set the current user correctly in job context

peetucket:
## Why was this change made?

When implementing the download report (as csv) job, I realized the current user (which is necessary for running a report) is not being set in a way that can be used.  We need the actual user model object with persisted groups in order to run the job correctly (not just the username persisted in params).

If this is merged, the download job (PR #1688) can be refactored to make use of it.

### DIFF
--- a/app/jobs/generic_job.rb
+++ b/app/jobs/generic_job.rb
@@ -33,7 +33,7 @@ class GenericJob < ActiveJob::Base
   def perform(_bulk_action_id, params)
     @groups = params[:groups]
     @pids = params[:pids]
-    @current_user = params[:user]
+    @current_user = bulk_action.user
   end
 
   def bulk_action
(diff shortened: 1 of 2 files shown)


app/jobs/generic_job.rb
```diff
@@ -33,7 +33,7 @@ class GenericJob < ActiveJob::Base
   def perform(_bulk_action_id, params)
     @groups = params[:groups]
     @pids = params[:pids]
-    @current_user = params[:user]
+    @current_user = bulk_action.user
```
jcoyne: I don't believe this is correct.  The params[:user] should be a User object.

peetucket:
indeed, i stumbled upon this when doing the download job when it wasn't working like I expected -- it doesn't seem to be a problem with other jobs (but downloads need the user set correctly)

PR #1695, opened 2019-10-04
allow custom date queries to work in reports

### LINKED ISSUE #1694: Running a report (in the view or download) where you enter in dates produces no results
**Describe the bug**
Faceting on specific date ranges produces no results in the report view.

**User Impact**
Users can't download specific date range reports.

**To Reproduce**
Select a date facet (such as "Earliest Accessioned") and then enter manual dates in the fields (e.g. 2019-10-01 and 2019-10-02).  Click Submit and you will get a results view with results, e.g.

https://argo.stanford.edu/catalog?f%5Baccessioned_earliest_dttsi%5D%5B%5D=%5B2019-10-01T00%3A00%3A00.000Z+TO+2019-10-02T23%3A59%3A59.000Z%5D

Then select "Report view" and you will get an empty grid and nothing to download.

**Expected behavior**
We should get the same results in the grid view

**Additional context**
perhaps related to #1693 which changed how the report view search results are build by leaving off faceting ?

peetucket:
## Why was this change made?

to fix #1694

### DIFF
--- a/app/search_builders/report_search_builder.rb
+++ b/app/search_builders/report_search_builder.rb
@@ -3,6 +3,10 @@
 class ReportSearchBuilder < Blacklight::SearchBuilder
   include Blacklight::Solr::SearchBuilderBehavior
   include Argo::AccessControlsEnforcement
+  include Argo::CustomSearch
+  include Argo::DateFieldQueries
 
-  self.default_processor_chain -= [:add_facetting_to_solr]
+  self.default_processor_chain -= [:add_facetting_to_solr] # remove faceting from reports
+
+  self.default_processor_chain += [:add_date_field_queries] # ensure date field queries work
 end
--- a/spec/controllers/report_controller_spec.rb
+++ b/spec/controllers/report_controller_spec.rb
@@ -33,6 +33,13 @@ RSpec.describe ReportController, type: :controller do
       data = JSON.parse(response.body)
       expect(data['rows'].length).to eq(10)
     end
+    it 'returns data for custom date searches (i.e. with user custom entered dates)' do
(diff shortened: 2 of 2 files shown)


app/search_builders/report_search_builder.rb
```diff
@@ -3,6 +3,17 @@
...
   include Argo::AccessControlsEnforcement
+  include Argo::CustomSearch
+  include Argo::DateFieldQueries
+  include Argo::ProfileQueries
+
+  self.default_processor_chain -= [:add_facetting_to_solr] # remove faceting from reports
+
+  self.default_processor_chain += [
+    :add_access_controls_to_solr_params,
+    :pids_only,
```
jcoyne: Why is this needed?

app/search_builders/report_search_builder.rb
```diff
@@ -3,6 +3,17 @@
...
+    :add_date_field_queries,
+    :add_profile_queries
```
jcoyne: Why is this needed?

peetucket: not sure, maybe they aren't ?

app/search_builders/report_search_builder.rb
peetucket: removed

app/search_builders/report_search_builder.rb
peetucket: removed

peetucket:
Quick test on argo-stage confirms this fixes the issue with custom dates.

peetucket:
added test and confirm it fails without this change

PR #1698, opened 2019-10-07
enforce view restrictions for reports

peetucket:
## Why was this change made?

reports/downloads are allowing users to see items for which they should not have permissions -- add back in the query limitations into the new SearchBuilder

we confirmed this is currently a problem and that this fix works on argo-stage

### DIFF
--- a/app/search_builders/report_search_builder.rb
+++ b/app/search_builders/report_search_builder.rb
@@ -8,5 +8,8 @@ class ReportSearchBuilder < Blacklight::SearchBuilder
 
   self.default_processor_chain -= [:add_facetting_to_solr] # remove faceting from reports
 
-  self.default_processor_chain += [:add_date_field_queries] # ensure date field queries work
+  self.default_processor_chain += [
+    :add_access_controls_to_solr_params, # enforce restrictions
+    :add_date_field_queries # ensure date field queries work
+  ]
 end
--- a/spec/search_builders/report_search_builder_spec.rb
+++ b/spec/search_builders/report_search_builder_spec.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ReportSearchBuilder do
+  subject { report_search_builder.with(user_params) }
+
+  let(:user_params) { {} }
+  let(:solr_params) { {} }
+  let(:context) { CatalogController.new }
+
+  let(:report_search_builder) { described_class.new(context) }
+
+  describe '#initialize' do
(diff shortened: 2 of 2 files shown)


app/search_builders/report_search_builder.rb
```diff
@@ -8,5 +8,9 @@ class ReportSearchBuilder < Blacklight::SearchBuilder
 
   self.default_processor_chain -= [:add_facetting_to_solr] # remove faceting from reports
 
-  self.default_processor_chain += [:add_date_field_queries] # ensure date field queries work
+  self.default_processor_chain += [
+    :add_access_controls_to_solr_params, # enforce restrictions
+    :add_date_field_queries,   # ensure date field queries work
```
mjgiarlo: kill the comma, methinks

peetucket: yup, and a blank line -- fixed, rebased, pushed

app/search_builders/report_search_builder.rb
```diff
@@ -8,5 +8,8 @@ class ReportSearchBuilder < Blacklight::SearchBuilder
 
   self.default_processor_chain -= [:add_facetting_to_solr] # remove faceting from reports
 
-  self.default_processor_chain += [:add_date_field_queries] # ensure date field queries work
+  self.default_processor_chain += [
+    :add_access_controls_to_solr_params, # enforce restrictions
+    :add_date_field_queries # ensure date field queries work
```
mjgiarlo: @peetucket unless you're pressing to get this deployed, it might be useful to add a spec as well. if you've already considered and decided not to, happy to hear you out

peetucket: i just added a simple spec that mimics what is there for the existing `SearchBuilder` - it confirms the processor chains have the necessary methods ... there are existing specs that confirm the correctness of those processor chains

building a feature or controller spec is another possibility, but would require confirming that the fixtures have the necessary data to support the checking of varying permissions

mjgiarlo: ok. approving!

spec/search_builders/report_search_builder_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  subject { report_search_builder.with(user_params) }
+
+  let(:user_params) { {} }
+  let(:solr_params) { {} }
+  let(:context) { CatalogController.new }
+
+  let(:report_search_builder) { described_class.new(context) }
+
+  describe '#initialize' do
+    it 'has add_access_controls_to_solr_params in chain once' do
```
jcoyne: Why did you count rather than just `include?`

PR #1702, opened 2019-10-15
touchups for formatting and wording clarity

jmartin-sul:
## Why was this change made?

to address some minor rough edges in README wording and formatting.

it was the only thing that was updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Note that `bundle install` may complain if MySQL isn't installed.  You can eithe
 ### Install components
 ## Run the servers
 
-```
+```bash
 docker-compose up -d
 ```
 
@@ -41,13 +41,13 @@ change the value of `ROLES` to `sdr:administrator-role`
 
 If you want to use the rails console use:
 
-```
+```bash
 docker-compose run --rm web rails console
 ```
 
 If you want to run background jobs, which are necessary for spreadsheet bulk uploads and indexing to run:
 
-```
+```bash
 docker-compose run web bin/delayed_job start
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,20 +87,19 @@ This will allow you to view rails output in real-time.  You can also add 'byebug
 
 If you run into errors related to the version of bundler when building the `web` container, that likely means you need to pull a newer copy of the base Ruby image specified in `Dockerfile`, e.g., `docker pull ruby:{MAJOR}.{MINOR}-stretch`.
 
-Also, if you run into webpacker related issues, you may need to manually install yarn and compile webpacker in your Docker container (or local laptop if you running that way):
+Also, if you run into webpacker related issues (e.g. [`Node Sass could not find a binding for your current environment`](https://github.com/sul-dlss/argo/pull/1678)), you may need to manually install yarn and compile webpacker from within your Docker container (or natively from your laptop, if that's how you're running the web service, instead of using Docker):
```
justinlittman: For docker is this done at build time or run time (i.e., shell into container)?

jcoyne:
@jmartin-sul are you able to address @justinlittman's questions?

README.md
jmartin-sul: my recollection is that the below commands would use the container to output artifacts to the host filesystem.  which is to say, not shelling into the running container, just using a transient container to output assets.  but my recollection is fuzzy at this point.

sorry for the late response, replying unintentionally but slowly got pushed way down my TODO list.  but @jcoyne also prodded me a couple times recently to follow up.

jmartin-sul: also plausible that the above line and the below two commands can get removed -- might've been made moot by https://github.com/sul-dlss/argo/pull/1686 (see also https://github.com/sul-dlss/argo/pull/1678#discussion_r329572955 and https://github.com/sul-dlss/argo/pull/1678#issuecomment-536619578).  i'd advocate slightly for leaving it since peter and i both found it helpful at one point, but i don't feel strongly.

PR #1708, opened 2019-10-23
fix typo in html for refresh mods button

### LINKED ISSUE #1707: "Refresh MODs" button in "Bulk Update" does nothing
**Describe the bug**
If you go to the Bulk Update (https://argo.stanford.edu/report/bulk), get a list of druids in the box and click "Refresh MODs", nothing happens.  We should expect a box to appear below confirming the selection and allowing this action to proceed.  Other bulk update buttons may also be affected.

**User Impact**
Users are unable to perform the require bulk actions.

**To Reproduce**
Described above.

**Expected behavior**
Described above.

**Additional context**
First reported October 21 in #dlss-aaas.

peetucket:
## Why was this change made?

to fix the issue with the refresh mods button HTML ID not matching the JS code (needs to match https://github.com/sul-dlss/argo/blob/master/app/javascript/packs/bulk.js#L287)

fixes #1707 

n/a

### DIFF
--- a/app/views/report/bulk.html.erb
+++ b/app/views/report/bulk.html.erb
@@ -58,7 +58,7 @@ ul li{
     <strong>Options that require versioning</strong>
     <div class="row bulk-buttons-row">
       <div class="col-sm-2">
-        <button class="bulk_button btn-block" class="update_buttons" id="referesh-mods-button">Refresh MODS</button>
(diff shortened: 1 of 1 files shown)


jcoyne:
Thanks @peetucket!

PR #1710, opened 2019-10-28
Tests run without dor-services-app

jcoyne:
## Why was this change made?

We don't need this server up to run tests.

n/a


jmartin-sul:
i guess sqlite and webmock cover those needs, respectively?

jmartin-sul:
registering similar surprise 😮, but build passes, so merging...

jcoyne:
@ndushay the database is only to support dor-services-app and we don't have any tests that need that.

ndushay:
@jcoyne thanks for the explanation.  Perhaps we should have tests that need dor-services-app, eventually.

mjgiarlo:
I have a vague recollection that we *used to* have tests that needed dor-services-app, but I could be mistaken.

PR #1711, opened 2019-10-28
Use postgres 11

jcoyne:
## Why was this change made?

dor-services-app is not yet compatible with postgres 12

N/a

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -65,7 +65,7 @@ services:
     ports:
       - 3002:3000
   db:
-    image: postgres
+    image: postgres:11
      # No ports shared externally, so that this doesn't conflict with the postgres
      # server that TravisCI starts up.
      # ports:
@@ -73,7 +73,7 @@ services:
     environment:
       - POSTGRES_PASSWORD=sekret
     volumes:
-      - ./postgres-data:/var/lib/postgresql/data
+      - postgres-data:/var/lib/postgresql/data
 
   workflow:
     image: suldlss/workflow-server:latest
@@ -101,3 +101,4 @@ services:
 #      - 6379:6379
 volumes:
   node_modules:
+  postgres-data:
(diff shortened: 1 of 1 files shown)


jcoyne:
@ndushay who would this documentation be for?  I'm not exactly sure why we'd want to document this.

jmartin-sul:
documentation outside the docker-compose.yml that this PR modifies might be overkill, but a comment/link explaining the pinning would be nice.  submitted a review w/ such a comment, for you to merge or not as you see fit.

also, fwiw, it looks like this was _just_ fixed?  https://github.com/ged/ruby-pg/issues/302#issuecomment-546942134
> This is fixed in libpq and will be part of PostgreSQL-12.1: [postgres/postgres@ba19a6b](https://github.com/postgres/postgres/commit/ba19a6b73c5bd771d8864171ede03503a9ff564e)

jcoyne:
@jmartin-sul this pg container is a dependency of dor-services-app, and the libpg change needs to be released in 12.1 (which hasn't happened yet) and then it needs to be packaged by alpine linux and then the dor-services-app image needs to be rebuilt with that alpine package installed.   This is all getting very complex and it's easier to say that this PR aligns the pg version with what is supported by dor-services-app.

jmartin-sul:
> @jmartin-sul this pg container is a dependency of dor-services-app, and the libpg change needs to be released in 12.1 (which hasn't happened yet) and then it needs to be packaged by alpine linux and then the dor-services-app image needs to be rebuilt with that alpine package installed. This is all getting very complex and it's easier to say that this PR aligns the pg version with what is supported by dor-services-app.

@jcoyne, just updated my comment suggestion to use your suggested wording.

as with my prior suggestion, feel free to merge without if you're still not a fan of the new suggestion.  totally up to you.

jmartin-sul:
(i did it as a commit suggestion, so if you do like it, you can just squash it in and merge this yourself.  but again, NBD if you decline to include it, IMHO.)

PR #1712, opened 2019-10-28
checksum_report_job: slightly better message to honeybadger on 404

ndushay:
## Why was this change made?

Because errors like https://app.honeybadger.io/projects/54415/faults/38904205 are confusing;  this will give a bit more context to the 404.

## Was the documentation updated?

n/a

PR #1713, opened 2019-10-28
checksum job honeybadger notifications only need raise

ndushay:
## Why was this change made?

I realized PR #1712 was klunkier than necessary;  this is better.  Hoping to get context of errors to the calling app for less confusion.

n/a


app/jobs/checksum_report_job.rb
```diff
@@ -31,20 +31,18 @@ def perform(bulk_action_id, params)
...
         # if we couldn't connect to preservation API, we have an issue
-        Honeybadger.notify message
         raise message
       rescue StandardError => e
         bulk_action.update(druid_count_fail: pids.length)
         message = "#{Time.current} ChecksumReportJob creation failed #{e.class} #{e.message}"
         log.puts(message)
         # if we couldn't write out the file, we have an issue
-        Honeybadger.notify message
         raise message
```
jcoyne: Do we need custom messages here?  Can we just re-raise `e`?

PR #1718, opened 2019-10-30
change purge action from get to delete

peetucket:
## Why was this change made?

The purge action is currently a GET operation.  This is hazardous and it should be a DELETE.  Discovered after work done in #1716 

Also, fixed the problem where a purge from bulk actions was not showing the status correctly (it was silently redirecting to the home page after the purge was done and then showing the status as a 404)

Confirmed working as expected on localhost (both bulk action and single purge from item detail page) and also confirmed a GET request to the purge method no longer works.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -53,7 +53,7 @@ Metrics/AbcSize:
 # Offense count: 7
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 406
+  Max: 410
 
 # Offense count: 7
 Metrics/CyclomaticComplexity:
--- a/app/components/action_button.html.erb
+++ b/app/components/action_button.html.erb
@@ -1,1 +1,1 @@
-<%= link_to label, url, class: "btn button btn-primary #{'disabled' if disabled?}", data: button_data %>
+<%= link_to label, url, method: method, class: "btn button btn-primary #{'disabled' if disabled?}", data: button_data %>
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -279,7 +279,11 @@ class ItemsController < ApplicationController
     ActiveFedora.solr.conn.commit
 
     respond_to do |format|
-      format.any { redirect_to '/', notice: params[:id] + ' has been purged!' }
+      if params[:bulk]
+        format.html { render plain: 'Purged.' }
+      else
+        format.any { redirect_to '/', notice: params[:id] + ' has been purged!' }
+      end
     end
   end
 
--- a/app/javascript/packs/bulk.js
+++ b/app/javascript/packs/bulk.js
@@ -32,6 +32,10 @@ function process_patch(druids, action_url, req_params, success_string) {
 	process_request(druids, action_url, 'PATCH', req_params, success_string, show_buttons, show_buttons);
 }
 
+function process_delete(druids, action_url, req_params, success_string) {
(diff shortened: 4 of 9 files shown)


app/components/action_button.html.erb
```diff
@@ -1 +1 @@
-<%= link_to label, url, class: "btn button btn-primary #{'disabled' if disabled?}", data: button_data %>
+<%= link_to label, url, method: "#{method}", class: "btn button btn-primary #{'disabled' if disabled?}", data: button_data %>
```
jcoyne: I think you can drop the interpolation here? `method: method` is easier, right?

peetucket: yes, good call

.rubocop_todo.yml
```diff
@@ -53,7 +53,7 @@ Metrics/AbcSize:
 # Offense count: 7
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 406
+  Max: 410
```
peetucket: this got longer as the ItemsContoller got longer so I could fix the redirect issue with the bulk action call

app/javascript/packs/bulk.js
```diff
@@ -32,6 +32,10 @@ function process_patch(druids, action_url, req_params, success_string) {
 	process_request(druids, action_url, 'PATCH', req_params, success_string, show_buttons, show_buttons);
 }
 
+function process_delete(druids, action_url, req_params, success_string) {
+	process_request(druids, action_url, 'DELETE', req_params, success_string, show_buttons, show_buttons);
```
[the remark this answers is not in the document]
peetucket: Note: if we add the 'use strict' statement, there are issues in the JS that cause problems (namely: undeclared variables).  I left that alone for now.

app/javascript/packs/bulk.js
```diff
@@ -43,7 +47,7 @@ function set_content_type(druids){
 }
 
 function purge(druids){
-	process_get(druids, purge_url, "Purged");
+	process_delete(druids, purge_url, null, "Purged");
```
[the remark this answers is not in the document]
peetucket: see comment above

app/controllers/items_controller.rb
```diff
@@ -279,7 +279,11 @@ def purge_object
     ActiveFedora.solr.conn.commit
 
     respond_to do |format|
-      format.any { redirect_to '/', notice: params[:id] + ' has been purged!' }
+      if params[:bulk]
+        format.html { render plain: 'Purged.' }
+      else
+        format.any { redirect_to '/', notice: params[:id] + ' has been purged!' }
+      end
```
peetucket: this fixes a related problem that currently exists, where the user sees a bogus "not found" error message when bulk purging items

jcoyne: Can you add a test for this case?

peetucket: test added to `items_controller_spec`

spec/controllers/items_controller_spec.rb
```diff
@@ -67,23 +67,31 @@
       end
 
       it 'redirects to root and flashes a confirmation notice when successful' do
-        post 'purge_object', params: { id: @pid }
```
peetucket: it seems like these tests should have been failing before since purge was a GET ... I don't think the verbs are being respected here by rspec

jcoyne: I think this is one of the reasons that controller tests are deprecated and request specs are preferred now.

PR #1733, opened 2019-11-22
virtual object creator: increase poll interval from 3 to 10 seconds

ndushay:
## Why was this change made?

To reduce network traffic to dor-services-app

n/a


ndushay:
I thought about putting it in a config, but since there's a possibility the approach may change to webhooks I figured it was easier to leave it as a constant in the class.   Thanks for at least calling it out as a proper constant declaration.

ndushay:
@andrewjbtw just want you to be aware of this change.   It should get deployed on Monday.

PR #1744, opened 2019-12-12
files_controller.preserved does something nice with Preservation::Client::NotFoundError

ndushay:
## Why was this change made?

Previous method of getting "preserved" content files from sdr-services-app (via dor-services-client) returned nil when the file didn't exist.  preservation-client v1.0.0 throws a Preservation::Client::NotFoundError when an API call to preservation-catalog returns 404. This PR traps for that error and returns an informative error page.

I also made sure that all usage of preservation-client v1.0.0 here copes at least as well with non-happy path responses as it did dor-services-client did.

n/a


app/controllers/files_controller.rb
```diff
@@ -30,6 +30,10 @@ def preserved
     response.headers['Content-Disposition'] = "attachment; filename=#{filename}"
     response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
     self.response_body = file_content
+  rescue Preservation::Client::NotFoundError => e
+    response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
+    response.status = :not_found
```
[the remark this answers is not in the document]
ndushay: it seemed to be needed from empirical testing.

app/controllers/files_controller.rb
```diff
@@ -30,6 +30,10 @@ def preserved
     response.headers['Content-Disposition'] = "attachment; filename=#{filename}"
     response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
     self.response_body = file_content
+  rescue Preservation::Client::NotFoundError => e
+    response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
```
jcoyne: Why are you setting this header?

ndushay: b/c it was set above?  should it be left empty?

jcoyne: I don't know why it's useful for a 404.

app/controllers/files_controller.rb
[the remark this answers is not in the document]
ndushay: nm - i misread your question.  removing duplication.

app/controllers/files_controller.rb
ndushay: i'll remove it.

ndushay: removed.

app/controllers/files_controller.rb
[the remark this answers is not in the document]
ndushay: de-dupped.

ndushay:
@jcoyne @mjgiarlo I believe I've addressed your concerns;  thanks for the improvements.

mjgiarlo:
> @jcoyne @mjgiarlo I believe I've addressed your concerns; thanks for the improvements.

thanks! feel free to merge when :green_heart:

PR #1746, opened 2019-12-13
Encode the filename when downloading from preservation

### LINKED ISSUE #1745: File downloaded from preservation core via Argo have truncated filenames
**Describe the bug**
When downloading files from preservation core via Argo, filenames are truncated to include only the string before the first space.

**User Impact**
Besides the filenames not matching the original, without extensions some operating systems do not choose the correct application to open these files. On OSX, the default assumption seems to be that files without extensions should open in a text editor.

The files themselves seem to be fine, provided you render them in the correct application for the file type.

**To Reproduce**
1. Go to https://argo.stanford.edu/view/druid:bt292qp8660
2. Click on the link to download "Annika Mulaney Thesis Final Version.pdf"
3. Choose the link that downloads from preservation (not Stacks).
4. The downloaded file will be named "Annika"

Other examples:
https://argo.stanford.edu/view/druid:cs953px2715 (jpg)
https://argo.stanford.edu/view/druid:bb007hx5508 (Rightslink printable license)

**Expected behavior**
Downloaded files have their full name and extension.

**Additional context**
Stacks download links work correctly, but some files exist only in preservation.

aaron-collier:
## Why was this change made?

fixes #1745 by encoding the filename in the headers when downloading through argo from the preservation link.

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -27,7 +27,7 @@ class FilesController < ApplicationController
     authorize! :view_content, @object
     file_content = Preservation::Client.objects.content(druid: @object.pid, filepath: filename, version: params[:version])
     response.headers['Content-Type'] = 'application/octet-stream'
-    response.headers['Content-Disposition'] = "attachment; filename=#{filename}"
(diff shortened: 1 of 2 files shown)


peetucket:
that's pretty much what I was thinking it would be

PR #1753, opened 2019-12-17
Display a useful message to the user when the file is not in preservation

jcoyne:
## Why was this change made?

Previously this exception was uncaught which sent the user to the system status page.

n/a

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -9,6 +9,15 @@ class FilesController < ApplicationController
     raise ArgumentError, 'Missing file parameter' if filename.blank?
 
     @available_in_workspace = Dor::Services::Client.object(params[:item_id]).files.list.include?(filename)
+    @has_been_accessioned = Dor::Config.workflow.client.lifecycle('dor', params[:item_id], 'accessioned')
+
+    if @has_been_accessioned
+      begin
+        @last_accessioned_version = Preservation::Client.objects.current_version(params[:item_id])
+      rescue Preservation::Client::NotFoundError
+        return render status: :unprocessable_entity, plain: "Preservation has not yet received #{params[:item_id]}"
+      end
+    end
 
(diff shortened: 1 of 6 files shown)


app/controllers/files_controller.rb
```diff
@@ -9,6 +9,15 @@ def index
     raise ArgumentError, 'Missing file parameter' if filename.blank?
 
     @available_in_workspace = Dor::Services::Client.object(params[:item_id]).files.list.include?(filename)
+    @has_been_accessioned = Dor::Config.workflow.client.lifecycle('dor', params[:item_id], 'accessioned')
+
+    if @has_been_accessioned
+      begin
+        @last_accessioned_version = Preservation::Client.objects.current_version(params[:item_id])
+      rescue Preservation::Client::NotFoundError
```
mjgiarlo: Worth considering a block-level `rescue` here instead of the explicit `begin`/`end`? Don't care **much** about it but it'd drop 2LOC :laughing:

app/views/files/index.html.erb
```diff
@@ -20,11 +20,10 @@
     </p>
   <% end %>
 
-  <%# TODO: has_been_accessioned? is deprecated, migrate to async call %>
-  <% if file['preserve'] == 'yes' && has_been_accessioned?(@object.pid) %>
+  <% if file['preserve'] == 'yes' && @has_been_accessioned %>
```
mjgiarlo: get outta here, logic! :clap:

app/controllers/files_controller.rb
jcoyne: Begin and end define a block, so this is a block-level rescue.  I'm not sure I understand what you're after.

mjgiarlo: Oh, my bad. I misread the `if` block here as a "block" block. :tea:

PR #1760, opened 2019-12-20
Pass version to the workflow service

jcoyne:
## Why was this change made?

This allows the workflow service to skip the callback to dor-services-app


jcoyne:
This work was completed in a different PR.

Issue #1757, opened 2019-12-19
New workflow steps aren't indexed in Argo facets, workflow grid


andrewjbtw:
**Describe the bug**
Three new workflow steps have been added to the accessionWF since October:

- publish-complete
- shelve-complete
- preservation-ingest-initiated

These workflow steps do not appear in the Argo facets for workflows, which are:

- Workflows (WPS)
- Workflows (WSP)
- Workflows (SWP)

![Screen Shot 2019-12-19 at 9 41 05 AM](https://user-images.githubusercontent.com/8975583/71196268-915af800-2244-11ea-8abd-645f73392a5b.png)

The other two facets are harder to screenshot, since they arrange the order of Workflow, Status, Process differently.

The three new steps are listed on the "All Workflows" grid but the values are all 0:

![Screen Shot 2019-12-19 at 9 44 52 AM](https://user-images.githubusercontent.com/8975583/71196416-e139bf00-2244-11ea-9bdf-08b2471381d2.png)

**User Impact**

As an administrator, I use the facets and the workflow grid to track down steps that are in error, or that have been waiting or queued for an unusually long time.

The specific case where I noticed this issue was while investigating objects that might have been affected by the production problems this week related to the preservation API. I wanted to see if any items have been stuck at "waiting" for a long time at preservation-ingest-initiated but I can't facet on that step.

**Expected behavior**
The facets and the workflow grid list all of the workflow steps and statuses accurately.

**Additional context**
The steps and statuses are being reported on the individual druid pages when you view the accessionWF from those pages, so this might just be an indexing problem.

jcoyne:
I believe the problem lies here: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/workflow/document.rb#L51-L63

```ruby
druid =  "druid:bq653yd1233"
all_wf = Dor::Config.workflow.client.all_workflows_xml druid
nodeset = Nokogiri::XML(all_wf).xpath('/workflows/workflow')
workflows =       nodeset.map { |wf_node| Dor::Workflow::Document.new wf_node.to_xml }
accession = workflows.first
# This does not return the correct processes:
accession.processes
```

PR #1761, opened 2019-12-20
Update dor-services to get new workflow indexing

jcoyne:
## Why was this change made?

Ref #1757


jcoyne:
@ndushay yeah, the changes are just stub changes to match the changes in dor-services.

PR #1763, opened 2019-12-20
Trap unexpected preservation client errors

### LINKED ISSUE #1756: trap for preservation-client errors and handle appropriately in argo
PR #1753 traps for NotFoundError for ONE call to preservation-client, but it doesn't trap for other specific pres-client errors or even their superclass. 

argo uses preservation-client in at least 3 ways:
- to retrieve preserved content file
- to get current_version
- to get checksums report
- opening/closing versions (via dor-services-app and dor-services-client, I believe ... but same idea applies re: error trapping and handling)

Ideally, we could:
- get error details to the user when appropriate 
   -  e.g. if pres-client has an UnexpectedResponse or other error or unexpected status code  (i.e. not a 200) , the user could see "unable to get response from Preservation for (druid)" or something instead of "something went wrong" 500 page.
- have a honeybadger notification when appropriate
- log appropriate error messages

See https://github.com/sul-dlss/preservation-client/blob/master/lib/preservation/client.rb#L30-L39

mjgiarlo:
Fixes #1756

## Why was this change made?

To gain additional visibility into preservation-client errors, we now log them and notify Honeybadger about them.

no

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -16,6 +16,11 @@ class FilesController < ApplicationController
         @last_accessioned_version = Preservation::Client.objects.current_version(params[:item_id])
       rescue Preservation::Client::NotFoundError
         return render status: :unprocessable_entity, plain: "Preservation has not yet received #{params[:item_id]}"
+      rescue Preservation::Client::Error => e
+        message = "Preservation client error getting current version of #{params[:item_id]}: #{e}"
+        logger.error(message)
+        Honeybadger.notify(message)
+        return render status: :internal_server_error, plain: message
       end
     end
 
@@ -40,7 +45,12 @@ class FilesController < ApplicationController
     response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
     self.response_body = file_content
   rescue Preservation::Client::NotFoundError => e
-    render(plain: "Preserved file not found: #{e}", status: :not_found)
+    render status: :not_found, plain: "Preserved file not found: #{e}"
+  rescue Preservation::Client::Error => e
+    message = "Preservation client error getting content of #{filename} for #{@object.pid} (version #{params[:version]}): #{e}"
+    logger.error(message)
+    Honeybadger.notify(message)
+    render status: :internal_server_error, plain: message
   end
 
   private
--- a/app/jobs/checksum_report_job.rb
+++ b/app/jobs/checksum_report_job.rb
@@ -26,30 +26,30 @@ class ChecksumReportJob < GenericJob
         csv_report = Preservation::Client.objects.checksums(druids: pids)
         File.open(report_filename, 'w') { |file| file.write(csv_report) }
         bulk_action.update(druid_count_success: pids.length) # this whole job is run in one call, so it either all succeeds or fails
-      rescue Preservation::Client::NotFoundError, Preservation::Client::UnexpectedResponseError => e
-        bulk_action.update(druid_count_fail: pids.length)
-        message = "#{Time.current} ChecksumReportJob got error from Preservation Catalog API: #{e.class} #{e.message}"
-        log.puts(message)
-        # honeybadger and other notifications should happen at prescat level, but not getting good message
-        raise message
-      rescue Preservation::Client::ConnectionFailedError => e
-        bulk_action.update(druid_count_fail: pids.length)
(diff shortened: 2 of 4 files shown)


app/controllers/files_controller.rb
```diff
@@ -16,6 +16,11 @@ def index
         @last_accessioned_version = Preservation::Client.objects.current_version(params[:item_id])
       rescue Preservation::Client::NotFoundError
         return render status: :unprocessable_entity, plain: "Preservation has not yet received #{params[:item_id]}"
+      rescue Preservation::Client::Error => e
+        message = "Preservation client error getting current version of #{params[:item_id]}: #{e}"
+        Rails.logger.error(message)
+        Honeybadger.notify(message)
+        return render status: :internal_server_error, plain: message
```
mjgiarlo: Is this the right status here? Or should it also be a 422? Or... ?

app/controllers/files_controller.rb
```diff
@@ -40,7 +45,12 @@ def preserved
     response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
     self.response_body = file_content
   rescue Preservation::Client::NotFoundError => e
-    render(plain: "Preserved file not found: #{e}", status: :not_found)
+    render status: :not_found, plain: "Preserved file not found: #{e}"
+  rescue Preservation::Client::Error => e
+    message = "Preservation client error getting content of #{filename} for #{@object.pid} (version #{params[:version]}): #{e}"
+    Rails.logger.error(message)
+    Honeybadger.notify(message)
+    render status: :internal_server_error, plain: message
```
mjgiarlo: Is this the right status here? Or should it also be a 422? Or... ?

app/jobs/checksum_report_job.rb
```diff
@@ -26,31 +26,31 @@ def perform(bulk_action_id, params)
...
-        log.puts(message)
-        # if we couldn't connect to preservation API, we have an issue
-        raise message
       rescue StandardError => e
         bulk_action.update(druid_count_fail: pids.length)
-        message = "#{Time.current} ChecksumReportJob creation failed #{e.class} #{e.message}"
-        log.puts(message)
-        # if we couldn't write out the file, we have an issue
-        raise message
+        message = exception_message_for(e)
```
mjgiarlo: I did some collapsing here since the only behavior varying in each block was the error message.

app/jobs/checksum_report_job.rb
```diff
@@ -26,31 +26,31 @@ def perform(bulk_action_id, params)
...
+        log.puts(message) # this one goes to the user via the bulk action log
+        Rails.logger.error(message) # this is for later archaeological digs
```
mjgiarlo: This is a change in behavior introduced in this branch: instead of raising the error, log it and notify Honeybadger.

app/controllers/files_controller.rb
```diff
@@ -16,6 +16,11 @@ def index
         @last_accessioned_version = Preservation::Client.objects.current_version(params[:item_id])
       rescue Preservation::Client::NotFoundError
         return render status: :unprocessable_entity, plain: "Preservation has not yet received #{params[:item_id]}"
+      rescue Preservation::Client::Error => e
+        message = "Preservation client error getting current version of #{params[:item_id]}: #{e}"
+        Rails.logger.error(message)
```
jcoyne: I think `logger` is already a method on controller so you can omit `Rails.`

app/controllers/files_controller.rb
jcoyne: I think this is fine.

app/controllers/files_controller.rb
```diff
@@ -40,7 +45,12 @@ def preserved
     response.headers['Last-Modified'] = Time.now.utc.rfc2822 # HTTP requires GMT date/time
     self.response_body = file_content
   rescue Preservation::Client::NotFoundError => e
-    render(plain: "Preserved file not found: #{e}", status: :not_found)
+    render status: :not_found, plain: "Preserved file not found: #{e}"
+  rescue Preservation::Client::Error => e
+    message = "Preservation client error getting content of #{filename} for #{@object.pid} (version #{params[:version]}): #{e}"
+    Rails.logger.error(message)
```
jcoyne: `Rails.` not needed.

app/jobs/checksum_report_job.rb
jcoyne: Omit `Rails.`

mjgiarlo:
@jcoyne Got 'em all. Thanks!

Issue #1728, opened 2019-11-20
Virtual object bulk actions restart unexpectedly


andrewjbtw:
**Describe the bug**
I recently ran a virtual object bulk action job on a large CSV. In preparation for this job, the parent objects had been registered and accessioned to version 1. As a result of the job, successful virtual objects should have only been updated to version 2. But at some point the job began to restart itself from the top row, and by the time I made changes to stop it from repeatedly remaking virtual objects it had made up to 5 versions of some objects but had not updated others at all.

The full CSV can be found attached to this ticket: https://jirasul.stanford.edu/jira/browse/LEGACY-4586

Most of the parent objects from the CSV can be found in Argo with this search: https://argo.stanford.edu/catalog?f%5Btag_ssim%5D%5B%5D=Project+%3A+Stanford+Band+%28Series+3+Parents%29

You can see that some objects in that search now have as many as 6 versions, while most objects were not changed at all.

Note that the parent objects that are now on version 2 were not modified by the bulk action. In order to prevent the bulk action job from repeatedly running on the same objects over the weekend, I changed the rights to all parents to citation-only, which meant creating new versions for the rights update. 

**Timeline**

As best as I can tell, the bulk action did not start re-running the top rows until a number of hours had passed. Times below are approximate, but I think accurate to within 30 minutes:

***Friday***

2019-11-15 3:30 PM: started the bulk action job

2019-11-15 10:00 PM: checked on the items and the first few parents (maybe 10-15?) now had a version 2. Items looked correct. There were some SSL/network errors that I cleared on the workflow grid.

***Saturday***

2019-11-16 8:00 AM: checked again and saw that there was a set of parent objects with 3-5 versions. Checked on the "Opened" items and concluded that the bulk action was running on the top rows of the CSV again. Set parents to citation-only in attempt to prevent further repetition before we all came back to work Monday.

2019-11-18: Saw that the bulk action was still running, requested that we stop the job by killing the process. Even if it wasn't still creating objects, it was still checking their statuses to evaluate whether they could be merged.

**User Impact**
As long as a virtual object bulk action might run multiple times on the same object in a CSV, it is too risky to run it with CSVs with many rows, especially CSVs with long rows (many children). Virtual object bulk actions will need to be broken up into many parts.

**To Reproduce**
1. Submit a virtual object CSV with multiple rows and a large number of child druids on each row. It might require at least 10 rows and 1000 druids distributed across these rows to trigger this behavior.
2. Watch to see if it restarts.

[...]

PR #1775, opened 2020-01-23
an attempt to avoid the problem of mysql going away

peetucket:
Note: this branch should be tested on stage first

## Why was this change made?

another attempt to address #1728

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -39,3 +39,4 @@ Metrics/BlockLength:
     - 'config/routes.rb'
     - 'spec/**/*'
     - 'lib/tasks/*.rake'
+    - 'app/jobs/create_virtual_objects_job.rb'
--- a/app/jobs/create_virtual_objects_job.rb
+++ b/app/jobs/create_virtual_objects_job.rb
@@ -58,6 +58,7 @@ class CreateVirtualObjectsJob < GenericJob
       end
 
       errors = VirtualObjectsCreator.create(virtual_objects: virtual_objects)
+      ActiveRecord::Base.clear_active_connections!
       if errors.empty?
         bulk_action.increment!(:druid_count_success, virtual_objects.length)
         log.puts("#{Time.current} #{SUCCESS_MESSAGE}: #{parent_ids_from(virtual_objects).to_sentence}")


.rubocop_todo.yml
```diff
@@ -50,6 +50,12 @@ Lint/UnusedMethodArgument:
 Metrics/AbcSize:
   Max: 82
 
+# Offense count: 1
```
peetucket: This was needed to make rubocop happy for the change in the PR.  Not sure about the other changes to the file.

.rubocop.yml
```diff
@@ -39,3 +39,4 @@ Metrics/BlockLength:
     - 'config/routes.rb'
     - 'spec/**/*'
     - 'lib/tasks/*.rake'
+    - 'app/jobs/create_virtual_objects_job.rb'
```
peetucket: This was needed to make rubocop happy for the change in the PR.

app/jobs/create_virtual_objects_job.rb
```diff
@@ -57,6 +57,7 @@ def perform(bulk_action_id, params)
         virtual_objects.reject! { |virtual_object| problematic_druids.include?(virtual_object[:parent_id]) }
       end
 
+      ActiveRecord::Base.clear_active_connections!
       errors = VirtualObjectsCreator.create(virtual_objects: virtual_objects)
       if errors.empty?
```
justinlittman: You'll want to take a closer look at the code, but I think might want to place it after `VirtualObjectsCreator.create`, as that is the long-running process.

justinlittman:
I'd suggest that we deploy this branch to stage rather than merging.

app/jobs/create_virtual_objects_job.rb
peetucket: I presume this is the long running line: https://github.com/sul-dlss/argo/blob/master/app/services/virtual_objects_creator.rb#L25

I wonder if instead we need to reconnect MySQL here during the polling?  :  https://github.com/sul-dlss/argo/blob/master/app/services/virtual_objects_creator.rb#L43

peetucket: We need to only deploy argo to stage to test this correct (i.e. no other projects need to be coordinate)?

justinlittman: Yes, Argo only.

justinlittman: If my understanding is correct, VirtualObjectsCreator just polls dor-services-app; it doesn't hit the db.

peetucket: Yeah, but I was thinking that if it sits in that method for a really long time running `redo` while it waits for the response, that is where the connection may get dropped.  But sure, I guess reconnecting once we exit the loop is fine too.

peetucket: move reconnect below job

mjgiarlo: @peetucket Can you add a comment above this line so we can remember why we added it? This is great. Thanks!

PR #1785, opened 2020-01-29
Remove workflows datastream from the datastreams table

jcoyne:
## Why was this change made?

We're deprecating the workflows datastream because it is an external datastream (proxy) and has no analog in a post-Fedora 3 vision.

Part of https://github.com/sul-dlss/dor-services-app/issues/601 

n/a


andrewjbtw:
Took a look on stage. Looks good to me.

PR #1787, opened 2020-01-30
Show the cocina data model on the show page

jcoyne:
@andrewjbtw would something like this be desirable?  This is the data representation we are going to be moving to as we migrate away from Fedora.  Currently we have only mapped over the things we need to run the robots, so this is incomplete, but it might be helpful in showing how we're doing the mapping.  I'm curious to get your thoughts.

Screenshot_2020-01-30 A thing - Argo

## Why was this change made?

As we move to a cocina based repository, argo users might want to see data as it is represented.

n/a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,7 +68,7 @@ gem 'rubyzip'
 gem 'blacklight', '~> 6.0'
 gem 'blacklight-hierarchy', '~> 2.0'
 gem 'dor-services', '~> 9.0'
-gem 'dor-services-client', '~> 4.0'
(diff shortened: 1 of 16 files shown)


justinlittman:
I'd suggest that it would be useful for devs.

jmartin-sul:
to andrew's concern about this being likely to puzzle many regular argo users:  we could just show it to argo super admins, and default it to collapsed.

jcoyne:
This is now behind a feature flag, so I don't think it'll confused anyone.

PR #1788, opened 2020-01-30
Show events on the item show page

jcoyne:
@andrewjbtw here is the event data. Should we stick it on the show page?
Screen Shot 2020-01-30 at 1 12 12 PM

## Why was this change made?

So we have a mechanism for using the event data.

n/a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,7 +68,7 @@ gem 'rubyzip'
 gem 'blacklight', '~> 6.0'
 gem 'blacklight-hierarchy', '~> 2.0'
 gem 'dor-services', '~> 9.0'
-gem 'dor-services-client', '~> 4.4'
+gem 'dor-services-client', '~> 4.7'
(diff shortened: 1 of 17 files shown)


andrewjbtw:
Is it possible to pull in user ids if an action was done by a person? I'm thinking mainly of registration and open/close versions. 

I don't know if it's possible to record ids when a system (like preassembly) is doing the open and close, but I would like to have that information recorded when it is available. That was about the only thing the old events datastream stored.

andrewjbtw:
In terms of actually showing this on the Argo page, I'd like @astridu's input on the design.

jcoyne:
Sure. I’ll make a ticket to record user on version events in dor-services-app

jcoyne:
This has been wrapped in a feature flag (`beta=1` in the URI params), so I think we can merge it now.

PR #1810, opened 2020-02-11
Avoid a race condition around accession of APOs

### LINKED ISSUE #1809: Create an APO has a race condition
**Describe the bug**
APO form kicks off the accessionWF in https://github.com/sul-dlss/argo/blob/master/app/forms/apo_form.rb#L26

However, it hasn't yet created the descMetadata, which happens https://github.com/sul-dlss/argo/blob/master/app/forms/apo_form.rb#L35

**User Impact**
descriptive_metadata Workflow step fails with "Dor::Services::Client::UnexpectedResponse: Unprocessable Entity: 422 (druid:rv794hh9709 had no resolvable identifiers: ["uuid:2f631e50-4cf3-11ea-962c-005056a7041d"])"
Very few APOs get created in Argo, so there isn't many people impacted right now.

**To Reproduce**
Run the integration tests.  May have to run multiple times as sometimes argo is faster than the workflow.

**Expected behavior**
descriptive_metadata should be created before the workflow is kicked off.

jcoyne:
## Why was this change made?
Do the writing first, then run the robots
Fixes #1809 

n/a

### DIFF
--- a/app/forms/apo_form.rb
+++ b/app/forms/apo_form.rb
@@ -21,13 +21,16 @@ class ApoForm < BaseForm
     @errors.empty?
   end
 
-  # Copies the values to the model and saves and indexes
+  # Copies the values to the model, saves and indexes, and starts workflow.
   def save
+    @needs_accession_workflow = false
     find_or_register_model
     sync
     add_default_collection
     model.save
     model.update_index
+    # Kick off the accessionWF after all updates are complete.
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@ndushay :speech_balloon: 
> it feels wrong to start the accessionWF before registering the object??

maybe I am reading it wrong... is the register on line 27 and the accessionWF kick-off on line 33?

ndushay:
I'm guessing I'm not looking at the full code context broadly enough;  it's already merged and we'll soon confirm that the code is fine (no complaints while using) or it's not (errors).  Laying my concern to rest.

PR #1811, opened 2020-02-14
Register using cocina

jcoyne:
## Why was this change made?
Testing how we can update with the cocina data model.

TODO:
- [ ] a way to add arbitrary tags.


app/controllers/dor/objects_controller.rb
```diff
@@ -1,3 +1,4 @@
+
```
mjgiarlo: :keyboard: :fire:

jcoyne:
Superseded by #1865

PR #1812, opened 2020-02-14
Sort versions as numbers in lifecycle display

mjgiarlo:
## Why was this change made?

It is surprising to scroll to the bottom of an object with 30+ versions and see the last entry, which one would reasonably assume to be the latest version, show up as version 9.

I didn't see an obvious place to add a test. Happy to take a suggestion, but I'm not inclined to sink a lot more time into this branch, and sorting of versions right now is doubleplusungood.

no

### DIFF
--- a/app/views/catalog/_show_milestones.html.erb
+++ b/app/views/catalog/_show_milestones.html.erb
@@ -4,27 +4,21 @@
   </thead>
   <tbody>
     <%
-      ## TODO: Pick a real date format for the date to_s calls.
-      ## Do it here and NOT in the model!!
       version_hash = document.get_versions
       milestones = document.milestones
-      milestones.keys.sort.each do |version|
-        steps=milestones[version]
-        first=true
+      milestones.keys.sort_by { |version_number_string| version_number_string.to_i }.each do |version|
+        steps = milestones[version]
         steps.each_pair do |k,m| -%>
-        <% if first
-            first=false
-        -%>
-          <tr style="display:none" class="last_step<%= version%>">
-            <td><span style="cursor: pointer;" onclick='$(".version<%= version%>").toggle();$(".last_step<%= version%>").toggle();'">+</span><%= version%></td>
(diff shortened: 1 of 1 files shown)


app/views/catalog/_show_milestones.html.erb
```diff
@@ -4,27 +4,21 @@
   </thead>
   <tbody>
     <%
-      ## TODO: Pick a real date format for the date to_s calls.
```
mjgiarlo: This looks bogus/outdated.

app/views/catalog/_show_milestones.html.erb
```diff
@@ -4,27 +4,21 @@
...
-      ## Do it here and NOT in the model!!
       version_hash = document.get_versions
       milestones = document.milestones
-      milestones.keys.sort.each do |version|
-        steps=milestones[version]
-        first=true
```
mjgiarlo: Variable is no longer used.

app/views/catalog/_show_milestones.html.erb
```diff
@@ -4,27 +4,21 @@
...
+      milestones.keys.sort_by { |version_number_string| version_number_string.to_i }.each do |version|
+        steps = milestones[version]
         steps.each_pair do |k,m| -%>
-        <% if first
-            first=false
```
mjgiarlo: OK, it's used here, but it's useless.

app/views/catalog/_show_milestones.html.erb
```diff
@@ -4,27 +4,21 @@
...
-        -%>
-          <tr style="display:none" class="last_step<%= version%>">
-            <td><span style="cursor: pointer;" onclick='$(".version<%= version%>").toggle();$(".last_step<%= version%>").toggle();'">+</span><%= version%></td>
+          <tr style="display:none" class="last_step<%= version %>">
```
mjgiarlo: Changes below here all whitespace/formatting-related.

PR #1822, opened 2020-02-19
Upgrade to dor-services 9

jcoyne:
## Why was this change made?

To stay consistent with the library used throughout the SDR.

n/a


jmartin-sul:
i didn't realize argo was still doing any (let alone so much!) direct updating of the solr index.

PR #1833, opened 2020-02-28
Remove ivar from test

jcoyne:
## Why was this change made?
Ivars are discouraged in rspec tests.

n/a

If so, please confirm:
- [ ] change was tested on stage    and/or
- [ ] test added to sul-dlss/infrastructure-integration-test


ndushay:
approved, but tests aren't passing.

PR #1841, opened 2020-03-04
We still need to deploy to -a and -b while they are in the pool

aaron-collier:
## Why was this change made?

This can be reverted once the old servers are removed from the lb pool.


config/deploy/prod.rb
```diff
@@ -1,7 +1,9 @@
-# frozen_string_literal: true
+ # frozen_string_literal: true
```
jcoyne: remove space

aaron-collier: Whoops, new keyboard is taking some getting used to... my bad. Fixed.

PR #1847, opened 2020-03-10
Deprecate the embargoable methods.

jcoyne:
## Why was this change made?

They are deprecated, because they are just extra indirection.

n/a
no


jmartin-sul:
@ndushay said

>  bad title for this PR

good catch, i took the liberty of amending the title (sorry @jcoyne if i got that wrong, feel free to re-fix if so)

PR #1855, opened 2020-03-18
Guard against an empty REMOTE_USER header

cbeer:


jcoyne:
This is because shib on centos 7 passes a blank header rather than omitting it.

PR #1856, opened 2020-03-18
Add a script to export OCR files for a list of barcodes

jcoyne:
## Why was this change made?

Fixes https://github.com/sul-dlss/google-books/issues/146

n/a

no

Tested on stage.

### DIFF
--- a/app/services/ocr_exporter.rb
+++ b/app/services/ocr_exporter.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+# This exports OCR files that are in preservation for a provided list of barcodes
+# This is used by researchers who want to get a dump of OCR files from the google books project
+class OCRExporter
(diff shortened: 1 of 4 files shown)


app/services/ocr_exporter.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def export
+    File.open(filename, 'r').each do |line|
+      druid = finder.find_druid(line.chomp)
+      unless druid
+        warn("no druid found for #{line.chomp}")
+        next
+      end
+
+      obj = Dor::Services::Client.object(druid).find
```
justinlittman: Skip if file already exists? Will make a retry faster.

jcoyne: How do we know what files are expected to exist without loading this?

Issue #1874, opened 2020-03-27
Download descriptive metadata bulk action fails on large sets (9000+ druids)


andrewjbtw:
**Describe the bug**
The descriptive metadata bulk action fails when a request contains more than ~10000 druids. Watching the bulk action progress on the server, it looks like the job times out at 8 minutes and stops. But on the bulk action interface, it appears that the job is still in progress even thought it has stopped:

<img width="1021" alt="Screen Shot 2020-03-27 at 1 15 14 PM" src="https://user-images.githubusercontent.com/8975583/77796731-11a3ad00-702d-11ea-92de-87725bde6567.png">

The log lists all the druids processed up to the point where the job stopped and does not report an errror.

**User Impact**
Users cannot download descriptive metadata for large collections with one bulk action. This is currently a blocker for @arcadiafalcone 

**To Reproduce**
1. Go to Bulk Actions
2. Select download descriptive metadata
3. Enter a list of 10,000+ druids
4. Wait until bulk action stalls.

**Expected behavior**
This bulk action should continue to run until it has processed every druid.

**Additional context**
It looks like this job does not make use of background changes that were implemented to make virtual object jobs handle 10,000+ druids. Virtual objects used to run into an 8 minute limit but no longer have that issue.

justinlittman:
Starting to suspect it might be a memory problem:
```

2020.03.30 12:58:30 INFO  -- [delayed_job:workers:delayed_job.1] check:memory(<1500Mb) [*1765Mb] => Fail
2020.03.30 12:58:30 ERROR -- [delayed_job:workers:delayed_job.1] NOTIFY: Bounded memory(<1500Mb): [*1765Mb] send to [:restart]
2020.03.30 12:58:30 INFO  -- [delayed_job:workers:delayed_job.1] schedule :restart (bounded memory(<1500Mb))
2020.03.30 12:58:30 INFO  -- [delayed_job:workers:delayed_job.1] => restart  (bounded memory(<1500Mb))
2020.03.30 12:58:30 INFO  -- [delayed_job:workers:delayed_job.1] switch :restarting [:up => :restarting] bounded memory(<1500Mb)
2020.03.30 12:58:30 INFO  -- [delayed_job:workers:delayed_job.1] executing: `bundle exec bin/delayed_job restart -i 1` with restart_timeout: 90.0s and restart_grace: 10.0s
[...]

PR #1877, opened 2020-03-30
Adds reset of db connections for DescmetadataDownloadJob.

justinlittman:
refs #1874

## Why was this change made?
DB connection was dropping since this was a long-running job.

No.

No.


justinlittman:
This does not fix the problem so closing.

PR #1880, opened 2020-03-31
Increased amount of memory available to a worker.

justinlittman:
refs  #1874

## Why was this change made?
To try to resolve issues with DescMetadata Download jobs failing.

No

No


justinlittman:
This does not fix the problem.

PR #1886, opened 2020-04-01
Remove temp fixture files that are in an ignored directory

mjgiarlo:
## Why was this change made?

I do not think these should have been added to version control, and they prevent systematic cleaning of this directory without making changes to the workspace.

no

no


mjgiarlo:
Damn, weird; Argo *does* depend on these files. :unamused:  Closing.

PR #1887, opened 2020-04-01
Reindex tags immediately after bulk update

### LINKED ISSUE #1884: Bulk updates for tags do not appear immediately
**Describe the bug**
Tags that are updated via the bulk updates interface are not reindexed after the update is run. The effect is to make it look like the tags were never updated. After the items are reindexed, the new tags appear.

**User Impact**
Users can be left thinking that their tag updates didn't work when the tags were actually updated correctly. They have to take an extra step of reindexing before they can use the tags.

**To Reproduce**
1. Go to the bulk updates page
2. Choose "Paste a druid list"
3. Paste in one or more druids into the box
4. Choose "Tags" under "Options that do not require versioning"
5. A "Change Tags" tags interface will appear below
6. Add a new tag or modify existing tags. Note that tags should be tab-separated.
7. Go to the druids that you updated.
8. Look at the tags. They'll probably be the old ones.
9. Reindex the druids.
10. See the new tags appear.

**Expected behavior**
Tag changes should be applied immediately.

**Additional context**
When editing tags for individual items, the tag changes appear immediately. Only bulk updates have this issue.

mjgiarlo:
Fixes #1884

## Why was this change made?

To fix a bug.

No.

Yes, but not in a way that should require extra testing.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -455,7 +455,12 @@ class ItemsController < ApplicationController
   end
 
   def reindex
-    Argo::Indexer.reindex_pid_remotely(@object.pid) unless params[:bulk]
+    # Skip reindexing all bulk operations *except* bulk tag operations which do
+    # require immediate reindexing since they do not touch Fedora (and thus do
+    # not send messages to Solr)
+    return if params[:bulk] && params[:tags].nil?
+
+    Argo::Indexer.reindex_pid_remotely(@object.pid)
   end
 
   # ---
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@jcoyne @justinlittman The CI build for this branch passed, btw: https://travis-ci.org/github/sul-dlss/argo/builds/669700138 Not sure why the check is showing as "expected." This PR is good to go AFAICT.

Do we want to test this in stage? Or is it straightforward enough to push out there given the current confusing behavior in prod?

PR #1889, opened 2020-04-01
Update rubocop with --auto-gen-config

aaron-collier:
## Why was this change made?

Updates rubocop and removed unnecessaily disabled cops.

N/A


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2019-12-16 11:00:09 -0600 using RuboCop version 0.74.0.
+# on 2020-04-01 08:40:11 -0700 using RuboCop version 0.80.0.
```
jcoyne, line 3: This isn't the version in the Gemfile.lock, is it?

aaron-collier: That's odd, we pin 0.74.0, why would running it locally use 0.80.0?

aaron-collier: Updating to 0.80.x

jcoyne: you probably ran `rubocop` rather than `bundle exec rubocop`

aaron-collier: Oh yup. Need to get into that habit.

mjgiarlo: Incidentally, this is one of the reasons why I still use `rvm` with per-project gemsets; I almost never have to use `bundle exec` because it does the right thing.

PR #1890, opened 2020-04-01
Periodically commit zip to avoid memory issues with DescMetadata Down…

justinlittman:
…load job.

refs #1874

## Why was this change made?
Reduce the memory used by this operation so eye doesn't kill it.

No.

Tested on prod.


jmartin-sul:
neat!  didn't know about that feature.

PR #1899, opened 2020-04-10
Update sass to 6.0

jcoyne:
## Why was this change made?
 Sass-rails 5 is eol

no

no


jcoyne:
@justinlittman no

PR #1901, opened 2020-04-10
Add WorkflowGridComponent

jcoyne:
## Why was this change made?

Refactor away from helpers and toward components

no

no


app/views/report/_workflow_grid.html.erb
```diff
@@ -1 +1,2 @@
-<%= render_workflow_grid %>
+<% # facet_tree is from blacklight-hierarchy %>
```
mjgiarlo, line 1: Remove space between `%` and `#` to make erblint happy?

PR #1905, opened 2020-04-14
Extract components for workflow-table and workflow-table-process

### LINKED ISSUE #1895: Index the "started" status on the workflow grid
The "started" status is not indexed in the workflow grid:

techmd-gb

**User Impact**
You have to infer which items are "started" based on how many are "waiting" at the next step.

**Additional information**

Now that items don't seem to get stuck at "queued", it may be best to replace "queued" with "started" to avoid further cluttering the grid.

(I noticed while investigating https://github.com/sul-dlss/technical-metadata-service/issues/129)

jcoyne:
## Why was this change made?

Follow the component pattern. Displayed started rather than queued. Fixes #1895

no

no


app/components/workflow_table_process_component.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class WorkflowTableProcessComponent < ApplicationComponent
+  # TODO
```
justinlittman, line 4: ?

spec/components/workflow_table_process_component_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe WorkflowTableProcessComponent, type: :component do
+  subject(:component) { described_class.new(workflow_table_process: [process, 'desc'], name: name, data: data) }
+  # it "renders something useful" do
```
justinlittman, line 7: Cruft?

PR #1910, opened 2020-04-15
Set the proper redis path in docker-compose

aaron-collier:
## Why was this change made?

Adds the redis url in docker compose to dor-services-app so that queues work properly.

N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,28 +72,6 @@ services:
       - dor-indexing-app
       - suri
       - redis
-
-  dor-services-worker:
-    image: suldlss/dor-services-app:latest
-    environment:
-      DATABASE_NAME: dor-services-app
-      DATABASE_USERNAME: postgres
-      DATABASE_PASSWORD: sekret
-      DATABASE_HOSTNAME: db
-      DATABASE_PORT: 5432
-      SOLR_URL: http://solr:8983/solr/argo
-      SETTINGS__DOR_INDEXING__URL: http://dor-indexing-app:3000/dor
-      SETTINGS__SOLR__URL: http://solr:8983/solr/argo
-      SETTINGS__FEDORA_URL: http://fedoraAdmin:fedoraAdmin@fcrepo:8080/fedora
-      SETTINGS__SURI__URL: http://suri:3000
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -67,6 +67,7 @@ services:
       SETTINGS__FEDORA_URL: http://fedoraAdmin:fedoraAdmin@fcrepo:8080/fedora
       SETTINGS__SURI__URL: http://suri:3000
       SETTINGS__WORKFLOW_URL: http://workflow:3000
+      REDIS_URL: redis://redis:6379/
     depends_on:
       - db
```
mjgiarlo, line 72: should `- redis` be added as a dependency here too?

aaron-collier: Yeah, a quick add is coming in too to add the worker to docker-compose as well.

docker-compose.yml
```diff
@@ -72,7 +72,32 @@ services:
...
       - dor-indexing-app
       - suri
+      - redis
 
+  dor-services-worker:
+    image: suldlss/dor-services-app:latest
+    ports:
+      - 3006:3000
```
mjgiarlo, line 80: I wonder if we can exclude ports since the queue UI ought to be available via dor-services-app image.

aaron-collier: .... that may be what I was just trying to resolve. Thanks!

jcoyne:
This seems to be moving even further from Argo. How far do we want the docker compose here to go? Should this be in a different repo just for the whole infrastructure?

aaron-collier:
@jcoyne I can see that... Maybe I should move it into integration test? Not sure that's the right place for it either. Without these changes, it's not much use to me right now however. So I'm open to suggestion.

jcoyne:
@aaron-collier I'm just seeing that the tests in Travis CI have to pull down a lot of images just to run the tests.  it doesn't seem like it has to be that way.

aaron-collier:
I hadn’t thought of that. Good point. I’ll follow up with this first thing.

aaron-collier:
@jcoyne dropped dor-services-worker from this PR

PR #1921, opened 2020-04-22
Reduce size of docker image by basing on Alpine

mjgiarlo:
## Why was this change made?

Faster images mean faster development and testing. Also brings consistency with our other images.

Includes:
* Add puma binstubs …
  * As part of this, exclude all bin scripts from Rubocop checks since they are generated files and rubocop can render them inoperable.
  * Also, re-generate rubocop TODOs while I am at it.
* Move puma command, invoked via binstub, straight into Dockerfile so it is the default command for the image. 
* Fix the documentation to use binstubs as well. (Ran into errors invoking rake, and preferring binstubs over bundle exec.)

Yes

No


Dockerfile
```diff
@@ -1,26 +1,33 @@
...
+RUN bundle config build.nokogiri --use-system-libraries && \
+  bundle config set without 'production' && \
+  bundle install
 
 COPY package.json yarn.lock ./
 RUN yarn install
+
+COPY . .
+
+CMD ["bin/puma", "-C", "config/puma.rb", "config.ru"]
```
aaron-collier, line 33: Naive question, does this not need the be prefixed with "bundle exec" as we've seen in other locations?

mjgiarlo: Discussed on slack and determined this was not a problem

PR #1923, opened 2020-04-23
set default agreement on edit apo page

### LINKED ISSUE #787: Problem with population of dropdown lists in APO editing screen
Per Geoff Willard:

If you take an APO (any APO probably), you’ll notice that when you bring up the edit screen the dropdown values for ‘Agreement’ and ‘Default Object Rights’ are always SPEC/UA and World respectively. If you make changes and save, the changes will persist, and the APO will respond correctly on the registration screen. However, if you go back into the APO to edit it, you may be fooled thinking you forgot to set the correct agreement or default rights. Or worse, you go in to make some other change, save it, and then don’t realize that you’ve inadvertently changed the agreement and default rights. Strange!

peetucket:
## Why was this change made?

fixes #787

n/a

no

### DIFF
--- a/app/forms/apo_form.rb
+++ b/app/forms/apo_form.rb
@@ -65,6 +65,12 @@ class ApoForm < BaseForm
     model.administrativeMetadata.ng_xml.xpath('//registration/workflow/@id').to_s
   end
 
+  def agreement
+    return if new_record?
+
+    model.agreement
+  end
+
   def use_license
     return '' if new_record?
 
--- a/app/views/apo/_form.html.erb
+++ b/app/views/apo/_form.html.erb
@@ -8,7 +8,7 @@
 
   <div class="form-group">
     <label>Agreement</label><br>
-    <%= select_tag :agreement, options_for_select(agreement_options) %>
+    <%= select_tag :agreement, options_for_select(agreement_options, @form.agreement) %>
   </div>
 
   <div class="form-group">
--- a/spec/forms/apo_form_spec.rb
+++ b/spec/forms/apo_form_spec.rb
@@ -114,6 +114,12 @@ RSpec.describe ApoForm do
       it { is_expected.to eq 'digitizationWF' }
     end
 
+    describe '#agreement' do
+      subject { instance.agreement }
+
+      it { is_expected.to eq agreement.pid }
+    end
+
     describe '#use_license' do
       subject { instance.use_license }
 


app/forms/apo_form.rb
```diff
@@ -65,6 +65,12 @@ def default_workflow
     model.administrativeMetadata.ng_xml.xpath('//registration/workflow/@id').to_s
   end
 
+  def agreement
+    return if new_record?
```
jcoyne, line 69: why do we need to check if it's a new record or not? Can we just `delegate :agreement, to: :model`?

peetucket:
There may be a separate issue where setting an agreement object in dor-services here for an object that already has an agreement is not clearing the old association from RELS-EXT causing a duplicate agreement object in RELS-EXT and ambiguity: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/admin_policy_object.rb 

Still trying to verify, but not sure where this actually happens.

app/forms/apo_form.rb
peetucket: makes sense

app/forms/apo_form.rb
```diff
@@ -7,6 +7,7 @@ class ApoForm < BaseForm
   DEFAULT_MANAGER_WORKGROUPS = %w(developer service-manager metadata-staff).freeze
 
   attr_reader :default_collection_pid
+  delegate :agreement, to: :model
```
jcoyne, line 10: It looks like this form was never intended to be used in the edit use case, because the `@model` is not set until `save` is called.

peetucket: Ahh, so that explains why it was done via an accessor method like the other attributes (and the way I had it originally).  When you call edit on the apo controller (https://github.com/sul-dlss/argo/blob/master/app/controllers/apo_controller.rb#L24-L28), it passes the model into the initializer where it gets set here: https://github.com/sul-dlss/argo/blob/master/app/forms/base_form.rb#L12-L15 and then delegation works, but on the new page, the model is not sent to the initializer and you get a test failure.

peetucket: reverting back to previous version without delegation

app/forms/apo_form.rb
peetucket: after second look, delegation won't work currently since we don't have an associated model on the new action

PR #1926, opened 2020-04-23
Add dissertation ID to the report view

### LINKED ISSUE #1433: Add dissertationid as a report view column option
As the ETD service manager, I need to pull reports from Argo for ETDs frequently. Dissertations are commonly identified by <dissertationid>. When I pull an Argo report for ETDs, I need to have the option to include <dissertationid> as a column header/with accompanying information. This information is contained in the DC datastream. Here is an example:

Screen Shot 2019-06-17 at 1 02 49 PM

aaron-collier:
## Why was this change made?

Fixes #1433

Adds a `Dissertation ID` column to the report output.

Screen Shot 2020-04-23 at 3 09 15 PM

N/A

No.


app/models/report.rb
```diff
@@ -141,6 +141,12 @@ class << self
         proc: lambda { |doc| doc.preservation_size },
         solr_fields: [SolrDocument::FIELD_PRESERVATION_SIZE],
         sort: false, default: true, width: 50, download_default: false
+      },
+      {
+        field: :dissertation_id, label: 'Dissertation ID',
+        proc: lambda { |doc| doc[:identifier_ssim].map { |id| id.include?('dissertationid') ? id.split(/:/).last : nil }.join },
```
aaron-collier, line 147: Note, the final `join` is required here because later in the code any array is joined by `;` characters, leaving the output here like `00000001234;;` so flattening into a string avoids the appending of `;` characters.

fedora_conf/data/load_order
```diff
@@ -49,3 +49,4 @@ dd327qr3670.xml
 fg464dn8891.xml
 pb873ty1662.xml
 qq613vj0238.xml
+bb007hx5508.xml
```
jcoyne: Can we write a test so that we don't have to load more fixtures?  These are very hard to manage and they're slow to load.

aaron-collier: Sorry, that was cruft and has been removed.

app/models/report.rb
jcoyne: use `filter { |id| id.include?('dissertationid')}.map { |id| id.split(/:/).last }` and then you won't have to `join`.

PR #1978, opened 2020-05-05
Set the page to 1 when downloading

### LINKED ISSUE #794: Report View: Download problems with druid list
In a collection with more than x items, if you go to "report view" from anything other than page 1 of the paginated results, then on-screen the list of druids looks correct but, when downloaded, the first y druids are replaced by the druids at the top of page z (whichever page you were on in the results view), leading to a list of druids that is missing y number of druids and has y number of duplicate druids.

[See Lynn's screenshots below for an example. The original example in this ticket was a search that no longer returns results.]

aaron-collier:
## Why was this change made?

Fixes #794 - without forcing the page to 1 on download, the current page param is sent and used as the initial page.

N/A

No.

### DIFF
--- a/spec/controllers/report_controller_spec.rb
+++ b/spec/controllers/report_controller_spec.rb
@@ -50,7 +50,7 @@ RSpec.describe ReportController, type: :controller do
         expect(response).to have_http_status(:ok)
         pids = JSON.parse(response.body)['druids']
         expect(pids).to be_a(Array)
-        expect(pids.length > 1).to be_truthy
+        expect(pids.length).to be > 1
         expect(pids.first).to eq('br481xz7820')
       end
     end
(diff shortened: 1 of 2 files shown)


aaron-collier:
@jcoyne added a quick test to the existing spec.

spec/controllers/report_controller_spec.rb
```diff
@@ -73,6 +73,17 @@
...
         expect(data[1].first).to eq('br481xz7820') # first data row starts with pid
       end
+
+      it 'downloads valid CSV data when not on the first page' do
+        get :download, params: { fields: ' ', page: 3 }
+        expect(response).to have_http_status(:ok)
+        expect(response.header['Content-Disposition']).to eq('attachment; filename=report.csv')
+        data = CSV.parse(response.body)
+        expect(data.first.length).to eq(26)
+        expect(data.length > 1).to be_truthy
```
jcoyne, line 83: can you use `expect(actual).to be >  expected`

Issue #165, opened 2015-11-11
Predictive Text for Tags on registration form


blalbrit:
When a user is on the registration page, they have currently have 4 tag boxes that they can interact with. The number of potential tags is overwhelming, so the user desire is to have auto-complete enabled so that they can quickly filter down to a desired tag or tag set. This would be similar to the functionality in the project tag box.

blalbrit:
Suggest backlogging this until a future iteration of Argo can address user interface/experience needs wholistically.

andrewjbtw:
Edited the title to clarify that this is for the registration form. A separate ticket could apply to the "edit tags" interface if we want to apply the predictive text to all individual tag edits.

PR #1993, opened 2020-05-08
Disable tag update

jcoyne:
## Why was this change made?

This temporarily disables tag updates until the table can be normalized: https://github.com/sul-dlss/dor-services-app/pull/854

This is being done so that #165 can be done.

no

no


mjgiarlo:
Heads-up, @andrewjbtw

PR #1994, opened 2020-05-08
Changes ocr_exporter to work with zips.

justinlittman:
refs https://github.com/sul-dlss/google-books/issues/511

## Why was this change made?
Switch from exporting by page to exporting by zip.

No

No

### DIFF
--- a/app/services/ocr_exporter.rb
+++ b/app/services/ocr_exporter.rb
@@ -2,6 +2,7 @@
 
 # This exports OCR files that are in preservation for a provided list of barcodes
 # This is used by researchers who want to get a dump of OCR files from the google books project
+# The OCR files for a book are zipped.
 class OCRExporter
   # @param [String] filename the input file name
   # @param [String] directory the output directory path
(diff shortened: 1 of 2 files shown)


app/services/ocr_exporter.rb
```diff
@@ -36,9 +37,8 @@ def export
 
   def filenames(object)
     object.structural.contains.map do |file_set|
-      # TODO: using label in place of filename for now: https://github.com/sul-dlss/google-books/issues/444
-      file_set.structural.contains.find { |file| file.hasMimeType == 'text/plain' }&.label
-    end.compact
+      file_set.structural.contains.map { |file| file.filename.match?(/-gb-txt.zip/) ? file.filename : nil }
```
jcoyne, line 40: You can simplify by using `filter` here rather than `map`

justinlittman: Still need `map`, so I'd call it a toss-up on "simplify". Changed.

app/services/ocr_exporter.rb
```diff
@@ -36,9 +37,8 @@ def export
 
   def filenames(object)
     object.structural.contains.map do |file_set|
-      # TODO: using label in place of filename for now: https://github.com/sul-dlss/google-books/issues/444
-      file_set.structural.contains.find { |file| file.hasMimeType == 'text/plain' }&.label
-    end.compact
+      file_set.structural.contains.filter { |file| file.filename.match?(/-gb-txt.zip/) }.map(&:filename)
+    end.flatten.compact
```
jcoyne, line 41: do we still need compact?

justinlittman: Nope.

app/services/ocr_exporter.rb
```diff
@@ -36,9 +37,8 @@ def export
 
   def filenames(object)
     object.structural.contains.map do |file_set|
```
jcoyne, line 39: change `map` to `flat_map` and we can drop `flatten`

PR #2003, opened 2020-05-11
Add a "back to seach" button on bulk actions index

### LINKED ISSUE #997: After completing a bulk action, provide option to return to previous search
" it would be useful if there were a way to go back to the previous search once having completed a bulk action."

From @nullhandle

aaron-collier:
## Why was this change made?

Fixes #997 

This adds a "Back to search" button to the top right of the Bulk actions list and utilizes the last search used before bulk actions so that it returns after a new bulk action is created.

Screen Shot 2020-05-11 at 1 50 40 PM

N/A

No.

### DIFF
--- a/app/views/bulk_actions/index.html.erb
+++ b/app/views/bulk_actions/index.html.erb
@@ -1,7 +1,16 @@
 <div class='container'>
 
   <h1>Bulk Actions</h1>
-  <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-success' %>
+  <div class="container">
+    <div class="row">
+      <div class="col-md-10">
+        <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-success' %>
+      </div>
+      <div class="col-md-2">
+        <%= link_back_to_catalog label: 'Back to search', class: 'btn btn-default' %>
+      </div>
+    </div>
+  </div>
 
   <table class='table'>
     <thead>
--- a/app/controllers/bulk_actions_controller.rb
+++ b/app/controllers/bulk_actions_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class BulkActionsController < ApplicationController
+  include Blacklight::SearchContext
+
   before_action :set_bulk_action, only: %i[destroy file]
 
   rescue_from ActiveRecord::RecordNotFound, with: -> { render plain: 'Record Not Found', status: :not_found }


app/views/bulk_actions/index.html.erb
```diff
@@ -1,7 +1,14 @@
 <div class='container'>
 
   <h1>Bulk Actions</h1>
-  <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-success' %>
+  <div>
+    <div style="float: left">
+      <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-success' %>
+    </div>
+    <div style="float: right">
+      <%= link_to 'Back to search', @last_search_url, class: 'btn btn-primary' %>
```
jcoyne, line 9: why is this a primary action?  Seems like it should be a `btn-default` (`btn-secondary` in bootstrap 4)

app/views/bulk_actions/index.html.erb
```diff
@@ -1,7 +1,14 @@
 <div class='container'>
 
   <h1>Bulk Actions</h1>
-  <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-success' %>
+  <div>
+    <div style="float: left">
```
jcoyne, line 5: Why does this need to float?  Can you use the grid layout instead?

app/views/bulk_actions/index.html.erb
aaron-collier: Sure, was just going for a different color, but happy to change it.

app/views/bulk_actions/index.html.erb
aaron-collier: Sure, this was quick, but look look at a grid layout.

app/views/bulk_actions/index.html.erb
```diff
@@ -1,7 +1,16 @@
...
 
   <h1>Bulk Actions</h1>
-  <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-success' %>
+  <div class="container">
+    <div class="row">
+      <div class="col-md-10">
+        <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-success' %>
+      </div>
+      <div class="col-md-2">
+        <%= link_to 'Back to search', @last_search_url, class: 'btn btn-default' %>
```
jcoyne, line 10: Could you use `link_back_to_catalog` here and avoid all the controller logic?

PR #2008, opened 2020-05-12
Port "Tags" bulk update to bulk action

### LINKED ISSUE #2007: [EPIC] Port "Tags" bulk update to bulk action
Part of #1914 

Port the updating of tags from a bulk update to a bulk action

- [ ] Unrevert the [reversion PR](https://github.com/sul-dlss/argo/pull/2053) (note `NOTE`s in code)
- [ ] Update the related infrastructure integration test. (see [draft PR](https://github.com/sul-dlss/infrastructure-integration-test/pull/96) as a start)

aaron-collier:
## Why was this change made?

Fixes #2007

Draft - getting started with this and expect cruft until complete.


app/jobs/set_tags_job.rb
```diff
@@ -0,0 +1,37 @@
...
+  def perform(bulk_action_id, params)
+    super
+
+    with_bulk_action_log do |log_buffer|
+      log_buffer.puts("#{Time.current} Starting #{self.class} for BulkAction #{bulk_action_id}")
+      update_druid_count
+
+      pids.map { |druid_with_tag| druid_with_tag.split(DELIM) }.each do |pid|
+        log_buffer.puts("#{Time.current} #{self.class}: Attempting to set tags for #{pid[0]} (bulk_action.id=#{bulk_action_id})")
+        set_tags(pid[0], pid[1], log_buffer)
```
mjgiarlo, line 19: Isn't this dropping all but the first tag? I ask because I'm not sure what the value of `pids` is on line 17. Is it just whatever the user puts in the textarea (via the default bulk action UI)? In that case, I assume pids looks something like `["druid:1\tTag : One\tTag : Two", "druid:2\tTag : 2\tTag : 3"]`

That means after the `#map` operation, the value of `pid` in the `#each` block looks like `["druid:1", "Tag : One", "Tag : Two"]` and then `["druid:2", "Tag : 2", "Tag : 3"]`. So `pid[0]` will be the druid and `pid[1]` will be the first tag. The other tags are ignored.

Would you consider something like:

```ruby
# ...
pids.map { |druid_with_tag| druid_with_tag.split("\t") }.each do |druid, *tags| 
  log_buffer.puts("#{Time.current} #{self.class}: Attempting to set tags for #{druid} (bulk_action.id=#{bulk_action_id})")
  set_tags(druid, log_buffer, *tags)
end
# ...

private

def set_tags(druid, log_buffer, *tags)
# ...
end
```

:question:

aaron-collier: It is (at the moment) - I made a comment in slack that I needed to address this issue, yes. It's incoming ;-)

app/javascript/packs/bulk.js
```diff
@@ -293,10 +293,6 @@ Blacklight.onLoad(()=>{
 	$('#apply-apo-defaults-button').on('click', () => $('#apply_apo_defaults').show(400))
 	$('#add-workflow-button').on('click', () => $('#add_workflow').show(400))
 	$('#close-versions-button').on('click', () => $('#close').show(400))
-  $('#show_tags').on('click', () => {
```
mjgiarlo, line 296: It looks like there may still be a bunch of tag-related code in bulk.js. I wonder if anything else can/should be ditched here.

app/views/report/bulk.html.erb
```diff
@@ -138,23 +131,14 @@ ul li{
...
-</div>
 <div class="bulk_operation" id="source_id">
   <h1>Change source id</h1>
   <p>This needs a list of pids and source ids, with a tab between the pid and the corresponding source id. E.g.:<br>
   druid:bc088fn5010  fuller:M1090_S15_B02_F12_0007
   </p>
   <p>This formatting is easier to perform by using Excel or a text editor, and then copying the result to this text area.</p>
   <textarea style="width:50%" id="source_ids" name="source_ids"></textarea><br>
-  <span class="bulk_button" id="set_source_id" name="set_source_id">Update source ids</span>
+  <button class="bulk_button" id="set_source_id" name="set_source_id">Update source ids</button>
```
mjgiarlo, line 141: Does this affect the UI? Or is it for Capybara purposes?

app/controllers/items_controller.rb
```diff
@@ -210,22 +209,6 @@ def catkey
     end
   end
 
-  def tags_bulk
```
mjgiarlo, line 213: Now that this is no longer a concern in the items controller, I wonder if lines 446-449 can change to be:

```ruby
return if params[:bulk]
```

app/views/bulk_actions/new.html.erb
```diff
@@ -84,6 +85,14 @@
           </span>
         </div>
 
+        <div role='tabpanel' class='tab-pane' id='SetTagsJob'>
```
mjgiarlo, line 88: Would you consider adding test coverage for the changes to this file to `spec/views/bulk_actions/new.html.erb_spec.rb`?

spec/jobs/set_tags_job_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  let(:object_client1) { instance_double(Dor::Services::Client::Object, administrative_tags: tags_client1) }
+  let(:object_client2) { instance_double(Dor::Services::Client::Object, administrative_tags: tags_client2) }
+  let(:tags_client1) { instance_double(Dor::Services::Client::AdministrativeTags) }
+  let(:tags_client2) { instance_double(Dor::Services::Client::AdministrativeTags) }
+
+  before do
+    allow(Dor::Services::Client).to receive(:object).with(druid1).and_return(object_client1)
+    allow(Dor::Services::Client).to receive(:object).with(druid2).and_return(object_client2)
+    allow(tags_client1).to receive(:replace).with(tags: tags1)
```
mjgiarlo, line 23: Consider moving these stubs into lines 17 and 18, like the stubs on lines 15 and 16?

spec/jobs/set_tags_job_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  after do
+    FileUtils.rm('foo.txt')
+  end
+
+  it 'adds tags to pids in list' do
+    expect(tags_client1).to have_received(:replace).with(tags: tags1)
+    expect(tags_client2).to have_received(:replace).with(tags: tags2)
+    expect(Argo::Indexer).to have_received(:reindex_pid_remotely).twice
```
mjgiarlo, line 40: :+1: dig the `#twice` usage here.

app/jobs/set_tags_job.rb
```diff
@@ -0,0 +1,40 @@
...
+  private
+
+  def set_tags(druid, log_buffer, *tags)
+    Dor::Services::Client.object(druid).administrative_tags.replace(tags: tags)
+    # tags require immediate reindexing since they do not touch Fedora (and thus do
+    # not send messages to Solr)
+    Argo::Indexer.reindex_pid_remotely(druid)
+
+    log_buffer.puts("#{Time.current} #{self.class}: Successfully set tags for #{druid} (bulk_action.id=#{bulk_action.id})")
+    bulk_action.increment(:druid_count_success).save
```
mjgiarlo, line 35: Can you add test coverage around expected values for `druid_count_success`?

app/jobs/set_tags_job.rb
```diff
@@ -0,0 +1,40 @@
...
+  rescue StandardError => e
+    log_buffer.puts("#{Time.current} #{self.class}: Unexpected error for #{druid} (bulk_action.id=#{bulk_action.id}): #{e}")
+    bulk_action.increment(:druid_count_fail).save
```
mjgiarlo, line 38: Can you add test coverage for the error case? In particular that the log buffer received the expected message and that the values for `druid_count_success` and `druid_count_fail` are expected?

app/views/report/bulk.html.erb
aaron-collier: Yes, the test fails with `span` because of "click_button".

mjgiarlo: OK, great!

PR #2016, opened 2020-05-12
Extract a VersionMilestonesComponent

jcoyne:
## Why was this change made?

This allows us to test the milestones view. It also fixes the collapse behavior which is broken in prod:
Screen Shot 2020-05-12 at 10 53 46 AM

no

no


jcoyne:
@ndushay There are no differences

ndushay:
> @ndushay There are no differences

that is awesome ... and communicating it, especially with pictures, isn't a bad thing ...

PR #2022, opened 2020-05-12
Document usage of eye to manage background workers

### LINKED ISSUE #768: README:  add info about eye's use in job monitoring
we use [eye](https://github.com/kostya/eye) to monitor delayed_job workers on the boxes to which argo's deployed.  we should mention at least basic info about that in the dependencies section of README.md.

mjgiarlo:
Fixes #768

## Why was this change made?

This commit better documents how Argo uses the eye gem to manage DelayedJob-based background workers. As part of this work, also:

* Stop using the capistrano3-delayed-job gem to manage delayed_job, since it does not have knowledge of eye. Instead, define our own eye-based Capistrano tasks in lib/capistrano/tasks/delayed_job.rake.
* Move custom Argo tasks to lib/capistrano/tasks/*.rake
* Use eye to start, restart, stop, and view status of delayed_job workers

Yes.

No, it does not. I tested the tasks individually, and in the usual deployment flow (both deploy and rollback scenarios), in the QA environment and everything worked flawlessly.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -33,41 +33,6 @@ set :linked_dirs, %w[log config/settings tmp/pids tmp/cache tmp/sockets vendor/b
 # Default value for keep_releases is 5
 # set :keep_releases, 5
 
-namespace :deploy do
-  # execute the eye script located in the deployed argo's bin directory, since
-  # eye may not be installed system-wide.  load the delayed_job_workers.eye config,
-  # which should monitor workers for memory consumption (restarting them individually
-  # if and when they exceed the configured threshold).
-  desc "stop/start eye, config for monitoring the deployment's delayed_job workers"
-  after :restart, :load_eye_dj_config do
-    on roles(:app) do
-      within release_path do
-        # :delayed_job_workers is set by the env specific cap configs.  it won't
-        # yet be set when this task is defined (though it will be by the time it's
-        # executed).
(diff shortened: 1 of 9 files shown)


mjgiarlo:
See also related DevOpsDocs change: https://github.com/sul-dlss/DevOpsDocs/pull/421

config/deploy.rb
```diff
@@ -33,41 +33,6 @@
...
-          execute :'./bin/eye', :quit
-          # avoid spaces in the command name, see http://capistranorb.com/documentation/getting-started/tasks/
-          execute :'./bin/eye', :load, :'config/eye/delayed_job_workers.eye'
-        end
-      end
-    end
-  end
-
-  desc 'Run rake yarn:install'
-  task :yarn_install do
```
jcoyne, line 61: Why is this being removed?  Is it unnecessary?

mjgiarlo: @jcoyne Not removed, but moved to its own `lib/capistrano/tasks/yarn.rake` to keep deploy.rb slim and have custom tasks defined in their own files.

mjgiarlo:
@jmartin-sul I will deploy to stage as another test.

jmartin-sul:
@mjgiarlo said:
> @jmartin-sul ~I will deploy to stage as another test.~
> 
> Confirmed it works as expected in stage. Thanks!

thanks!  and sorry, i had actually missed that you'd deployed to QA and tested there.  i'd have just merged had i noticed that.

PR #2037, opened 2020-05-13
The registration form is drawn by the registration controller

jcoyne:
## Why was this change made?

The form belongs with the controller.

- manually on laptop

n/a


aaron-collier:
@jcoyne needs a rebase

PR #2039, opened 2020-05-13
Add autocomplete of tags on the registration form

### LINKED ISSUE #165: Predictive Text for Tags on registration form

jcoyne:
## Why was this change made?

Fixes #165 

- manually on laptop

none

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ gem 'rubyzip'
 gem 'blacklight', '~> 6.0'
 gem 'blacklight-hierarchy', '~> 2.0'
 gem 'dor-services', '~> 9.4'
-gem 'dor-services-client', '~> 6.0'
+gem 'dor-services-client', '~> 6.3'
 gem 'dor-workflow-client', '~> 3.19'
 gem 'mods_display'
 gem 'okcomputer' # monitors application and its dependencies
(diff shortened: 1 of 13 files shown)


andrewjbtw:
This looks good except for one odd situation where the box turns red and the form gives an error.

To reproduce that error:

1. Start typing a tag until the suggestions pop up.
2. Hit enter without choosing a value. The text should turn partly red.
Screen Shot 2020-06-05 at 2 41 25 PM
3. Now choose a valid tag. The whole text will be red.
Screen Shot 2020-06-05 at 2 42 02 PM
4. Try to register. An error message will pop up.
Screen Shot 2020-06-05 at 2 49 06 PM

If you then clear the box and re-try by selecting the valid tag first, the problem goes away. Ideally, the box would re-set from the red error state as soon as the valid tag is selected. I discovered this accidentally by hitting enter when only one value was in the dropdown because I thought that would autofill it.

jcoyne:
@andrewjbtw I believe this is fixed, ping the infra team when you want this redeployed.

andrewjbtw:
Tested on stage and confirmed that this is fixed. I don't see any other issues with tag autocomplete so I think this is good to go.

PR #2044, opened 2020-05-15
add object to resource type list

### LINKED ISSUE #1906: Type "Object" missing from resource types in "Set content type"
In the "Set content type" options, for resources, the type "object" (used when a PDF is accessioned into a "book" object) is missing, making it impossible to use "Set content type" to update that resource to "document" or "file".

User Impact: forces the user to hand-edit the contentMD to achieve the desired results.

**To Reproduce**
1. Go to an item of type book that contains a PDF
2. Click on "Set content type" (after preparing the object)
3. Scroll down to 'Old resource type'
4. See that there is no type "Object"

**Expected behavior**
Expected behavior is to represent existing resource types within the drop down so that all can be edited if necessary.

**Screenshots**
Screenshot 2020-04-15 08 25 27

Screenshot 2020-04-15 08 25 11

peetucket:
## Why was this change made?

(re) fixes #1906 

previously done in #1912, but that PR was closed without merging

tested in -stage on 5/15/2020

none

### DIFF
--- a/lib/constants.rb
+++ b/lib/constants.rb
@@ -13,7 +13,7 @@ module Constants
 
   CONTENT_TYPES = %w[image book file map media document 3d].freeze
 
-  RESOURCE_TYPES = %w[image page file audio video document 3d].freeze
+  RESOURCE_TYPES = %w[image page file audio video document 3d object].freeze
 
   RELEASE_TARGETS = [
     %w[Searchworks Searchworks],


jmartin-sul:
@peetucket, actually, how about a follow on to update the related test:  https://github.com/sul-dlss/argo/blob/ee7aaec0e630cfba94fc428fb035c44211968e59/spec/lib/constants_spec.rb#L11-L14

peetucket:
> @peetucket, actually, how about a follow on to update the related test:
> 
> https://github.com/sul-dlss/argo/blob/ee7aaec0e630cfba94fc428fb035c44211968e59/spec/lib/constants_spec.rb#L11-L14

Good point, see #2046

Issue #924, opened 2017-01-23
Support Argo home page message or alert


LynnMcRae:
Support the ability to put up a temporary message to users on the home page to alert them to current problems, upcoming outages, etc.  Something an SDR admin could do by altering a resource without the need for deployment/restart or other developer/devOps involvement. 

This has been discussed in the past as a need for a general facility for potentially any of our apps, particularly those where processing is involved -- Hydrus availability issues, ETDs, Mirador -- but also as a means to make announcements about, say, important new features.

andrewjbtw:
This would be extremely useful not only for temporary alerts but for links to release notes and documentation of new functionality. We are making significant changes to Argo but have no way other than Slack or email to notify people.

astridu:
![image](https://user-images.githubusercontent.com/42216982/81609592-031b1600-938d-11ea-9d7b-826dac3d6062.png)

1. Display separate boxes for each message.
2. Yellow border and exclamation point (fa-exclamation-circle [&#xf06a;]): EAAB01, fill is 10% transparency
3. Black 'x' (fa-remove (alias) [&#xf00d;]) on right side of box allows user to close the message after viewing it. Message will not display again during this session, but will appear the next time user logs in.
4. Secondary messages appear in a bulleted list below yellow boxes.

jcoyne:
@astrid are you working on designs for how  one sets/removes/changes these messages?

astridu:
I am not. Is there a ticket for that, or should I put it in here?

jcoyne:
I think putting it in here is fine.

peetucket:
I could probably start working on this mocking out some UI.  Unless there is some generalized library tool we can use for this purpose (or we want to build one, which I think is not likely), I think it makes sense to put these messages in a simple local db table in argo, with a simple CRUD interface for admins to edit what's there, and a rails partial to show any active messages?

astridu:
Sorry for the delay, I just created a design for the admin UI that creates/removes/changes the messages.
![image](https://user-images.githubusercontent.com/42216982/81859289-db0ced80-9519-11ea-9818-92a58dd4d5ab.png)

astridu:
And Andrew approved.

jcoyne:
@astridu does that indicate we can remove the "feedback" and "Indexer Backlog: $x" items from the header?
[...]

PR #2045, opened 2020-05-15
Content blocks

jcoyne:


spec/models/content_block_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ContentBlock, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
jmartin-sul, line 6: cruft, or still left to do?

spec/requests/content_blocks_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  describe 'GET /content_blocks' do
+    before do
+      create(:content_block)
+      create(:content_block)
+    end
+
+    it 'displays the content blocks' do
+      get content_blocks_path
+      expect(response.body).to include 'Message alerts'
+    end
```
jmartin-sul, line 22: might be nice to also test that secondary alerts display below primary alerts, with their specific header text

jcoyne: we never test page layout in a request test. In fact it would be impossible to do so because CSS is not loaded.

spec/models/content_block_spec.rb
jcoyne: Completed.

app/models/content_block.rb
```diff
@@ -0,0 +1,28 @@
...
+
+class ContentBlock < ApplicationRecord
+  ORDINAL_STRING = { 1 => 'Primary', 2 => 'Secondary' }.freeze
+  scope :expired, -> { where('end_at < ?', current_time) }
+  scope :unexpired, -> { where('end_at >= ?', current_time) }
+  scope :active, -> { where('start_at < ? AND end_at >= ?', current_time, current_time) }
+  scope :primary, -> { where(ordinal: 1) }
+  scope :secondary, -> { where(ordinal: 2) }
+
+  validates :ordinal, presence: true, inclusion: { in: [1, 2] }
```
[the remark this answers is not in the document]
jcoyne: we have constraints on the database, so adding checking in the app would be redundant.

app/javascript/style/variables.scss
```diff
@@ -8,6 +8,8 @@
 $barley-corn: #b3995d;
 $floral-white: #f9f6ef;
 $tahuna-sands: #d2c295;
+$tahuna-sands: #d2c295;
```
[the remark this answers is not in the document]
jcoyne: Good catch. I've removed it.

app/models/content_block.rb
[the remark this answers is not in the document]
jmartin-sul: ah ok, i either missed the non-null db constraints yesterday, or those got added after i'd loaded the page (since i see GH saying "changed since last view").  apologies, agreed the db constraints are sufficient.

spec/requests/content_blocks_spec.rb
jmartin-sul: ah, good point

andrewjbtw:
@jcoyne I've created a specific workgroup for being able to edit these messages: sdr:argo-message-editors

jcoyne:
@andrewjbtw the problem is that prior to this pull request, we have given members of `sdr:administrator-role` the ability to do everything in Argo.  Are you asking me to remove that ability?  Do you want to create workgroups for all other types of access?

jcoyne:
@ndushay on the issue: https://github.com/sul-dlss/argo/issues/924#issuecomment-629711308

andrewjbtw:
So unfortunately, we're going to have to put this on hold until administrative permissions are worked out, which might not be for a long time. The problem with this feature vs. other features people have admin access to is that these messages are displayed prominently on the front page for everyone and too many (100+) people would have full edit access under the current permissions scheme. The risk is low that people would abuse their access, but it's the kind of thing where it would be hard to explain why it was left open if something did happen.

It's true that admin access means that many users could cause other types of problems, but I don't think any other change people can make would be displayed to all users as soon as they load Argo, and there are other permissions (like APOs) that limit the scope of what people can do. What we really need to do is rationalize permissions, which is something I'll be working on, first by mapping out the current workgroup situation. As I understand it, Argo was originally intended to be adminstrator-only, with not a large user base, which is probably why people kept getting added to the admin roles.

PR #2046, opened 2020-05-16
remove test for constants

peetucket:
## Why was this change made?

~~to update the test~~ to remove a test that goes with #2044 (that is not useful)

it is a test only

no


peetucket:
Yeah, it is an odd test.  And adding the new entry didn't cause to fail either (because it was an "includes")

jmartin-sul:
yeah, i could go either way on that.  i noticed right after merging #2044 that there was no modified test, and reflexively suggested modifying the most related looking one i found.  but agree this test isn't _super_ meaningful and probably ends up being more of a "are you sure you meant to remove this?" (since as peter points out, it's an `include` test, and doesn't fail on additions).

there is also this test that makes sure the `RESOURCE_TYPES` all actually get rendered:  https://github.com/sul-dlss/argo/blob/ee7aaec0e630cfba94fc428fb035c44211968e59/spec/views/content_types/_content_type.html.erb_spec.rb#L28-L30

so... up to @peetucket, i'd say.  remove if you think it's meaningless, merge as-is if you think it's worthwhile.

PR #2059, opened 2020-05-19
Remove unnecessary form field

jcoyne:
## Why was this change made?
This makes the form less confusing.

automated tests

none

### DIFF
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -11,7 +11,7 @@ class UploadsController < ApplicationController
   # POST /apos/:apo_id/uploads
   def create
     directory_name = Time.zone.now.strftime('%Y_%m_%d_%H_%M_%S_%L')
-    output_directory = File.join(Settings.bulk_metadata.directory, params[:druid], directory_name)
+    output_directory = File.join(Settings.bulk_metadata.directory, params[:apo_id], directory_name)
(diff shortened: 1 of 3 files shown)


app/controllers/uploads_controller.rb
```diff
@@ -11,7 +11,7 @@ def new
   # POST /apos/:apo_id/uploads
   def create
     directory_name = Time.zone.now.strftime('%Y_%m_%d_%H_%M_%S_%L')
-    output_directory = File.join(Settings.bulk_metadata.directory, params[:druid], directory_name)
+    output_directory = File.join(Settings.bulk_metadata.directory, params[:apo_id], directory_name)
```
ndushay, line 14: what if you're working with multiple druids for the same APO?

jcoyne: I'm not following.  Can you be more specific on how you would interact with Argo in order to do that?

ndushay: I wasn't thinking big picture - I was looking at this code change and wondering if we would ever have a problem because we are using the apo id for a directory name rather than the actual desired druid.  I don't know if the directory is exposed to the user, or how the user actions get to this code.

jcoyne: The parameter that was `druid` was always holding the same value as `apo_id`

Issue #183, opened 2015-11-13
Download all files in an object


rosy1280:
On an item page, a user should be presented with a way to download all the files currently listed in the contents section of an ojbect.

Here's a random druid with a content section so you can see what I mean. https://argo.stanford.edu/view/jz198xp6935

jcoyne:
I think this looks like a good way of doing the download: https://github.com/WeTransfer/zip_tricks

but I'm not sure what the user interface should look like.  Should this only be for download from the workspace or only from preservation? @astridu can you help?

astridu:
![image](https://user-images.githubusercontent.com/42216982/81346946-14ec7880-9070-11ea-8213-aaaadaad19e8.png)

Add the font awesome download icon (fa-download [&#xf019;]) plus 'Download all files' link in the gray 'Content' bar.

jcoyne:
@astridu Should there be a way to indicate if it's downloading from preservation or workspace?  Or if where they are downloading from preservation or workspace?

astridu:
Could they select if they want to download from preservation or workspace?

jcoyne:
I think it would be possible if they had uploaded new files that hadn't been reaccessioned. Maybe @andrewjbtw can help understand the cases we want to support.

andrewjbtw:
Huh, I was misunderstanding what "workspace" meant here. Somehow all these months I've read it as a category, not a location. As in "Preservation" and "Stacks" are both "Workspaces". It never occurred to me that "Workspace" is supposed to me the /dor workspace.

I think we should just take "Workspace" out entirely as an option. I've never seen anything downloadable from "Workspace" and I don't think it's a case we should support.

The two locations we enable for download are Stacks and Preservation. I think those should be separate options because there are use cases for downloading only the access files (Stacks).

andrewjbtw:
The Stacks link will appear for files that have shelve=yes. Also, Stacks enforces the no-download rules on everyone, even Argo users. I don't think we can do anything about that in the short run.

jcoyne:
@andrewjbtw I don't think argo has access to the stacks mount, so downloading from there would not be possible.

andrewjbtw:
It's already possible, but it must be following the public links via http. That must explain why the access controls are enforced.

mjgiarlo:
@jcoyne @andrewjbtw Where does that leave this ticket? Is download from preservation sufficient to satisfy this need for now? I assume the "Contents" that are being displayed are what's been preserved.

[...]

PR #2062, opened 2020-05-20
Remove download from workspace

jcoyne:
## Why was this change made?

Per Andrews request: https://github.com/sul-dlss/argo/issues/183#issuecomment-625550644

Test suite

none


jcoyne:
Andrew approved of this change in standup today.

PR #2063, opened 2020-05-20
Upgrade dor-services to 9.4.0 so that agreements on APOs are replaceable

### LINKED ISSUE #2061: When setting an agreement from the APO edit form, any previous agreement is not removed from RELS-EXT
When fixing #787, I noticed that if you select a new agreement when editing an APO, the new agreement gets added to RELS-EXT but the previous agreement does not get removed, leaving with you multiple agreements like this in the RELS-EXT:

`<hydra:referencesAgreement rdf:resource="info:fedora/druid:xf765cv5573"/>`

The edit APO form in Argo just picks the first agreement found, making it look like it didn't change.

The agreement is set in Argo here: https://github.com/sul-dlss/argo/blob/master/app/forms/apo_form.rb#L42

The bug is likely in dor-services gem itself, some spots to explore:

https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/admin_policy_object.rb#L33-L37

jcoyne:
## Why was this change made?

Fixes #2061 

Tested on stage

none

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,7 +64,7 @@ gem 'rubyzip'
 # Stanford/Hydra related gems
 gem 'blacklight', '~> 6.0'
 gem 'blacklight-hierarchy', '~> 2.0'
-gem 'dor-services', '~> 9.0'
+gem 'dor-services', '~> 9.4'
 gem 'dor-services-client', '~> 6.0'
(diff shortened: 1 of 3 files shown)


peetucket:
just pushed a commit to fix the tests

PR #2066, opened 2020-05-22
Download all files from preservation

### LINKED ISSUE #183: Download all files in an object

jcoyne:
## Why was this change made?

Allows downloading all files from preservation on argo. Fixes #183 

Tested on stage.

none


ndushay:
should there be an alert box to users when the content is > threshold size?   To make sure they meant to click it?

andrewjbtw:
Yes - if we don't put a limit on it, we should have a confirmation. I used to work with a system where a single click could start a massive download and it was surprisingly easy to unintentionally do it.

peetucket:
Looks like a small rebase is needed so we can merge and move forward with sul-dlss/dor-services-app#877

PR #2070, opened 2020-05-27
Add default tag to rights for set rights modal in item view

### LINKED ISSUE #446: Update setting rights forms to show current APO's default
Originally described in a follow on comment in https://github.com/sul-dlss/argo/issues/429#issue-125500022

Lynn McRae said:

>Ideal would be to match the presentation shown under Item Registration (including the indication of which setting matches the APO default), but that is secondary. It is only important to replace the "none" with the proper term.  The "Set object rights" bulk operation drop-down may share this problem, but that button is not currently active so I can't verify.

screen shot 2016-01-07 at 2 27 01 pm
screen shot 2016-01-07 at 2 28 09 pm

aaron-collier:
## Why was this change made?

Fixes #446 

This adds "(APO default)" to the proper rights selection in the modal for setting the rights on an item. If no APO is present, it leaves the list untouched.

Before:
Screen Shot 2020-05-28 at 10 04 12 AM

After:

Screen Shot 2020-05-28 at 9 50 48 AM

Unit test.

N/A

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -295,7 +295,7 @@ class ItemsController < ApplicationController
 
     apo_object = Dor.find(@apo)
     @form = AccessForm.new(cocina, apo_object)
-    
+
     respond_to do |format|
       format.html { render layout: !request.xhr? }
     end
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -292,9 +292,10 @@ class ItemsController < ApplicationController
     rescue Dor::Services::Client::UnexpectedResponse
       cocina = NilModel.new(params[:id])
     end
-    @form = AccessForm.new(cocina)
-    @rights_list = rights_list_for_apo(@apo)
 
+    apo_object = Dor.find(@apo)
+    @form = AccessForm.new(cocina, apo_object)
+    
     respond_to do |format|
       format.html { render layout: !request.xhr? }
     end
@@ -346,27 +347,6 @@ class ItemsController < ApplicationController
     return if params[:bulk] && params[:tags].nil?
 
     Argo::Indexer.reindex_pid_remotely(@object.pid)
-  end
-
-  def rights_list_for_apo(apo_id)
-    return Constants::REGISTRATION_RIGHTS_OPTIONS if apo_id.blank?
-
-    apo_object = Dor.find(apo_id)
-    default_opt = apo_object.default_rights || 'citation-only'
-
-    # iterate through the default version of the rights list.  if we found a default option
-    # selection, label it in the UI text and key it as 'default' (instead of its own name).  if
-    # we didn't find a default option, we'll just return the default list of rights options with no
-    # specified selection.
-    result = []
-    Constants::REGISTRATION_RIGHTS_OPTIONS.each do |val|
-      result << if default_opt == val[1]
-                  ["#{val[0]} (APO default)", val[1]]
-                else
-                  val
-                end
-    end
-    result
   end
 
   # ---
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -352,8 +352,7 @@ class ItemsController < ApplicationController
     return Constants::REGISTRATION_RIGHTS_OPTIONS if apo_id.blank?
 
     apo_object = Dor.find(apo_id)
-    default_opt = apo_object.default_rights
-    default_opt = 'citation-only' if default_opt == 'none'
+    default_opt = apo_object.default_rights || 'citation-only'
 
     # iterate through the default version of the rights list.  if we found a default option
     # selection, label it in the UI text and key it as 'default' (instead of its own name).  if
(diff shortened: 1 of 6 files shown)


app/controllers/items_controller.rb
```diff
@@ -347,6 +348,28 @@ def reindex
     Argo::Indexer.reindex_pid_remotely(@object.pid)
   end
 
+  def rights_list_for_apo(apo_id)
+    return Constants::REGISTRATION_RIGHTS_OPTIONS if apo_id.blank?
+
+    apo_object = Dor.find(apo_id)
+    default_opt = apo_object.default_rights
+    default_opt = 'citation-only' if default_opt == 'none'
```
ndushay, line 356: could this be:

```
default_opt = apo_object.default_rights || 'citation-only'
```

mjgiarlo: @ndushay @aaron-collier Maybe I'm not reading this right or I have an incomplete understanding, but the "before" code I'm seeing in the snippet above would set `default_opt` to `'citation-only'` if already set to `'none'`. The suggested change only sets to `'citation-only'` if `default_opt` is falsey, but `'none'` is truthy. What am I missing?

app/controllers/items_controller.rb
```diff
@@ -347,6 +348,27 @@ def reindex
     Argo::Indexer.reindex_pid_remotely(@object.pid)
   end
 
+  def rights_list_for_apo(apo_id)
```
jcoyne, line 351: Can this be moved into `AccessForm`?

app/controllers/items_controller.rb
```diff
@@ -347,6 +348,27 @@ def reindex
...
+    return Constants::REGISTRATION_RIGHTS_OPTIONS if apo_id.blank?
+
+    apo_object = Dor.find(apo_id)
```
jcoyne, line 354: But maybe this can be in the controller and passed to the form.

app/views/items/_rights.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= form_with model: @form, url: url_for(controller: :items, action: :set_rights, id: @object.pid) do |f| %>
   <div class='form-group'>
-    <%= f.select :rights, Constants::DEFAULT_RIGHTS_OPTIONS, {}, class: 'form-control' %>
+    <%= f.select :rights, @rights_list, {}, class: 'form-control' %>
```
jcoyne, line 3: then you can do `f.object.rights_list`

spec/controllers/items_controller_spec.rb
```diff
@@ -220,6 +220,61 @@
...
 
+  describe '#rights' do
+    before do
+      allow(Dor::Services::Client).to receive(:object).with(pid).and_return(object_client)
+    end
+
+    let(:object_client) { instance_double(Dor::Services::Client::Object, find: cocina) }
+    let(:cocina) { instance_double(Cocina::Models::DRO, version: 1) }
+
+    context 'without an assigned apo' do
```
jcoyne, line 231: All objects have an APO.  It is required.  This is not a case we would need to deal with.

aaron-collier: Is that guaranteed? One of the fixture objects doesn't have an APO (@apo is empty, not nil). Happy to remove if there isn't a reason we'd get there.

spec/controllers/items_controller_spec.rb
```diff
@@ -223,15 +223,55 @@
...
 
-    it 'renders the modal with rights selector' do
-      get :rights, params: { id: pid }
-      expect(response).to have_http_status(:ok)
+    context 'without an assigned apo' do
+      before do
+        get :rights, params: { id: pid }
+      end
+
+      let(:apo) { nil }
```
mjgiarlo, line 236: Hmm, where is this variable being used?

aaron-collier: In `@rights_list = rights_list_for_apo(@apo)`

mjgiarlo: I mean in that block of tests. I don't see references to `apo` until the next `context` block.

mjgiarlo: OK, I think I see now. Here? https://github.com/sul-dlss/argo/blob/amcollie_446/spec/controllers/items_controller_spec.rb#L20

aaron-collier: That's the one

spec/controllers/items_controller_spec.rb
jcoyne: The Ur-APO is the only object that is permitted to not have an APO.

app/controllers/items_controller.rb
ndushay: it was me missing stuff, maybe?

aaron-collier: I think your suggestion is safe, as 'none' isn't anything that is shown, and elsewhere in the code 'citation-only' is displayed as 'none. So I think this is just fine. Done.

aaron-collier:
@jcoyne I've moved things into the form, I think this addresses your request now. Thanks.

PR #2092, opened 2020-06-02
correctly show bulk job running status even if errors are present

### LINKED ISSUE #2068: Clearly indicate when a MODS bulk load is still running
**Describe the bug**
When a MODS bulk upload runs into errors while still processing, the interface displays the status as "system errors" before the job has been completed, when in fact more druids may still be getting processed.

**User Impact**
A status of "System errors" implies that a job has been completed, which can cause confusion if the job is not complete. This may result in 

1. Users not knowing when a job is complete so that they can start a new upload job that might fix the errors
2. Users starting a new job too early, while the previous job is still running.

**To Reproduce**
1. Go to the page for an APO
2. Run a bulk upload that generates errors part way through
3. Watch the druid count increment even after the status is changed to "system errors"

**Expected behavior**
The status should show that the job is still in process until it's completed.

**Additional context**
Reported by @arcadiafalcone in argo-feedback, using this APO as an example:
https://argo.stanford.edu/apos/druid:mf043wz1042/bulk_jobs

peetucket:
## Why was this change made?

fixes #2068 

This should show the status of bulk jobs correctly even if error are shown.

Still need to figure this out, maybe deploy to stage and run a couple example jobs, one with errors, one without.

None needed.

### DIFF
--- a/app/views/bulk_jobs/_bulk_index_table.html.erb
+++ b/app/views/bulk_jobs/_bulk_index_table.html.erb
@@ -8,25 +8,20 @@
       <th>Druids</th><th>Success</th><th /><th />
     </tr>
     <% @bulk_jobs.each do |job| %>
+      <% job_status = 'not started'
+         job_status = 'in progress' if job.key?('dir')
+         job_status = 'completed' if job['argo.bulk_metadata.bulk_log_job_complete']
+         job_status += ' (with system errors)' if job.key?('error') %>
       <tr>
         <td><%= job['argo.bulk_metadata.bulk_log_job_start'] %></td>
         <td><%= job['argo.bulk_metadata.bulk_log_user'] %></td>
         <td><%= job['argo.bulk_metadata.bulk_log_input_file'] %></td>
         <td><%= job['argo.bulk_metadata.bulk_log_note'] %></td>
-        <td>
-          <% if(job['argo.bulk_metadata.bulk_log_job_complete']) %>
-            completed
-          <% else %>
-            in process
-          <% end %>
-          <% if(job.has_key?('error')) %>
-            (with system errors)
-          <% end %>
-        </td>
+        <td><%= job_status %></td>
         <td><%= job['argo.bulk_metadata.bulk_log_record_count'] %></td>
         <td><%= job['argo.bulk_metadata.bulk_log_druids_loaded'] %></td>
 
-        <% if(job.has_key?('dir')) %>
+        <% if(job.key?('dir')) %>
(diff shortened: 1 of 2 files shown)


app/views/bulk_jobs/_bulk_index_table.html.erb
```diff
@@ -14,13 +14,14 @@
...
         <td>
-          <% if(job.has_key?('error')) %>
-            system errors
-          <% elsif(job['argo.bulk_metadata.bulk_log_job_complete']) %>
+          <% if(job['argo.bulk_metadata.bulk_log_job_complete']) %>
             completed
           <% else %>
             in process
           <% end %>
+          <% if(job.has_key?('error')) %>
```
jcoyne, line 22: Should this block be nested in the block on line 18?  Would anything ever be "In process with system errors"?

peetucket: > Would anything ever be "In process with system errors"?

Yes, apparently that is possible and that is the gist of the ticket that spawned this work (#2068) ... sorry forgot to add to the PR description originally.

I presume what is happening is that some of the rows in the spreadsheet succeed and some do not, those the system errors state while still in progress.

jcoyne: Can you write tests for these cases? You may consider doing a ViewComponent, as it's easier to test, but not required.

peetucket: i did a bit of refactoring of the view (but not a component), add a new status, and then added a test to the existing view test to test the status display

spec/views/bulk_jobs/_bulk_index_table.html.erb_spec.rb
```diff
@@ -62,4 +84,13 @@
     expect(rendered).to have_css 'div#bulk-upload-table table tr td', text: 'error:  job log dir not found', count: 3
     expect(rendered).to have_css 'div#bulk-upload-table table tr', text: 'error:  job log dir not found', count: 1
   end
+
+  it 'shows the status' do
+    render
+    expect(rendered).to have_css 'div#bulk-upload-table table tr td', text: /\Anot started\z/, count: 1
+    expect(rendered).to have_css 'div#bulk-upload-table table tr td', text: /\Acompleted\z/, count: 1
+    expect(rendered).to have_css 'div#bulk-upload-table table tr td', text: /\Ain progress\z/, count: 1
```
peetucket, line 92: using regex here to ensure the string is an exact match in the column

PR #2095, opened 2020-06-03
Do not display the edit xml link for datastreams that should not be editable

### LINKED ISSUE #1913: Remove "edit xml" button from datastreams people don't edit
remove from everything except RELS-EXT, descMetadata, identityMetadata, contentMetadata, versionMetadata, rightsMetadata(?)

This will prevent editing in places people don't need to edit and hopefully prevent errors.

aaron-collier:
## Why was this change made?

Fixes #1913 

When included in the settings as editable:

Screen Shot 2020-06-03 at 3 05 48 PM

When not included in the settings as editable:

Screen Shot 2020-06-03 at 3 06 11 PM

Unit test

config/settings.yml updated (no need to update shared_configs)

### DIFF
--- a/app/views/catalog/ds.html.erb
+++ b/app/views/catalog/ds.html.erb
@@ -21,9 +21,7 @@
         # If the user is a repo admin or is permitted to manage both content and
         # rights, display raw xml editing
     %>
-      <%= if can?(:manage_item, @obj) &&
-             @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) &&
-             Settings.editable_datastreams.include?(params[:dsid])
+      <%= if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid])
             link_to "Edit #{params[:dsid]}", ds_solr_document_path(params[:id], params[:dsid], edit: 'true'), title: params[:dsid], data: { blacklight_modal: 'preserve' }
         end %>
     <% content = if params[:dsid] == 'full_dc'
--- a/app/views/catalog/ds.html.erb
+++ b/app/views/catalog/ds.html.erb
@@ -22,10 +22,7 @@
         # rights, display raw xml editing
     %>
       <%= if can?(:manage_item, @obj) &&
-             (
-               @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) ||
-               params[:dsid] == 'RELS-EXT'
-             ) &&
+             @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) &&
              Settings.editable_datastreams.include?(params[:dsid])
             link_to "Edit #{params[:dsid]}", ds_solr_document_path(params[:id], params[:dsid], edit: 'true'), title: params[:dsid], data: { blacklight_modal: 'preserve' }
         end %>
--- a/app/views/catalog/ds.html.erb
+++ b/app/views/catalog/ds.html.erb
@@ -21,9 +21,9 @@
         # If the user is a repo admin or is permitted to manage both content and
         # rights, display raw xml editing
     %>
-      <%= if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid])
-            link_to "Edit #{params[:dsid]}", ds_solr_document_path(params[:id], params[:dsid], edit: 'true'), title: params[:dsid], data: { blacklight_modal: 'preserve' }
-        end %>
+      <% if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid]) %>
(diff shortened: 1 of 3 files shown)


app/views/catalog/ds.html.erb
```diff
@@ -26,7 +26,8 @@
                @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) ||
                params[:dsid] == 'RELS-EXT'
              ) &&
-             params[:dsid] != 'workflows'
+             params[:dsid] != 'workflows' &&
```
jcoyne, line 29: Can't this line be removed now as it's not in the list?

aaron-collier: Sure. dropped.

app/views/catalog/ds.html.erb
```diff
@@ -26,7 +26,7 @@
                @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) ||
                params[:dsid] == 'RELS-EXT'
              ) &&
```
jcoyne, line 28: Can you drop this conditional too, all of the datastreams you specified match this condition.

app/views/catalog/ds.html.erb
```diff
@@ -22,11 +22,8 @@
         # rights, display raw xml editing
     %>
       <%= if can?(:manage_item, @obj) &&
-             (
-               @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) ||
-               params[:dsid] == 'RELS-EXT'
-             ) &&
-             params[:dsid] != 'workflows'
+             @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) &&
```
jcoyne, line 25: This line can go too.  All of the things in editable_datastreams meet this condition.

aaron-collier: Oh, that's great. Will simplify the test a bit as well. Thanks. Commit incoming.

app/views/catalog/ds.html.erb
```diff
@@ -21,12 +22,7 @@
         # If the user is a repo admin or is permitted to manage both content and
         # rights, display raw xml editing
     %>
-      <%= if can?(:manage_item, @obj) &&
-             (
-               @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) ||
-               params[:dsid] == 'RELS-EXT'
-             ) &&
-             params[:dsid] != 'workflows'
+      <%= if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid])
```
jcoyne, line 25: the if should be after the "link_to" or make the if be in its own erb block. Then you can remove the rubocop disable stuff.

app/views/catalog/ds.html.erb
```diff
@@ -22,9 +21,7 @@
         # If the user is a repo admin or is permitted to manage both content and
         # rights, display raw xml editing
     %>
-      <%= if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid])
-            link_to "Edit #{params[:dsid]}", ds_solr_document_path(params[:id], params[:dsid], edit: 'true'), title: params[:dsid], data: { blacklight_modal: 'preserve' }
-        end %>
+      <%= render 'edit_datastream_xml' if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid]) %>
```
jcoyne, line 24: I think a partial here is giving a lot of unnecessary indirection?  Why not just do the link_to here?  Isn't this easier to read?
```suggestion
      <% if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid]) %>
        <%= link_to "Edit #{params[:dsid]}", ds_solr_document_path(params[:id], params[:dsid], edit: 'true'), title: params[:dsid], data: { blacklight_modal: 'preserve' } %>
      <% end %>
```

aaron-collier: Oh, I see, I misunderstood your earlier comment to mean I should extract that into a partial, not just a plan erb block... Whoops.

app/views/catalog/ds.html.erb
```diff
@@ -21,14 +21,9 @@
...
     %>
-      <%= if can?(:manage_item, @obj) &&
-             (
-               @obj.datastreams[params[:dsid]].respond_to?(:ng_xml) ||
-               params[:dsid] == 'RELS-EXT'
-             ) &&
-             params[:dsid] != 'workflows'
-            link_to "Edit #{params[:dsid]}", ds_solr_document_path(params[:id], params[:dsid], edit: 'true'), title: params[:dsid], data: { blacklight_modal: 'preserve' }
-        end %>
+      <%= if can?(:manage_item, @obj) && Settings.editable_datastreams.include?(params[:dsid]) %>
```
jcoyne, line 24: for `if` blocks there should not be an `=` prefix for the erb as this doesn't return a value.

aaron-collier: I really shouldn't code while having trouble sleeping... fixed.

PR #2099, opened 2020-06-05
versions should not be closable until an incompleted, active assemblyWF has completed.

### LINKED ISSUE #2093: The accessionWF should not run if there's an active assemblyWF
**Describe the bug**
It is possible to run the accessionWF when the assemblyWF is still active. This can happen if an item is already at v1 or higher. 

**User Impact**
This is a serious problem because the accessionWF deletes files from the active workspace. So if an item has an active assemblyWF and the accessionWF will run, then the files that the assemblyWF is supposed to process will be deleted. 

At a minimum, this means that users will need to start over to accession the item. The worst case scenario is that data is actually lost - for example, if a user has already deleted their copy because they thought accessioning was complete.

**To Reproduce**
1. Find an item that has already been accessioned once and has an active assemblyWF. For example: https://argo-stage.stanford.edu/view/druid:kt849tp3820
2. See that the status is "Opened"
3. See that the button for "Close version" is active on the left
4. If you close the version, the accessionWF will run

**Expected behavior**
The assemblyWF is a part of accessioning, even if it's not part of the accessionWF. The accessionWF should only run if the assemblyWF has been completed.

**Additional context**
If this is the first version of the object and it has not been accessioned yet, there is a different problem. In this situation, the object status remains "Registered" until the accesionWF runs. That is also a serious problem, to be ticketed on its own - #2094

ndushay:
## Why was this change made?

Part of #2093 - versions should not be closable until an incompleted, _active_ assemblyWF has completed.

wrote rspec tests.

n/a


.rubocop.yml
```diff
@@ -28,6 +28,9 @@ Layout/LineLength:
   IgnoredPatterns:
     - '\A\s*#' # allow long comments
 
+Lint/DeprecatedOpenSSLConstant:
+  Enabled: true
+
```
ndushay, line 33: I got a message that this needed to be set ... so I set it.

jcoyne:
I believe this is just handling "should we show the button?" but we probably would also need this validation where the version is actually closed: https://github.com/sul-dlss/dor-services-app/blob/38aaaea039f928d9410513346d3654b52dec8bb8/app/services/version_service.rb#L82-L84

peetucket:
> Agreed with Justin. I approve and think this should prevent the specific issue in the ticket, but we should also prevent it from happening in dor-services-app in the spot that Justin mentioned.

Which could be done in the same way (raise an exception if the assemblyWF:accessioning-initiate step is in waiting)...

ndushay:
I agree with @jcoyne ... but that will be a separate PR in a separate code base.  I will change this so it doesn't close the ticket and add a comment the above comment to the ticket, too.  Do you feel @andrewjbtw should test this in stage, or are you confident I got this right in argo?

peetucket:
I feel confident this looks ok.

ndushay:
From Andrew in slack with me just now:

> I don’t think I need to review it (in stage) - it should be pretty clear to test.

ndushay:
merging this, and will work on dor-services-app PR soon.

PR #2103, opened 2020-06-05
Draw datastreams without depending on ds_specs_ssim

### LINKED ISSUE #2100: Display of datastreams should not depend on ds_specs_ssim
Once done, remove this from the dor_indexing_app.

jcoyne:
## Why was this change made?

This makes the dataflow a lot easier to follow as there's no indirection via dor_indexing_app. This also allows for a smaller solr index. Fixes #2100

Tested locally on my laptop

n/a

### DIFF
--- a/app/components/datastream_row.rb
+++ b/app/components/datastream_row.rb
@@ -2,40 +2,38 @@
 
 # Draws a row for a datastream
 class DatastreamRow < ApplicationComponent
-  # @param [SolrDocument] document the Solr document model for the item
-  # @param [Hash] attributes the datastream attributes
-  def initialize(document:, attributes:)
-    @document = document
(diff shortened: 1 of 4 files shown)


justinlittman:
Can you provide some screenshots to provide context?

jcoyne:
@justinlittman there are no visual changes.

justinlittman:
Even if it doesn't result in visual changes, it does involve rendering.

jcoyne:
@justinlittman If nothing is changing, what would a picture show you?

justinlittman:
The context for the change. (That is, what part of the app if affected.)

jcoyne:
Picture attached above.

PR #2107, opened 2020-06-08
Remove dead code

jcoyne:
## Why was this change made?
This code is never used

Test suite

n/a


jcoyne:
Merging because tests are passing on travis, but Github is not showing the status.

PR #2111, opened 2020-06-08
remove all suri settings from argo (no longer used)

ndushay:
## Why was this change made?

To reduce confusion about suri configuration settings

test suite

shared_configs for qa, for stage, config files in app (see Files changed), and a PR for shared_configs for prod:  please merge sul-dlss/shared_configs/pull/1381 after this is merged.

### DIFF
--- a/config/initializers/dor_config.rb
+++ b/config/initializers/dor_config.rb
@@ -17,14 +17,6 @@ Dor.configure do
     url Settings.solrizer_url
   end
 
-  suri do
-    mint_ids     Settings.suri.mint_ids
-    id_namespace Settings.suri.id_namespace
-    url          Settings.suri.url
-    user         Settings.suri.user
-    pass         Settings.suri.pass
-  end
-
   stacks do
     local_workspace_root Settings.stacks.local_workspace_root
   end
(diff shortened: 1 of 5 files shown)


config/initializers/dor_config.rb
```diff
@@ -17,12 +17,9 @@
     url Settings.solrizer_url
   end
 
+  # SURI url currently only used for okcomputer;  it's essential for registering objects
```
jcoyne, line 20: I think we can remove the OKComputer check too. There is no direct dependency.

jcoyne:
It looks like a druid needs to be provided to `spec/features/item_view_metadata_spec.rb`

config/initializers/dor_config.rb
ndushay: will do.

ndushay:
Thanks, Justin.  I was hoping for the failing spec to magically go away, but a great steer works, too.  :-)

jcoyne:
We know about great steer in Minnesota

spec/features/item_view_metadata_spec.rb
```diff
@@ -125,6 +125,7 @@
       let(:dro_struct) { instance_double(Cocina::Models::DROStructural, contains: []) }
 
       before do
+        allow(item).to receive(:pid).and_return('fc123ky4567')
```
jcoyne, line 128: I'd suggest fixing it by providing the pid to the item (with the druid: prefix).

ndushay:
this is ready for re-review;  after merging, please merge sul-dlss/shared_configs/pull/1381

Issue #2104, opened 2020-06-05
Release bulk action jobs are not completing


andrewjbtw:
Multiple users are reporting that the Release objects bulk action is not completing. The bulk actions interface shows the jobs are processing but nothing seems to be happening.

Bulk actions known to have stalled in the past two days are

ReleaseObjectJob_10030
ReleaseObjectJob_10044

**Additional information**
I have not been able to determine a cause, but I found this in HB:
https://app.honeybadger.io/projects/49894/faults/62649217

One possibility is that the bulk action is running into problems when a list of druids contains a combination of items that have and have not been released once before. This can happen because users may submit a whole collection for release where some but not all items were previously released. There's no easy way to exclude the already-released items.

ndushay:
looked at HB error, and the log file is there, and it has the right permissions, and it is linked to at the directory level, not the individual file level ... like all the other logs.  

Is it possible delayed job runs somehow with diff uid ????

ndushay:
another thought:  what if job was running during log rotation -- would that cause the job to fail?  The timestamps of the errors in HB would suggest this is unlikely.

Unless the delayed job process itself held on to the file point to the workflow log file, and when that file was rotated, the HB time of occurrence somehow pertains to ... the delayed job process start time?

aaron-collier:
So this is strange, the file not found error in WorkflowClient happens if the file is empty. Once I dropped a test log line the action ran fine.

ndushay:
After pairing with Aaron, our thoughts:

We ssh'ed to argo-prod-02, and experimented in the rails console.  Aaron got the same failure when executing the line (https://github.com/sul-dlss/argo/blob/master/app/jobs/release_object_job.rb#L61) from `release_object_job` when the file was empty:

```
WorkflowClientFactory.build.create_workflow_by_name(current_druid, 'releaseWF', version: object.current_version)
```

However, when he went through the steps from WorkflowClientFactory (https://github.com/sul-dlss/argo/blob/master/app/services/workflow_client_factory.rb#L6-L7) , there was no problem, even tho the log file was the same, and still empty:

```
    logger = Logger.new(Settings.workflow.logfile, Settings.workflow.shift_age)
    Dor::Workflow::Client.new(url: Settings.workflow_url, logger: logger, timeout: Settings.workflow.timeout)
```

We also ran rails console on prod and were able to write log messages to an empty workflow_service.log file:

```
[...]

PR #2127, opened 2020-06-18
Switch from DelayedJob to Sidekiq

jcoyne:
Needs to have a redis hooked up in puppet.

## Why was this change made?

DelayedJob is not stopping jobs consistently (see #2104) and it is not consistent with the rest of our architecture.

To be tested on stage

Needs to have REDIS_URL set by puppet or shared configs.


README.md
```diff
@@ -198,29 +198,7 @@ and in development or testing mode:
...
-00:00 delayed_job:status
-      01 ./bin/eye info delayed_job
-      01 command :info, objects not found!
-      01 command :info, objects not found!
-    ✘ 01 lyberadmin@argo-qa-a.stanford.edu 0.801s
-    ✘ 01 lyberadmin@argo-qa-b.stanford.edu 0.813s
-```
-
-This means the eye daemon was shutdown and did not restart. To start it back up, invoke `cap {ENV} delayed_job:reload`. Note that this is for reloading eye, not for restarting the workers. You should not have to invoke this task with any regularity; it is for recovering from unusual circumstances only.
+Argo uses systemd to manage and monitor its Sidekiq-based background job workers in all deployed environments.
```
mjgiarlo, line 201: How would you feel about including a link to https://github.com/sul-dlss/dlss-capistrano#sidekiq-via-systemd here?

app/jobs/modsulator_job.rb
```diff
@@ -32,8 +31,7 @@ def perform(apo_id, uploaded_filename, output_directory, user, groups, filetype
     File.open(log_filename, 'a') do |log|
       start_log(log, user, original_filename, note)
 
-      # If a modsulator request fails, the job will fail and automatically be
-      # retried (see config/initializers/delayed_job.rb)
+      # If a modsulator request fails, the job will fail and automatically be  retried
```
mjgiarlo, line 34: Trivial niggle: there's an extra space between `be` and `retried`

lib/capistrano/tasks/delayed_job.rake
```diff
@@ -1,77 +0,0 @@
-# frozen_string_literal: true
```
mjgiarlo, line 1: :sob:

mjgiarlo:
@jcoyne see https://github.com/sul-dlss/puppet/pull/5550. Feel free to un-draft and coordinate w/ @sul-dlss/operations on getting the merged so it's in sequence w/ merging/deploying this PR. Cool?

jcoyne:
After I deployed this to stage, I had to manually kill eye and delayed_job.

mjgiarlo:
> After I deployed this to stage, I had to manually kill eye and delayed_job.

Maybe run `cap ENV delayed_job:stop` immediately before deploying? No biggie either way. Thanks for letting us know.

PR #2130, opened 2020-06-18
Set the correct attribute for data confirmation

jcoyne:
## Why was this change made?
currently we have the `confirm` attribute set, but we want this to be `data-confirm`

Locally.

n/a


andrewjbtw:
Confirming that this works. I think download all files is ready to go.

PR #2131, opened 2020-06-22
Remove unused file

jcoyne:
## Why was this change made?
This file is not used.

Test suite

n/a


jcoyne:
@ndushay we have two files that have the same name, and only one is used.  The file you pointed to is in the `app/views/shared`, but that is not the one I deleted.

ndushay:
@jcoyne I did look for the file in my argo repo and didn't see it:

```sh
 $ ll app/views/shared/
total 24
drwxr-xr-x+  5 ndushay  staff  160 Apr 28 12:46 .
drwxr-xr-x+ 24 ndushay  staff  768 Jun 11 09:03 ..
-rw-r--r--   1 ndushay  staff  407 Sep 17  2019 _ajax_modal.html.erb
-rw-r--r--   1 ndushay  staff   31 Apr  3 11:02 _footer.html.erb
-rw-r--r--   1 ndushay  staff  946 Apr 28 12:46 _header_navbar.html.erb
```

jcoyne:
@ndushay maybe your checkout is not up to date. That was added in the migration to blacklight 7.

Files are here: https://github.com/sul-dlss/argo/blob/master/app/views/_user_util_links.html.erb
and https://github.com/sul-dlss/argo/blob/master/app/views/shared/_user_util_links.html.erb

PR #2136, opened 2020-06-23
Enable browser to download files streamed from preservation

mjgiarlo:
## Why was this change made?

To fix a bug reported by @oceanofsound and confirmed by @andrewjbtw. The bug has two angles:

1. Argo's files controller begins writing to the response stream before sending the headers, which results in the browser not interpreting the request as a download (but as an inline request, which will render the file in the browser instead of prompting for download location).
2. Once the request has nearly finished, the response stream is not properly closed (possibly because of writing to the response after streaming has begun) and a 500 error is raised.

This branch fixes both bugs by sending headers before writing to the response stream and ensuring the response stream is closed.

Note to @sul-dlss/infrastructure-team reviewers: I'm not sure how to test this in a way that is both resilient and useful. Happy to get feedback on this, but I would like to fix this production bug ASAP so I am not inclined to sink a bunch of time into unit tests.

Tested in stage.

None.

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -35,21 +35,34 @@ class FilesController < ApplicationController
 
   def preserved
     authorize! :view_content, @cocina_model
-    Preservation::Client.objects.content(druid: @cocina_model.externalIdentifier,
-                                         filepath: filename,
-                                         version: params[:version],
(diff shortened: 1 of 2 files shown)


peetucket:
Rubocop issues...

PR #2138, opened 2020-06-23
[Refactor] Create a BulkActionForm

jcoyne:
No functional changes

## Why was this change made?

This separates responsibility for the form rendering from the model.  This means the model no longer needs to have virtual properties, and can be responsible solely for persistence.

Test suite

n/a

### DIFF
--- a/app/controllers/bulk_actions_controller.rb
+++ b/app/controllers/bulk_actions_controller.rb
@@ -14,21 +14,16 @@ class BulkActionsController < ApplicationController
 
   # GET /bulk_actions/new
   def new
-    @bulk_action = BulkAction.new
+    @form = BulkActionForm.new(BulkAction.new, groups: current_user.groups)
     @last_search = session[:search].present? ? searches_from_history.find(session[:search]['id']) : searches_from_history.first
   end
 
   # POST /bulk_actions
   def create
     # Since the groups aren't persisted, we need to pass them here.
-    @bulk_action = BulkAction.new(
-      bulk_action_params.merge(user: current_user,
-                               groups: current_user.groups,
-                               pids: pids_with_prefix(bulk_action_params[:pids]))
-    )
+    @form = BulkActionForm.new(BulkAction.new(user: current_user), groups: current_user.groups)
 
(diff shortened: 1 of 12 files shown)


app/controllers/bulk_actions_controller.rb
```diff
@@ -14,21 +14,16 @@ def index
 
   # GET /bulk_actions/new
   def new
-    @bulk_action = BulkAction.new
+    @form = BulkActionForm.new(BulkAction.new, groups: current_user.groups)
```
mjgiarlo, line 17: I wonder whether the `BulkActionForm` ought to be responsible for instantiating a `BulkAction`, given a user instance. From that user instance, the form can create the instance and get the groups. I don't think it's a big deal, though.

app/views/bulk_actions/new.html.erb
```diff
@@ -1,13 +1,13 @@
...
-    <% if @bulk_action.errors.any? %>
+  <%= form_for(@form, html: { 'data-controller' => 'bulk_actions' }) do |f| %>
+    <% if @form.errors.any? %>
       <div id="error_explanation">
-        <h2><%= pluralize(@bulk_action.errors.count, 'error') %> prohibited this bulk_action from being saved:</h2>
+        <h2><%= pluralize(@form.errors.count, 'error') %> prohibited this bulk_action from being saved:</h2>
 
         <ul>
-        <% @bulk_action.errors.full_messages.each do |message| %>
+        <% @form.errors.each do |message| %>
```
mjgiarlo, line 10: Are the "full" messages still yielded here?

app/controllers/bulk_actions_controller.rb
jcoyne: I don't want to do that because it changes the pattern that all forms use (and the pattern that reform uses)

mjgiarlo: Makes sense.

app/views/bulk_actions/new.html.erb
jcoyne: I've restored this.

PR #2141, opened 2020-06-24
Move the Download All button to the right

### LINKED ISSUE #2134: Small fixes to new "download all files" option
**Describe the bug**
1. The "download all files" option in Argo is appearing on the left-hand side of the "Contents" title bar. But in stage testing it had appeared on the right-hand side. I think it's easier to see if it's on the right:

Screen Shot 2020-06-23 at 12 25 31 PM

2. Clicking on the option collapses the "Contents" list. If possible, clicking should leave the list open. This may have happened in testing but I didn't notice it.

jcoyne:
And prevent the accordion from collapsing.

Now looks like this:
Screen Shot 2020-06-25 at 10 19 18 AM

## Why was this change made?
Fixes #2134

Tested locally

n/a

### DIFF
--- a/app/components/download_all_button_component.html.erb
+++ b/app/components/download_all_button_component.html.erb
@@ -1,0 +1,5 @@
+<span class="section-head-link">
+  <small>
+    <%= link_to 'Download all files', download_item_files_path(document), link_options %>
+  </small>
+</span>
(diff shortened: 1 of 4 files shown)


justinlittman:
Can you provide a screenshot?

jcoyne:
@justinlittman added.

PR #2142, opened 2020-06-25
[Refactor] Begin removing fixtures

jcoyne:
No functional changes.

## Why was this change made?

Fixtures is a Fedora 3 concept that we need to distance ourselves from.  This also helps by moving the dependencies of each test into the tests themselves rather than depending on the preexisting contents of the repository.

Test suite

n/a


fedora_conf/data/druids
```diff
@@ -19,7 +19,6 @@ druid:bw071ht4530   MÃ©moires sur la vie privÃ©e de Marie-Antoinette, reine
 druid:bt855nf7073   English land and English landlords: an enquiry into the origin and character of the English land system, with proposals for its reform
```
jcoyne, line 19: eventually we will be able to get rid of all of these.

PR #2155, opened 2020-06-29
Call the workflow service directly

jcoyne:
## Why was this change made?
This gives us up to date information immediately and removes the necessity of the workflow service to index this field

Tested on stage

n/a


spec/services/workflow_service_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+            <process version="2" priority="0" note="" lifecycle="" laneId="default" elapsed="0.0" attempts="0" datetime="2019-01-28T20:48:17+00:00" status="completed" name="approve"/>
+            <process version="2" priority="0" note="" lifecycle="" laneId="default" elapsed="0.0" attempts="0" datetime="2019-01-28T20:48:18+00:00" status="completed" name="start-assembly"/>
+          </workflow>
+          <workflow objectId="druid:ab123cd4567" id="versioningWF">
+            <process version="2" priority="0" note="" lifecycle="opened" laneId="default" elapsed="" attempts="0" datetime="2019-01-28T20:48:16+00:00" status="completed" name="start-version"/>
+            <process version="2" priority="0" note="" lifecycle="" laneId="default" elapsed="" attempts="1" datetime="2019-01-28T20:48:16+00:00" status="completed" name="submit-version"/>
+            <process version="2" priority="0" note="" lifecycle="" laneId="default" elapsed="" attempts="1" datetime="2019-01-28T20:48:16+00:00" status="completed" name="start-accession"/>
+          </workflow>
```
ndushay, line 56: does this need to be so long?   seems like 2-3 WF would be enough for test, no?

jcoyne: Probably not, this was just copied from https://github.com/sul-dlss/dor_indexing_app/blob/3be90132d5d3c830b0ca65b2594cc4072890a51d/spec/indexers/workflows_indexer_spec.rb#L12-L62 to make sure we did the equivalent correctly.

jcoyne: I've made that change.

ndushay: thank you!

PR #2156, opened 2020-06-29
Add a field type for structured values.

jcoyne:
## Why was this change made?

This will enable structured data to be indexed, without having to use delimiters as we currently do.

Test suite.

solr_conf/conf/schema.xml


jcoyne:
@ndushay I'm following what the access team is doing here: https://github.com/sul-dlss/sul-solr-configs/blob/master/searchworks-prod/schema.xml#L273

mjgiarlo (approved):

PR #2162, opened 2020-06-30
Componentize facets to avoid Blacklight deprecation

jcoyne:
## Why was this change made?

Not using components is deprecated and issues warnings.

Test suite and locally.

n/a


app/components/date_choice_facet_component.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= render(@layout.new(facet_field: @facet_field)) do |component| %>
+  <% component.with(:label) do %>
+    <%= @facet_field.label %>
+  <% end %>
+  <% component.with(:body) do %>
+    <ul class="facet-values list-unstyled">
+      <%= render_facet_limit_list @facet_field.paginator, @facet_field.key %>
+    </ul>
```
cbeer, line 8: I think ideally you'd be able to do:

```suggestion
    <%= render Blacklight::FacetFieldListComponent.new(@facet_field, layout: false)
```

I don't know if the `more >` link that's in there for backwards compat. presents a problem (and maybe the backwards compat needs to be explicit instead of keying off the layout?)

peetucket:
rebased and force-pushed to get latest bug fixes on master branch .... planning a test deploy to prod to see if it improves speed (seems fine on -stage as a first test)

peetucket:
It all looks the same to me on -stage.

ndushay:
any infrastructure-integration-test code that exercises these facets?   (I doubt it)

peetucket:
I don't think there are integration tests that specifically target the facets in argo.

peetucket:
Verified as faster on -stage and -prod

PR #2168, opened 2020-07-06
Revert "Revert "Inline the render_ds_profile_header"" 

jcoyne:
## Why was this change made?
 
Re-add the changes that were previously added in #2157 and reverted in #2165. Now with tests and fixes.

Test suite and tested on stage.

n/a


spec/controllers/datastreams_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    allow(item).to receive_messages(save: nil)
+    allow(Argo::Indexer).to receive(:reindex_pid_remotely)
+  end
+
+  let(:pid) { 'druid:bc123df4567' }
+  let(:item) { Dor::Item.new pid: pid }
+  let(:user) { create(:user) }
+  let(:state_service) { instance_double(StateService, allows_modification?: true) }
+
+  describe '#datastream_update' do
```
mjgiarlo, line 18: Should this now be `'#update'` instead?

jcoyne: Yes, fixed.

mjgiarlo: :beers:

PR #2174, opened 2020-07-06
Switch to using an APO factory rather than a fixture

jcoyne:
## Why was this change made?
We're trying to get rid of FOXML fixtures because they are coupled to Fedora 3, which we are trying to decouple from

Test suite

n/a


spec/controllers/registrations_controller_spec.rb
```diff
@@ -244,6 +244,7 @@
 
       let(:col_ids_for_apo) { ['druid:pb873ty1662'] }
       let(:apo_id) { 'druid:fg464dn8891' }
+      # let(:apo) { instance_double(Dor::AdminPolicyObject) }
```
ndushay, line 247: remove this?

jcoyne: Thank you. I will remove.

PR #2179, opened 2020-07-09
clean up remaining cruft from delayed_job

jmartin-sul:
* drop supporting database table
* clean up stray references in code and comments
* a couple opportunistic touchups on `ModsulatorClient` (better constructor param comment, DRY up logging a bit)

## Why was this change made?

happened to look at the schema and notice the `delayed_jobs` table, which seemed like cruft in light of the switch to sidekiq.

unit tests (i wasn't able to get the unit tests working locally just now, would like to pair on that if someone has time, but this change seemed safe and limited enough that i figured i'd throw up a PR and see if CI passed)

none

### DIFF
--- a/app/services/modsulator_client.rb
+++ b/app/services/modsulator_client.rb
@@ -25,7 +25,7 @@ class ModsulatorClient
 
   # @param [String] uploaded_filename The full path to the Spreadsheet/XML file.
   # @param [String] pretty_filename A prettified version of the filename. Modsulator writes this to xmlDocs[@sourceFile].
-  # @param [#puts] log the logger to write to
+  # @param [#puts(String)] log handle to output stream from caller (e.g. to provide user with job-specific log file)
   def initialize(uploaded_filename:, pretty_filename:, log:)
     @uploaded_filename = uploaded_filename
     @pretty_filename = pretty_filename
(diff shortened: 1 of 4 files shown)


app/services/modsulator_client.rb
```diff
@@ -69,20 +69,15 @@ def call_remote_modsulator(url, content_type)
...
-    log.puts "argo.bulk_metadata.bulk_log_nonexistent_file #{e.message}"
+    log_request_error(e, url, 'argo.bulk_metadata.bulk_log_nonexistent_file')
   rescue Errno::EACCES => e
-    delayed_log_url(e, url)
-    log.puts "argo.bulk_metadata.bulk_log_invalid_permission #{e.message}"
+    log_request_error(e, url, 'argo.bulk_metadata.bulk_log_invalid_permission')
   rescue Faraday::Error => e
-    delayed_log_url(e, url)
-    log.puts "argo.bulk_metadata.bulk_log_internal_error #{e.message}"
+    log_request_error(e, url, 'argo.bulk_metadata.bulk_log_internal_error')
```
jmartin-sul, line 78: diff coverage is failing because this is the only tested case out of these refactored error handling code paths.  these are all pretty similar calls, the tested usage works as it did before, and overall coverage seems to have gone up just a bit since a few lines of code were dropped.   but also, happy to add simple tests for the other error cases if people prefer.

PR #2181, opened 2020-07-10
Adds xsd and local imports in mods-3-6.xsd

### LINKED ISSUE #2180: Modsulator jobs failing with ERROR: attribute use (unknown), attribute 'ref'
**Describe the bug**
Modsulator jobs in stage and prod are failing with the error:

`1508:0: ERROR: attribute use (unknown), attribute 'ref': The QName value '{http://www.w3.org/XML/1998/namespace}lang' does not resolve to a(n) attribute declaration.`

**User Impact**
It's not possible to bulk update metadata.

**To Reproduce**
1. Go to an APO
2. Click on MODS bulk loads
3. Load a spreadsheet that will change the descriptive metadata on a druid.
4. See Argo report an error as "Exception"
5. Go to the /workspace/bulk folder on the Argo server and look at the log.txt file for that upload job
6. See the full error message

**Expected behavior**
Descriptive metadata gets updated.

**Additional context**
You won't see an error if the spreadsheet data matches what's already there. In that case, the modsulator job just skips the druid without attempting to update.

This issue on stage and prod was first reported this morning. It was not ocurring yesterday morning.

jermnelson:
## Why was this change made?
Fixes #2180 

testing suite

n/a

### DIFF
--- a/app/services/mods_validator.rb
+++ b/app/services/mods_validator.rb
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 class ModsValidator
-  SCHEMA = 'app/helpers/xml/mods-3-6.xsd'
+  SCHEMA = 'mods-3-6.xsd'
   def self.validate(doc)
-    xsd = Nokogiri::XML::Schema(File.read(SCHEMA))
-    xsd.validate(doc).map(&:message)
+    Dir.chdir('app/helpers/xml') do
+      xsd = Nokogiri::XML::Schema(File.open(SCHEMA))
+      xsd.validate(doc).map(&:message)
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/mods_validator.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 class ModsValidator
-  SCHEMA = 'app/helpers/xml/mods-3-6.xsd'
+  SCHEMA = 'mods-3-6.xsd'
```
jcoyne, line 4: Please revert this and the choir. They are not necessary

jermnelson: I think this required. Initially I didn't change any code in `mods_validator.rb` class but it failed both the build and when Andrew tested on stage.

jcoyne: @jermnelson I think this is because we haven't specified the full paths to the schemas in the XML, so

PR #2182, opened 2020-07-13
Update dependencies

sul-devops-team:


jcoyne:
This last test seems like a problem in dor-services-app. It is not returning a valid cocina model since it doesn't have an adminPolicy.

When I log in I can do this:
```
irb(main):009:0> apo = Dor.find('druid:db758kj5615')
irb(main):010:0> apo
=> #<Dor::AdminPolicyObject pid: "druid:db758kj5615", admin_policy_object_id: nil, agreement_object_id: nil>
```

Which is odd because Fedora shows that object to have an APO:

Screen Shot 2020-07-13 at 3 09 29 PM

mjgiarlo:
@jcoyne :speech_balloon: 
> 
> Which is odd because Fedora shows that object to have an APO:
> 
> Screen Shot 2020-07-13 at 3 09 29 PM

Not sure if related or not but I'd expect to see a bunch of XML namespaces declared as attrs on the `rdf:RDF` element. Malformed XML? Or perhaps just not rendered in the view you screenshotted.

jcoyne:
@mjgiarlo I think it's because it's not in the Solr index.

PR #2184, opened 2020-07-15
Run circle without docker-compose

jcoyne:
## Why was this change made?

This should result in faster build times and aligns with hydra_etd

test suite

circle.yml


aaron-collier:
👍 to this. If we get away from the machine image, I think we'll end up saving quite a bit of compute credits at circle.

PR #2194, opened 2020-08-01
show error message for malformed XML instead of raising an exception

### LINKED ISSUE #2183: No warning given anymore when saving malformed XML for datastream (on the client side)
**Describe the bug**
Per a report from one of our users, if malformed XML was entered when editing a datastream, a "warning would appear and it would not allow you to save your edits in the datastream until you fixed the XML."  But upon re-trying this just now, the user "can verify that the warning message is not showing up."  Instead, users get redirected to the library system status page (on prod, or a generic 500 error page on stage).  Also, the attempted edit isn't saved.

**User Impact**

Some of our datastreams can have lengthy and complex XML, and our systems sometimes necessitate editing them by hand.  As such, it's easy to produce malformed XML, and the loss of this check makes it easy to attempt to save bad XML, which results in getting bounced to a perplexing 500 error page and losing your edits.

**To Reproduce**
1. Open a datastream for editing (in the case we just saw, it was `descMetadata`)
2. Edit it so that the XML is no longer well formed (e.g. add an opening tag without a corresponding closing tag)
3. Try to save the datastream
4. Get a 500 error with no explanation (bouncing to system status page on prod, or generic 500 error page on stage).

**Expected behavior**

Upon attempting to save the bad XML, a warning is popped up, the XML as edited is retained but not saved, and the user is allowed to continue editing until well formed XML is produced.

**Screenshots**

e.g. you tried to close a `<mods>` open tag with a `</modsssssssss>` closing tag, hit update without realizing your mistake...

Screen Shot 2020-07-14 at 2 32 39 PM

and cryptic error message results...

Screen Shot 2020-07-14 at 2 32 56 PM

**Additional context**

peetucket:
## Why was this change made?

Fixes #2183 

Show error message on invalid XML when saving a datastream instead of of raising an exception.

Unit test updated.  Tested in browser in localhost.

### DIFF
--- a/app/controllers/datastreams_controller.rb
+++ b/app/controllers/datastreams_controller.rb
@@ -48,17 +48,18 @@ class DatastreamsController < ApplicationController
     raise ArgumentError, 'Missing content' if params[:content].blank?
 
     begin
-      # check that the content is well-formed xml
       Nokogiri::XML(params[:content], &:strict)
+      @object.datastreams[params[:id]].content = params[:content] # set the XML to be verbatim as posted
+      @object.save
+      Argo::Indexer.reindex_pid_remotely(@object.pid)
+      msg = 'Datastream was successfully updated'
     rescue Nokogiri::XML::SyntaxError
-      raise ArgumentError, 'XML is not well formed!'
+      # if the content is not well-formed xml, inform the user rather than raising an exception
+      error_msg = 'The datastream could not be saved due to malformed XML.'
     end
-    @object.datastreams[params[:id]].content = params[:content] # set the XML to be verbatim as posted
-    @object.save
-    Argo::Indexer.reindex_pid_remotely(@object.pid)
 
     respond_to do |format|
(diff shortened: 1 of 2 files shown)


app/controllers/datastreams_controller.rb
```diff
@@ -48,17 +48,21 @@ def update
...
       Nokogiri::XML(params[:content], &:strict)
+      @object.datastreams[params[:id]].content = params[:content] # set the XML to be verbatim as posted
+      @object.save
+      Argo::Indexer.reindex_pid_remotely(@object.pid)
+      msg = 'Datastream was successfully updated'
     rescue Nokogiri::XML::SyntaxError
-      raise ArgumentError, 'XML is not well formed!'
+      # if the the content is well-formed xml, inform the user
+      error_msg = 'The datastream could not be saved due to malformed XML.'
+    rescue StanfordError => e
```
jcoyne, line 59: why catch then? Can you just let it go?

peetucket: Yeah, you are right - doesn't make much sense.  Removed.

Issue #2200, opened 2020-08-05
[UBER] Set new CDL option (controlled digital lending) in object rights


peetucket:
We will have a new rights block for the fall 2020 term known as CDL (Controlled Digital Lending).  This will be a new form of access rights which will provide temporary access to materials for classes during the term.  Enforcement of access will occur via the PURL page with work done by the Access team, but we need to provide an ability to set the new rights in arbitrary objects.

The end result needs to be:

- Add a new option to the available object rights drop-down menu called "CDL".  This menu is shown on the item detail page when the user clicks the "Set Rights" button the left side of the page.
- This new option also needs to be available on the Bulk Actions page for the "Set object rights" action.
- When selected, this new option needs to be accepted as a valid choice and needs to be passed through the dor-services-app

Changes:
- [x]  ~~Update dor-services gem needs to recognize this new option and set the rights in the rightsMetadata datastream appropriately.  Ticket: sul-dlss/dor-services#712, PR: sul-dlss/dor-services#713~~
- [x] ~~Update rightsMetadata description in consul: https://consul.stanford.edu/display/chimera/Rights+metadata+--+the+rightsMetadata+datastream~~
- [x] ~~Update cocina models as needed  PR 1: sul-dlss/cocina-models#136~~
- [x] ~~File level access for cocina: https://github.com/sul-dlss/dor-rights-auth/pull/47#discussion_r476017934 and now a ticket: sul-dlss/cocina-models#140~~
- [x] ~~Update dor-rights-auth to recognize new cdl option in rights metadata.  Coordinate with Access team as needed, since they consume this gem too (see #cdl channel for communication about this).  Ticket: sul-dlss/dor-rights-auth#44, in progress PR: sul-dlss/dor-rights-auth#45~~
- [x] ~~Cut new release of dor-rights-auth gem, then bump it's version in dor-services gem and cut a new release of dor-services gem that uses it~~
- [x] ~~Update dor-services-app to map the new CDL option for the to  and from fedora classes.  Will need to bump to newer version of dor-rigthts-auth to get to the mapping to/from fedora working correctly.  Ticket: sul-dlss/dor-services-app#969, In progress PR: sul-dlss/dor-services-app#974~~
- [x] ~~Ensure indexing in argo shows the facet correctly for the new rights option.  This means dor_indexing_app need to be updated, possibly just requires bumping dor-indexing-app to use the new dor-rights-auth  Ticket: sul-dlss/dor_indexing_app#431~~
[...]

PR #2208, opened 2020-08-18
Changes needed for controlled digital lending

peetucket:
## Why was this change made?

To address #2200 

To get new CDL rights options showing in Argo.

~~HOLD for sul-dlss/dor-services-app#974 which should be deployed first~~

- new unit tests
- tested in stage

How do we know it's working?

- [x] Open a new version of an item.  Use the blue "Set rights" button to set the object rights to controlled digital lending.
- [x] Verify the rightsMetadata has the correct `<cdl>` clause as described in #2200.
- [x] Close the object and wait for it to be accessioned.
- [x] Look at the events and the cocina model and verify the rights/descriptions look correct.
- [x] Verify the public XML in PURL has the correct `<cdl>` clause in the rightsMetadata.
- [x] Look at a collection object, verify that CDL rights are not available to be set via the "Set rights" button.
- [x] Verify that setting rights back to world for a single CDL object works as expected.
- [x] Test a bulk update for updating rights for several items and verify the <cdl> clause is set correctly as noted in steps 1-4 above.
- [x] Test a bulk update for updating rights for several items back to world and verify the <cdl> clause is removed and world rights are set.

The consul page on rightsMetadata was updated


spec/requests/set_rights_spec.rb
```diff
@@ -129,5 +129,31 @@
         expect(Argo::Indexer).to have_received(:reindex_pid_remotely).with(pid)
       end
     end
+
+    context 'when setting to cdl' do
```
jcoyne, line 133: We should add a test for removing CDL.

jcoyne: Added.

jcoyne:
@peetucket currently this is showing CDL as an option for the set rights of collections.  That is not valid, so we must prevent that.

spec/requests/set_rights_spec.rb
```diff
@@ -58,175 +16,278 @@
...
-                'releaseDate' => '2021-02-11T00:00:00.000+00:00',
-                'access' => 'world'
+      context 'when setting cdl access' do
+        let(:updated_model) do
+          cocina_model.new(
+            {
+              'access' => {
+                'access' => 'citation-only',
+                'download' => 'none',
+                'controlledDigitalLending' => true,
```
peetucket, line 145: So access == 'citation-only' and controlledDigitalLending == true is how we model this here, 
but access == 'controlled digital lending' and controlled_digital_lending == true is how we interpret this coming from fedora ?

As shown in this test here...
https://github.com/sul-dlss/dor-services-app/blob/master/spec/services/cocina/from_fedora/dro_access_spec.rb#L109

Rights as "controlled digital lending" is coming from dor-services gem here:

https://github.com/sul-dlss/dor-services/blob/master/lib/dor/datastreams/rights_metadata_ds.rb#L165-L177

via dor-services-app here:

https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/from_fedora/access.rb#L45-L53

app/forms/access_form.rb
```diff
@@ -1,15 +1,26 @@
 # frozen_string_literal: true
 
 # A form object for updating the rights on an item
+# @abstract
 class AccessForm
   extend ActiveModel::Naming
-
-  # @param [Cocina::Models::DRO]
-  def initialize(model, apo)
+  # @param [Cocina::Models::DRO, Cocina::Models::Collection] model
```
mjgiarlo, line 7: Can you add `@param` documentation for new `default_rights` param?

app/forms/access_form.rb
```diff
@@ -1,15 +1,26 @@
...
+  def initialize(model, default_rights: nil)
```
mjgiarlo, line 8: Can this be `default_rights: 'citation-only'`? Then line 10 can be `@default_rights = default_rights`

app/forms/collection_rights_form.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+# This represents the form in the modal on the item page that appears when you
+# click "Set rights" in the sidebar
+class CollectionRightsForm < AccessForm
+  # @param [Cocina::Models::Collection] model
```
mjgiarlo, line 6: As above, add YARD to cover the `default_rights` param

app/forms/collection_rights_form.rb
```diff
@@ -0,0 +1,30 @@
...
+  def initialize(model, default_rights: nil)
+    super
+  end
+
+  # TODO: validate the rights that is set doesn't include CDL
```
mjgiarlo, line 11: For a follow-up PR or cruft?

app/forms/collection_rights_form.rb
```diff
@@ -0,0 +1,30 @@
...
+  def sync
+    access_additions = CocinaAccess.from_form_value(rights)
+    updated_access = model.access.new(access_additions.value!)
+
+    @model = model.new(access: updated_access)
+  end
+
+  # all the rights except CDL
+  def rights_list
+    super.filter { |i| i.last != 'cdl-stanford-nd' }
```
mjgiarlo, line 22: What is the significance of `last` here? It's looping over an array of arrays, and the last value in the inner array is where there's a label that could match?

app/forms/access_form.rb
```diff
@@ -1,15 +1,26 @@
...
     @model = model
-    @apo = apo
+    @default_rights = default_rights || 'citation-only'
+  end
+
+  # @param [HashWithIndifferentAccess] params the values from the form
+  # @option params [String] :rights the rights representation from the form (must be one of the keys in Dor::RightsMetadataDS::RIGHTS_TYPE_CODES, or 'default')
+  def validate(params)
+    rights = params[:rights]
+    return false unless valid_rights_options.include?(rights)
```
mjgiarlo, line 17: I find it a little surprising that the abstract base class is calling a method that only lives in the concrete child classes. I might expect to see that method also defined in this class but with a dumb return value or perhaps an exception raised.

app/forms/collection_rights_form.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  private
+
+  def valid_rights_options
```
mjgiarlo, line 27: Worth adding a comment that this is used by the parent class?

app/forms/dro_rights_form.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+# This represents the form in the modal on the item page that appears when you
+# click "Set rights" in the sidebar
+class DRORightsForm < AccessForm
+  # @param [Cocina::Models::DRO] model
```
mjgiarlo, line 6: YARD for `default_rights` pls

app/forms/dro_rights_form.rb
```diff
@@ -0,0 +1,42 @@
...
+            # Ensure files attached to dark objects are neither published nor shelved
+            file[:access].merge!(access: 'dark')
+            file[:administrative].merge!(shelve: false)
+          end
+        end
+      end
+    end
+  end
+
+  def valid_rights_options
```
mjgiarlo, line 39: Worth adding a comment that this is used by the parent class?

app/services/cocina_dro_access.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# This transcodes the access from the item set rights form to Cocina DROAccess schema
+class CocinaDROAccess
+  extend Dry::Monads[:maybe]
+
+  # @param [String] the rights representation from the form (must be one of the keys in Dor::RightsMetadataDS::RIGHTS_TYPE_CODES, or 'default')
```
mjgiarlo, line 7: Won't this tell YARD that the name of the param is `the` instead of `rights`?

app/services/cocina_dro_access.rb
```diff
@@ -0,0 +1,41 @@
...
+             }
+           elsif rights.end_with?('-nd') || %w[dark citation-only].include?(rights)
+             {
+               access: rights.delete_suffix('-nd'),
+               download: 'none',
+               controlledDigitalLending: false
+             }
+           elsif rights.start_with?('loc:')
+             {
+               access: 'location-based',
```
mjgiarlo, line 27: This line looks uncovered by tests, according to CodeClimate, FYI.

app/controllers/items_controller.rb
```diff
@@ -266,8 +266,9 @@ def rights
       cocina = NilModel.new(params[:id])
     end
 
-    apo_object = Dor.find(@apo)
-    @form = AccessForm.new(cocina, apo_object)
+    form_type = cocina.collection? ? CollectionRightsForm : DRORightsForm
```
mjgiarlo, line 269: These next couple lines look to be uncovered.

app/forms/collection_rights_form.rb
jcoyne: yes.

app/forms/collection_rights_form.rb
jcoyne: cruft, it's been done.

app/forms/access_form.rb
jcoyne: no, because they may pass in `default_rights: nil` and we still want the instance variable to get a value.

app/controllers/items_controller.rb
jcoyne: They were uncovered before too.

app/forms/access_form.rb
jcoyne: But then it would be impossible to have code coverage on that method.

app/services/cocina_dro_access.rb
jcoyne: fixed

app/forms/collection_rights_form.rb
jcoyne: done

app/forms/dro_rights_form.rb
jcoyne: added

app/forms/dro_rights_form.rb
jcoyne: added

app/forms/collection_rights_form.rb
jcoyne: added

app/forms/access_form.rb
jcoyne: Because you don't instantiate abstract classes and each concrete class overrides that method.

spec/requests/set_rights_spec.rb
peetucket: We made everything consistent in dor-services app... we now say access is 'citation-only' there.

app/forms/access_form.rb
peetucket: adding param documentation here

app/forms/access_form.rb
mjgiarlo: @jcoyne Sure, but it's not impossible in Ruby to instantiate an abstract class if we cared enough about code coverage thereof, right? No big deal here, though. I don't think we *need* a dummy implementation in the parent class, especially if we have a code comment here about the expectation that child classes implement this method. Cool with adding that?

app/services/cocina_dro_access.rb
mjgiarlo: Just an FYI, to be clear, not a blocker.

app/forms/access_form.rb
justinlittman: I added a comment, so resolving conversation.

app/services/cocina_dro_access.rb
peetucket: yeah, i went looking for where the other bits of this class might be tested so I could extend it, but it wasn't obvious to me

PR #2210, opened 2020-08-18
Set access using cocina data model

### LINKED ISSUE #2190: Setting rights in argo to "citation only" does not work.
**Describe the bug**
Cannot change rights to citation only in argo.

**User Impact**
Users cannot change rights to citation only for an object.

**To Reproduce**

1. Open an object for versioning (if needed).
2. Click "Set Rights" on the left side nav.
3. Select "Citation only".
4. Click "Update".
5. See an error message "Invalid new rights setting."

**Expected behavior**
Rights need to change.

**Additional context**

Tried on stage with this object which is open and easy to test with:

https://argo-stage.stanford.edu/catalog/druid:cc243mg0841

It looks like it is possible to update to "Stanford", so its not a general case problem, but seems like only certain values.  Noticed that dor-services gem expects a value of "none" when setting the rights, but Argo sends a value of "citation-only" in the edit view:

https://github.com/sul-dlss/dor-services/blob/master/lib/dor/datastreams/rights_metadata_ds.rb#L62-L75

Its not clear which one we want to change (looks like the value of "citation-only" is sprinkled throughout Argo).

jcoyne:
## Why was this change made?

This allows us to set to citation-only and it eliminates one place where we depend on the legacy ActiveFedora API. Fixes #2190

Tested on staging.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -213,8 +213,11 @@ class ItemsController < ApplicationController
 
   # This is called from the item page and from the bulk (synchronous) update page
   def set_rights
-    @object.read_rights = params[:access_form][:rights]
-    save_and_reindex
+    object_client = Dor::Services::Client.object(@object.pid)
+    dro = object_client.find
+    json = JSON.parse(dro.to_json)
+               .merge(CocinaAccess.from_form_value(params[:access_form][:rights]).deep_stringify_keys)
+    object_client.update(params: json)
 
     respond_to do |format|
       if params[:bulk]
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -213,13 +213,6 @@ RSpec.describe ItemsController, type: :controller do
     end
   end
 
-  describe '#set_rights' do
-    it 'sets an item to dark' do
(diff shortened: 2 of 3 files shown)


app/controllers/items_controller.rb
```diff
@@ -213,8 +213,11 @@ def refresh_metadata
 
   # This is called from the item page and from the bulk (synchronous) update page
   def set_rights
-    @object.read_rights = params[:access_form][:rights]
-    save_and_reindex
+    object_client = Dor::Services::Client.object(@object.pid)
+    dro = object_client.find
+    json = JSON.parse(dro.to_json)
```
mjgiarlo, line 218: Instead of `JSON.parse(dro.to_json)`, can you do `dro.to_h`? Isn't the thing in the `json` variable a hash anyway? I guess I'm wondering: why bother with the conversion to and then immediately from JSON if we have a cocina model that might/should support `to_h`?

jcoyne: No, because dro.to_h will return date objects and then if you send that to update it causes an error because the dates are not represented as strings.  I think we can look into how we improve this interaction, but I didn't want to get into it here.

mjgiarlo: OK, sounds good.

PR #2212, opened 2020-08-19
Display cocina by default

jcoyne:
But keep it collapsed

## Why was this change made?

Allow users to begin seeing the new data model.

Tested on staging.

Screen Shot 2020-08-19 at 2 41 56 PM


jcoyne:
Andrew signed off on this.

ndushay:
Thanks for the screen shot!  It helps a lot!

PR #2222, opened 2020-08-25
Disable publish/shelve in content metadata when setting rights to dark

### LINKED ISSUE #2220: Setting rights to "dark" in Argo is failing
**Describe the bug**
Setting an object's rights to "dark" is failing if any files in the object were previously published and/or shelved (i.e. there are files with "publish=yes" or "shelve=yes" in the contentMetadata).

Honeybadger error: https://app.honeybadger.io/projects/49894/faults/66958746

It looks like the system is validating that an item with rights set to dark does not have any shelved or published files. This is a good validation to have, but it will always fail if the system doesn't change the contentMetadata first when changing rights to dark. Users shouldn't be expected to edit the contentMetadata themselves to change the publish and shelve settings for all files, as that would require directly editing the XML. The act of setting rights to dark is supposed to apply contentMetadata changes automatically.

**User Impact**
It is not possible to set rights to dark via the set rights interface. This may result in delays in unpublishing purls. Time is usually a factor when purls are unpublished, since the most common reasons to set rights to dark are permissions or privacy issues, so a delay in applying these rights is a potentially serious problem.

**To Reproduce**
1. Go to an item in Argo that has 1) a published purl and 2) files available to access from the purl
2. Open the item for modification
3. Attempt to set rights to dark
4. The error will result in you being taken to the general library-status page
5. Check honeybadger

**Expected behavior**
Setting rights to dark should apply a change to the contentMetadata to set all files listed there to:

publish="no"
shelve="no"

This change should not affect the preserve status. That should remain "yes" or "no" according to how it was before the rights were set to dark.

**Additional context**
The expected behavior is how setting rights to dark has been applied since at least 2019, when I started at Stanford. Previously, when setting rights to dark, I'd see a corresponding change in the contentMetadata.

mjgiarlo:
Fixes #2220

## Why was this change made?

See #2220.

CI

None.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -214,7 +214,9 @@ class ItemsController < ApplicationController
     dro = object_client.find
     access_additions = CocinaAccess.from_form_value(params[:access_form][:rights])
     updated_access = dro.access.new(access_additions.value!)
-    updated = dro.new(access: updated_access)
+    structural_additions = CocinaStructural.from_form_value(params[:access_form][:rights], dro.structural)
+    updated_structural = dro.structural.new(structural_additions)
+    updated = dro.new(access: updated_access, structural: updated_structural)
     object_client.update(params: updated)
     reindex
 
--- a/app/services/cocina_structural.rb
+++ b/app/services/cocina_structural.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class CocinaStructural
+  # @param structural [Cocina::Models::DROStructural] the DRO structural metadata to modify
+  # @param rights [String] the rights representation from the form (must be one of the keys in Dor::RightsMetadataDS::RIGHTS_TYPE_CODES, or 'default')
+  # @return [Hash<Symbol,String>] a hash representing a subset of the Structural subschema of the Cocina model
+  def self.from_form_value(rights, structural)
+    # Convert to hash so we can mutate it
+    structural.to_h.tap do |structure_hash|
+      if rights == 'dark' && structural&.contains&.any?
+        structure_hash[:contains].each do |fileset|
+          fileset[:structural][:contains].each do |file|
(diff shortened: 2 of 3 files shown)


app/controllers/items_controller.rb
```diff
@@ -214,7 +214,9 @@ def set_rights
     dro = object_client.find
     access_additions = CocinaAccess.from_form_value(params[:access_form][:rights])
     updated_access = dro.access.new(access_additions.value!)
-    updated = dro.new(access: updated_access)
+    structural_additions = CocinaStructural.from_form_value(params[:access_form][:rights], dro.structural)
+    updated_structural = dro.structural.new(structural_additions.value!)
```
jcoyne, line 218: Won't this raise an exception if rights is world?

mjgiarlo: I don't know. I was following the example above and leaning on the existing tests.

jcoyne: The existing test is only testing the set to dark case, so can you add a "set to stanford only" case?

mjgiarlo: I'll add a test case for sure.

mjgiarlo: @jcoyne also, isn't this a problem for the CocinaAccess stuff you added above? it can also return `None()`.

jcoyne: in the CocinaAccess case `None` is an error. The form should always return something sensible.

jcoyne: Yes, it's also a problem there. I'll work on that.

jcoyne: The `None` case never occurs on this update form. It's only used for registration. See https://github.com/sul-dlss/argo/pull/2224

mjgiarlo: Aha, I see. OK, I've updated the approach to stop using monads. Take another look?

PR #2225, opened 2020-08-26
Rename sul-purl-test to sul-purl-stage

mejackreed:
sul-purl-test is being decommissioned for naming consistency. It will be
aliased to sul-purl-stage, but for visibility we should probably rename
it everywhere.


jcoyne:
Thank you @mejackreed

PR #2226, opened 2020-08-27
[HOLD] Bump dor-services gem to get new cdl rights option

jcoyne:
## Why was this change made?

To get new CDL rights options showing in Argo.

HOLD for sul-dlss/dor-services-app#974 which should be deployed first


jcoyne:
Duplicate of #2208

PR #2232, opened 2020-08-27
Set collection using the cocina API

jcoyne:
## Why was this change made?

The API allows changes to be validated and logged as events. This moves us away from Fedora 3.

Test suite.

n/a


app/controllers/items_controller.rb
```diff
@@ -47,14 +47,24 @@ def embargo_form
...
   def set_collection
-    @object.collections.each { |collection| @object.remove_collection(collection.pid) } # first remove any existing collections
-    if params[:collection].present?
-      @object.add_collection(params[:collection]) # collection provided, so add it
-      response_message = 'Collection successfully set.'
-    else
-      response_message = 'Collection(s) successfully removed.' # no collection selected from drop-down, so don't bother adding a new one
-    end
-    save_and_reindex
+    object_client = Dor::Services::Client.object(@object.pid)
```
mjgiarlo, line 50: There's a lot going on here that may be ripe for eventual extraction but I don't think that's necessary at this point.

mjgiarlo (approved):

PR #2234, opened 2020-08-27
Use route helpers

jcoyne:
## Why was this change made?
Route helpers are the idiomatic rails way of specifying a route.  This is easier to grep for too.


jcoyne:
Yes, the feature tests, request tests and view tests all exercise these changes.

PR #2236, opened 2020-08-28
Update catkey using the cocina api 

jcoyne:
## Why was this change made?
To move us away from Fedora 3

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -160,8 +160,14 @@ class ItemsController < ApplicationController
   end
 
   def catkey
-    @object.catkey = params[:new_catkey].strip
-    save_and_reindex
+    object_client = Dor::Services::Client.object(@object.pid)
+    dro = object_client.find
+    updated_identification = dro.identification.new(
+      catalogLinks: [{ catalog: 'symphony', catalogRecordId: params[:new_catkey].strip }]
+    )
+    updated = dro.new(identification: updated_identification)
+    object_client.update(params: updated)
+    reindex
 
     respond_to do |format|
       if params[:bulk]
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -160,7 +160,9 @@ class ItemsController < ApplicationController
   end
 
   def catkey
-    CatkeyService.update(@object.pid, params[:new_catkey].strip)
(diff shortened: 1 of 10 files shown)


app/controllers/items_controller.rb
```diff
@@ -141,8 +141,14 @@ def source_id
   end
 
   def catkey
-    @object.catkey = params[:new_catkey].strip
-    save_and_reindex
+    object_client = Dor::Services::Client.object(@object.pid)
+    dro = object_client.find
+    updated_identification = dro.identification.new(
+      catalogLinks: [{ catalog: 'symphony', catalogRecordId: params[:new_catkey].strip }]
```
justinlittman, line 147: Wouldn't this step on all existing catalogLinks? Existing source id?

jcoyne: Yes, that's the same behavior as the existing code: https://github.com/sul-dlss/argo/pull/2236/files#diff-2eb2bf110a48bee6c79d3bbb377a549eL144

justinlittman: Where is `@object.catkey` implemented?

jcoyne: it's part of the dor-services gem.
https://github.com/sul-dlss/dor-services/blob/4cf2fc6503060656a3485856109501d7b3a79011/lib/dor/models/abstract.rb#L79

jcoyne: It looks like we will have to store the previous catkeys though so that we can create blank catalog records for old catkeys in the release workflow. https://github.com/sul-dlss/dor-services/blob/master/lib/dor/datastreams/identity_metadata_ds.rb#L128

jcoyne:
Probably want to throw in an update to ManageCatkeyJob to this PR.

app/controllers/items_controller.rb
jcoyne: Turns out we don't because dor-services-app is calling the `catkey=` method for us. https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/object_updater.rb#L75

mjgiarlo (approved):

mjgiarlo:
@jcoyne Needs rebase.

mjgiarlo:
@jcoyne feel free to merge once :green_heart:

PR #2240, opened 2020-09-01
Spot check argo db for data

tallenaz:
## Why was this change made?

In addition to the okcomputer db schema check, we want to check for data loss. This is adapted from similar approaches in sul-pub and prescat

Deployed to stage, and it works as expected.

DevOpsDocs edits forthcoming.

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -40,10 +40,41 @@ class RubydoraCheck < OkComputer::PingCheck
   end
 end
 
+# check models to see if at least they have some data
+class TablesHaveDataCheck < OkComputer::Check
+  def check
+    msg = [
+      Bookmark,
+      BulkAction,
+      Search,
+      User
+    ].map { |klass| table_check(klass) }.join(' ')
+    mark_message msg
+  end
+
+  private
+
+  # @return [String] message
+  def table_check(klass)
+    # has at least 1 record
+    return "#{klass.name} has data." if klass.any?
+
+    mark_failure
+    "#{klass.name} has no data."
+  rescue ActiveRecord::RecordNotFound
+    mark_failure
+    "#{klass.name} has no data."
+  rescue => e # rubocop:disable Style/RescueStandardError
+    mark_failure
+    "#{e.class.name} received: #{e.message}."
+  end
+end
+
 # REQUIRED checks, required to pass for /status/all
 #  individual checks also avail at /status/<name-of-check>
 OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
 OkComputer::Registry.register 'rails_cache', OkComputer::GenericCacheCheck.new
+OkComputer::Registry.register 'feature-tables-have-data', TablesHaveDataCheck.new
 
 OkComputer::Registry.register 'active_fedora_conn', RubydoraCheck.new(client: ActiveFedora::Base.connection_for_pid(0))
 # fedora_url is covered by checking ActiveFedora::Base.connection_for_pid(0)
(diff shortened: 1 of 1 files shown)


jcoyne:
Has this sort of failure ever occurred? This seems like a strangely specific failure mode we're checking for.

tallenaz:
Not to my knowledge in argo. This sort of check does seem pretty important for cap and prescat. And we do want to be able to check for data loss where it's appropriate. Do we care, then, if these tables in argo lose data?

jcoyne:
@tallenaz yes, we care if data is lost, but this check doesn't check for that, it checks that these tables have at least one row in them.  That would be only a subset of the cases of data loss.  I think this is such a rare and catastrophic case to be not worth bothering checking for.  Furthermore, I don't think we care about most of these tables.  They are mostly blacklight builtins that we aren't using in Argo.

tallenaz:
We agree that we care about data loss, and we agree that this checks for a subset of cases of data loss. In my view, that makes this a case of "something is better than nothing". But it sounds like you have something better in mind, which I'd be happy to draw from. So I'd like to ask which tables in argo we care about, and how you think I could approach checking for a more central case or set of cases of data loss.

jcoyne:
@tallenaz This approach is like checking the office for tigers.  Yes, we don't want tigers in the office, but this is such an unusual occurrence, it's not something we bother doing.

tallenaz:
We agree this is a check for a catastrophic situation, but we disagree that this is a silly thing to be prepared for. Substitute earthquakes for tigers, for example.

jcoyne:
@tallenaz my point about about this being catastrophic (all the data is gone) is that we'd notice this through other means (honeybadger and user reports) and Nagios wouldn't be all that helpful.  Subtle data loss is more pernicious and we maybe wouldn't have another way to detect it.

config/initializers/okcomputer.rb
```diff
@@ -40,10 +40,41 @@ def profile
   end
 end
 
+# check models to see if at least they have some data
+class TablesHaveDataCheck < OkComputer::Check
+  def check
+    msg = [
+      Bookmark,
+      BulkAction,
+      Search,
```
jcoyne, line 49: In the case of searches, this is okay to not have any. In fact they are supposed to be cleared out periodically: https://github.com/projectblacklight/blacklight/blob/91d5e6fb1fe8589262c3b86c0e5a2b5963e1f431/lib/railties/blacklight.rake#L5-L7

config/initializers/okcomputer.rb
```diff
@@ -40,10 +40,41 @@ def profile
   end
 end
 
+# check models to see if at least they have some data
+class TablesHaveDataCheck < OkComputer::Check
+  def check
+    msg = [
+      Bookmark,
```
jcoyne, line 47: We don't use bookmarks:
```
 Bookmark.count
   (1.1ms)  SELECT COUNT(*) FROM `bookmarks`
=> 0
```

tallenaz:
Understood. Having some level of coverage through nagios is what ops is aiming for, even if it wouldn't be all that. I tacked on a commit addressing the last two comments.

jcoyne (approved):

config/initializers/okcomputer.rb
```diff
@@ -40,10 +40,39 @@ def profile
...
+  private
+
+  # @return [String] message
+  def table_check(klass)
+    # has at least 1 record
+    return "#{klass.name} has data." if klass.any?
+
+    mark_failure
+    "#{klass.name} has no data."
+  rescue ActiveRecord::RecordNotFound
```
jcoyne, line 62: How do we ever get into here?

jcoyne (changes requested):

jcoyne (approved):

PR #2277, opened 2020-10-09
Update facet field for data quality

jcoyne:
## Why was this change made?
For https://github.com/sul-dlss/dor_indexing_app/pull/442

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -122,8 +122,7 @@ class CatalogController < ApplicationController
     add_common_date_facet_fields_to_config! config
 
     # This will help us find records that need to be fixed before we can move to cocina.
-    config.add_facet_field 'source_id_errors_ssim', label: 'Source ID Errors', home: false,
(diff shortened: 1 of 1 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -122,8 +122,7 @@ class CatalogController < ApplicationController
     add_common_date_facet_fields_to_config! config
 
     # This will help us find records that need to be fixed before we can move to cocina.
-    config.add_facet_field 'source_id_errors_ssim', label: 'Source ID Errors', home: false,
```
ndushay, line 125: I get the addition but why the removal of this facet?  It's not mentioned in the PR description or the ticket.

jcoyne: It was renamed because it now has more than source id errors.

ndushay: ah!  thank you.

ndushay (approved):
approve pending green for tests.

PR #2280, opened 2020-10-13
Switch to use CocinaDroAccess when registering objects.

### LINKED ISSUE #2279: Can't register to CDL rights
**Describe the bug**
Registering an item with CDL rights isn't working. Attempts to register are generating this error:

`cdl-stanford isnt include enum in #/components/schemas/DROAccess/properties/access`

The error appears on the Argo registration form when you hover over the red alert triangle (see screenshot below).

**User Impact**
Digitization staff can't register objects to CDL rights. Objects have to be registered citation-only and changed later.

**To Reproduce**
1. Go to Argo registration form
2. Try to register with the "controlled digital lending (no-download)" rights
3. See error

**Expected behavior**
Registration should accept all the rights choices.

**Screenshots**
Screen Shot 2020-10-12 at 3 40 25 PM

**Additional context**
Changing the rights later seems to still be working. At least, I changed 3 items earlier in the day, before this error at registration was reported.

justinlittman:
closes #2279

## Why was this change made?
So that a user can select CDL when registering an object.

Stage.

NA


mjgiarlo (approved):
Great, quick fix! Thanks, @justinlittman

Issue #2308, opened 2020-11-04
Determine cause of Argo front page slowness


andrewjbtw:
**Describe the bug**
The Argo front page can take up to a minute to load. Maybe more if someone has not already gone through the Stanford login process, which always adds extra time.

**User Impact**
It slows users down and in a number of cases it's made people think Argo is actually down rather than slow. 

Individual item pages don't seem to be affected, and searches that don't return a lot of data in the facets seem to run fine.

**To Reproduce**
Go to the Argo front page and wait. Similar wait times affect the "expand to show all facets" display and the workflow grid.

**Expected behavior**
Argo isn't the fastest website, but it should load more quickly than a minute. 

**Additional context**
There's been some discussion about the cause of the slowness. Unfortunately, we've never benchmarked Argo performance during "normal" times so we do not have formal data for comparison.

There's a theory that the cause of the slowness is the number of tags that Argo has to load. This theory is based on the fact that when Argo is slow, the tags are the data that takes longest to load. Loading Argo without showing tags is much faster.

I do not think that we have sufficient evidence to support this theory because we did not measure Argo performance with and without tags during times when Argo has not been slow. The fact that Argo takes a long time to load tags when Argo is slow does not demonstrate that loading tags is the cause of Argo being slow. There could be a different underlying cause that slows down Argo, with slow loading tags being a symptom of that same cause, not the cause itself.

My experience with Argo this summer and fall suggests that tags are not the reason Argo has been slow. I use Argo every weekday, I load the workflow grid multiple times per day, and I always display the tags facet. If Argo had been uniformly slow all summer, I would have noticed it. Instead, Argo has been slow, then not slow, then slow again. No changes to tags have been made during these periods.

Consider the following, which is based on my experience plus user reports:

1. July: Argo took nearly a minute to load, very similar to the current issue
2. Early August up until the last week of October: Argo loaded much more quickly
3. Starting last week (late October) up until now: Argo back to loading slowly

That suggests to me that 

- a change made in early August sped up Argo performance and 
- a change made in late October slowed it down again

**Suggestion: look at componentized facets**

I went back through the change history, and this is what I found:

- Argo became noticeably faster after this pull request to componentize the facet configurations was deployed in early August: https://github.com/sul-dlss/argo/pull/2162/files (see discussion on Slack after the change was made: https://stanfordlib.slack.com/archives/C0EAZ09QF/p1596647909226600 )

[...]

PR #2309, opened 2020-11-05
Set the component option on newest facets

aaron-collier:
## Why was this change made?

Part of #2308 - a few of the newest facets did not have `component: true` set. 

N/A

N/A

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -161,10 +161,10 @@ class CatalogController < ApplicationController
     config.add_facet_field 'sw_language_ssim',           label: 'SW Language',          component: true, limit: 10, home: false
     config.add_facet_field 'mods_typeOfResource_ssim',   label: 'MODS Resource Type',   component: true, limit: 10, home: false
     # Adding the facet field allows it to be queried (e.g., from value_helper)
-    config.add_facet_field 'is_governed_by_ssim', if: false
-    config.add_facet_field 'is_member_of_collection_ssim', if: false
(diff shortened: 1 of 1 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -161,10 +161,10 @@ class CatalogController < ApplicationController
     config.add_facet_field 'sw_language_ssim',           label: 'SW Language',          component: true, limit: 10, home: false
     config.add_facet_field 'mods_typeOfResource_ssim',   label: 'MODS Resource Type',   component: true, limit: 10, home: false
     # Adding the facet field allows it to be queried (e.g., from value_helper)
-    config.add_facet_field 'is_governed_by_ssim', if: false
-    config.add_facet_field 'is_member_of_collection_ssim', if: false
-    config.add_facet_field 'tag_ssim', if: false
-    config.add_facet_field 'project_tag_ssim', if: false
+    config.add_facet_field 'is_governed_by_ssim', component: true, if: false
```
jcoyne, line 164: I don't understand why you would do this since they never render (i.e. `if: false`) ?

jcoyne (changes requested):

app/controllers/catalog_controller.rb
aaron-collier: Yeah, I'm more than happy to close this. Just testing ideas about the timing of when things slowed down and wasn't at all confident about this in particular.

aaron-collier:
Closing. Verified on prod that it did not effect anything.

PR #2318, opened 2020-11-30
Override Blacklight default error handler to give a more useful error…

### LINKED ISSUE #2306: Instead of "network error" show a descriptive message when a modal won't load because of a Cocina issue
**Describe the bug**
When an item doesn't have a valid Cocina representation, certain actions will pop up a window with a message that just says "Network error". These actions are:

- the "Set rights" blue button
- clicking on a file in the file list (under contents) to download a file

Since the issue isn't really a network error, show a message that indicates that it is related to Cocina.

**User Impact**
Users are blocked from the tasks they are trying to do, and are shown a misleading message. To the extent possible, a message like "network error" should be reserved for when there is a real network connection problem.

**To Reproduce**
I've made an object with a non-valid source id on stage for testing:
https://argo-stage.stanford.edu/view/druid:qv217cy9963

1. Go to the example item above
2. Click on "set rights" or try to download "testfile.txt"
3. See a window pop up with the message "network error".

**Expected behavior**
If the Cocina representation is the problem, the error message should indicate that.

**Screenshots**
This is the only message people see for this error:
Screen Shot 2020-11-03 at 12 29 54 PM

jcoyne:
… message

Upstream port at https://github.com/projectblacklight/blacklight/pull/2386

### Before

(smaller box with "Network Error" as the only information)

### After

image

## Why was this change made?

Ref #2306  (this is prerequisite for a fix)

Tested on staging on https://argo-stage.stanford.edu/view/druid:qv217cy9963 (try to set rights)

### DIFF
--- a/app/javascript/modules/blacklight-override.js
+++ b/app/javascript/modules/blacklight-override.js
@@ -1,0 +1,15 @@
+// Override Blacklights default error handler to give a more useful error message
+// Ported upstream as https://github.com/projectblacklight/blacklight/pull/2386
+Blacklight.modal.onFailure = function(jqXHR, textStatus, errorThrown) {
+  console.error('Server error:', this.url, jqXHR.status, errorThrown)
+  var contents =  `<div class="modal-header">
+            <div class="h5 modal-title">There was a problem with your request.</div>
+            <button type="button" class="blacklight-modal-close close" data-dismiss="modal" aria-label="Close">
+              <span aria-hidden="true">&times;</span>
+            </button></div>
+            <div class="modal-body"><p>Expected a successful response from the server, but got an error</p>
+            <pre>${this.type} ${this.url}` +
+            `\n${jqXHR.status}: ${errorThrown}</pre><div>`
+  $(Blacklight.modal.modalSelector).find('.modal-content').html(contents);
+  $(Blacklight.modal.modalSelector).modal('show');
+}
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -34,9 +34,11 @@ import 'modules/permission_list'
 import 'modules/populate_druids'
 import 'modules/sharing'
 import TagsAutocomplete from 'modules/tags_autocomplete'
-import Argo from  'argo'
+import Argo from 'argo'
 
 import 'blacklight-frontend/app/assets/javascripts/blacklight/blacklight'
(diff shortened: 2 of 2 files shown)


ndushay:
I think this is on the right track, but for the exceptions andrew mentions, getting the actual e.message from DSA via DSC, and displaying THAT message, would be even better.

jcoyne:
@ndushay right, but we can't do that until this part is done.

ndushay:
I would also wonder if this might hijack errors in an undesired way?  Andrew should probably poke at it in stage (?)

ndushay:
and there is certainly no reason a separate commit or two on this PR couldn't get more error info into a useful message for each spot Andrew called out in his ticket

app/javascript/modules/blacklight-override.js
```diff
@@ -0,0 +1,15 @@
+// Override Blacklights default error handler to give a more useful error message
+// Ported upstream as https://github.com/projectblacklight/blacklight/pull/2386
+Blacklight.modal.onFailure = function(jqXHR, textStatus, errorThrown) {
```
ndushay, line 3: what is textStatus?  Is it helpful in any way?

jcoyne: It is the variable jquery uses for the status: https://api.jquery.com/jquery.ajax/#jqXHR

ndushay: thanks.

jcoyne:
@ndushay I think that this PR stands on it's own.  This one change provides details that is more helpful and truthful than: "Network Error".  This is much broader change than Andrews request, so I would like this to go first and then follow up with specific changes for that issue.   I believe that small PRs are easier to review and reason about.

ndushay:
@jcoyne even if I fully agreed with you, I still think this PR, especially if it stands alone, should get Andrew's approval, as it might present this information in more places than he identified.

andrewjbtw:
Naomi's right that what I'd like the message to show is the underlying error from DSA (if that's the problem). I don't have a strong opinion on whether that should be one or more PRs, but that's where I'd like to get before closing the issue.

I do think this is an improvement over the vague "network error", which I think some people took to mean it might be a temporary connection issue that would get fixed on its own.

ndushay (approved):
If Andrew's ok w this as it stands, I'm ok too.

PR #2324, opened 2020-12-03
Add missing methods to the NilModel

### LINKED ISSUE #2323: [argo/stage] NoMethodError: undefined method `collection?' for #<NilModel:0x0000000007265988 @pid="druid:qv217cy9963">
## Backtrace

    line 267 of [PROJECT_ROOT]/app/controllers/items_controller.rb: rights

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49894/faults/69244926)

This is what was causing #2306

jcoyne:
## Why was this change made?
Fixes #2323

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -265,7 +265,6 @@ class ItemsController < ApplicationController
     end
 
     form_type = cocina.collection? ? CollectionRightsForm : DRORightsForm
-
     @form = form_type.new(cocina, default_rights: @object.admin_policy_object.default_rights)
 
     respond_to do |format|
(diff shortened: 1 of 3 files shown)


ndushay (approved):
Is this the fix for both the "set rights" and the "getting a file from the file list" on the original #2306?   If so ... cool.  

However, I think the "file" one might be different ... as I experimented unsuccessfully with adding access and collection to nil_model when I was FR and I think the "file" problem needed completely diff stuff.

jcoyne:
No, this is the fix for #2323, which is one contributing factor to #2306

PR #2325, opened 2020-12-03
Display an error when there is a bad cocina fetch

### LINKED ISSUE #2306: Instead of "network error" show a descriptive message when a modal won't load because of a Cocina issue

jcoyne:
## Why was this change made?

Fixes #2306

Tested on stage
Screen Shot 2020-12-03 at 4 40 19 PM

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -24,6 +24,13 @@ class ItemsController < ApplicationController
     set_governing_apo
   ]
 
+  rescue_from Dor::Services::Client::UnexpectedResponse do |exception|
+    md = /\((.*)\)/.match exception.message
+    detail = JSON.parse(md[1])['errors'].first['detail']
+    redirect_to solr_document_path(params[:id]),
+                flash: { error: "Unable to retrieve the cocina model: #{detail.truncate(200)}" }
+  end
+
   def purl_preview
     xml = Dor::Services::Client.object(params[:id]).metadata.descriptive
     @mods_display = ModsDisplayObject.new(xml)
--- a/spec/controllers/items_controller_spec.rb
+++ b/spec/controllers/items_controller_spec.rb
@@ -135,11 +135,12 @@ RSpec.describe ItemsController, type: :controller do
 
       context 'when the date is malformed' do
         before do
-          allow(embargo_service).to receive(:update).and_raise(Dor::Services::Client::UnexpectedResponse)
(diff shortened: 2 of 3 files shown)


peetucket (approved):

peetucket:
Janky test failure on CI?

app/controllers/items_controller.rb
```diff
@@ -24,6 +24,13 @@ class ItemsController < ApplicationController
     set_governing_apo
   ]
 
+  rescue_from Dor::Services::Client::UnexpectedResponse do |exception|
+    md = /\((.*)\)/.match exception.message
+    detail = JSON.parse(md[1])['errors'].first['detail']
+    redirect_to solr_document_path(params[:id]),
+                flash: { error: "Unable to retrieve the cocina model: #{detail.truncate(200)}" }
+  end
+
```
ndushay, line 33: would it be better for the DSC to have a specific error for validation failures?  Presumably with the DSA also having a specific error for validation failures?  Could be a separate issue.

jcoyne: I'm not sure I understand DSC always returns UnexpectedResponse when there is a response that is not a 200 or 404 : https://github.com/sul-dlss/dor-services-client/blob/34645ee6759eeca7ff6191b8c40d478d4a7edc3c/lib/dor/services/client/object.rb#L52

DSA always returns an ErrorResponse model in these cases: https://github.com/sul-dlss/dor-services-app/blob/master/openapi.yml#L1023-L1034

Why only handle a subset of the errors when we can handle all of them?

justinlittman (approved):

PR #2334, opened 2021-01-05
Remove unnecessary fixture

jcoyne:
## Why was this change made?
Decouple Argo from Fedora 3


README.md
```diff
@@ -74,24 +74,9 @@ CI runs a series of steps;  this the sequence to do it locally, along with some
...
-    RAILS_ENV=test bundle exec rake spec
-    ```
-
-    `rake argo:repo:load` loads test fixture objects into fedora/solr.   You will NOT be able to re-run this task successfully unless fixtures are no longer in the docker containers (e.g. if you `docker-compose down`).  That is because they are already loaded.  The error messages in your terminal output do not surface this cause, but that is likely at play if you see `"Rubydora::FedoraInvalidRequest: See logger for details"`, esp with `"Caused by: RestClient::InternalServerError: 500 Internal Server Error"` in there. In this case, run only `bundle exec rake spec` instead.
-
-    Note that `RAILS_ENV=test` should not be necessary when running `bundle exec rake spec` on its own.
-
-    Note further that ```docker-compose down -v``` will wipe out the volumes for solr & fedora and allows you to start afresh (with ```docker-compose pull```) if your test data is in a strange state.
+```
+RAILS_ENV=test bundle exec rake ci
```
mjgiarlo, line 78: Now that we no longer need to load fixtures, is the explicit `RAILS_ENV=test` still necessary? Or can we jettison that as well (here and in the CI configuration)?

lib/tasks/argo_testing.rake
```diff
@@ -3,20 +3,13 @@
...
 desc 'run specs and rubocop (for CI)'
 task default: %i[rubocop ci]
 
 desc 'run specs after loading up solr, fedora, etc.'
 task :ci do
   if Rails.env.test?
-    Rake::Task['argo:repo:load'].invoke # load 'em all!
     Rake::Task['spec'].invoke
   else
     system('RAILS_ENV=test rake ci')
```
mjgiarlo, line 15: I wonder if this can now be simplified to `Rake::Task['spec'].invoke` ?

mjgiarlo (commented):
This is lovely. I have a couple questions.

.circleci/config.yml
```diff
@@ -222,9 +222,6 @@ jobs:
             curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
             chmod +x ./cc-test-reporter
             ./cc-test-reporter before-build
-      - run:
```
mjgiarlo, line 225: Can we also nuke the RAILS_ENV declaration on line 17?

README.md
```diff
@@ -74,24 +74,9 @@ CI runs a series of steps;  this the sequence to do it locally, along with some
...
-    RAILS_ENV=test bundle exec rake spec
-    ```
-
-    `rake argo:repo:load` loads test fixture objects into fedora/solr.   You will NOT be able to re-run this task successfully unless fixtures are no longer in the docker containers (e.g. if you `docker-compose down`).  That is because they are already loaded.  The error messages in your terminal output do not surface this cause, but that is likely at play if you see `"Rubydora::FedoraInvalidRequest: See logger for details"`, esp with `"Caused by: RestClient::InternalServerError: 500 Internal Server Error"` in there. In this case, run only `bundle exec rake spec` instead.
-
-    Note that `RAILS_ENV=test` should not be necessary when running `bundle exec rake spec` on its own.
-
-    Note further that ```docker-compose down -v``` will wipe out the volumes for solr & fedora and allows you to start afresh (with ```docker-compose pull```) if your test data is in a strange state.
+```
+bin/rake
```
mjgiarlo, line 78: :clap:

mjgiarlo (approved):
Approve pending consideration of one more CI tweak.

lib/tasks/argo_testing.rake
```diff
@@ -3,25 +3,9 @@
...
-desc 'run specs after loading up solr, fedora, etc.'
-task :ci do
-  if Rails.env.test?
-    Rake::Task['argo:repo:load'].invoke # load 'em all!
-    Rake::Task['spec'].invoke
-  else
-    system('RAILS_ENV=test rake ci')
-  end
-end
+task default: %i[rubocop spec]
```
mjgiarlo, line 8: :clap:

PR #2337, opened 2021-01-06
Remove search form partial

jcoyne:
## Why was this change made?
This was overriding Blacklight behavior that we didn't want from old versions of blacklight.  Since blacklight 7, this is not necessary as blacklight does what we want.  This removes a bunch of deprecation warnings.


mjgiarlo (approved):

mjgiarlo:
@jcoyne does this need testing in QA or nah?

jcoyne:
I don't think that'll reveal anything different from what we see locally (no changes).

PR #2345, opened 2021-01-07
Prefer out of the box Rails environments (staging -> production)

mjgiarlo:
## Why was this change made?

To remove an artifact of a bygone age and increase consistency.

CI

None


config/deploy.rb
```diff
@@ -7,7 +7,10 @@
 ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }.call
 
 # Default deploy_to directory is /var/www/my_app
-set :deploy_to, '/home/lyberadmin/argo'
+set :deploy_to, '/opt/app/lyberadmin/argo'
```
mjgiarlo, line 10: These do not vary per env, so instead of cluttering up the individual env configs, do it once here.

lib/tasks/argo_testing.rake
```diff
@@ -1,12 +1,5 @@
 # frozen_string_literal: true
 
-require 'fileutils'
```
mjgiarlo, line 3: Neither of these requires are used here.

lib/tasks/argo_testing.rake
```diff
@@ -17,4 +10,16 @@ rescue LoadError
   end
 end
 
-require 'rspec/core/rake_task' if %w[test development].include? Rails.env
+begin
```
mjgiarlo, line 13: For consistency with above task.

peetucket (approved):

peetucket:
LGTM

PR #2369, opened 2021-01-21
Authorize with cocina; use legacy API

jcoyne:
## Why was this change made?

Currently we receive a deprecation warning in this method because we're doing authorization with ActiveFedora::Base object.  This prevents the deprecation.  This also uses the legacy metadata API to decouple from direct Fedora writes.


justinlittman (approved):

mjgiarlo (approved):
Should AB test in a deployed environment?

jcoyne:
@mjgiarlo I don't believe we have the infrastructure for A/B testing 😆

jcoyne:
Andrew said it looked good.

PR #2377, opened 2021-02-03
Pin zip_tricks to 5.3.1 in order to avoid breaking download all...

### LINKED ISSUE #2320: Argo download all files option fails with "Secure Connection Failed"
**Describe the bug**
The download all files option on an Argo show page fails with the following error (in Firefox, but Chrome also fails to download):

```
Secure Connection Failed

An error occurred during a connection to argo.stanford.edu. PR_CONNECT_RESET_ERROR

    The page you are trying to view cannot be shown because the authenticity of the received data could not be verified.
    Please contact the website owners to inform them of this problem.
```

**User Impact**
It's not possible to download all files from an object with a single click.

**To Reproduce**
1. Go to an Argo item that has files, e.g. https://argo.stanford.edu/view/druid:bb003dn0409
2. Click on download all files (in the right hand side of the "contents" header)
3. Wait
4. Continue to wait
5. See error

**Expected behavior**
A zip should start downloading after a few seconds.

**Additional context**
Could this be the result of something changing after the UIT maintenance shut down? I just noticed the issue this week, but I don't use it that often so I don't remember when it last worked for me.

aaron-collier:
## Why was this change made?

Fixes #2320 

Manually on prod

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,7 @@ gem 'rake'
 gem 'retries'
 gem 'ruby-prof'
 gem 'rubyzip'
-gem 'zip_tricks', '~> 5.3'
+gem 'zip_tricks', '5.3.1' # 5.3.1 is required as 5.4+ breaks the download all feature
 
 # Stanford related gems
 gem 'blacklight', '~> 7.13'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -609,7 +609,7 @@ GEM
     xpath (3.2.0)
       nokogiri (~> 1.8)
     zeitwerk (2.4.2)
-    zip_tricks (5.5.0)
+    zip_tricks (5.3.1)
 
 PLATFORMS
   ruby
@@ -681,7 +681,7 @@ DEPENDENCIES
   webdrivers
   webmock
   webpacker (~> 5.0)
-  zip_tricks (~> 5.3)
+  zip_tricks (= 5.3.1)
 
 BUNDLED WITH
    2.2.7


Gemfile
```diff
@@ -57,7 +57,7 @@ gem 'rake'
 gem 'retries'
 gem 'ruby-prof'
 gem 'rubyzip'
-gem 'zip_tricks', '~> 5.3'
+gem 'zip_tricks', '5.3.1'
```
mjgiarlo, line 60: Can you add a comment here about the pin?

mjgiarlo (approved):
Feel free to merge once you stuff a comment in the Gemfile. Thanks! Good work!

Gemfile
aaron-collier: Done.

jcoyne:
@aaron-collier have you reported this bug to the maintainer of zip_tricks?  Would you please do that and add a comment in the gemfile pointing at the issue?

mjgiarlo:
@jcoyne @aaron-collier do we know enough to write up an actionable, informative issue?

jcoyne:
@mjgiarlo I that that even letting them know that our use case broke between version X and version Y is helpful

mjgiarlo:
> @mjgiarlo I that that even letting them know that our use case broke between version X and version Y is helpful

🆗

jcoyne:
Here are the commits to check. https://github.com/WeTransfer/zip_tricks/compare/v5.3.1...v5.4.0 We could be extra helpful and bisect these.

jcoyne:
I'll bet you a 🍸  that it's https://github.com/WeTransfer/zip_tricks/commit/ccd64dbf25084235892e5f4bba82fb9ca7c13ef6

mjgiarlo:
> I'll bet you a  that it's [WeTransfer/zip_tricks@ccd64db](https://github.com/WeTransfer/zip_tricks/commit/ccd64dbf25084235892e5f4bba82fb9ca7c13ef6)

That's the commit I was most suspicious of, so not taking the bet.

PR #2409, opened 2021-03-01
Decouple workflows controller from Fedora 3

### LINKED ISSUE #2400: Retrieve pid and version of objects from dor-services-app in workflows_controller
Remove the retrieval of the objects from Fedora 3 in load_resource and retrieve pid and version from dor-services-app

jcoyne:
## Why was this change made?

Fixes #2400

### DIFF
--- a/app/components/workflow_process_row.rb
+++ b/app/components/workflow_process_row.rb
@@ -4,11 +4,11 @@
 class WorkflowProcessRow < ApplicationComponent
   # @param [Dor::Workflow::Response::Process] process_status the model for the WorkflowProcess
   # @param [Integer] index the row index
-  # @param [Dor::Item] item the repository object that the workflow is about
-  def initialize(process:, index:, item:)
+  # @param [Cocina::Models::DRO,Cocina::Models::Collection] cocina_object the repository object that the workflow is about
+  def initialize(process:, index:, cocina_object:)
     @process = process
     @index = index
-    @item = item
+    @cocina_object = cocina_object
   end
 
   delegate :pid, :workflow_name, :repository, :name, :status, :datetime,
@@ -24,5 +24,9 @@ class WorkflowProcessRow < ApplicationComponent
     status == 'error'
   end
 
-  attr_reader :process, :index, :item
+  def show_reset_button?
+    error? && can?(:manage_item, cocina_object)
+  end
+
+  attr_reader :process, :index, :cocina_object
 end
--- a/app/controllers/workflows_controller.rb
+++ b/app/controllers/workflows_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class WorkflowsController < ApplicationController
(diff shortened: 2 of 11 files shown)


app/components/workflow_process_row.rb
```diff
@@ -4,7 +4,7 @@
 class WorkflowProcessRow < ApplicationComponent
   # @param [Dor::Workflow::Response::Process] process_status the model for the WorkflowProcess
   # @param [Integer] index the row index
-  # @param [Dor::Item] item the repository object that the workflow is about
+  # @param [Cocina::Models::DRO] item the repository object that the workflow is about
   def initialize(process:, index:, item:)
```
mjgiarlo, line 8: I know it changes a public interface, but should `item` here change to `cocina_item` per our new cocina-models usage guidance?

app/controllers/workflows_controller.rb
```diff
@@ -59,6 +57,8 @@ def update
 
   # add a workflow to an object if the workflow is not present in the active table
   def create
+    cocina = maybe_load_cocina(params[:item_id])
```
mjgiarlo, line 60: Will this always be an item? Or can it be an APO or collection? I ask mostly to help guide what variable name we use here.

app/presenters/workflow_presenter.rb
```diff
@@ -3,9 +3,11 @@
 class WorkflowPresenter
   # @param [Object] view the rails view context
   # @param [WorkflowStatus] workflow_status
-  def initialize(view:, workflow_status:)
+  # @param [Cocina::Models::DRO] item the repository object that the workflow is about
+  def initialize(view:, workflow_status:, item:)
```
mjgiarlo, line 7: Should `item` be `cocina_item` or `cocina_object` instead?

mjgiarlo (commented):
Looks good overall, but have a question about variable naming with regard to cocina model instances.

app/controllers/workflows_controller.rb
jcoyne: It could be a Collection or a DRO, both have workflows.

mjgiarlo: OK. Should we call it `cocina_object` then?

jmartin-sul (commented):
agreed with @mjgiarlo, looks good to me overall, but might be nice to switch from `item` to `cocina_object` across the board

jcoyne:
This already uses cocina_object. Am I missing something?

app/presenters/workflow_presenter.rb
mjgiarlo: @jcoyne here's the one that's outstanding

jcoyne: Thank you. I've got it.

jmartin-sul (approved):
thanks for the ivar/param renaming!

mjgiarlo (approved):

PR #2410, opened 2021-03-01
use cocina based rights options for collections and apos

peetucket:
## Why was this change made?

The analog to changes made to fix a bug with setting the rights of an item in #2394 

This is the attempt to normalize the usage of default object rights and collection rights in all locations and stop using options that exist in the datastream.  We already do this for items, so expect it should be possible for collections and default object rights in APOs as well (and perhaps necessary).  

What is not clear is why we have not not done this yet (i.e. why/if we still need to use the `DEFAULT_RIGHTS_OPTIONS` in these locations in order to match the allowed values in the datastream at https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb) or if it will work using the cocina based `REGISTRATION_RIGHTS_OPTIONS` for items and `COLLECTION_RIGHTS_OPTIONS` for collections.  
This PR makes the changes assuming it will.

### DIFF
--- a/app/forms/access_form.rb
+++ b/app/forms/access_form.rb
@@ -5,14 +5,14 @@
 class AccessForm
   extend ActiveModel::Naming
   # @param [Cocina::Models::DRO, Cocina::Models::Collection] model
-  # @param [String] the default rights to assign to the object, from Constants::REGISTRATION_RIGHTS_OPTIONS (which is defined as RIGHTS_TYPE_CODES in dor-services)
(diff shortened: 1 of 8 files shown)


ndushay:
@peetucket is this made moot by @justinlittman 's wor?  If so please close and remove branch.

peetucket:
Closed in favor of #2718 (other substantive changes were made in other PRs, that PR just changes the comments).

PR #2412, opened 2021-03-01
Decouple collection membership form from Fedora 3

### LINKED ISSUE #2399: Remove legacy dependencies in value_helper
- replace Dor.find with Dor::Services::Client
- replace descMetadata calls with Cocina data

jcoyne:
## Why was this change made?

Fixes #2399

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -88,7 +88,13 @@ class ItemsController < ApplicationController
         format.html { render status: :ok, plain: response_message }
       else
         format.json do
-          new_collection_html = render_to_string('items/_collection_ui_line_item', formats: [:html], layout: false, locals: { col: Dor.find(params[:collection]) })
+          collection = Dor::Services::Client.object(params[:collection]).find
+          new_collection_html = render_to_string('items/_collection_ui_line_item', formats: [:html], layout: false,
+                                                                                   locals: {
+                                                                                     collection_label: collection.label,
+                                                                                     collection_id: collection.externalIdentifier,
+                                                                                     item_id: @object.id
+                                                                                   })
           render status: :ok, plain: { message: response_message, new_collection_html: new_collection_html }.to_json
         end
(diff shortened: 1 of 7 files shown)


app/helpers/value_helper.rb
```diff
@@ -1,17 +1,6 @@
...
 
 module ValueHelper
-  # Renderers
-
-  # Given a fedora object, fetch the preferred object title from MODs, if not found, notify Honeybadger and return Fedora label
-  # @return [String] object title, as pulled from descMetadata
-  def object_title(obj)
-    return obj.descMetadata.title_info.first.strip if obj.descMetadata.title_info.first.present?
-
-    Honeybadger.notify("No title found in descMetadata for collection #{obj.pid}")
```
jcoyne, line 11: The last occurance of this was two years ago, so it's not something we're hitting anymore: https://app.honeybadger.io/projects/49894/faults/45409599

app/controllers/items_controller.rb
```diff
@@ -88,7 +88,13 @@ def add_collection
         format.html { render status: :ok, plain: response_message }
       else
         format.json do
-          new_collection_html = render_to_string('items/_collection_ui_line_item', formats: [:html], layout: false, locals: { col: Dor.find(params[:collection]) })
+          collection = Dor::Services::Client.object(params[:collection]).find
+          new_collection_html = render_to_string('items/_collection_ui_line_item', formats: [:html], layout: false,
+                                                                                   locals: {
+                                                                                     collection_label: collection.label,
+                                                                                     collection_id: col.id,
```
aaron-collier, line 95: Should this be `collection.id`?

aaron-collier (changes requested):

app/controllers/items_controller.rb
jcoyne: Thank you. good catch!

aaron-collier (approved):

PR #2435, opened 2021-03-04
Test on ruby 2.7.2

jcoyne:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ orbs:
   node: circleci/node@2.1.1
 
 references:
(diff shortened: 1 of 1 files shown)


jmartin-sul (approved):
LGTM, feel free to merge.

is ruby getting upgraded on the VM?  i glanced at prod, and it still appears to be running `2.7.1p83`. (that was my only worry, and it's pretty minor, that's the only reason i didn't just merge myself)

jcoyne:
@jmartin-sul we want to test the upgrade before we upgrade, right?

jmartin-sul:
> @jmartin-sul we want to test the upgrade before we upgrade, right?

fair point 🙂

PR #2453, opened 2021-03-10
Bump underlying cocina model that uses publish attr

mjgiarlo:
## Why was this change made?

To test changes before they go to prod.

Will be tested in QA/stage.

None


mjgiarlo:
The flappy specs keep causing test failures :man_shrugging:

aaron-collier (approved):

Issue #2415, opened 2021-03-01
Decouple the catalog controller from fedora 3


aaron-collier:
- [ ] Remove the dependency on F3 for authorizing view_metadata in the catalog controller.

jcoyne:
Cocina doesn't keep a version history, so I'm not sure how you'd get that from cocina.

aaron-collier:
@jcoyne thank you, I've removed that item and will ticket the versioning discussion separately.

jcoyne:
@aaron-collier I think this ticket as written is blocked by that yet to be ticketed work around the version log.

PR #2468, opened 2021-03-11
Decouple CatalogController authorization from Fedora 3

jcoyne:
## Why was this change made?

Ref #2415

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -41,7 +41,7 @@ class Ability
     can %i[manage_item manage_desc_metadata manage_governing_apo view_content view_metadata], [NilModel, Cocina::Models::DRO] if current_user.manager?
     can :create, Cocina::Models::AdminPolicy if current_user.manager?
 
-    can %i[view_metadata view_content], [ActiveFedora::Base, Cocina::Models::DRO] if current_user.viewer?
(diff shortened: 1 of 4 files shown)


app/models/ability.rb
```diff
@@ -41,7 +41,7 @@ def grant_permissions
     can %i[manage_item manage_desc_metadata manage_governing_apo view_content view_metadata], [NilModel, Cocina::Models::DRO] if current_user.manager?
     can :create, Cocina::Models::AdminPolicy if current_user.manager?
 
-    can %i[view_metadata view_content], [ActiveFedora::Base, Cocina::Models::DRO] if current_user.viewer?
+    can %i[view_metadata view_content], [ActiveFedora::Base, Cocina::Models::DRO, Cocina::Models::Collection] if current_user.viewer?
```
jmartin-sul, line 44: i'm a little rusty on argo permission checking stuff:  i think this makes sense, but why no need to also add `Cocina::Models::AdminPolicy`?

jmartin-sul (commented):
lgtm, i think, just one question about a revised permission check

app/models/ability.rb
jcoyne: Sure, I've added that too.  This only affects people who have the `sdr:viewer-role` and I'm not sure who that is.

aaron-collier (approved):

PR #2477, opened 2021-03-11
guard against republushing objects that have not been published

### LINKED ISSUE #2384: Do not allow publish on "registered" objects
**Describe the bug**
It turns out it's possible to publish registered objects. This can happen accidentally via 

- bulk action publish
- bulk action SearchWorks release
- clicking on the publish blue button

**User Impact**
Registered items are in-progress by definition and should not be considered ready for a public purl. There's no easy way to remove the purl if it's published while an object is registered, as we don't have an "unpublish" action. Unpublishing usually only happens if the accessionWF publish step is run. To unpublish you essentially need to set the object rights to dark and run accessioning, which is not a good option for an item that is still only registered.

**Expected behavior**
Only items that have been accessioned should be publishable.

**Additional context**
I've made the mistake of accidentally publishing registered objects myself.

peetucket:
## Why was this change made?

Fixes #2384 

- before allowing a bulk action that republishes or releases object (which also would result in a publish), confirm that the object has actually been published before and abort if not
- rename the publish check route used by the "Republish" blue button to match what it does

- updated tests

### DIFF
--- a/app/presenters/buttons_presenter.rb
+++ b/app/presenters/buttons_presenter.rb
@@ -31,7 +31,7 @@ class ButtonsPresenter
            :refresh_metadata_item_path,
            :item_manage_release_path,
            :embargo_form_item_path,
-           :dor_path,
+           :dor_republish_path,
            to: :url_helpers
 
   def url_helpers
@@ -134,7 +134,7 @@ class ButtonsPresenter
 
   def republish_button
     {
-      url: dor_path(pid: pid),
+      url: dor_republish_path(pid: pid),
       label: 'Republish',
       check_url: workflow_service_published_path(pid),
       new_page: true
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -174,7 +174,7 @@ Rails.application.routes.draw do
   end
 
   namespace :dor do
-    match 'republish/:pid', action: :republish, via: %i[get post]
+    match 'republish/:pid', action: :republish, as: 'republish', via: %i[get post]
     match 'reindex/:pid',   action: :reindex, as: 'reindex', via: %i[get post]
     resources :objects, only: :create # we only implement create for object registration
   end
--- a/spec/jobs/release_object_job_spec.rb
+++ b/spec/jobs/release_object_job_spec.rb
@@ -3,7 +3,6 @@
 require 'rails_helper'
 
 RSpec.describe ReleaseObjectJob do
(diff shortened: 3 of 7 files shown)


app/presenters/buttons_presenter.rb
```diff
@@ -31,7 +31,7 @@ def initialize(manager:, solr_document:)
            :refresh_metadata_item_path,
            :item_manage_release_path,
            :embargo_form_item_path,
-           :dor_path,
+           :dor_republish_path,
```
peetucket, line 34: `dor_path` is not very descriptive since this is actually the republish path

config/routes.rb
```diff
@@ -174,7 +174,7 @@
   end
 
   namespace :dor do
-    match 'republish/:pid', action: :republish, via: %i[get post]
+    match 'republish/:pid', action: :republish, as: 'republish', via: %i[get post]
```
peetucket, line 177: as noted above, `dor_path` is not very descriptive since this is actually the republish path

jcoyne (changes requested):
Shouldn't this guard be in the dor-services-app API where it lets us "publish"?

peetucket:
> Shouldn't this guard be in the dor-services-app API where it lets us "publish"?

like here https://github.com/sul-dlss/dor-services-app/blob/main/app/jobs/publish_job.rb#L12  or here https://github.com/sul-dlss/dor-services-app/blob/main/app/controllers/objects_controller.rb#L90 ?

I guess if we do it in the controller, we could return an http status that indicated it didn't happen?  Not sure what the implications of that would be for consumers

jcoyne:
Yes, exactly. And that would prevent this from happening for any other API user.

spec/jobs/release_object_job_spec.rb
```diff
@@ -3,7 +3,6 @@
 require 'rails_helper'
 
 RSpec.describe ReleaseObjectJob do
```
peetucket, line 5: Note that the tests changed a lot less that it appears from the diffs.  I basically added two separate contexts: 'with already published objects' and 'when objects have never been published'.  The `Dor::Workflow::Client` instance then mocks the return value to verify the publish doesn't happen.  The nesting of the existing tests in a new context block and the addition of the new mock for `Dor::Workflow::Client` resulted in a messy diff.  The end state is really just a few new tests.

peetucket:
The failed test above is the flappy `spec/features/item_registration_spec.rb` spec

jcoyne (approved):

Issue #2438, opened 2021-03-04
Add error handling and display for when users are updating metadata


aaron-collier:
In the legacy update - handle errors and display to user

Anywhere a user provides or updates descriptive metadata, we need error handling if round trip validation fails

PR #2480, opened 2021-03-12
Handle datastream validation errors when updating via legacy metadata…

justinlittman:
… endpoint.

refs #2438

## Why was this change made?
Notify a user when she attempts to update a descMetadata datastream with invalid MODS.

Unit, QA

NA


justinlittman:
See https://github.com/sul-dlss/dor-services-app/pull/2504 for the dor-services-app side.

mjgiarlo (approved):

PR #2493, opened 2021-03-15
Prefer using dor-services-client over Fedora connections

### LINKED ISSUE #2396: Refactor descmetadata_download_job to decouple from fedora 3
- Replace call to fedora 3: https://github.com/sul-dlss/argo/blob/main/app/jobs/descmetadata_download_job.rb#L70
- Retrieve source descriptive metadata (https://github.com/sul-dlss/argo/blob/main/app/jobs/descmetadata_download_job.rb#L47) from dor-services-app (blocked by: https://github.com/sul-dlss/dor-services-app/issues/1885)

mjgiarlo:
Fixes #2396

## Why was this change made?

This refactors the `query_dor` method in the DescMetadataDownloadJob by replacing the call to `Dor.find` and `descMetadata` with corresponding dor-services-client calls. Note that I am not sure we still need the retry behavior that was patched in here previously, so I was inclined to make the smaller change and not remove this code. I might easily be convinced to do that, though.

CI

None

### DIFF
--- a/app/jobs/descmetadata_download_job.rb
+++ b/app/jobs/descmetadata_download_job.rb
@@ -33,18 +33,17 @@ class DescmetadataDownloadJob < GenericJob
   end
 
   def process_druid(current_druid, log, zip_file)
-    dor_object = query_dor(current_druid, log)
-    if dor_object.nil?
+    cocina_object = query_dor(current_druid, log)
+    if cocina_object.nil?
       bulk_action.increment(:druid_count_fail).save
       return
     end
-    cocina_object = Dor::Services::Client.object(current_druid).find
     unless ability.can?(:view_metadata, cocina_object)
       log.puts("#{Time.current} Not authorized for #{current_druid}")
       return
     end
 
-    desc_metadata = dor_object.descMetadata.content
+    desc_metadata = Dor::Services::Client.object(current_druid).metadata.mods
 
     write_to_zip(desc_metadata, current_druid, zip_file)
     log.puts("argo.bulk_metadata.bulk_log_bulk_action_success #{current_druid}")
@@ -64,11 +63,11 @@ class DescmetadataDownloadJob < GenericJob
   # Queries DOR for a given druid, attempting up to MAX_TRIES times in case of failure.
   # @param [String]   druid   The ID of the object to find
   # @param [File]     log     Log file to write to
-  # @return  The Dor::Item corresponding to the given druid, or nil if none was found.
+  # @return [Cocina::Models::DRO,Cocina::Models::Collection,Cocina::Models::AdminPolicy,nil] cocina model instance corresponding to the given druid, or nil if none was found.
   def query_dor(druid, log)
     attempts ||= MAX_TRIES
-    Dor.find druid
-  rescue RestClient::RequestTimeout
+    Dor::Services::Client.object(druid).find
+  rescue Faraday::TimeoutError
     if (attempts -= 1) > 0
       log.puts "argo.bulk_metadata.bulk_log_retry #{druid}"
       sleep(SLEEP_SECONDS**attempts) unless Rails.env.test?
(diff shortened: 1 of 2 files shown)


app/jobs/descmetadata_download_job.rb
```diff
@@ -64,11 +63,11 @@ def process_druid(current_druid, log, zip_file)
...
   # @param [String]   druid   The ID of the object to find
   # @param [File]     log     Log file to write to
-  # @return  The Dor::Item corresponding to the given druid, or nil if none was found.
+  # @return [Cocina::Models::DRO,Cocina::Models::Collection,Cocina::Models::AdminPolicy,nil] cocina model instance corresponding to the given druid, or nil if none was found.
   def query_dor(druid, log)
     attempts ||= MAX_TRIES
-    Dor.find druid
-  rescue RestClient::RequestTimeout
+    Dor::Services::Client.object(druid).find
+  rescue Dor::Services::Client::NotFoundResponse, Dor::Services::Client::UnexpectedResponse
```
mjgiarlo, line 70: Leave previous retry logic in place but use DSC-relevant exceptions. I could be convinced to remove this entirely but didn't seem necessary for the requested refactoring.

spec/jobs/descmetadata_download_job_spec.rb
```diff
@@ -18,21 +18,20 @@
...
     allow(Dor::Services::Client).to receive(:object).with(pid_list[0]).and_return(object_client1)
     allow(Dor::Services::Client).to receive(:object).with(pid_list[1]).and_return(object_client2)
-
-    allow(Dor).to receive(:find).with(pid_list[0]).and_return(item1)
-    allow(Dor).to receive(:find).with(pid_list[1]).and_return(item2)
   end
 
   after do
-    FileUtils.rm_rf(output_directory) if Dir.exist?(output_directory)
+    FileUtils.rm_rf(output_directory)
```
mjgiarlo, line 34: No need to check if dir exists when using `rm_rf` as this method will not complain if the dir doesn't exist. Prefer asserting over asking.

spec/jobs/descmetadata_download_job_spec.rb
```diff
@@ -42,9 +41,7 @@
   end
 
   describe 'start_log' do
-    let(:log) { double('log') }
-
-    before { allow(log).to receive(:flush) }
+    let(:log) { double('log', flush: nil) }
```
mjgiarlo, line 44: No need to do this in two lines when you can do it in one.

spec/jobs/descmetadata_download_job_spec.rb
```diff
@@ -134,21 +148,28 @@
...
       result = download_job.query_dor('druid:hj185xx2222', log)
       expect(result).not_to be_nil
       expect(log).not_to receive(:puts)
     end
 
-    it 'attempts three connections and logs failures' do
-      dor_double = class_double('Dor').as_stubbed_const(transfer_nested_constants: false)
-      expect(dor_double).to receive(:find).exactly(3).times.and_raise(RestClient::RequestTimeout)
-      expect(log).to receive(:puts).twice.with('argo.bulk_metadata.bulk_log_retry druid:123')
-      expect(log).to receive(:puts).once.with('argo.bulk_metadata.bulk_log_timeout druid:123')
```
mjgiarlo, line 149: No need to test this in two places.

app/jobs/descmetadata_download_job.rb
jcoyne: I agree. I don't think we need this, but fine.  However, I don't think that repeating on these exceptions is the equivalent behavior to the timeout.  The equivalent would be a Faraday::TimeoutError

jcoyne (changes requested):

app/jobs/descmetadata_download_job.rb
mjgiarlo: @jcoyne Good call. Made that change and pushed it. Thanks!

jcoyne (approved):

PR #2504, opened 2021-03-16
Use dor services client destroy method in Purge Job

### LINKED ISSUE #2475: Decouple purge_job from Fedora 3
https://github.com/sul-dlss/argo/blob/ba9a94a1f24e302ad2472da7110afc3e7b2a1354/app/jobs/purge_job.rb#L37-L38

jermnelson:
## Why was this change made?
Fixes #2475 

Adjusted purge_job_spec and ran the test suite

n/a


spec/jobs/purge_job_spec.rb
```diff
@@ -54,6 +54,8 @@
     allow(Dor).to receive(:find).with(pids[1]).and_return(item2)
```
jcoyne, line 54: Can you remove this?

jermnelson: Done, pushed up new commit.

Issue #2497, opened 2021-03-16
Refactor items_controller to decouple from fedora 3


aaron-collier:
This is complete when all references to `@object` are replaced with `@cocina`

PR #2516, opened 2021-03-17
use the API to apply admin policy defaults

jcoyne:
## Why was this change made?
This decouples from Fedora 3

Ref #2497

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -4,8 +4,6 @@
 class ItemsController < ApplicationController
   include ModsDisplay::ControllerExtension
   before_action :load_cocina, except: :purl_preview
-  before_action :create_obj, only: %i[apply_apo_defaults]
-
   before_action :authorize_manage!, only: %i[
     add_collection set_collection remove_collection
     mods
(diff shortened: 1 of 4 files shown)


jmartin-sul (approved):
lgtm, test failure seems like flappy specs

mjgiarlo (approved):

app/controllers/items_controller.rb
```diff
@@ -332,11 +332,6 @@ def tags_client
     Dor::Services::Client.object(@cocina.externalIdentifier).administrative_tags
   end
 
-  # Filters
```
jmartin-sul, line 335: `save_and_reindex` still refers to `@object`, but `save_and_reindex` looks like it can be excised also, since nothing seems to call it anymore

jmartin-sul (changes requested):
there's a bit more cruft removal that can be done now that `ItemsController#create_obj` is getting removed.

app/controllers/items_controller.rb
jcoyne: Yes, good call. I removed it.

jmartin-sul: thanks!

jmartin-sul (approved):

PR #2521, opened 2021-03-18
Decouple versionsController from Fedora 3

### LINKED ISSUE #2401: Refactor versions_controller to use dor-services-app metadata.version endpiont
TBD

jcoyne:
## Why was this change made?
Fixes #2401

Tested on QA.

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class VersionsController < ApplicationController
-  before_action :create_obj, only: %i[open_ui close_ui]
+  before_action :load_and_authorize_resource
 
   def open_ui
     respond_to do |format|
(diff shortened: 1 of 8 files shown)


app/controllers/versions_controller.rb
```diff
@@ -10,15 +10,17 @@ def open_ui
...
 
   def close_ui
-    @description = @object.datastreams['versionMetadata'].current_description
-    @tag = @object.datastreams['versionMetadata'].current_tag
-
-    # do some stuff to figure out which part of the version number changed when opening
-    # the item for versioning, so that the form can pre-select the correct significance level
-    @changed_significance = which_significance_changed(get_current_version_tag(@object), get_prior_version_tag(@object))
+    versions = Dor::Services::Client.object(params[:item_id]).versions.inventory
+    previous_version, current_version = versions.sort(&:versionId).last(2)
```
mjgiarlo, line 14: Are the versions not sorted by version ID already?

app/controllers/versions_controller.rb
```diff
@@ -96,8 +77,8 @@ def which_significance_changed(cur_version_tag, prior_version_tag)
     nil
   end
 
-  # Filters
-  def create_obj
-    @object = Dor.find params[:item_id]
+  def load_and_authorize_resource
+    @cocina = Dor::Services::Client.object(params[:item_id]).find
```
mjgiarlo, line 81: Is this always an item or can it vary? Either way, should we adopt our new naming scheme here and call this variable `@cocina_object` etc.?

spec/views/versions/_close_ui.html.erb_spec.rb
```diff
@@ -3,11 +3,12 @@
 require 'rails_helper'
 
 RSpec.describe 'versions/_close_ui.html.erb' do
-  let(:object) { double('object', pid: 'druid:abc123') }
+  before do
+    @cocina = instance_double(Cocina::Models::DRO, externalIdentifier: 'druid:abc123')
```
mjgiarlo, line 7: I'm curious why we need an ivar assignment here over, say, an `assign()` statement.

app/controllers/versions_controller.rb
jcoyne: This is what the code in dor-services is doing( https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/version_metadata_ds.rb#L230) , so I just thought I ought to do it here, although they always do seem to be sorted.

spec/views/versions/_close_ui.html.erb_spec.rb
jcoyne: `assign()` is deprecated and using ivars is the preferred way.

app/controllers/versions_controller.rb
mjgiarlo: It's safe enough, nbd. No harm done in making the sort explicit on this side.

spec/views/versions/_close_ui.html.erb_spec.rb
mjgiarlo: Huh, OK! Good to know, thanks.

mjgiarlo (approved):

PR #2522, opened 2021-03-18
Load the datastream inventory using the API

jcoyne:
## Why was this change made?

Decouples CatalogController from Fedora

Tested on stage.

Before:

Screen Shot 2021-03-18 at 2 36 59 PM

After:

Screen Shot 2021-03-18 at 2 39 14 PM

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -20,62 +20,6 @@ RSpec.describe CatalogController, type: :controller do
     end
   end
 
-  describe '#show' do
-    let(:druid) { 'druid:fakepid' }
-    let(:item) { instance_double(Dor::Item, datastreams: {}) }
-
-    before do
-      ActiveFedora::SolrService.add(id: druid,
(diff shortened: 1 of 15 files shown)


jcoyne:
@andrewjbtw there's a slight UX change in this one (datastream version and control_group dropped) . Do you approve?

spec/controllers/catalog_controller_spec.rb
```diff
@@ -20,62 +20,6 @@
     end
   end
 
-  describe '#show' do
```
jcoyne, line 23: I removed this whole test because it's duplicated by the `spec/feature/item_view_metadata_spec.rb`

andrewjbtw:
@jcoyne can the version number be included without having to stay coupled? As long as we still have Fedora it's useful to see when the number is incremenented. The control group can go.

If it's too much trouble to keep the version number, it can go too. We still have the Fedora web UI and I'm almost certainly the only Fedora version user.

jcoyne:
@andrewjbtw no problem I just need to update once https://github.com/sul-dlss/dor-services-client/pull/212 is merged and released.

jcoyne:
After making the updates, we see the version again:

Screen Shot 2021-03-18 at 5 10 57 PM

aaron-collier (approved):

PR #2561, opened 2021-03-22
Refactor updating the cocina model

jcoyne:
There is now a singe class that is responsible for applying updates to the model


app/services/model_updater.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+# Applies updates Cocina Models
+class ModelUpdater
+  # @param [Cocina::Models::DRO] model the orignal state of the model
+  # @param [Hash<Symbol,String>] attributes the values to update. Pass nil as a value to clear the current field.
+  # @return [Cocina::Models::DRO] the model with updates applied
+  def self.update(model, attributes)
```
justinlittman, line 8: For your consideration: I find the extra indirection of self.update(<with attributes>) confusing. Without reading the code, it isn't immediately apparent which attributes it will accept and how they are handled.

I'd suggest that more focused methods would make updates clearer. For example self.update_catkey().

jcoyne: I've split those parts out into a ChangeSet class.  This is the pattern used by Valkyrie (https://github.com/samvera/valkyrie/wiki/ChangeSets-and-Dirty-Tracking)

justinlittman: Thanks -- this pattern is clearer to me. Does this belong in cocina-models or dor-services-client so that it is available to all apps that might want to update models?

app/models/item_change_set.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+# Represents a set of changes to an item.
```
justinlittman, line 3: Is this intended only for updating Cocina::Models::DRO (i.e., not APOs or Collections)? If so, can you make it explicit?

app/services/item_change_set_persister.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+# Writes updates Cocina Models
```
justinlittman, line 3: Also, is this intended only for updating Cocina::Models::DRO (i.e., not APOs or Collections)? If so, can you make it explicit?

justinlittman (changes requested):

app/services/model_updater.rb
jcoyne: I don't think so, the change-set is tightly coupled to the argo interactions.

app/models/item_change_set.rb
jcoyne: I'm not sure what terminology is explicit in Argo.  I thought "item" was explicit.
Screen Shot 2021-03-23 at 5 42 26 PM

app/services/item_change_set_persister.rb
jcoyne: This class is intended to encapsulate the implementation of the API, so that "DRO" is not relevant from the outside.  All you need to know is that this takes an item and writes it via the API.

justinlittman (approved):

PR #2562, opened 2021-03-23
allow book reading direction to be set with content type updates

### LINKED ISSUE #2534: use "book (ltr)" and "book (rtl)" in the Argo "set content type" menus
This is the last bit of work remaining from https://github.com/sul-dlss/argo/issues/2427

In the Argo registration form and in the Preassembly app, we've decided to include book reading order in the content type menus even though reading order is a subset of the "book" content type.

In Argo the content type menu still shows the generic "book", so it isn't possible to switch the reading direction of an existing book from within Argo itself without editing the datastream.

Screen Shot 2021-03-19 at 4 54 57 PM

We should add "book (ltr)" and "book (rtl)" to this menu for now. Future interface improvements could give us a way to select reading order without treating it like a content type, but this strategy will work for our current system.

peetucket:
## Why was this change made?

~~NOTE: If would best to review, merge and deploy sul-dlss/dor-services-app#2605 first, though if done out of order, this would just result in a `<bookData>` node being left behind in contentMetada if you switched from a book type to a non book type.~~ <-- merged

Fixes #2534 - allow book type content type updates to set reading direction like they do at registration

Note: with this PR the reading direction will be set/removed correctly in the cocina model, but it does not get removed from contentMetadata if you change from a book type to a non-book type.  As noted above, this problem is fixed in sul-dlss/dor-services-app#2605, in the part of the code that translates the cocina model data back into contentMetadata.

Updated tests and verified on argo-qa

### DIFF
--- a/app/controllers/content_types_controller.rb
+++ b/app/controllers/content_types_controller.rb
@@ -48,20 +48,16 @@ class ContentTypesController < ApplicationController
   end
 
   def old_content_type
-    if @cocina_object.type == 'http://cocina.sul.stanford.edu/models/book.jsonld' && @cocina_object.structural.hasMemberOrders[0]&.viewingDirection == 'right-to-left'
+    if @cocina_object.type == Cocina::Models::Vocab.book && @cocina_object.structural.hasMemberOrders[0]&.viewingDirection == 'right-to-left'
       # if we have a book type, we need to also check the viewing direction
       'book (rtl)'
     else
-      Constants::CONTENT_TYPES.key(@cocina_object.type)
+      Constants::CONTENT_TYPES.key(@cocina_object.type) || ''
     end
-  rescue StandardError
-    ''
   end
 
   def old_resource_type
-    Constants::RESOURCE_TYPES.key(@cocina_object.structural.contains.first.type)
-  rescue StandardError
-    ''
+    Constants::RESOURCE_TYPES.key(@cocina_object.structural.contains&.first&.type) || ''
   end
 
   def new_content_type
--- a/app/controllers/content_types_controller.rb
+++ b/app/controllers/content_types_controller.rb
@@ -48,7 +48,7 @@ class ContentTypesController < ApplicationController
   end
 
   def old_content_type
-    if @cocina_object.type == Cocina::Models::Vocab.book && @cocina_object.structural.hasMemberOrders[0]&.viewingDirection == 'right-to-left'
+    if @cocina_object.type == Cocina::Models::Vocab.book && @cocina_object.structural.hasMemberOrders&.first&.viewingDirection == 'right-to-left'
       # if we have a book type, we need to also check the viewing direction
       'book (rtl)'
     else
@@ -57,6 +57,8 @@ class ContentTypesController < ApplicationController
   end
 
   def old_resource_type
+    return '' if [Cocina::Models::Vocab.collection, Cocina::Models::Vocab.admin_policy].include? @cocina_object.type
+
     Constants::RESOURCE_TYPES.key(@cocina_object.structural.contains&.first&.type) || ''
   end
 
--- a/app/controllers/content_types_controller.rb
+++ b/app/controllers/content_types_controller.rb
@@ -40,6 +40,14 @@ class ContentTypesController < ApplicationController
     {}.tap do |attributes|
       attributes[:type] = Constants::CONTENT_TYPES[params[:new_content_type]] if content_type_should_change?
       attributes[:structural] = structural_with_resource_type_changes if resource_types_should_change?
+
+      # The book content type is a special case, since we need to deal with the rtl vs ltr viewing direction attribute update
+      case params[:new_content_type]
+      when 'book (ltr)'
+        attributes[:structural][:hasMemberOrders] = [{ viewingDirection: 'left-to-right' }]
+      when 'book (rtl)'
+        attributes[:structural][:hasMemberOrders] = [{ viewingDirection: 'right-to-left' }]
+      end
     end
   end
 
--- a/app/controllers/content_types_controller.rb
+++ b/app/controllers/content_types_controller.rb
@@ -38,21 +38,51 @@ class ContentTypesController < ApplicationController
 
   def cocina_update_attributes
     {}.tap do |attributes|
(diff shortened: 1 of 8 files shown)


lib/constants.rb
```diff
@@ -50,13 +50,14 @@ module Constants
   ].freeze
 
   CONTENT_TYPES = {
-    'image' => Cocina::Models::Vocab.image,
-    'book' => Cocina::Models::Vocab.book,
+    'book (ltr)' => Cocina::Models::Vocab.book,
+    'book (rtl)' => Cocina::Models::Vocab.book,
```
peetucket, line 54: add reading direction options for the books, and reorder this list to match what is shown in the registration options

app/helpers/registration_helper.rb
```diff
@@ -25,17 +25,8 @@ def apo_list(permission_keys)
...
-      'Map',
-      'Media',
-      '3D',
-      'Document',
-      'Geo',
-      'Webarchive-seed'
-    ]
+    # the content types selectable in registration are capitalized but match what is available when setting content types
+    #  in the bulk and item detail page for setting content types (exception: '3d' needs to go '3D')
+    Constants::CONTENT_TYPES.keys.map { |content_type| content_type == '3d' ? '3D' : content_type.capitalize }
```
peetucket, line 30: refactor to reuse the constants list, which should now be identical

spec/features/item_registration_spec.rb
```diff
@@ -44,6 +44,7 @@
 
     find("option[value='goobiWF']")
     select 'goobiWF', from: 'workflow_id'
+    select 'Book (ltr)', from: 'content_type'
```
peetucket, line 47: we should explicitly select the content type in this test so that if we change the order in the drop-down list, the test still works

app/helpers/registration_helper.rb
peetucket, line 30: refactor to reuse the constants list, which should now be identical

spec/helpers/registration_helper_spec.rb
```diff
@@ -33,6 +33,12 @@
     end
   end
 
+  describe '#valid_content_types' do
+    it 'returns the expected content types in the expected order' do
+      expect(valid_content_types).to eq ['Book (ltr)', 'Book (rtl)', 'File', 'Image', 'Map', 'Media', '3D', 'Document', 'Geo', 'Webarchive-seed']
+    end
+  end
```
peetucket, line 40: this verifies that our refactoring returns the expected list in the expected order

app/views/content_types/_content_type.html.erb
```diff
@@ -3,33 +3,23 @@
...
-  <% old_content_type = begin
-                          Constants::CONTENT_TYPES.key(@cocina_object.type)
-                        rescue StandardError
-                          ''
-                        end %>
-  <% old_resource_type = begin
-                           Constants::RESOURCE_TYPES.key(@cocina_object.structural.contains.first.type)
-                         rescue StandardError
-                           ''
-                         end %>
```
peetucket, line 15: moved this to the controller both to refactor and because of new logic with the retrieving the old content type

app/controllers/content_types_controller.rb
```diff
@@ -38,13 +38,51 @@ def render_error(msg)
...
   end
 
+  def old_content_type
+    if @cocina_object.type == 'http://cocina.sul.stanford.edu/models/book.jsonld' && @cocina_object.structural.hasMemberOrders[0]&.viewingDirection == 'right-to-left'
+      # if we have a book type, we need to also check the viewing direction
+      'book (rtl)'
+    else
+      Constants::CONTENT_TYPES.key(@cocina_object.type)
+    end
+  rescue StandardError
```
jcoyne, line 57: what sort of error are we expecting here? Can you get more specific than "StandardError"?

spec/support/cocina_matchers.rb
```diff
@@ -18,6 +20,10 @@ def match_cocina_type?(actual, expected)
     actual.type == expected[:content_type]
   end
 
+  def match_cocina_viewing_direction?(actual, expected)
+    actual.structural.hasMemberOrders.map(&:viewingDirection).all? { |viewing_direction| viewing_direction.in?(expected[:viewing_direction]) }
```
jcoyne, line 24: `expected[:viewing_direction]` is not a set, so why use `in?`?

lib/constants.rb
jcoyne: I don't think you can do this because there are other places in the code where we call `CONTENT_TYPES#key` and now the values are no longer unique.

jcoyne (changes requested):

app/controllers/content_types_controller.rb
```diff
@@ -38,13 +38,51 @@ def render_error(msg)
...
+      attributes[:structural] = if resource_types_should_change?
+                                  structural_with_resource_type_changes
+                                else
+                                  @cocina_object.structural.new(hasMemberOrders: member_orders)
+                                end
     end
   end
 
+  def old_content_type
+    if @cocina_object.type == 'http://cocina.sul.stanford.edu/models/book.jsonld' && @cocina_object.structural.hasMemberOrders[0]&.viewingDirection == 'right-to-left'
```
jcoyne, line 51: Can you use `Cocina::Models::Vocab.book` here instead?

app/controllers/content_types_controller.rb
```diff
@@ -38,13 +38,51 @@ def render_error(msg)
...
+      # if we have a book type, we need to also check the viewing direction
+      'book (rtl)'
```
jcoyne, line 53: Can you add a test for this case and for (ltr)?  It's not immediately clear in `constants.rb` that the order of the members is important, but it is to this method.

jcoyne (changes requested):

app/controllers/content_types_controller.rb
peetucket: this was copypasta from the view, but not really needed, will ditch the rescue

app/controllers/content_types_controller.rb
peetucket: Yes, good point.  I added several tests to the `content_types_controller_spec` to verify the `old_content_type` method works as expected in various scenarios

jcoyne (approved):

PR #2577, opened 2021-03-25
Use the factories for creating test objects

jcoyne:
## Why was this change made?
This helps work around a problem where the ur-apo gets removed from the index


jmartin-sul (approved):
i learned some things about factorybot from this PR

PR #2578, opened 2021-03-25
Removed a sort field that doesn't exist

jcoyne:
## Why was this change made?

This was added in https://github.com/sul-dlss/argo/commit/893a1eb1cabe58623e2320b76af68adfa6fd3a46 but we never actually indexed into that field


ndushay (approved):
I can't disagree with this assertion.  I tried.  Not merging due to temporary moratorium on merges until dep updates are deployed.

PR #2579, opened 2021-03-25
Stop searching field that is not populated

jcoyne:
## Why was this change made?

Keep it clean.


ndushay (approved):
I also tried to disprove this statement and failed.

Not merging due to temporary moratorium on updating main until dep updates are deployed.

PR #2588, opened 2021-03-26
Display the full dublin core

### LINKED ISSUE #2587: "View Full Dublin Core" is broken
Screen Shot 2021-03-26 at 10 38 24 AM

## Backtrace

    line 4 of [PROJECT_ROOT]/app/views/datastreams/dc.html.erb: block (2 levels) in _app_views_datastreams_dc_html_erb___80685122880831431_4240440
    line 3 of [PROJECT_ROOT]/app/views/datastreams/dc.html.erb: block in _app_views_datastreams_dc_html_erb___80685122880831431_4240440
    line 2 of [PROJECT_ROOT]/app/components/blacklight_modal_component.html.erb: call

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49894/faults/78577927)

jcoyne:
Extracted the metadata service components out into their own controller

## Why was this change made?
Fixes #2587


app/controllers/datastreams_controller.rb
```diff
@@ -4,21 +4,15 @@ class DatastreamsController < ApplicationController
...
-  end
+  before_action :show_aspect, only: %i[show edit]
 
   def edit
     @content = @object_client.metadata.datastream(params[:id])
     render layout: !request.xhr?
   end
 
   def show
-    if params[:dsid] == 'full_dc'
```
mjgiarlo, line 21: Can you say more about the reason for this change?

mjgiarlo (approved):

app/controllers/datastreams_controller.rb
jcoyne: This was broken, if you go to "View Full Dublin Core" and then click "View XML" it was trying to hit this path, but missing because some other PR changed the parameter names.

PR #2592, opened 2021-03-26
Decouple the collection form from Fedora 3

### LINKED ISSUE #2421: Refactor collection_form to decouple from fedora 3
Once a collection is created, retrieve the metadata from cocina instead of fedora 3 (Dor.find)

In order to decouple the [collection_form](https://github.com/sul-dlss/argo/blob/main/app/forms/collection_form.rb#L35), the abstract must be able to be written to cocina (stored as a descriptive value in Note with  type: `summary`, label: `Abstract`)

jcoyne:
## Why was this change made?

Fixes #2421

### DIFF
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -3,7 +3,11 @@
 # Inspired by Reform, but not exactly reform
 # This is for the collection form and it's only used for create, not for update
 # as it registers a new object (of type collection) on each call to `#save`
-class CollectionForm < BaseForm
+class CollectionForm
+  def initialize
+    @errors = ActiveModel::Errors.new(self)
+  end
+
   # @param [HashWithIndifferentAccess] params the parameters from the form
   # @return [Boolean] true if the parameters are valid
   def validate(params)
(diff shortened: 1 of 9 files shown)


app/forms/collection_form.rb
```diff
@@ -14,35 +18,20 @@ def validate(params)
...
 
-  # @return [Dor::Collection] registers the Collection
+  # @return [Cocina::Models::Collection] registers the Collection
   def register_model
-    response = Dor::Services::Client.objects.register(params: cocina_model)
-    WorkflowClientFactory.build.create_workflow_by_name(response.externalIdentifier, 'accessionWF', version: '1')
-    # Once it's been created we populate it with its metadata
-    Dor.find(response.externalIdentifier)
+    Dor::Services::Client.objects.register(params: cocina_model).tap do |response|
+      WorkflowClientFactory.build.create_workflow_by_name(response.externalIdentifier, 'accessionWF', version: '1')
```
mjgiarlo, line 33: I get that this is the code we already had here, but also I wonder about creating accessionWF this way vs. using `Dor::Services::Client.object(response.externalIdentifier).accession.start(workflow: 'accessionWF')`. When should we reach for one over the other?

app/forms/collection_form.rb
```diff
@@ -56,7 +45,7 @@ def cocina_model
       }
     }
 
-    reg_params[:description] = build_description if params[:collection_title].present? && params[:collection_abstract].present?
+    reg_params[:description] = build_description if params[:collection_title].present? || params[:collection_abstract].present?
```
mjgiarlo, line 48: Can you say more about how this change, making the expected descriptive attrs more lenient, factors in?

app/forms/collection_form.rb
jcoyne: I don't know. I didn't look into that. Would you like to create an issue?

app/forms/collection_form.rb
jcoyne: I believe this was a bug. However this code was never activated because we haven't had descriptive being updated in prod, so nobody noticed.

app/forms/collection_form.rb
mjgiarlo: I don't know if it's issue-worthy. There's a ton of `WorkflowClientFactory` in Argo, so I'm not sure there's much value in replacing this one call with a DSC call.

mjgiarlo (approved):
LGTM. Do we want this tested in stage before merge or nah?

jcoyne:
I think we can get a whole bunch of stuff merged and then present it to Andrew. I think that's a more effective use of everyones time.

app/forms/collection_form.rb
andrewjbtw: FWIW, I don't know of a route to this form for an existing collection. Updates to collection description would go through modsulator/the datastream itself.

jcoyne: It's only used for registration of new collections (via a new APO).

PR #2593, opened 2021-03-26
Decouple the registration front end from Fedora

### LINKED ISSUE #2586: Decouple registration form from fedora 3
Currently the ObjectsController#create returns a location header with a path to fedora. The registration javascript then parses the druid from this header. We must decouple this from the Fedora path entirely so that we can remove `ApplicationHelper#fedora_base` https://github.com/sul-dlss/argo/blob/ba9a94a1f24e302ad2472da7110afc3e7b2a1354/app/helpers/application_helper.rb#L8-L10

jcoyne:
It was creating links to Fedora, now create links to the show page instead

## Why was this change made?

Fixes #2586

### DIFF
--- a/app/javascript/registration/grid.js
+++ b/app/javascript/registration/grid.js
@@ -5,21 +5,14 @@ import 'jquery-ui/ui/widgets/dialog'
 import 'jquery-ui/ui/widgets/progressbar'
 
 var gridContext = function() {
-  var druidFormatter = function(val, opts, rowObj) {
-    if (val.trim() != '') {
-      var href = dor_path + "objects/druid:" + val.trim();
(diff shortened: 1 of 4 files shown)


app/javascript/registration/grid.js
```diff
@@ -5,21 +5,14 @@ import 'jquery-ui/ui/widgets/dialog'
 import 'jquery-ui/ui/widgets/progressbar'
 
 var gridContext = function() {
-  var druidFormatter = function(val, opts, rowObj) {
```
jcoyne, line 8: This block was dead code, the real "druidFormatter" is defined farther down in the file.

jmartin-sul (approved):

PR #2595, opened 2021-03-26
Remove unnecessary mocks from tests

jcoyne:


jmartin-sul (approved):

jmartin-sul:
build failure is just because of a codeclimate complaint about a 0.1% coverage drop (and then when you click through for details, it only complains about a coverage drop in app/jobs/descmetadata_download_job.rb, which should be unaffected by this change to spec/forms/collection_form_spec.rb)

PR #2597, opened 2021-03-27
Check for authorization using Cocina models

jcoyne:

### DIFF
--- a/app/views/shared/_user_util_links.html.erb
+++ b/app/views/shared/_user_util_links.html.erb
@@ -49,7 +49,7 @@
       </a>
       <div class="dropdown-menu" aria-labelledby="registerDropdown">
         <%= link_to 'Register Items', registration_path, class: 'dropdown-item' %>
-        <% if can? :create, Dor::AdminPolicyObject %>
+        <% if can? :create, Cocina::Models::AdminPolicy %>
           <%= link_to 'Register APO', new_apo_path, class: 'dropdown-item' %>
         <% end %>
       </div>


ndushay (changes requested):
I think this might need to be vetted by Andrew or is there already an integration test for this permission?

jcoyne:
@ndushay we never grant the `Dor::AdminPolicyObject` permission in Ability, so the only way this ever works is when you are an admin and get `:manage, :all` permissions, thus there is nothing that Andrew could test and there is no way to write an integration test for it.

ndushay:
wouldn't a test be:  if you have admin permission, you can see the Register APO link, and if you don't have admin permission, you can't see the link?   That seems testable

jcoyne:
@ndushay like this? https://github.com/sul-dlss/argo/blob/main/spec/features/apo_spec.rb#L111

ndushay (approved):
The spec you reference has a reference to Dor.find, by the way - https://github.com/sul-dlss/argo/blob/main/spec/features/apo_spec.rb#L88

I think what I'm concerned about is ... gee this one line change looks fine in isolation, but ... so do a lot of changes and we have broken a lot of stuff in production by not foreseeing the implications of changes.   

Really, what I was trying to say was:  if you pause and step back ... is there a way to test that this has no deleterious affects?  That cocina truly does have the correct information this change depends on;  that we have tested that the change works both when permissions are supposed to be granted and when they aren't?   That we have test code to check both those conditions if at all possible?

You are better placed to know the answers to those questions than I am, and I trust you.

jcoyne:
@ndushay 

The Dor find in that spec gets removed in a different PR that decouples the APO from from Fedora 3.

I don't see a way to test this beyond what we already have since we give the administrators wildcard access.  I do know that the existing code fails here when we decouple from dor-services. If you have suggestions on what I should do, I'd be pleased to entertain them.

PR #2603, opened 2021-03-29
Transfer safeurl from dor-services

jcoyne:
## Why was this change made?

So we can decouple from dor-services

### DIFF
--- a/app/components/external_links_component.rb
+++ b/app/components/external_links_component.rb
@@ -13,7 +13,7 @@ class ExternalLinksComponent < ViewComponent::Base
   end
 
   def dor_link
-    link_to 'Fedora UI', File.join(safeurl, "objects/#{document.id}"),
+    link_to 'Fedora UI', File.join(fedora_url_without_credentials, "objects/#{document.id}"),
             target: '_blank', rel: 'noopener', class: 'nav-link'
   end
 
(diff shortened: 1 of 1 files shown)


app/components/external_links_component.rb
```diff
@@ -44,4 +44,12 @@ def released_to_searchworks?
   private
 
   attr_reader :document
+
+  def safeurl
```
ndushay, line 48: "safeurl" isn't a great name -- maybe repository_safeurl or dor_safeurl or storage_safeurl ???

ndushay (changes requested):

app/components/external_links_component.rb
jcoyne: I'd like to keep it the same as we had before in dor-services so that we can easily track down the origin of the code: https://github.com/sul-dlss/dor-services/blob/1743ea8c144467d0167c333d34cfb62b9d2eb7ec/lib/dor/static_config/fedora_config.rb#L27

If you want to change it later, I think that could be a follow on PR.

ndushay: i don't see how keeping the method name the same will actually help us track stuff in another coe base, and espec since I'm suggesting we keep safeurl in the name and just contextualize WHICH url it is in a file named "external_links_component"

mjgiarlo: @jcoyne @ndushay as a compromise, we could consider leaving the name `safeurl` since it establishes a connection to the dor-services code **and** dropping a comment above the new method definition to call out this context further?

ndushay (approved):

PR #2604, opened 2021-03-29
Import xml prettify from dor-services

jcoyne:
## Why was this change made?

This previously was a monkey-patch on Nokogiri::XML.  This helps us decouple from dor-services

### DIFF
--- a/app/controllers/datastreams_controller.rb
+++ b/app/controllers/datastreams_controller.rb
@@ -13,7 +13,7 @@ class DatastreamsController < ApplicationController
 
   def show
     if params[:id] == 'full_dc'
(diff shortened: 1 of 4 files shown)


ndushay:
Does this require Andrew review??

ndushay (changes requested):
marking as request changes to note that this may require andrew review ... not sure.

andrewjbtw:
Is this what displays the datastream and workflow XML display modals? I think it should be ok to merge but I'll be sure to look at those displays on stage when I review.

ndushay (approved):
as long as Andrew is happy, I'm happy.

PR #2605, opened 2021-03-29
Remove dead code

jcoyne:
## Why was this change made?
The Dor::Exception is not raised in this block


ndushay (approved):

ndushay:
merging as it seems Dor::Exception is caught in dor-services-app version service and not passed to dor-services-client.  :-P

PR #2613, opened 2021-03-30
Reduce drag when working locally in development

mjgiarlo:
## Why was this change made?

* Bump dependencies and fix `Dockerfile` regarding mimemagic
  * Without these changes, I could not bring up the `web` container
* Make DSA container create the UR-APO and add a role to the web container
  * This role grants the logged in user (via `ROLES` env var) permission to use any created APOs when registering items.

Both of these changes are included to make it easier to create fixture data.

CI, locally

docker-compose

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       NODE_ENV: development
       RAILS_LOG_TO_STDOUT: 'true'
       REMOTE_USER: blalbrit@stanford.edu
-      ROLES: sdr:administrator-role
+      ROLES: sdr:administrator-role;sdr:service-manager
       # Allow bulk action logs to be written (can't write to /tmp)
       SETTINGS__BULK_METADATA__DIRECTORY: '/app/tmp'
       SETTINGS__BULK_METADATA__TEMPORARY_DIRECTORY: '/app/tmp/tmp'
@@ -61,6 +61,7 @@ services:
       RAILS_LOG_TO_STDOUT: 'true'
       SECRET_KEY_BASE: 769171f88c527d564fb65b4b7ef712d5ae9761a21e26a41cd7c88eb0af89c74f857b9be4089119f71cf806dfc8bf9d9d2f0df91c00b119c96f462b46ebf43b0f
       SOLR_URL: http://solr:8983/solr/argo
+      SETTINGS__ENABLED_FEATURES__CREATE_UR_ADMIN_POLICY: 'true'
       SETTINGS__DOR_INDEXING__URL: http://dor-indexing-app:3000/dor
       SETTINGS__SOLR__URL: http://solr:8983/solr/argo
       SETTINGS__FEDORA_URL: http://fedoraAdmin:fedoraAdmin@fcrepo:8080/fedora


Gemfile.lock
```diff
@@ -313,6 +313,9 @@ GEM
     mime-types (3.3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2021.0225)
+    mimemagic (0.3.10)
```
jcoyne, line 316: why did you add mimemagic as a dependency?

jcoyne (changes requested):

Gemfile.lock
jmartin-sul: from glancing at `Gemfile`, my take is that this came along for the ride with the other updates?

how would you feel about approving/merging this to keep things moving, and ticketing mimemagic cleanup for later?  or, are you worried this'll cause trouble when deployed?  i have no guess on that last point either way.

jmartin-sul (approved):

Gemfile.lock
mjgiarlo: @jmartin-sul it shouldn't take too long or be too hairy to pluck this out and try again, so nbd

jmartin-sul: ah ok, cool

mjgiarlo: @jmartin-sul @jcoyne done.

jmartin-sul (approved):

Gemfile.lock
jmartin-sul: approved again 🙂

mjgiarlo: @jcoyne another look?

Dockerfile
```diff
@@ -7,6 +7,8 @@ RUN apk add --update --no-cache \
   libxslt-dev \
   # needed for mysql2 dependency
   mariadb-dev \
+  # needed because mimemagic-a-palooza
+  shared-mime-info \
```
jcoyne, line 11: Is this still needed since mimemagic has been removed?

jcoyne (changes requested):

Dockerfile
mjgiarlo: apparently not. removing. thanks!

mjgiarlo:
@jcoyne have another look?

docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
       NODE_ENV: development
       RAILS_LOG_TO_STDOUT: 'true'
       REMOTE_USER: blalbrit@stanford.edu
-      ROLES: sdr:administrator-role
+      ROLES: sdr:administrator-role;sdr:service-manager
```
jcoyne, line 16: I don't understand why this change is necessary. This role isn't one we use in https://github.com/sul-dlss/argo/blob/main/app/models/user.rb#L9

jcoyne (changes requested):

docker-compose.yml
mjgiarlo: @jcoyne they're used here: https://github.com/sul-dlss/argo/blob/main/app/forms/apo_form.rb#L7 Without adding this role, if you create an APO as admin, you can't even *see* it in the "register items" view.

jcoyne (approved):

PR #2615, opened 2021-03-31
Decouple from the license services that are part of dor-services

### LINKED ISSUE #2601: Decouple from Dor::CreativeCommonsLicenseService

jcoyne:
## Why was this change made?
fixes #2601
This allows us to decouple from Fedora 3

### DIFF
--- a/app/forms/apo_form.rb
+++ b/app/forms/apo_form.rb
@@ -83,9 +83,7 @@ class ApoForm < BaseForm
   end
 
   def license_options
-    [['-- none --', '']] +
-      options_for_use_license_type(Dor::CreativeCommonsLicenseService, use_license) +
-      options_for_use_license_type(Dor::OpenDataLicenseService, use_license)
+    [['-- none --', '']] + options_for_use_license_type(use_license)
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/forms/apo_form.rb
```diff
@@ -100,16 +98,42 @@ def tags_client
...
+                  uri: 'http://opendatacommons.org/licenses/by/1.0/' },
+    'odc-odbl' => { label: 'Open Data Commons Open Database License 1.0',
+                    uri: 'http://opendatacommons.org/licenses/odbl/1.0/' },
+    'by' => { label: 'Attribution 3.0 Unported',
+              uri: 'https://creativecommons.org/licenses/by/3.0/' },
+    'by-sa' => { label: 'Attribution Share Alike 3.0 Unported',
+                 uri: 'https://creativecommons.org/licenses/by-sa/3.0/' },
+    'by_sa' => { label: 'Attribution Share Alike 3.0 Unported',
+                 uri: 'https://creativecommons.org/licenses/by-sa/3.0/',
+                 deprecation_warning: 'license code "by_sa" was a typo in argo, prefer "by-sa"' },
```
mjgiarlo, line 114: I know you extracted this from dor-services, but I wonder if we can drop this? Could be a follow-up.

app/forms/apo_form.rb
```diff
@@ -100,16 +98,42 @@ def tags_client
...
-        ["#{term.label} (#{term.deprecation_warning})", term.key]
-      elsif term.deprecation_warning.nil?
-        [term.label, term.key]
+  def options_for_use_license_type(current_value)
+    LICENSES.map do |key, attributes|
+      if key == current_value && attributes.key?(:deprecation_warning)
+        ["#{attributes.fetch(:label)} (#{attributes.fetch(:deprecation_warning)})", key]
+      elsif !attributes.key?(:deprecation_warning)
+        [attributes.fetch(:label), key]
       end
```
mjgiarlo, line 137: If we drop the `by_sa` entry above, we can clean this up. If not, NBD.

mjgiarlo (approved):
Have a question for you to consider, and pending that, :heavy_check_mark:

app/forms/apo_form.rb
jcoyne: That seems good to me. I'll drop that change.

mjgiarlo: Also checked in QA and stage and _nada_. :beers:

app/forms/apo_form.rb
jcoyne: Should we keep this code so that we can switch to use URIs and then deprecate all the 3.0 licenses?

mjgiarlo: @jcoyne meh. ok w/ whatever.

mjgiarlo: @jcoyne still :heavy_check_mark: so I leave that to you!

PR #2634, opened 2021-04-02
Remove unused constant

jcoyne:
## Why was this change made?
We no longer use this constant.

### DIFF
--- a/lib/constants.rb
+++ b/lib/constants.rb
@@ -3,24 +3,6 @@
 ##
 # A module for including constants throughout the Argo application
 module Constants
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

justinlittman (changes requested):
So I think this is handled by https://github.com/sul-dlss/argo/pull/2692, which fixes all of the rights options for collections, not just CDL. Let's discuss to see if this PR is necessary.

jcoyne:
@justinlittman I'm fine with #2692 going first, but I really want to excise DEFAULT_RIGHTS_OPTIONS as it's confusing. I don't know if you want to add that to your PR or we do that here.

justinlittman (approved):

aaron-collier (approved):

PR #2651, opened 2021-04-08
Adds managing barcodes to manage catkey bulk action.

### LINKED ISSUE #2550: Add manage barcode bulk action

justinlittman:
closes #2550

## Why was this change made?
The people want to manage their barcodes.

Unit, qa

NA


app/controllers/bulk_actions_controller.rb
```diff
@@ -57,7 +57,10 @@ def bulk_action_params
       :pids,
       manage_release: %i[tag what who to],
       set_governing_apo: [:new_apo_id],
-      manage_catkeys: [:catkeys],
+      set_catkeys_and_barcodes: %i[
+        catkeys catkeys_option
+        barcodes barcodes_options
```
mjgiarlo, line 62: Should `barcodes_options` be `barcodes_option` for consistency w/ catkeys, licenses, and such?

app/forms/bulk_action_form.rb
```diff
@@ -30,16 +30,21 @@ def sync
...
+  # rubocop:disable Metrics/PerceivedComplexity
   def csv_as_string
     csv_file = create_virtual_objects&.fetch(:csv_file) ||
                register_druids&.fetch(:csv_file) ||
-               import_tags&.fetch(:csv_file)
+               import_tags&.fetch(:csv_file) ||
+               set_catkeys_and_barcodes&.fetch(:csv_file)
 
-    # Short-circuit if request is not related to creating virtual objects
+    # Short-circuit if no csv file
```
mjgiarlo, line 41: Good eye. Thanks for adjusting this!

app/forms/bulk_action_form.rb
```diff
@@ -30,16 +30,21 @@ def sync
...
   delegate :action_type, :description, to: :model
 
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
   def csv_as_string
     csv_file = create_virtual_objects&.fetch(:csv_file) ||
                register_druids&.fetch(:csv_file) ||
-               import_tags&.fetch(:csv_file)
+               import_tags&.fetch(:csv_file) ||
+               set_catkeys_and_barcodes&.fetch(:csv_file)
```
mjgiarlo, line 39: Is `:csv_file` an arg for catkeys & barcodes? If so, I'd've expected to see it in the permitted params in the controller (which currently only has catkeys/_option and barcodes/_option.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,131 @@
+# frozen_string_literal: true
+
+##
+# Job to update/add catkey/barcodes to objects
+class SetCatkeysAndBarcodesJob < GenericJob
+  attr_reader :catkeys, :barcodes
```
mjgiarlo, line 6: Are these used? I didn't see any lines assigning a corresponding ivar.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    update_pids, catkeys, barcodes = params_from(params)
+
+    with_bulk_action_log do |log|
+      log.puts("#{Time.current} Starting ManageCatkeyJob for BulkAction #{bulk_action_id}")
+      update_druid_count
+      update_pids.each_with_index do |current_druid, i|
+        change_set = ItemChangeSet.new do |change|
+          change.catkey = catkeys[i] if catkeys
+          change.barcode = barcodes[i] if barcodes
```
mjgiarlo, line 29: Should these conditions be checking e.g. `catkeys` or `catkeys[i]`?

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,131 @@
...
+  # @option params [String] :barcodes list of barcodes to be associated 1:1 with pids in order
+  # @option params [String] :barcodes_option option to update the barcodes
+  # @option params [String] :csv_file CSV string
+  def perform(bulk_action_id, params)
+    super
+
+    update_pids, catkeys, barcodes = params_from(params)
+
+    with_bulk_action_log do |log|
+      log.puts("#{Time.current} Starting ManageCatkeyJob for BulkAction #{bulk_action_id}")
```
mjgiarlo, line 24: `s/ManageCatkeyJob/SetCatkeysAndBarcodesJob/` ?

mjgiarlo (commented):
Some early comments. (Not yet done reviewing but thought I'd send these in the interest of early feedback, if helpful. If not, disregard!)

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: `catkeys`. A null indicates that catkeys was not selected for updating.

justinlittman:
@mjgiarlo Thanks for the early comments! They helped me identify some bugs.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+class SetCatkeysAndBarcodesJob < GenericJob
+  ##
+  # A job that allows a user to specify a list of pids and a list of catkeys to be associated with these pids
+  # @param [Integer] bulk_action_id GlobalID for a BulkAction object
+  # @param [Hash] params additional parameters that an Argo job may need
+  # @option params [Array] :pids required list of
+  # @option params [String] :catkeys list of catkeys to be associated 1:1 with pids in order
+  # @option params [String] :catkeys_option option to update the catkeys
+  # @option params [String] :barcodes list of barcodes to be associated 1:1 with pids in order
+  # @option params [String] :barcodes_option option to update the barcodes
```
mjgiarlo, line 14: Should the `_option` params include `use_` on the front of them, to match what's permitted in the controller above?

app/jobs/set_catkeys_and_barcodes_job.rb
mjgiarlo: @justinlittman Ah, I see, it uses the `use_*_option` params (unless CSV file used instead, in which case the other args are ignored). Thanks.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+      log.puts("#{Time.current} Catkey/barcode added/updated/removed successfully")
+    rescue StandardError => e
+      log.puts("#{Time.current} Catkey/barcode failed #{e.class} #{e.message}")
+      bulk_action.increment(:druid_count_fail).save
+      nil
+    end
+  end
+
+  def dig_from_params_if_option_set(params, key)
+    params.dig(:set_catkeys_and_barcodes, key).split("\n").map(&:strip).map(&:presence) if params.dig(:set_catkeys_and_barcodes, "use_#{key}_option".to_sym) == '1'
```
mjgiarlo, line 65: Can you say more about the case(s) `.split("\n").map(&:strip).map(&:presence)` is meant to handle? Are you trying to split multi-line input into an array of strings without whitespace padding on either side?

If so, I wonder if `params.dig(:set_catkeys_and_barcodes, key).split` might suffice:

```ruby
[1] pry(main)> " 1 \n 2 \n 3 ".split("\n").map(&:strip).map(&:presence)
=> ["1", "2", "3"]
[2] pry(main)> " 1 \n 2 \n   \n 3 ".split("\n").map(&:strip).map(&:presence)
=> ["1", "2", nil, "3"]
[3] pry(main)> " 1 \n 2 \n   \n 3 ".split
=> ["1", "2", "3"]
```

And doesn't the presence function in there mean you may have `nil`s to deal with somewhere?

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+          barcodes ||= []
+          barcodes << row['Barcode'].presence
+        end
+      end
+    else
+      catkeys = dig_from_params_if_option_set(params, :catkeys)
+      barcodes = dig_from_params_if_option_set(params, :barcodes)
+      update_pids = pids
+    end
+    [update_pids, catkeys, barcodes]
```
mjgiarlo, line 89: Take this with a grain of salt: as I'm looking at the code in this job, and looking at the screenshot, I am wondering if implementing this feature as two bulk actions (one for CSVs, one using textareas) might not afford both 1) more easily grokkable code due to reduced responsibilities, and 2) a more streamlined user experience. Tagging @andrewjbtw and @astridu on this. The downside, of course, is that it means yet more bulk actions.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+      log.puts("#{Time.current} Finished SetCatkeysAndBarcodesJob for BulkAction #{bulk_action_id}")
+    end
+  end
+
+  private
+
+  def update_catkey_and_barcode(current_druid, change_set, log)
+    log.puts("#{Time.current} Beginning SetCatkeysAndBarcodesJob for #{current_druid}")
+    cocina = Dor::Services::Client.object(current_druid).find
```
mjgiarlo, line 39: What would you think about naming this variable `cocina_object` per our new handy-dandy cocina-models naming convention?

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+  # @option params [String] :barcodes_option option to update the barcodes
+  # @option params [String] :csv_file CSV string
+  def perform(bulk_action_id, params)
+    super
+
+    update_pids, catkeys, barcodes = params_from(params)
+
+    with_bulk_action_log do |log|
+      log.puts("#{Time.current} Starting SetCatkeysAndBarcodesJob for BulkAction #{bulk_action_id}")
+      update_druid_count
```
mjgiarlo, line 23: Does `update_druid_count` work as expected when going the CSV route? I *think* it could handle both cases if you change this to `update_druid_count(update_pids.count)`.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+  end
+
+  private
+
+  def update_catkey_and_barcode(current_druid, change_set, log)
+    log.puts("#{Time.current} Beginning SetCatkeysAndBarcodesJob for #{current_druid}")
+    cocina = Dor::Services::Client.object(current_druid).find
+
+    unless ability.can?(:manage_item, cocina)
+      log.puts("#{Time.current} Not authorized for #{current_druid}")
```
mjgiarlo, line 42: Should this increment the failed count via `bulk_action.increment(:druid_count_fail).save` ?

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+      return
+    end
+
+    log_update(change_set, log)
+
+    begin
+      object_client = Dor::Services::Client.object(current_druid)
+      cocina_item = object_client.find
```
mjgiarlo, line 50: Don't you already have this stuffed in `cocina` above?

app/models/item_change_set.rb
```diff
@@ -10,11 +10,12 @@ class ItemChangeSet
...
+    barcode
   ].freeze
 
-  def initialize
-    @changes = {}
-    yield self
+  def initialize(changes = {})
+    @changes = changes
+    yield self if block_given?
   end
```
mjgiarlo, line 19: It looks like this change may be motivated by ease of testing, which... it's a simple and small class, so NBD, but I wonder if it's worth supporting two interaction modes for our change sets. An alternative to adding new initialization behavior for easier testing may be a new change set matcher?

Pinging @jcoyne on this for his thoughts.

app/models/item_change_set.rb
```diff
@@ -35,4 +36,11 @@ def initialize
   def changed?
     @changes.any?
   end
+
+  def ==(other)
+    PROPERTIES.each do |property|
+      return false if public_send(property) != other.public_send(property)
+    end
+    true
+  end
```
mjgiarlo, line 45: Cool! Can you say more about how this is used?

app/services/item_change_set_persister.rb
```diff
@@ -65,14 +66,12 @@ def updated_access(updated)
...
+    identification_props[:barcode] = barcode if barcode_changed?
     if catkey_changed?
-      catalog_links = []
-      catalog_links << { catalog: 'symphony', catalogRecordId: catkey } if catkey
-      updated_identification = updated_identification.new(catalogLinks: catalog_links)
+      identification_props[:catalogLinks] = catkey ? [{ catalog: 'symphony', catalogRecordId: catkey }] : nil
     end
-
-    updated.new(identification: updated_identification)
+    updated.new(identification: identification_props.compact.presence)
```
mjgiarlo, line 75: Can you say more about this refactoring? This looks related to a conversation we had on the bulk license action PR, but maybe not.

spec/support/cocina_matchers.rb
```diff
@@ -48,3 +48,16 @@ def match_contained_cocina_types?(actual, expected)
...
+  match do |actual|
+    expected.all? do |expected_key, expected_value|
+      # NOTE: there's no better method on Hash that I could find for this.
+      #        #include? and #member? only check keys, not k/v pairs
+      actual.identification.to_h.any? do |actual_key, actual_value|
+        actual_key == expected_key && actual_value == expected_value
+      end
+    end
+  end
+end
```
mjgiarlo, line 63: :beers:

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+                "Catkey updated to #{change_set.catkey}."
+              else
+                'Catkey removed.'
+              end
+    end
+    if change_set.barcode_changed?
+      msgs << if change_set.barcode
+                "Barcode updated to #{change_set.barcode}."
+              else
+                'Barcode removed.'
```
mjgiarlo, line 124: This looks like it may be uncovered in the tests.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+  end
+  # rubocop:enable Style/GuardClause
+
+  def version_message(change_set)
+    msgs = []
+    if change_set.catkey_changed?
+      msgs << if change_set.catkey
+                "Catkey updated to #{change_set.catkey}."
+              else
+                'Catkey removed.'
```
mjgiarlo, line 117: This looks like it may be uncovered in the tests.

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+      state_service = StateService.new(current_druid, version: cocina_item.version)
+      open_new_version(cocina_item.externalIdentifier, cocina_item.version, version_message(change_set)) unless state_service.allows_modification?
+      ItemChangeSetPersister.update(cocina_item, change_set)
+
+      bulk_action.increment(:druid_count_success).save
+      log.puts("#{Time.current} Catkey/barcode added/updated/removed successfully")
+    rescue StandardError => e
+      log.puts("#{Time.current} Catkey/barcode failed #{e.class} #{e.message}")
+      bulk_action.increment(:druid_count_fail).save
+      nil
```
mjgiarlo, line 60: This looks like it may be uncovered in the tests.

mjgiarlo: Do you think we want HB notifications for failures or meh?

app/jobs/set_catkeys_and_barcodes_job.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  private
+
+  def update_catkey_and_barcode(current_druid, change_set, log)
+    log.puts("#{Time.current} Beginning SetCatkeysAndBarcodesJob for #{current_druid}")
+    cocina = Dor::Services::Client.object(current_druid).find
+
+    unless ability.can?(:manage_item, cocina)
+      log.puts("#{Time.current} Not authorized for #{current_druid}")
+      return
```
mjgiarlo, line 43: This looks like it may be uncovered in the tests.

mjgiarlo (commented):
I have a few niggly questions and some comments, most of which shouldn't be too burdensome. But for this:

> [a]s I'm looking at the code in this job, and looking at the screenshot, I am wondering if implementing this feature as two bulk actions (one for CSVs, one using textareas) might not afford both 1) more easily grokkable code due to reduced responsibilities, and 2) a more streamlined user experience. ... The downside, of course, is that it means yet more bulk actions.

Tagging @andrewjbtw and @astridu on this for their views.

app/jobs/set_catkeys_and_barcodes_job.rb
astridu: Does that mean instead of 1 option for bar codes, it's 2 options: "Manage catkeys/bar codes by entering manually", "Manage catkeys/bar codes by uploading CSV"?

mjgiarlo: @astridu I think so, yeah. When we move to the new design, I think this could work as a single bulk action but with the current UI... there's a lot going on on the page.

app/jobs/set_catkeys_and_barcodes_job.rb
astridu: That's ok with me.

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: Yes. Fixed.

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: I added a comment to make clearer, but need to preserve blank lines as nils to indicate that a barcode/catkey should be removed.

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: Good catch. I just copied this from the original job and missed this opportunity.

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: Fixed.

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: Fixed.

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: Yes, removed. How did I not catch this when I copied from original job?

app/models/item_change_set.rb
justinlittman: It wasn't testing per se. Just a more compact syntax for initializing.

app/models/item_change_set.rb
justinlittman: It is used to compare equality of changesets in specs, e.g., https://github.com/sul-dlss/argo/blob/c1f49ec56bc236a8fe3fd4a7963e021c3ea4580c/spec/jobs/set_catkeys_and_barcodes_job_spec.rb#L94-L96

app/services/item_change_set_persister.rb
justinlittman: `catalogLinks: nil` (which is dropped in compact) instead of `catalogLinks: []`, which is not.

While logically equivalent, this causes problems whenever we try to compare cocina objects. (I think this is going to cause us problems down the road with cocina.)

app/jobs/set_catkeys_and_barcodes_job.rb
justinlittman: The original job didn't. (HB notify or provide test coverage.) I'll add test coverage.

app/jobs/set_catkeys_and_barcodes_job.rb
mjgiarlo: @justinlittman ah, thank you!

app/jobs/set_catkeys_and_barcodes_job.rb
mjgiarlo: :1st_place_medal:

app/jobs/set_catkeys_and_barcodes_job.rb
andrewjbtw: It's ok with me too. But I wonder if we should take this chance to segment the dropdown menu into "Bulk actions that need a CSV" and "Bulk actions that don't". I think this would have been the only one to combine the two.

app/models/item_change_set.rb
mjgiarlo: Aha, thanks

app/jobs/set_catkeys_and_barcodes_job.rb
andrewjbtw: After some preliminary testing I'm now more in favor of separating the CSV from the form. I got tripped up a few times with a CSV in terms of figuring out when to use a form field and when not to. It was mostly inattention on my part (not remembering that I'd clicked "populate from search" before adding my CSV) but I can see it confusing others too.

mjgiarlo (approved):
I have some (minor!) refactoring questions around `ItemChangeSetPersister#update_identification` and `ItemChangeSet#==` but given how many times you've pushed to this branch already, and given that this implements an important new feature, I'm inclined to approve and maybe toss a small refactoring PR at Argo once tested and merged to get your impressions. If that works for you!

andrewjbtw:
Functionally, everything is working in both the CSV and non-CSV actions, as far as I can tell.

A couple comments on the UI:

On the CSV: the text area for listing druids doesn't need to be there. Since everything will come from the CSV, that text area and the "Populate from search" button should be removed.

For the form: in the background within the text areas, I think it would be helpful to have text that indicates: "Check the box and leave this area blank to remove existing [barcodes/catkeys]" There's similar text in the license/statement bulk action.

It does seem like we now have pseudo-A/B testing with the two separate actions, and maybe we'll learn that we only need one over time.

justinlittman:
@andrewjbtw Made fixes.

PR #2661, opened 2021-04-13
remove pids_only param from search context when returning to search view

### LINKED ISSUE #2645: "Back to search" link on Bulk Action page sometimes returns a "pids-only" search
**Describe the bug**
This is an odd one. Under some circumstances, the "Back to search" link on the Bulk actions page links to a "pids-only" search result. The link is here on the "Bulk actions" page:

Screen Shot 2021-04-06 at 1 32 01 PM

When you first load the page, that link takes you back to your most recent search. But after submitting a bulk action, the page is reloaded and the link then goes to a "pids-only" search:

Screen Shot 2021-04-06 at 1 10 15 PM

**User Impact**
Since the pids-only search just lists druids, there's no link for people to follow to get back to their items.

**To Reproduce**
1. Go to bulk actions
2. Click on new bulk action
3. Choose something simple and non-destructive, like reindex
4. Enter a druid in the box
5. Submit the bulk action
6. Look at the "back to search" link again

**Expected behavior**
Back to search should be a regular search.

**Additional context**
This has probably been a problem for a while but not noticed. I don't think it's related to the work cycle.

peetucket:
## Why was this change made?

fixes #2645 - remove pids_only param when linking back to search results from a bulk action

Localhost browser and qa

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -243,6 +243,11 @@ class CatalogController < ApplicationController
     end
   end
 
+  # do not add the pids_only search param to the blacklight search history (used in bulk actions only)
+  def blacklisted_search_session_params
+    super << :pids_only
+  end
+
   # This overrides Blacklight to pass context to the search service
   # @return [Hash] a hash of context information to pass through to the search service
   def search_service_context


app/helpers/bulk_action_helper.rb
```diff
@@ -9,6 +9,12 @@ def render_bulk_action_type(bulk_action)
     render partial: 'default'
   end
 
+  # link back to the search results page, and remove the pids_only param from the search context
+  def back_to_search_results
+    current_search_session.try(:query_params)&.delete('pids_only')
```
jcoyne, line 14: why do you need a `try` here?

jcoyne: Would it be easier to not store `pids_only` in the search history?  That would involve overwriting ` find_or_initialize_search_session_from_params` on the controller to call super with params.except('pids_only')`   I can't think of any time where we'd want that.

peetucket: I think I found a better way using this method I just found in blacklight while poking around: https://github.com/projectblacklight/blacklight/blob/master/app/controllers/concerns/blacklight/search_context.rb#L104-L107

peetucket: Testing in argo-qa, seems to work fine.

jcoyne (approved):
Nice find!

PR #2663, opened 2021-04-14
Refactor: Set an embargo using the update API

jcoyne:
## Why was this change made?

The update API is now mature enough where we can replace and retire the embargo update API.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -121,13 +121,8 @@ class ItemsController < ApplicationController
                          flash: { error: 'Invalid date' }
     end
 
-    object_client = Dor::Services::Client.object(@cocina.externalIdentifier)
-    existing_embargo = @cocina.access.embargo
-    updated_embargo = existing_embargo.new(releaseDate: params[:embargo_date])
-    updated_access = @cocina.access.new(embargo: updated_embargo)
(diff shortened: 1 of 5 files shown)


app/controllers/items_controller.rb
```diff
@@ -114,10 +114,22 @@ def mods
...
+    begin
+      params[:embargo_date].to_date
+    rescue Date::Error
+      return redirect_to solr_document_path(params[:id]),
+                  flash: { error: "Invalid date" }
+    end
+
     object_client = Dor::Services::Client.object(@cocina.externalIdentifier)
-    object_client.embargo.update(embargo_date: params[:embargo_date], requesting_user: current_user.to_s)
+    existing_embargo = @cocina.access.embargo# || Cocina::Models::Embargo.new(access: 'world')
```
justinlittman, line 125: Cruft?

justinlittman (commented):
Should this use the changeset / persister pattern?

app/controllers/items_controller.rb
jcoyne: Yep, removed.

jcoyne:
@justinlittman absolutely right. I just added that in the second commit.

justinlittman (approved):

PR #2668, opened 2021-04-14
Add a feature test for open and close version

jcoyne:
Depends on https://github.com/sul-dlss/dor-services-app/pull/2694

## Why was this change made?

Because we don't have a feature test for these things


ndushay (approved):
LGTM.

mjgiarlo (approved):

PR #2672, opened 2021-04-14
prevent closing a version when invalid cocina

### LINKED ISSUE #2298: Prevent closing objects that lack a valid cocina representation
A benefit of Cocina is that we can track the validity of an object. An object without a valid Cocina representation will hit an error in the accessionWF. Since closing a version from Argo kicks off the accessionWF, we can block the close action if the Cocina representation is not valid.

Benefits: this will leave the object in an opened state, which allows users to take all available actions to fix the issue. Once the object is closed and the accessionWF has started, some actions are blocked and it becomes harder to fix issues.

Challenge: we should show a message if the object can't close because of a Cocina model issue. The usual practice of graying out the "close" button to prevent closing is less desirable here because it does not provide any feedback on why the object can't close.

peetucket:
## Why was this change made?

HOLD - Pending approval by @andrewjbtw 

Fixes #2298 - do not allow a user to close an object (in the UI) if the cocina is invalid.  See the UI screenshot below in the comments.

**Important note**: this does not prevent closing via a bulk action.

Existing tests and verified on QA too

Find objects that are in this state for testing:
QA: https://argo-qa.stanford.edu/catalog?f[data_quality_ssim][]=Cocina+conversion+failed
Stage:  https://argo-stage.stanford.edu/catalog?f[data_quality_ssim][]=Cocina+conversion+failed
Prod: https://argo.stanford.edu/catalog?f[data_quality_ssim][]=Cocina+conversion+failed

To see in action, try this object:

https://argo-qa-a.stanford.edu/view/druid:bg954kx8787

* Open the `identityMetadata` and remove the sourceId element (if it exists) to make it invalid.
* Add the sourceId element back to make it valid again: e.g. ` <sourceId source="petucket">test-04182021abcdef</sourceId>`

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -206,7 +206,7 @@ class CatalogController < ApplicationController
     _deprecated_response, @document = search_service.fetch(params[:id])
 
     @cocina = maybe_load_cocina(params[:id])
-    flash[:error] = 'Warning: this object cannot currently be represented in the Cocina model.' if @cocina.instance_of?(NilModel)
+    flash[:alert] = 'Warning: this object cannot currently be represented in the Cocina model.' if @cocina.instance_of?(NilModel)
 
     authorize! :view_metadata, @cocina
 
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -206,6 +206,8 @@ class CatalogController < ApplicationController
     _deprecated_response, @document = search_service.fetch(params[:id])
 
     @cocina = maybe_load_cocina(params[:id])
+    flash[:error] = 'Warning: this object cannot currently be represented in the Cocina model.' if @cocina.instance_of?(NilModel)
+
     authorize! :view_metadata, @cocina
 
     object_client = Dor::Services::Client.object(params[:id])
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -78,7 +78,7 @@ class VersionsController < ApplicationController
   end
 
   def load_and_authorize_resource
-    @cocina_object = Dor::Services::Client.object(params[:item_id]).find
+    @cocina_object = maybe_load_cocina(params[:item_id])
     authorize! :manage_item, @cocina_object
   end
 end
--- a/app/views/versions/_close_ui.html.erb
+++ b/app/views/versions/_close_ui.html.erb
@@ -1,23 +1,23 @@
-<%= form_with(url: close_item_versions_path(@cocina_object.externalIdentifier), method: :post) do %>
-  <div class='form-group'>
-    <label for="significance">Type</label>
-    <select id="significance" name="significance" class='form-control'>
-      <%# loop through the significance levels and pre-select the one that was chosen when opening the version %>
-      <% @significance_selected.keys.each do |significance| %>
-        <option value="<%= significance.to_s %>"<%= @significance_selected[significance] ? ' selected' : '' -%>><%= significance.to_s.capitalize -%></option>
-      <% end %>
-    </select>
-  </div>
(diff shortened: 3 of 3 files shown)


peetucket:
This PR changes the "close window" pop-up window UI to hide the close button and show a message when an object's cocina model cannot be created:

**Current close window version:** 

Screen Shot 2021-04-14 at 1 33 46 PM

**When invalid cocina:**

Screen Shot 2021-04-14 at 1 33 13 PM

app/controllers/items_controller.rb
```diff
@@ -115,7 +115,7 @@ def show
     authorize! :view_metadata, @cocina
 
     respond_to do |format|
-      format.json  { render json: @cocina }
+      format.json { render json: @cocina }
```
peetucket, line 118: rubocop complaint

app/controllers/versions_controller.rb
```diff
@@ -78,7 +78,7 @@ def which_significance_changed(current_tag, previous_tag)
   end
 
   def load_and_authorize_resource
-    @cocina_object = Dor::Services::Client.object(params[:item_id]).find
+    @cocina_object = maybe_load_cocina(params[:item_id])
     authorize! :manage_item, @cocina_object
```
peetucket, line 82: this method is in `application_controller` and loads the object but prevents an exception from being thrown if the cocina representation fails

app/controllers/catalog_controller.rb
```diff
@@ -206,6 +206,8 @@ def show
     _deprecated_response, @document = search_service.fetch(params[:id])
 
     @cocina = maybe_load_cocina(params[:id])
+    flash[:error] = 'Warning: this object cannot currently be represented in the Cocina model.' if @cocina.instance_of?(NilModel)
```
jcoyne, line 209: should this be `flash[:warn] = ` as it's a warning, not an error.

app/controllers/items_controller.rb
jcoyne: Thank you.

app/views/versions/_close_ui.html.erb
```diff
@@ -12,7 +12,12 @@
     <label for="description">Version description</label><br>
     <textarea id="description" name="description" class='form-control'><%= @description %></textarea>
   </div>
-  <button type='submit' class='btn btn-primary'>
-    Close Version
-  </button>
+  <% if @cocina_object.instance_of?(NilModel) %>
+    <div class='alert alert-danger'>This object cannot currently be represented in the Cocina model and thus should
```
jcoyne, line 16: I'm wondering if it makes sense to show any of the form. I think that is a little bit confusing that  it lets you can enter values, but you can't save them.  Also consider `alert-warn` rather than `alert-danger`.

app/controllers/catalog_controller.rb
peetucket: yes, that's better - i changed it

peetucket:
Updated the screenshots after switching to yellow warnings instead of red error notices, and hiding the whole close form instead of just the button.

jcoyne (approved):

andrewjbtw:
I really like this approach, both the banner on the show page and the message in the modal. 

For the message in the modal, I'd like to give people info on how to get help since they may not know what specific error to look out for and fix. Since there's space in the modal for a longer message:
```
This object cannot currently be represented in the Cocina model and thus cannot be closed. Please check the object for possible metadata errors, such as a missing title or source id. If you are unsure of what the issue is, you can ask for help on the #dlss-aaas Slack channel or email argo-feedback@lists.stanford.edu
```

For the banner message, I think the current message is good. There's less space there, so I'm hesitant to put in a long message.

peetucket:
I updated the message in the modal and verified on QA.

PR #2677, opened 2021-04-15
Adds manage embargo bulk action.

### LINKED ISSUE #2671: Add "Manage embargo" bulk action
This bulk action goes with #2009 (the single-object case).

The goal is to allow users to:

- Create embargoes for a set of objects
- Update embargoes for a set of objects

**Example use case**

University Archives has a set of meeting minutes that need to be embargoed for 15 years from date of the meeting. Each meeting has its own druid. They would like to set the embargo by supplying a list of

- druids
- release dates
- rights to be applied upon release

Example: https://jirasul.stanford.edu/jira/browse/LEGACY-4765

**Functionality**

This would be best done via CSV upload rather than through a form, since the bulk use case is one where the data is likely to vary (different dates or rights for different druids).

Headers:

- druid
- release_date (in YYYY-MM-DD format)
- rights

All three fields are required. Also, adding or extending an embargo requires versioning, so objects either need to be open already and/or opened automatically by running this action.

**Removing an embargo**

In the beginning, I don't think we need to support lifting an embargo via the bulk action by simply removing it. Instead, we should retain the existing practice where an embargo is lifted via setting it to release on the current day. But I could be convinced otherwise.

justinlittman:
closes #2671

## Why was this change made?
Support embargo management from Argo.

Unit, QA

NA

### DIFF
--- a/app/jobs/manage_embargoes_job.rb
+++ b/app/jobs/manage_embargoes_job.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+##
+# Job to update/add embargoes to objects
+class ManageEmbargoesJob < GenericJob
+  ##
+  # A job that allows a user to provide a spreadsheet for managing embargoes.
+  # @param [Integer] bulk_action_id GlobalID for a BulkAction object
+  # @param [Hash] params additional parameters that an Argo job may need
+  # @option params [String] :csv_file CSV string
+  def perform(bulk_action_id, params)
+    super
+
+    with_bulk_action_log do |log|
+      update_pids, embargo_release_dates, rights = params_from_csv(params)
+      log.puts("#{Time.current} Starting ManageEmbargoesJob for BulkAction #{bulk_action_id}")
+      update_druid_count(count: update_pids.count)
+      update_pids.each_with_index do |current_druid, i|
+        update_embargo(current_druid, embargo_release_dates[i], rights[i], log) unless current_druid.nil?
+      end
+      log.puts("#{Time.current} Finished ManageEmbargoesJob for BulkAction #{bulk_action_id}")
+    end
+  end
+
+  private
+
+  def params_from_csv(params)
+    druids = []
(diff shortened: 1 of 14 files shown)


justinlittman:
Awaiting further clarification of requirements for embargo rights.

justinlittman:
Tested on QA by @andrewjbtw

app/jobs/manage_embargoes_job.rb
```diff
@@ -0,0 +1,101 @@
...
+      open_new_version(cocina_object.externalIdentifier, cocina_object.version, msg) unless state_service.allows_modification?
+
+      changes = {}
+      changes[:embargo_release_date] = embargo_release_date if embargo_release_date
+      changes[:embargo_access] = rights if rights
+      unless changes.empty?
+        change_set = ItemChangeSet.new(cocina_object)
+        change_set.validate(changes)
+        change_set.save_model
+        # ItemChangeSet.new(cocina_object).validate(changes).save_model unless changes.empty?
```
jcoyne, line 88: cruft?

app/jobs/manage_embargoes_job.rb
```diff
@@ -0,0 +1,101 @@
...
+      state_service = StateService.new(current_druid, version: cocina_object.version)
+      open_new_version(cocina_object.externalIdentifier, cocina_object.version, msg) unless state_service.allows_modification?
+
+      changes = {}
+      changes[:embargo_release_date] = embargo_release_date if embargo_release_date
+      changes[:embargo_access] = rights if rights
+      unless changes.empty?
+        change_set = ItemChangeSet.new(cocina_object)
+        change_set.validate(changes)
+        change_set.save_model
```
jcoyne, line 87: You should call `save` here, `save_model` is an internal method called by save.

jcoyne (changes requested):

jcoyne (approved):

PR #2678, opened 2021-04-15
Adds new method for determining option for default object rights

### LINKED ISSUE #2631: Edit APO form loads the incorrect default object rights in the menu when rights are location-based
**Describe the bug**
When the default object rights on an APO are location-based, the Edit APO form pre-populates the access rights dropdown with the value of "World".

APO https://argo-qa.stanford.edu/view/druid:vr838wn4827 has rights as location: Archive of Recorded Sound ("ars"):

Screen Shot 2021-04-02 at 1 53 51 PM

But when loading the Edit APO form, the menu shows "world" selected:
Screen Shot 2021-04-02 at 1 55 26 PM

**User Impact**
It would be easy to accidentally change the default object rights to world when editing the APO to change another value.

**Expected behavior**
The currently selected value should populate the menu.

jermnelson:
## Why was this change made?
Fixes #2631

New unit tests

n/a

### DIFF
--- a/app/forms/apo_form.rb
+++ b/app/forms/apo_form.rb
@@ -35,7 +35,7 @@ class ApoForm
   def populate_from_model
     self.title = model.description.title.first.value
     self.agreement_object_id = model.administrative.referencesAgreement
-    self.default_rights = model.administrative.defaultAccess&.access
+    self.default_rights = default_object_rights
     self.use_statement = model.administrative.defaultAccess&.useAndReproductionStatement
     self.copyright_statement = model.administrative.defaultAccess&.copyright
     self.use_license = model.administrative.defaultAccess&.license
@@ -79,19 +79,6 @@ class ApoForm
     'none'
   end
 
-  # rubocop:disable Metrics/CyclomaticComplexity
-  # rubocop:disable Metrics/PerceivedComplexity
-  def default_object_rights
-    return default_rights if model.nil?
-    return "loc:#{model.administrative.defaultAccess.readLocation}" if default_rights == 'location-based'
-    return 'cdl-stanford-nd' if model.administrative&.defaultAccess&.controlledDigitalLending
-    return "#{default_rights}-nd" if model.administrative&.defaultAccess&.download == 'none' && default_rights.in?(%w[stanford world])
-
-    default_rights
-  end
-  # rubocop:enable Metrics/CyclomaticComplexity
-  # rubocop:enable Metrics/PerceivedComplexity
-
   private
 
   def default_collections
@@ -112,6 +99,24 @@ class ApoForm
         [attributes.fetch(:label), attributes.fetch(:uri)]
       end
     end.compact # the options block will produce nils for unused deprecated entries, compact will get rid of them
+  end
+
+  def default_object_rights
+    return default_rights if model.nil?
+
+    check_default_object_rights || default_rights
+  end
+
+  def check_default_object_rights
+    default_access = model.administrative&.defaultAccess
+
+    return if default_access.nil?
+
+    access = default_access.access
+
+    return "loc:#{default_access.readLocation}" if access == 'location-based'
+    return 'cdl-stanford-nd' if default_access.controlledDigitalLending
+    return "#{access}-nd" if default_access.download == 'none' && access.in?(%w[stanford world])
   end
 
   def manage_permissions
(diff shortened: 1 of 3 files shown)


app/forms/apo_form.rb
```diff
@@ -79,6 +79,19 @@ def collection_radio
     'none'
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
```
jermnelson, line 82: Open to suggestions on how to refactor this method to remove these rubocop exceptions.

app/forms/apo_form.rb
```diff
@@ -79,6 +79,19 @@ def collection_radio
...
+  # rubocop:disable Metrics/PerceivedComplexity
+  def default_object_rights
+    return default_rights if model.nil?
+    return "loc:#{model.administrative.defaultAccess.readLocation}" if default_rights == 'location-based'
+    return 'cdl-stanford-nd' if model.administrative&.defaultAccess&.controlledDigitalLending
```
jcoyne, line 87: why do you need &. in this line but not on the prior line?

app/forms/apo_form.rb
jcoyne: how about doing something like `default_access = model.administrative.defaultAccess` and using that local var throughout. That will likely reduce complexity substantially.

jermnelson: Thanks! That changed helped drop the complexity to 8/7. Moved some of the code to the private `check_default_object_rights` method and now passes.

app/forms/apo_form.rb
jermnelson: Thanks for the catch; updated in latest commit.

app/forms/apo_form.rb
```diff
@@ -79,6 +79,14 @@ def collection_radio
     'none'
   end
 
+  def default_object_rights
+    return default_rights if model.nil?
+
+    default_object_rights_option = check_default_object_rights
```
jcoyne, line 85: can you just return `check_default_object_rights || default_rights`?

app/forms/apo_form.rb
```diff
@@ -101,6 +109,14 @@ def options_for_use_license_type(current_value)
     end.compact # the options block will produce nils for unused deprecated entries, compact will get rid of them
   end
 
+  def check_default_object_rights
+    default_access = model.administrative&.defaultAccess
```
jcoyne, line 113: How about handling the case where default_access is `nil` first? Then you can get rid of the rest of the `&.` in this method.

app/forms/apo_form.rb
jermnelson: 👍

app/forms/apo_form.rb
jermnelson: Done, thanks!

app/forms/apo_form.rb
jcoyne, line 85: Do we need this allocation? Can you just move the call to the last line?

jermnelson: Done

app/views/apo/_form.html.erb
```diff
@@ -103,7 +103,7 @@
 
   <div class="form-group">
     <%= f.label :default_rights, "Default Object Rights" %>
-    <%= f.select :default_rights, options_for_select(Constants::REGISTRATION_RIGHTS_OPTIONS, @form.default_rights), {}, class: 'form-control' %>
+    <%= f.select :default_rights, options_for_select(Constants::REGISTRATION_RIGHTS_OPTIONS, @form.default_object_rights), {}, class: 'form-control' %>
```
jcoyne, line 106: It's conventional for the existing value to come from a method with the same name as the form field (as they were previously).  Can you align these?

jcoyne (changes requested):

app/views/apo/_form.html.erb
jcoyne: I think you might want to keep the ERB as it was and change this line: https://github.com/sul-dlss/argo/blob/48b8c859a723eba4092f50870c6de1d6e2560ffc/app/forms/apo_form.rb#L38 to call the `default_object_rights` method you defined.

jermnelson: Done, also made `default_object_rights` private and adjusted the tests accordingly

jcoyne (approved):

PR #2682, opened 2021-04-16
Restore missing remove icon on the tags facet

jcoyne:
## Why was this change made?
This went missing in the Blacklight 7 upgrade.

Before
Screen Shot 2021-04-16 at 3 58 24 PM
After

Screen Shot 2021-04-16 at 4 02 38 PM

### DIFF
--- a/app/javascript/argo.js
+++ b/app/javascript/argo.js
@@ -7,6 +7,8 @@ import JSONRenderer from 'controllers/json_renderer'
 import Tokens from 'controllers/tokens'
 import WorkflowGrid from 'controllers/workflow_grid_controller'
 import { Application } from 'stimulus'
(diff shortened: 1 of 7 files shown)


app/javascript/argo.js
```diff
@@ -7,6 +7,8 @@ import JSONRenderer from 'controllers/json_renderer'
 import Tokens from 'controllers/tokens'
 import WorkflowGrid from 'controllers/workflow_grid_controller'
 import { Application } from 'stimulus'
+import BlacklightHierarchyController from 'blacklight-hierarchy/app/assets/javascripts/blacklight/hierarchy/blacklight_hierarchy_controller'
```
aaron-collier, line 10: This is kinda gross, but it is what it is ;-)

aaron-collier (approved):

PR #2687, opened 2021-04-19
Add a mechanism for registering an agreement

### LINKED ISSUE #2357: Add ability to create agreement objects
To support accessioning agreement objects (see https://github.com/sul-dlss/pre-assembly/issues/743 ) we need to be able to register them. This ticket is to add the ability to register agreements. Currently, we can only register items or APOs. To register agreements we will need either a new form or a modification to the existing item registration form.

Being able to register agreements will help us detach the agreement creation process from Hydrus, which it currently uses: https://consul.stanford.edu/display/SDRA/Instructions+for+Creating+Agreements+Objects

This issue can be closed when both of these are done:
- agreements can be registered
- agreements can be uploaded (which "needs" rails 6 upgrade)

jcoyne:
## Why was this change made?
Fixes #2357

Tested locally, added a feature test and tested on stage.

Screen Shot 2021-04-26 at 5 28 14 PM

### DIFF
--- a/app/forms/agreement_form.rb
+++ b/app/forms/agreement_form.rb
@@ -3,10 +3,14 @@
 class AgreementForm < Reform::Form
   property :title, virtual: true
   property :source_id, virtual: true
-  property :agreement_file, virtual: true
+  property :agreement_file1, virtual: true
+  property :agreement_file2, virtual: true
 
+  validates :title, presence: true
   validates :source_id, format: { with: /\A\w+:\w+\z/,
-                                  message: 'must have a colon' }
+                                  message: 'must have a single colon in the middle' },
+                        presence: true
+  validates :agreement_file1, presence: true
 
   def persisted?
     false
@@ -21,8 +25,10 @@ class AgreementForm < Reform::Form
   end
 
   def save_model
-    @model = Dor::Services::Client.objects.register(params: model)
+    @model = Result.new(RegisterAgreement.register(model: model, uploaded_files: agreement_files))
   end
+
+  Result = Struct.new(:externalIdentifier)
 
   def new_resource(title:, source_id:)
     Cocina::Models.build_request(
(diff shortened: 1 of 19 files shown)


app/forms/agreement_form.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class AgreementForm < Reform::Form
+  property :title, virtual: true
+  property :source_id, virtual: true
+  property :agreement_file, virtual: true
+
+  validates :title, presence: true
+  validates :source_id, format: { with: /\A\w+:\w+\z/,
+                                  message: 'must have a colon' },
```
peetucket, line 10: If we want to be precise, the message is technically "must have a single colon in the middle".

peetucket:
Are you going to add the close version to the PR as noted in the slack thead? https://stanfordlib.slack.com/archives/C09M7P91R/p1618943506454400

I registered a new agreement in -stage (https://argo-stage.stanford.edu/view/druid:xv514tw3686), and without closing, it doesn't run through `accessionWF`, which I believe means it will not be preserved or made versionable.

jcoyne:
@peetucket I've flipped the bit so that new uploads are immediately accessioned.

andrewjbtw:
This looks good. I have two questions:

1. Could the box for the upload show the name of the selected file? It still shows "Agreement file" even after a file has been chosen.
Screen Shot 2021-04-22 at 2 50 26 PM

2. Would it be possible to allow upload of two files? Or does that add too much complexity? Often agreements include a .docx (editable for potential revisions) and a .pdf (the currently active, signed version).

Both are nice-to-haves. Additional files can be added with preassembly if need be.

jcoyne:
The two files will be easy. Unfortunately file upload fields are notoriously difficult to style.  Would you settle for less style and more functionality?

andrewjbtw:
That's fine with me. This is a case where the functionality is intended for so few users that having it work for now is more important than the display styling. Would it be simpler to simply indicate that a file was successfully uploaded and leave out the file name?

jcoyne:
@andrewjbtw I've added a second file to the form.

aaron-collier (approved):

PR #2692, opened 2021-04-20
Updates cocina model for new access/embargo modeling.

justinlittman:
## Why was this change made?
Tighter modeling of access/embargoes, including collection access.

Unit

NA

### DIFF
--- a/app/services/cocina_dro_access.rb
+++ b/app/services/cocina_dro_access.rb
@@ -13,14 +13,16 @@ class CocinaDroAccess
 
     data = if rights == 'cdl-stanford-nd'
              {
-               access: 'citation-only',
+               access: 'stanford',
                download: 'none',
+               readLocation: nil,
                controlledDigitalLending: true
              }
            elsif rights.end_with?('-nd') || %w[dark citation-only].include?(rights)
              {
                access: rights.delete_suffix('-nd'),
                download: 'none',
+               readLocation: nil,
                controlledDigitalLending: false
              }
            elsif rights.start_with?('loc:')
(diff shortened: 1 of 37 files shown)


app/services/cocina_dro_access.rb
```diff
@@ -13,14 +13,16 @@ def self.from_form_value(rights)
 
     data = if rights == 'cdl-stanford-nd'
              {
-               access: 'citation-only',
+               access: 'stanford',
```
justinlittman, line 16: This is a mistake. CDL should be access = stanford.

lib/constants.rb
```diff
@@ -39,15 +39,7 @@ module Constants
...
-    %w[Stanford stanford],
-    ['Location: Special Collections', 'loc:spec'],
-    ['Location: Music Library', 'loc:music'],
-    ['Location: Archive of Recorded Sound', 'loc:ars'],
-    ['Location: Art Library', 'loc:art'],
-    ['Location: Hoover Library', 'loc:hoover'],
-    ['Location: Media & Microtext', 'loc:m&m'],
-    ['Dark (Preserve Only)', 'dark'],
-    ['Citation Only', 'citation-only']
+    ['Dark (Preserve Only)', 'dark']
```
justinlittman, line 42: World and dark only for collections.

spec/controllers/content_types_controller_spec.rb
```diff
@@ -18,9 +18,7 @@
       'version' => 1,
       'type' => content_type,
       'externalIdentifier' => pid,
-      'access' => {
-        'access' => 'world'
-      },
+      'access' => {},
```
justinlittman, line 21: You'll see this change in many places. It is just using some valid value (in this case, the default dark), rather than an invalid value (world access without a download attribute). The actual valid value doesn't matter.

jcoyne (approved):

jcoyne (approved):

PR #2693, opened 2021-04-20
Refactor change sets to use reform-rails

jcoyne:
## Why was this change made?

This allows us to use a standard pattern rather than inventing our own.  This is more maintainable as we have less code to maintain and it is a library we are using in other projects such as h2.  Now fewer classes need to be aware of Persisters.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -37,8 +37,9 @@ class ItemsController < ApplicationController
   end
 
   def set_collection
-    change_set = ItemChangeSet.new(collection_ids: Array(params[:collection].presence))
-    ItemChangeSetPersister.update(@cocina, change_set)
+    change_set = ItemChangeSet.new(@cocina)
+    change_set.validate(collection_ids: Array(params[:collection].presence))
+    change_set.save
     reindex
 
     response_message = if params[:collection].present?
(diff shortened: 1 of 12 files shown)


justinlittman:
After reading the code, I'm unclear how this is a win. Before, I could understand change sets / persisters by reading the ruby code. Now I have to read / understand reform before reading the change sets / persisters code.

Is there any features of reform that we're taking advantage of here that wasn't already implemented in the existing pattern? (Other than less classes needing to be aware of persisters? A `save` method could easily be added to existing persisters to address this.)

jcoyne:
This eliminates a bunch of meta programming that we're doing that historically has been a source of trouble. We're already using reform-rails in h2, so having a single gem to be aware of across projects makes it easier to switch between projects.  I think having one project using reform and having another project that uses something that looks almost like reform but isn't is more or a challenge. 

Another advantage is that this will provide consistency between change sets within Argo itself.  Currently some change sets take a model and others take attributes, this is inconsistent and confusing.  This means we adapt to the established patterns and don't have to spend time discovering what the patterns should be.

aaron-collier:
I'm getting 9 of 11 integration tests failing on this branch. perhaps related to the other services, but wondering how we can coordinate making sure this doesn't break that test suite when deployed.

jcoyne:
@aaron-collier that doesn't make a lot of sense because the feature specs in argo are passing and this shouldn't affect any external service integrations.  Something changed in main maybe?

ndushay:
Putting in the obvious:   this needs testing and it blocks the fix of allowing catkey updates in collections.

aaron-collier:
Integration tests are passing so my above failures were most likely unrelated.

aaron-collier (approved):

andrewjbtw:
Not sure if it's directly related to reform, but this PR appears to be on stage right now and I'm seeing that the blue button "Manage release" modal is missing the radio buttons and dropdown menu
Screen Shot 2021-04-27 at 4 57 36 PM

andrewjbtw:
Looks like it was in this morning's prod deploy but I had missed it earlier, and now it's already fixed.

PR #2643, opened 2021-04-06
Upgrade to rails 6.0

jcoyne:
## Why was this change made?
Fixes #2625

PR #2706, opened 2021-04-22
Upgrade to Rails 6.1

jcoyne:
## Why was this change made?
Keep up to date on dependencies while we're in a workcycle.
Follows on from #2643 and #2625


config/environments/production.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'active_support/core_ext/integer/time'
```
jmartin-sul, line 3: was the inclusion of this `require` in prod and test configs but not the dev config intentional, or an oversight?  i'm not totally sure what extensions this provides, but a bit of cursory searching indicates that making the same inclusion in all envs is advised? https://github.com/rails/rails/commit/c8c549750da40cd540ad1064b0b75e5b4e32d3f1

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,4 +3,6 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += %i[
+  passw secret token _key crypt salt certificate otp ssn
```
jmartin-sul, line 7: no need to keep `:password` in the list?

jmartin-sul (commented):
a couple non-blocking questions, but otherwise LGTM

config/environments/production.rb
jcoyne: Yes, I think this was a merge problem. I'll add that to development.rb too.

jmartin-sul (approved):
approved, but refraining from merging to allow some testing (just automated integration tests?  manual testing also?)

ndushay:
Holding up for access rights?   or combine???

jcoyne:
No blockers from me. I'm leaving this for the TLs and FR to figure out what order to deploy/test stuff in.

aaron-collier (approved):

PR #2713, opened 2021-04-27
[HOLD] Select collections in the current APO

### LINKED ISSUE #2380: When a collection is moved to a new APO, add the collection to the default collections of the new APO and remove it from the defaults of the old collection APO
Argo supports moving a collection object from one APO to another through the blue button and bulk job interfaces. When moving a collection from APO A to APO Z:

- [ ] the collection druid should be removed as a default collection from the administrativeMetadata datastream of APO A  
- [ ] the collection druid should be added as a default collection to the administrativeMetadata datastream of APO Z 

The administrativeMetadata is what is used to populate the list of collections for each APO in the dropdown menu on the item registration page.

Currently, when moving a collection from APO A to APO Z, the administrativeMetadata is left unchanged for both APOs. This leads users to edit the datastreams manually to make these changes. If the changes happen automatically, there would be no need for these edits.

related to https://github.com/sul-dlss/argo/issues/2354

jcoyne:
## Why was this change made?

Stop managing a separate parallel list of collections in the administrative metadata for AdminPolicies.

Fixes #2380


andrewjbtw:
I am regrettably coming to the conclusion that we need to keep the manual collection setting process that we have. I've made a note to myself to come back to this to go through the reasoning.

andrewjbtw:
I'm finally closing this as not actionable at this point. I would love for this to be automatic but I did a review of registration workflows and there are cases where people register to collections that are not actually governed by the APO they are using. To support those cases, they need to be able to manually add collections to the registration list for that APO.

Web archives are the main user of this functionality, where a seed item might be registered to a collection that is not governed by the Seed APO, in which case the collection is manually added to the registration list for the Seed APO. But there are a couple of other cases where we still need to be able to manually manage the lists of collections that appear on the registration dropdown menus. The collections governed by the Hydrus Ur-APO are another case.

I would still like to simplify this but I think we need to review how SDR relates collections, items, and APOs. It has not been a strictly enforced hierarchy, so we can't assume that an item and the collection that contains that item share the same APO.

PR #2721, opened 2021-04-28
Restores tests to set_catkeys_and_barcodes_job_spec.

justinlittman:
## Why was this change made?
Tests were removed in https://github.com/sul-dlss/argo/pull/2693/files#diff-9a7dc28d99cfc95d006d7f19b549215a08e149e9a9a050c1f05f4b6bd1019896L89-L91

Unit

NA

### DIFF
--- a/spec/jobs/set_catkeys_and_barcodes_job_spec.rb
+++ b/spec/jobs/set_catkeys_and_barcodes_job_spec.rb
@@ -63,6 +63,10 @@ RSpec.describe SetCatkeysAndBarcodesJob do
   let(:object_client2) { instance_double(Dor::Services::Client::Object, find: item2) }
   let(:object_client3) { instance_double(Dor::Services::Client::Object, find: item3) }
 
+  let(:change_set1) { instance_double(ItemChangeSet, validate: true, model: item1, changed?: true) }
+  let(:change_set2) { instance_double(ItemChangeSet, validate: true, model: item2, changed?: false) }
+  let(:change_set3) { instance_double(ItemChangeSet, validate: true, model: item3, changed?: true) }
+
   before do
     allow(subject).to receive(:bulk_action).and_return(bulk_action_no_process_callback)
     allow(Dor::Services::Client).to receive(:object).with(pids[0]).and_return(object_client1)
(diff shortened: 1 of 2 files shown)


spec/jobs/set_catkeys_and_barcodes_job_spec.rb
```diff
@@ -75,6 +79,10 @@
     end
 
     context 'when catkey and barcode selected' do
+      before do
+        allow(ItemChangeSet).to receive(:new).and_return(change_set1, change_set2, change_set3)
```
jmartin-sul, line 83: neat, didn't know about this construct for multiple returns

jcoyne (approved):

spec/jobs/set_catkeys_and_barcodes_job_spec.rb
```diff
@@ -87,11 +95,14 @@
             }
           }
         expect(subject).to receive(:with_bulk_action_log).and_yield(buffer)
-        expect(subject).to receive(:update_catkey_and_barcode).with(ItemChangeSet, buffer)
-        expect(subject).to receive(:update_catkey_and_barcode).with(ItemChangeSet, buffer)
-        expect(subject).to receive(:update_catkey_and_barcode).with(ItemChangeSet, buffer)
+        expect(subject).to receive(:update_catkey_and_barcode).with(change_set1, buffer)
+        expect(subject).not_to receive(:update_catkey_and_barcode).with(change_set2, buffer)
```
jmartin-sul, line 99: so what's being tested here is how `change_setN#changed?` is acted upon, right?  as opposed to, e.g., whether the `ItemChangeSet` correctly determines if there's a change or not between `item1`'s barcode value and `barcodes[0]` differ?

jmartin-sul (commented):
one small question for my own edification, since this is already merged and i'd have approved anyway

spec/jobs/set_catkeys_and_barcodes_job_spec.rb
justinlittman: It is primarily to test that the correct change set was constructed as that will determine how the cocina model is updated.

PR #2722, opened 2021-04-28
Create a seed task

jcoyne:
## Why was this change made?

Makes setup easier.


jmartin-sul (approved):

jmartin-sul:
seems like build should pass after rebase on `main` now that https://github.com/sul-dlss/argo/pull/2723 is merged

jmartin-sul:
tested branch locally, works for me

Issue #2009, opened 2020-05-12
Provide interface for creating (not just updating) embargos in Argo - and for setting access upon embargo release


andrewjbtw:
Users (particularly University Archives) need to be able to set embargos for items. Currently, there is no user-facing UI for this in Argo. Instead embargos need to be set via adding an embargo datastream administratively. As far as I know, this means directly interacting with Fedora.

There are two components to being able to set embargos in Argo:

1. Adding the ability to create an embargo from an Argo item page.
2. Adding the ability to add/edit embargos in bulk.

Users are currently able to set embargos from the Hydrus and ETD apps. If an item already has an embargo set, Argo provides a button on the left side of the item page that allows you to manage it. If an embargo is not set, the button does not appear.

And remember:  there's no "embargo" without "argo"

related to (could be concurrent with?) Argo #2674 - providing all allowed access levels upon embargo release.

ndushay:
In addition to any unit and feature tests in Argo, this should get an infrastructure-integration-test:

- create an embargo, then check to see that it is correctly in index, in the (Fedora) rights.xml and embargo.xml, and in the cocina access.xml (?), and displays correct date on Argo item page, and is correct in the purl xml.
- update the embargo to a diff date and check that all of the above changed
- ?? remove an embargo ??  and check that all of the above changed.

Some of the model tests exist in the integration-tests:
- hydrus embargo test
- h2 embargo test

jcoyne:
@ndushay while I understand the need to ensure the pieces of the system work together, I'm unsure why this feature would required a new integration test. We already have 4 separate tests that show that dor-services-app sends embargo data to purl.  We can within the bounds of the Argo test suite show that this sends the correct data to dor-services-app.  This feature change does not touch any other integration beyond dor-services-app.   Integration tests are very slow and we must be judicious over which tests we add there.

andrewjbtw:
For the single object interface, we can make use of the space on the blue button list that's currently used by the "Update embargo" button:

<img width="242" alt="Screen Shot 2021-04-14 at 8 50 44 AM" src="https://user-images.githubusercontent.com/8975583/114740413-ad6a8e80-9cfe-11eb-823d-246b96a5f99f.png">

That button currently only appears when an embargo is active. Changes to make for the button:

- show it all the time
- if the there's no embargo, label it "Create embargo"
- if there is an embargo, label it "Manage embargo"
- if it can only have one name all the time, label it "Manage embargo"

[...]

PR #2725, opened 2021-04-28
Extract embargo controller from ItemsController

jcoyne:
## Why was this change made?
This refactor will enable us to add new embargoes in #2009


aaron-collier:
Closing as the work was also in https://github.com/sul-dlss/argo/pull/2728

PR #2728, opened 2021-04-29
Allow new embargoes to be created

### LINKED ISSUE #2009: Provide interface for creating (not just updating) embargos in Argo - and for setting access upon embargo release

jcoyne:
## Why was this change made?

Fixes #2009


app/controllers/embargos_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  def new
+    @change_set = change_set_class.new(@cocina)
+
+    respond_to do |format|
+      format.html { render layout: !request.xhr? }
+    end
+  end
+
+  def edit
+    @change_set = change_set_class.new(@cocina)
```
aaron-collier, line 15: CodeClimate is reporting that this method doesn't have coverage.

aaron-collier:
LGTM pending consideration of the one area that isn't showing coverage, which may help the reported tiny drop in coverage.

aaron-collier (approved):

PR #2732, opened 2021-04-29
Disable save button after submitting edit tags form

### LINKED ISSUE #2729: Screen does not update when updating object tags
Reported by @amyehodge :

```
when attempting to update a tag on an item. I clicked save and it didn’t look like it did anything, but when I closed the window and refreshed the page, the tag showed as updated. 
```

mjgiarlo:
Fixes #2729

## Why was this change made?

This provides users immediate feedback that clicking the button worked as expected and suggest to them that the page transition should happen momentarily. It also prevents double-submits. Note that I attempted to add test coverage to the edit_tags feature spec, but I cannot inspect the button element after clicking the button because the selenium driver says the reference is stale. Does not seem worth the time to futz with Capybara to test such a tiny change.

Peek 2021-04-29 11-40

CI

None

### DIFF
--- a/app/javascript/controllers/nested_form_controller.js
+++ b/app/javascript/controllers/nested_form_controller.js
@@ -17,4 +17,11 @@ export default class extends Controller {
     item.querySelector("input[name*='_destroy']").value = 1
     item.style.display = 'none'
   }
+
+  disableSubmit(event) {
+    const submitButton = event.currentTarget
+    submitButton.disabled = true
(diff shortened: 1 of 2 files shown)


ndushay:
marked HOLD as it shouldn't be merged to main until after @aaron-collier does the deploy of main ...  I think.

ndushay (approved):
very tidy!

ndushay:
(approved for merging by Andrew and Aaron in slack)

PR #2733, opened 2021-04-29
disable editing of source ID button for non items

### LINKED ISSUE #2727: Unable to change source id on collection
It is possible to create a collection without a source id.  Then if you click "Change source id" it results in an error:

Screen Shot 2021-04-28 at 10 13 49 PM
https://app.honeybadger.io/projects/49894/faults/78981903

peetucket:
## Why was this change made?

Fixes #2727 - non-items do not typically have sourceIDs (hydrus being one exception), and do not need to have the source ID editing enabled since it may cause odd behavior and there is no real use case for it

Added a new unit test


app/components/sidebar_controls_component.rb
```diff
@@ -70,7 +70,8 @@ def edit_collections
 
   def change_source_id
     render ActionButton.new(url: source_id_ui_item_path(id: pid),
-                            label: 'Change source id')
+                            label: 'Change source id',
+                            disabled: !item?)
```
jcoyne, line 74: Do we want to show this button at all on non-item objects?

jmartin-sul (approved):

app/components/sidebar_controls_component.rb
peetucket: So return from the method if its not an item, instead of disabling? 

Seems ok to me, though I'm not sure that's been our pattern in the past (where we instead turn buttons off if you can't do them).  What do you think @andrewjbtw ?

jcoyne: There are many buttons that just don't show if this is not the right type of object. I don't think there is consistency in the way this has been done.  I think the disabled is confusing to people because they don't always know why it's disabled.

peetucket: ok, will defer to andrew's preference

andrewjbtw: I agree with Justin. The buttons are already different for each item type and just leaving it out for collection seems fine.

I documented the existing blue button behavior before the work cycle as preparation for going to new page designs. It's definitely not consistent. We didn't really think about redesigning the sidebar as it already is since we were hoping to leave the buttons behind completely: https://consul.stanford.edu/pages/viewpage.action?pageId=163481988

peetucket: ok, updated the PR and tests to hide the button instead of disabling it

spec/components/sidebar_controls_component_spec.rb
```diff
@@ -56,6 +57,11 @@
         expect(rendered.css('a').size).to eq 15
       end
 
+      it 'does not include the change source id button for a collection (or any non-item)' do
+        allow(doc).to receive(:item?).and_return(false)
+        expect(rendered.css("a[href='/items/druid:zt570qh4444/source_id_ui']").size).to eq 0
+      end
+
```
peetucket, line 64: this new test is now basically the same as one below for APOs, so maybe we should ditch it

spec/components/sidebar_controls_component_spec.rb
```diff
@@ -85,6 +91,7 @@
       end
 
       it 'renders the appropriate default buttons for an apo' do
+        allow(doc).to receive(:item?).and_return(false)
```
jcoyne, line 94: If this is an APO, why does it need this method to be stubbed?

spec/components/sidebar_controls_component_spec.rb
```diff
@@ -94,10 +101,10 @@
         expect(rendered.css("a[href='/items/druid:zt570qh4444/workflows/new']").inner_text).to eq 'Add workflow'
         expect(rendered.css("a[href='/dor/republish/druid:zt570qh4444']").inner_text).to eq 'Republish'
         expect(rendered.css("a.disabled[data-confirm][data-method='delete'][href='/items/druid:zt570qh4444/purge']").inner_text).to eq 'Purge'
-        expect(rendered.css("a[href='/items/druid:zt570qh4444/source_id_ui']").inner_text).to eq 'Change source id'
+        expect(rendered.css("a[href='/items/druid:zt570qh4444/source_id_ui']").size).to eq 0
```
jcoyne, line 104: I think you can just remove this line, because we test that there are only 11 buttons and we test for all of them.

spec/components/sidebar_controls_component_spec.rb
```diff
@@ -33,6 +33,7 @@
       let(:catkey) { nil }
 
       it 'creates a hash with the needed button info for an admin' do
+        allow(doc).to receive(:item?).and_return(true)
```
jcoyne, line 36: this shouldn't be necessary as it already returns true due to https://github.com/sul-dlss/argo/pull/2733/files#L29

spec/components/sidebar_controls_component_spec.rb
jcoyne: I would agree

app/components/sidebar_controls_component.rb
peetucket: ahh, missed the solr document setup means i don't need to stub the item method...made all of the changes requested

PR #2734, opened 2021-04-29
Add download attribute to download links

### LINKED ISSUE #2731: Downloading reports from the bulk actions list changes the URL in the browser so that a refresh of the page redownloads the report
**Describe the bug**
Each user has access to a page that lists all bulk actions they previously ran. Certain bulk action produce reports that can be downloaded from this page. After downloading a report, the URL in the browser address bar is updated to show the download link. This results in the report being re-downloaded when the user refreshes the page.

**User Impact**
Causes some confusion. Users will refresh this page to check on the progress of running jobs, so a "clean" refresh is a smoother experience.

**To Reproduce**
1. Go to https://argo.stanford.edu/bulk_actions
2. Run a Download Descriptive Metadata job
3. Wait for it to finish
4. See the download link appear on the list
Screen Shot 2021-04-29 at 10 28 57 AM

5. Download the metadata 

6. See that the browser URL now points to the download link
Screen Shot 2021-04-29 at 10 33 53 AM
7. Reload the page and see the file reload again

**Expected behavior**
Downloading shouldn't change the URL.

jcoyne:
See https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/download

## Why was this change made?
Fixes #2731

### DIFF
--- a/app/views/bulk_actions/index.html.erb
+++ b/app/views/bulk_actions/index.html.erb
@@ -34,7 +34,7 @@
             <%= bulk_action.druid_count_success %> /
             <%= bulk_action.druid_count_fail %>
           </td>
-          <td><%= link_to('Log', file_bulk_action_path(bulk_action.id, filename: Settings.bulk_metadata.log)) %></td>
+          <td><%= link_to('Log', file_bulk_action_path(bulk_action.id, filename: Settings.bulk_metadata.log), download: true) %></td>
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

peetucket:
I think there are some more links on the same page that needs this (to the reports for some bulk actions): https://github.com/sul-dlss/argo/pull/2736

PR #2753, opened 2021-05-25
Default APO form to the correct current license

jcoyne:
## Why was this change made?
Previously it was trying to do a lookup on the license code, but the value it was sending in was the URI, so it was never matching.  This also simplifies the licenses table

### DIFF
--- a/app/forms/apo_form.rb
+++ b/app/forms/apo_form.rb
@@ -92,8 +92,8 @@ class ApoForm
   # entries, unless the current value is a deprecated entry, in which case, include that entry with the
   # deprecation warning in a parenthetical.
   def options_for_use_license_type(current_value)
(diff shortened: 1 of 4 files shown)


ndushay:
Would you please add a before and after picture of the license form?

ndushay:
also wondering if @andrewjbtw will want to sign off.

And ... what is with the code-climate diff coverage?

ndushay:
aha - just noticed `spec/forms/bulk_action_form_spec.rb` is checking that the pulldown is correctly populated - yay!

jcoyne:
@ndushay there is no change in the form, it only has the correct "selected" value.

jcoyne:
Codeclimate is off it's rocker.

ndushay (approved):

PR #2760, opened 2021-06-07
[HOLD] Update cocina to 0.60.0

jcoyne:


jcoyne:
This is already done.

PR #2769, opened 2021-06-28
Stop using the deprecated release_tags api

### LINKED ISSUE #2740: Decouple from the releaseTags API
Use the cocina API instead.

Depends on https://github.com/sul-dlss/dor-services-app/issues/2834 (for writing)

jcoyne:
## Why was this change made?

This is now available via the cocina api.

Fixes #2740

### DIFF
--- a/app/jobs/release_object_job.rb
+++ b/app/jobs/release_object_job.rb
@@ -53,18 +53,42 @@ class ReleaseObjectJob < GenericJob
   end
 
   def create_release_tag(bulk_action, cocina, log)
-    Dor::Services::Client.object(cocina.externalIdentifier).release_tags.create(
-      to: manage_release['to'],
-      who: manage_release['who'],
-      what: manage_release['what'],
-      release: string_to_boolean(manage_release['tag'])
-    )
+    object_client = Dor::Services::Client.object(cocina.externalIdentifier)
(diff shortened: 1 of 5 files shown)


app/jobs/release_object_job.rb
```diff
@@ -67,6 +63,34 @@ def create_release_tag(bulk_action, cocina, log)
     false
   end
 
+  # Return a copy of the existing cocina model with a new tag appended
+  def model_with_new_release_tag(cocina)
```
justinlittman, line 67: Should this be using a change set?

jcoyne: I don't think it needs to, but we could make a new change set object just for this sort of change. I think the indirection between having another class might make it slightly harder to follow than this simpler approach.

justinlittman: I don't want to delay this ticket, so happy to have this discussion sync. Couldn't it just be added to an existing change set class?

jcoyne: I'm not keen to have one set of functionality in a change set class that is used by one consumer and then a disjoint set of functionality in the same class used by another consumer.

justinlittman (approved):

PR #2774, opened 2021-06-29
Updated webpacker.

justinlittman:
## Why was this change made?
Latest is greatest.

QA

Gemfile, package.json


jcoyne:
@justinlittman is this ready to go?

justinlittman:
No. I don't think I ever got this to work.

PR #2776, opened 2021-07-06
Allows updating APO for collection.

### LINKED ISSUE #2764: Set governing APO is hitting errors
**Describe the bug**
Attempts to change the APO of objects in Argo are hitting errors:
https://app.honeybadger.io/projects/49894/faults/79350352
```
NoMethodError: undefined method `barcode' for #<#<Class:0x0000000008810968>:0x00007fb6b0c4f040>
```

This is happening for both items and collections. And it happens both from the single-item blue button and the bulk action

**User Impact**
It's not possible to change the APO for objects without editing the RELS-EXT.

**To Reproduce**
1. Go to an item, like this test item: https://argo.stanford.edu/view/druid:mc733pr7221
2. Click on the "set governing APO" blue button
3. Choose a new APO like "SDR Graveyard"
4. See nothing happen in Argo but HB gets an error

For bulk actions, go to New Bulk Action --> Update governing APO and try the same change. The same error will be sent to the bulk action log.

**Expected behavior**
The APO for the object(s) gets updated.

**Additional context**
I don't remember seeing this error in the Argo/Fedora workcycle but it might have been something we didn't catch in testing. The earliest instances of the HB error are from mid-May but no one seems to have reported this back then.

justinlittman:
closes #2764

## Why was this change made?
Allow a user to update APO for collection. Existing code was treating a Cocina collection like a DRO.

Unit

NA

### DIFF
--- a/app/services/collection_change_set_persister.rb
+++ b/app/services/collection_change_set_persister.rb
@@ -18,6 +18,7 @@ class CollectionChangeSetPersister
     updated = model
     updated = update_identification(updated) if changed?(:source_id) || changed?(:catkey) || changed?(:barcode)
     updated = updated_access(updated) if access_changed?
+    updated = updated_administrative(updated) if changed?(:admin_policy_id)
     object_client.update(params: updated)
   end
 
@@ -25,7 +26,7 @@ class CollectionChangeSetPersister
 
   attr_reader :model, :change_set
 
-  delegate :license, :copyright_statement, :use_statement, :catkey, :changed?, to: :change_set
+  delegate :admin_policy_id, :license, :copyright_statement, :use_statement, :catkey, :changed?, to: :change_set
 
   def access_changed?
     changed?(:copyright_statement) || changed?(:license) || changed?(:use_statement)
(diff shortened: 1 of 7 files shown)


app/services/collection_change_set_persister.rb
```diff
@@ -18,14 +18,15 @@ def update
     updated = model
     updated = update_identification(updated) if changed?(:source_id) || changed?(:catkey) || changed?(:barcode)
     updated = updated_access(updated) if access_changed?
+    updated = updated_administrative(updated) if changed?(:admin_policy_id)
```
jcoyne, line 21: This to me seems like a disjoint change from the other members of this change set.  I would think this should be a separate change set.

justinlittman: Say more. This follows the same pattern as the ItemChangeSet (which already updates the APO for a DRO).

jcoyne: There are two ways we can take this design, one is that a change set is a proxy object and capable of setting all of the properties.  The other way is that a change set represents an operation on an object.  If the latter, then we'd have separate types of change sets for the different operations the app provides (e.g. change the APO of a Collection)

jcoyne: I asked the Valkyrie developers (where this pattern originated) and their rule of thumb is 1 change set per form.

justinlittman: What does that buy us except a proliferation of change set classes and boilerplate code? Do we create parallel change set persisters as well?

jcoyne (approved):

PR #2777, opened 2021-07-06
Update cocina-models to 0.61.0

jcoyne:
This is blocked by #2769


jcoyne:
This won't pass until https://github.com/sul-dlss/dor-services-app/pull/2900 is merged which will trigger the container to be updated.

justinlittman (approved):

PR #2787, opened 2021-07-22
Update cocina-models to 0.61.3

jcoyne:
## Why was this change made?
To support the doi test prefix


peetucket (approved):
Pending tests passing

jmartin-sul (approved):

PR #2791, opened 2021-08-03
skip and log invalid druids in modsulator

### LINKED ISSUE #2789: Modsulator job should not retry when encountering incorrectly formatted druids in the input
**Describe the bug**
Bulk metadata uploads in Argo accept input files as XLSX or CSV. One column in this input file asks for the druid, formatted without the "druid:" prefix. ([Modsulator field-level instructions](https://consul.stanford.edu/display/DIGMETADATA/Replayable+spreadsheet+field-level+guidelines))

When the value in this column is incorrectly formatted (includes the "druid:" prefix, or is not a druid at all), the Modsulator job hits an error because no druid can be found for that row: https://app.honeybadger.io/projects/49894/faults/80627999 

This error triggers sidekiq's automatic retries and the job keeps retrying repeatedly even though it will never succeed since there's no way to correct the data at this point in the process.

Screen Shot 2021-07-27 at 12 29 00 PM

In the Argo interface, the job appears to be in progress and the error is not reported back to the user.

Screen Shot 2021-07-27 at 12 28 30 PM

Instead of retrying repeatedly, the job should either:

- Skip each row that has a bad value and continue to the end of the input file. Log the rows that were in error.

or

- Fail the job and report the issue back to the user. Ideally, log the rows that had the errors.

I think the first option was what used to happen before there were automatic retries, but I'm not 100% sure.

**User Impact**
Modsulator jobs appear to run indefinitely without an explanation. If users can see the error, they can correct their spreadsheet/CSV. People are left guessing what the problem is.

**To Reproduce**
1. Create a replayable spreadsheet where at least one row does not have a valid value for the druid
2. Submit this via Bulk MODS upload
3. See the error in HB and Sidekiq
4. See the job never complete in Argo

**Expected behavior**
The Modsulator job should process valid rows and report an error for invalid ones.

peetucket:
## Why was this change made?

Fixes #2789 -- when modsulator finds an invalid druid, it should skip and log it rather than try to send to dor-services-app, where it will fail and then be retried endlessly

Updated the fixture data to include a bogus druid and verified that the tests continue to run and that the log file has the expected output.


Gemfile
```diff
@@ -68,6 +68,7 @@ gem 'blacklight', '~> 7.18'
 gem 'blacklight-hierarchy', '~> 5.1'
 gem 'dor-services-client', '~> 7.0'
 gem 'dor-workflow-client', '~> 3.19'
+gem 'druid-tools'
```
peetucket, line 71: note that this gem is already included in argo because it is a dependency of the preservation-client gem, but this ensures we will have it even if that dependency goes away at some point

justinlittman (approved):

PR #2799, opened 2021-08-11
Set default what when releasing collection.

### LINKED ISSUE #2798: Dry::Struct::Error in ReleaseObjectJob # perform
See https://app.honeybadger.io/projects/49894/faults/80458990

justinlittman:
closes #2798

## Why was this change made?
Prevent user error.

### DIFF
--- a/app/components/manage_release/collection_form_component.html.erb
+++ b/app/components/manage_release/collection_form_component.html.erb
@@ -3,7 +3,7 @@
     <label>What should be released/withdrawn?</label>
     <div class='radio'>
       <label class='radio-inline'>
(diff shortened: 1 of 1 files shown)


jcoyne (approved):
@andrewjbtw  is "Collection" the right default for this radio, or should it be "self"?

andrewjbtw:
Probably? It depends on what the releaser wants to do. I suspect we have more collections released to "collection" than to "self" but it's a costlier mistake to accidentally release "collection" when you didn't want to.

jcoyne:
@andrewjbtw I think having something selected that prevents system errors is better than what we had before.  If this is the wrong choice, then it's trivial to change it to have "self" be the default.

PR #2804, opened 2021-08-24
no exception if an object has no collections

### LINKED ISSUE #2800: NoMethodError: undefined method `first' for nil:NilClass
See https://app.honeybadger.io/projects/49894/faults/79467245

peetucket:
## Why was this change made?

Fixes #2800 - though to be honest it's unclear how you can hit this point, because it's unclear why you'd get to the "remove a collection" endpoint for an object without any collections.  See the details on the ticket.

Added a new test

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -78,7 +78,7 @@ class ItemsController < ApplicationController
   end
 
   def remove_collection
-    new_collections = @cocina.structural.isMemberOf - [params[:collection]]
+    new_collections = Array(@cocina.structural.isMemberOf) - [params[:collection]]
     change_set = ItemChangeSet.new(@cocina)
     change_set.validate(collection_ids: new_collections)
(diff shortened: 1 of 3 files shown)


justinlittman (approved):

andrewjbtw:
Could the modal be hitting the remove collections endpoint even without a dropdown menu? The "remove" text is there even when there isn't a collection
Screen Shot 2021-08-24 at 9 16 08 AM

I've seen the HB error for this before and found it puzzling because as far as I've been able to tell no functionality has been affected.

peetucket:
This exception could also possibly be triggered from the "Set Collection" bulk action, which replaces any existing collection membership with the one chosen.  Perhaps it is hitting the remove collection endpoint for any of the objects chosen in the bulk action, and some of these have no collections to begin with.  The HB exception won't cause any weird side effects, but this PR would just turn it into a no-op with no exception.

PR #2814, opened 2021-09-16
Show the user a helpful error if they use an invalid tag

### LINKED ISSUE #2813: Show user a message when they enter a tag that does not use the valid tag format
Argo should show a message to users when they attempt to submit a tag that does not use the valid pattern for tags. Currently, the system reports to Honeybadger when an invalid tag is submitted but the user does not see this message. Instead, they get sent to the library system status page.

Tags in Argo must include the pattern:

`[term] : [term]`

It's legal to have more than one colon in a hierarchy, but at least one colon is required.

jcoyne:
## Why was this change made?
Fixes #2813

Deployed to qa for Andrew to test

Screen Shot 2021-09-16 at 2 40 36 PM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,7 @@ gem 'rest-client'
 gem 'retries'
 gem 'ruby-prof'
 gem 'rubyzip'
-gem 'turbo-rails', '~> 0.5.9'
+gem 'turbo-rails', '~> 0.7.13'
 gem 'zip_tricks', '5.3.1' # 5.3.1 is required as 5.4+ breaks the download all feature
 
 # Stanford related gems
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -540,7 +540,7 @@ GEM
     thor (1.1.0)
     trailblazer-option (0.1.1)
     ttfunk (1.4.0)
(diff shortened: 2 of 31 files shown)


Gemfile
```diff
@@ -60,7 +60,7 @@ gem 'rest-client'
 gem 'retries'
 gem 'ruby-prof'
 gem 'rubyzip'
-gem 'turbo-rails', '~> 0.5.9'
+gem 'turbo-rails', '~> 0.7.13'
```
ndushay, line 63: is this upgrade necessary for this change?   31 files changed is a lot for a new feature.

ndushay (approved):
I have no objections to this save:
- my comment about whether turbo-rails upgrade should be bundled into this PR
- please run the integration tests for argo before merging.

Gemfile
jcoyne: I made this change when I encountered this bug https://github.com/hotwired/turbo/issues/396.  While this didn't fix the bug I was looking at (I worked around it), this is how Andrew tested it, so to be respectful of his time, I'd rather not separate this change and ask him to test again.

PR #2818, opened 2021-09-20
Use the EditModalComponent for new collections

jcoyne:
## Why was this change made?
Without that component, the code breaks with the error:
```
Response has no matching <turbo-frame id="modal-frame"> element
```

### DIFF
--- a/app/views/collections/new.html.erb
+++ b/app/views/collections/new.html.erb
@@ -1,67 +1,67 @@
-<div class='modal-header'>
-  <h3 class='modal-title'>Create Collection</h3>
-  <button type='button' class='close' data-dismiss='modal' aria-hidden='true'>×</button>
-</div>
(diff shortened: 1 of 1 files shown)


app/views/collections/new.html.erb
```diff
@@ -1,67 +1,67 @@
-<div class='modal-header'>
-  <h3 class='modal-title'>Create Collection</h3>
-  <button type='button' class='close' data-dismiss='modal' aria-hidden='true'>×</button>
-</div>
+<%= render EditModalComponent.new do |component| %>
+  <% component.header { 'Create Collection' } %>
+  <% component.body do %>
+    <div data-controller="collection-editor">
```
jcoyne, line 4: This stuff just changed indent level. No code changes happen inside this div.

peetucket (approved):

ndushay (approved):

PR #2827, opened 2021-10-12
Remove jqueryui autocomplete

jcoyne:
## Why was this change made?

Jquery-ui is obsolete and challenging to work with because it overlaps in responsibility with bootstrap.


jcoyne:
@ndushay I will have @andrewjbtw test, but I think he would prefer we get all the related changes (#2827, #2828, #2829, and more to come) done before then.

PR #2836, opened 2021-10-19
Update preservation_client

jcoyne:


jmartin-sul:
argo's `preservation-client` dependency is now on 4.0.0, so i think this is no longer needed (reopen if that's a mistake, of course)

PR #2837, opened 2021-10-20
Start using the new show layout

jcoyne:
## Why was this change made?
To match designs in https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/444556663/comments

Before

Screen Shot 2021-10-20 at 8 26 51 AM

After

Screen Shot 2021-10-20 at 08 42 58-fullpage


app/helpers/argo_helper.rb
```diff
@@ -9,7 +9,7 @@ module ArgoHelper
   def container_classes
     return super if controller_name == 'apo'
 
-    'container-fluid'
+    action_name == 'show' ? 'container' : 'container-fluid'
```
jcoyne, line 12: This is run on every page, so I don't know why code climate sees no coverage here.

justinlittman (approved):

PR #2841, opened 2021-10-20
Overview component

jcoyne:
## Why was this change made?

Show the overview table on the show page.


justinlittman (approved):
Keep the empty test file?

PR #2842, opened 2021-10-20
Details component

jcoyne:
## Why was this change made?

To make the show page display the details as defined in the mockups.

Screen Shot 2021-10-20 at 16 53 16-fullpage

### DIFF
--- a/app/components/details_component.html.erb
+++ b/app/components/details_component.html.erb
@@ -1,0 +1,74 @@
+<table class="table table-sm mb-5 caption-header">
+    <caption>Details</caption>
+    <tbody>
+      <tr>
+        <th class="col-3" scope="row">Object type</th>
+        <td>
+            <%= object_type %>
+        </td>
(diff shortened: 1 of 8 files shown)


andrewjbtw:
Looking at this on my screen in the staging environment, I think moving the buttons to the top like that takes up too much screen for a first glance. (Also, it's a lot of red to see at once.) If we're going to have a transitional period between the buttons on the sidebar and removing the blue buttons altogether, we should spend some time on a transitional design.

I think the grid display is a huge improvement over the current display. Ideally, it would be closer to the top even if those fields are not yet editable in-line.

jcoyne:
@andrewjbtw a good number of those buttons will convert to pencil icons (e.g. "Change source id", "Set content type"), so we will be able to have a lot less red.

jcoyne:
Yep, we're going to be adding more functionality to this component that we want to test.

PR #2843, opened 2021-10-21
argo modals should not re-appear after redirect

### LINKED ISSUE #2832: Most recently edited modal pops up after clicking Republish button and can't be closed
**Describe the bug**
Under some circumstances, clicking on the "Republish" blue button on an Argo show page causes the most recently used modal to pop back up and it cannot be closed without reloading the page.

I made a video of the behavior. If you click on "Republish" without popping up any modals, everything works as expected and you just get a banner message about the republish action. But if you edit a datastream or use another modal, then the modal pops back up after you click on "Republish".

https://user-images.githubusercontent.com/8975583/137553005-08be1e4f-af7d-4d47-9922-a21ab7573460.mov

**User Impact**
This has been reported by a few users who have had to reload the page or open the druid in a new window after clicking on Republish and then not being able to close the modal.

**To Reproduce**
1. Go to an Argo show page
2. Click on a blue button that pops up a modal
3. Enter some data and submit it
4. Click republish button
5. See modal pop up again

**Expected behavior**
The Republish button should trigger a republish action in the background and reload the Argo page with a message saying the republish was successful. I believe the republish is still happening in this situation and the issue is just the modal popping up.

**Additional context**
As far as I can tell, if you open a modal but don't submit a change, this pop-up behavior will not happen. But I haven't checked that extensively. I have tested both datastream edits and other edits (such as the "open version" modal) and the modal pops up in both cases.

peetucket:
## Why was this change made?

Fixes #2832 -- argo modals re-appear after a re-direct and should not.  After playing around in localhost, it appears to me that the browser seems to be caching the version of the catalog show page that has the modal in it, and then when you re-direct back to the page after performing a reindex or republish, the browser is re-opening the modal.  Took a dumb attempt at just making the redirect URL change after a reindex or republish by adding a param and the problem went away in localhost.  

Is there a better way to do this?  There may be other redirects that cause this as well.

Tried doing this too at the top of the `show` method in the catalog controller, but that didn't have an effect even though I could see it appear in the headers with dev tools:

`response.headers["Cache-Control"] = "no-cache, no-store"`

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -28,4 +28,11 @@ document.addEventListener("turbo:load", async () => {
   // Start argo after free-jqgrid has been loaded
   new Argo().initialize()
 })
+
+ window.addEventListener('turbo:before-cache', function () {
+  // Close any lingering open modal windows
+  $('.modal').modal('hide');
+  $('.modal-backdrop').remove();
+});
+
 import '@hotwired/turbo-rails'


app/javascript/application.js
```diff
@@ -28,4 +28,19 @@ document.addEventListener("turbo:load", async () => {
   // Start argo after free-jqgrid has been loaded
   new Argo().initialize()
 })
+
+window.addEventListener('turbo:before-cache', function () {
+  // Close any lingering open modal windows
+  // $('.modal:visible').modal('hide');
+  // $('.modal-backdrop').remove();
+  if (document.body.classList.contains('modal-open')) {
+    $('.modal')
```
jcoyne, line 37: Please avoid using jquery.

jcoyne: Is there a reason you aren't using the bootstrap modal API?

mjgiarlo: At this point, I'm at the "adapt what was in the branch before by adding copypasta" phase. It's complicated by not being obvious how to reproduce this locally. I'll check the BS4 docs.

mjgiarlo: But I'm wondering if we can add more `data-dismiss="modal"` to modal buttons that submit instead?

mjgiarlo: @jcoyne I'll be pushing stuff to this branch here and there. feel free to disregard until it's undrafted. I don't think it'll be fruitful to code review what's here until then.

jcoyne: I would expect the only bit of code needed would be `$('.modal').modal('hide')`

peetucket: yeah, i think the overlay is the `$('.modal-backdrop').remove();`

mjgiarlo: I'm finding it hard to reconcile two things: "don't use JQuery" and "use the Bootstrap 4 API". Looking at the BS4 docs, it appears their examples all use JQuery. 🤷🏻‍♂️

PR #2851, opened 2021-11-02
Update to the latest version of turbo-rails

jcoyne:
## Why was this change made?
Keep up to date. This is the same version that h2 uses.


mjgiarlo (approved):

mjgiarlo:
@jcoyne You think this needs testing in a deployed env or nah?

jcoyne:
I think that our feature tests take care of that.

PR #2852, opened 2021-11-04
Prevent turbo from caching the catalog show page

### LINKED ISSUE #2832: Most recently edited modal pops up after clicking Republish button and can't be closed

mjgiarlo:
## Why was this change made?

Fixes #2832

This is the best I could do for #2832. 🤷🏻‍♂️ 

Note that this commit does not fix the missing republish flash notice. @jcoyne and I think this is unrelated and probably bears more work to hotwire-ify how Argo handles flash.

QA env

None.

### DIFF
--- a/app/views/catalog/show.html.erb
+++ b/app/views/catalog/show.html.erb
@@ -1,5 +1,6 @@
 <% @page_title = t('blacklight.search.show.title',
   document_title: truncate(strip_tags(document_presenter(@document).html_title).html_safe, length: 50), application_name: application_name).html_safe %>
+
 <% content_for(:head) { render_link_rel_alternates } %>
 
 <%= render_document_main_content_partial %>
--- a/app/views/datastreams/edit.html.erb
+++ b/app/views/datastreams/edit.html.erb
@@ -2,7 +2,7 @@
   <% component.header { params[:id] } %>
   <% component.body do %>
     <% if can?(:manage_item, @cocina) %>
-      <%= form_tag item_datastream_path(@cocina.externalIdentifier, params[:id]), id: 'xmlEditForm', method: :patch do %>
+      <%= form_tag item_datastream_path(@cocina.externalIdentifier, params[:id]), data: { turbo: false }, id: 'xmlEditForm', method: :patch do %>
           <input type="hidden" name="id" value="<%= @cocina.externalIdentifier %>">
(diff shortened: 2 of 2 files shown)


app/views/catalog/show.html.erb
```diff
@@ -1,6 +1,11 @@
 <% @page_title = t('blacklight.search.show.title',
   document_title: truncate(strip_tags(document_presenter(@document).html_title).html_safe, length: 50), application_name: application_name).html_safe %>
-<% content_for(:head) { render_link_rel_alternates } %>
+
+<% content_for(:head) do %>
+  <meta name="turbo-visit-control" content="reload">
```
jcoyne, line 5: This seems like a really big stick.  Is it possible to put `data-turbo="false"` on the affected forms instead?   I think this is more difficult to reason about as we don't have a clue as to where these ajax requests may be originating.

jcoyne (changes requested):

app/views/catalog/show.html.erb
mjgiarlo: changed.

jcoyne (approved):

PR #2881, opened 2021-12-06
Convert the spreadsheet jquery plugin to stimulus

jcoyne:
## Why was this change made?

This avoids using an event that will not be emitted in Blacklight 8


mjgiarlo (approved):
w00t. Merge when 🟢

PR #2882, opened 2021-12-07
use bootstrap classes rather than custom styles

jcoyne:


app/views/bulk_jobs/_bulk_index_table.html.erb
```diff
@@ -1,48 +1,46 @@
-<div id="bulk-upload-table">
-  <table class="table table-hover">
+<table class="table table-hover my-3">
```
jcoyne, line 1: a lot of this is just whitespace changes, you may prefer to review while hiding those changes.

mjgiarlo (approved):

PR #2902, opened 2021-12-14
Apply the proper bootstrap 5 classes for dropdowns and form spacing

### LINKED ISSUE #2901: (bootstrap 5) Some dropdown menus are not showing an icon to indicate they're menus
Some of the dropdown menus are not showing an indication that they are dropdowns. I'm not sure if there's an easy way to find them all, but I saw these on the blue button modals:

Set rights:
Screen Shot 2021-12-13 at 2 46 15 PM

Edit collections:
Screen Shot 2021-12-13 at 2 36 46 PM

I also noticed this on all three menus on the Register/Edit APO form:

Agreement dropdown:
Screen Shot 2021-12-13 at 5 51 21 PM

See also the "default object rights" and the "license" dropdown on the same form.

**Expected behavior**

Dropdowns in prod have had a little arrow on the right to show they are menus, such as:

Screen Shot 2021-12-13 at 5 52 29 PM

jcoyne:
## Why was this change made?
Fixes #2901

### DIFF
--- a/app/views/bulk_actions/forms/_set_license_and_rights_statements_form.html.erb
+++ b/app/views/bulk_actions/forms/_set_license_and_rights_statements_form.html.erb
@@ -1,34 +1,34 @@
 <% f.fields_for :set_license_and_rights_statements do |license_and_rights_statements| %>
-  <div class="form-group mb-0">
+  <div class="mb-3">
     <%= license_and_rights_statements.check_box :use_statement_option %>
-    <%= license_and_rights_statements.label :use_statement_option, 'Update use and reproduction statement.', class: 'ml-2' %>
+    <%= license_and_rights_statements.label :use_statement_option, 'Update use and reproduction statement.', class: 'ms-2' %>
 
     <%= link_to 'Sample use and reproduction statements.', 'https://consul.stanford.edu/display/APO/Sample+Access+Condition+statements', target: '_blank', rel: 'noopener' %>
   </div>
 
-  <div class="form-group ml-4 mb-5">
+  <div class="ms-4 mb-5">
     <%= license_and_rights_statements.text_area :use_statement, class: 'form-control', rows: 2, placeholder: 'Check the box and leave this area blank to remove existing use and reproduction statements' %>
   </div>
 
-  <div class="form-group mb-0">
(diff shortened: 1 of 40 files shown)


app/views/bulk_actions/forms/_set_license_and_rights_statements_form.html.erb
```diff
@@ -1,34 +1,34 @@
 <% f.fields_for :set_license_and_rights_statements do |license_and_rights_statements| %>
-  <div class="form-group mb-0">
+  <div class="mb-3 mb-0">
```
ndushay, line 2: does it make sense for it to have multiple mb settings?

ndushay (changes requested):
Basically looks good, but a couple of quibbles, and I think a problem spotted with multiple mb settings in one file.

Wow - this hits a lot of occurrences of "form-group" and "form-control" -- how confident are you that the changes you're using works in all these spots?  

form-group going to mb-3  (should it always be mb-3?)
form-control to form-select  (I understand this less, but it looks like form-control might be in bootstrap 5?  and I don't know if this mapping is correct for all the cases)

https://stackoverflow.com/questions/67655096/bootstrap-5-form-group-form-row-form-inline

https://stackoverflow.com/questions/68466581/bootstrap-5-position-floating-label-inside-form-control-with-width-auto

ndushay (approved):

ndushay:
approved but tests failing.  Merge when green.

jcoyne:
> Wow - this hits a lot of occurrences of "form-group" and "form-control" -- how confident are you that the changes you're using works in all these spots?

I'm completely confident that `form-group` does nothing in bootstrap 5.  The main purpose here was to make the dropdowns indicate they are dropdowns (the down cheveron), which is why the `form-control` was changed to `form-select`

> form-group going to mb-3 (should it always be mb-3?)

`.mb-3` provides the same spacing as form-group did before.

> form-control to form-select (I understand this less, but it looks like form-control might be in bootstrap 5? and I don't know if this mapping is correct for all the cases)

form-control doesn't work on `<select>` it needs to be `form-select` now.

PR #2904, opened 2021-12-14
Remove unnecessary override

jcoyne:
## Why was this change made?
This is currently in Blacklight and it is no longer necessary to patch it here


justinlittman (approved):
Approve pending green.

PR #2910, opened 2021-12-14
Migrate populating druids from jquery to stimulus

jcoyne:
## Why was this change made?
So that our javascript is layed out more consistently.


justinlittman (approved):
Approve pending green.

PR #2943, opened 2022-01-06
Update isMemberOf to array while parsing CSV

### LINKED ISSUE #2941: bulk registration from CSV fails when a collection is included
**Describe the bug**
The "Register druids" bulk action does not work if a collection is included in the CSV. Including a collection results in the following error:

RegisterDruidsJob: #/components/schemas/RequestDROStructural/properties/isMemberOf expected array, but received String: "druid:bk024qs1808"

If no collection is supplied, then registration works correctly.

**User Impact**
It is not possible to register a large batch of druids and include a collection. The user who reported this is trying to register around 10,000 druids.

**To Reproduce**
1. Go to Bulk actions --> New bulk action
2. Click on Register Druids
3. Upload a CSV. The CSV I tested with includes the following data
```
APO,Collection,Rights,Initial Workflow,Content Type,Project Name,Catkey,Source ID,Label,Tags,Tags,Tags
druid:sc012gz0974,druid:bk024qs1808,world,registrationWF,file,bulk registration,444,collection:array,collection array,csv : test,Project : two
```
4. See error in the bulk action log

**Expected behavior**
It should be possible to include a collection during bulk registration

**Additional context**
The registration form does appear to be allowing registration with a collection selected.

aaron-collier:
## Why was this change made?

Fixes #2941

### DIFF
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -50,7 +50,7 @@ class RegistrationCsvConverter
     }
 
     structural = {}
-    structural[:isMemberOf] = row['Collection'] if row['Collection']
+    structural[:isMemberOf] = [row['Collection']] if row['Collection']
     model_params[:structural] = structural
     model_params[:access] = access(row['Rights']) if row.fetch('Rights') != 'default'
     model_params[:administrative][:partOfProject] = row['Project Name'] if row['Project Name'].present?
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
Can you add a test that exercises this?

aaron-collier:
@jcoyne yes, looking at the tests now. I wanted to get the PR up as I don't seem to have a lot of success testing argo locally and wanted it to run the tests.

jcoyne (approved):

PR #2945, opened 2022-01-07
Replace registration job access selection with CocinaDroAccess

### LINKED ISSUE #2944: Bulk registration fails on "stanford no-download" and "cdl" rights statuses
**Describe the bug**
The register druids bulk action is not able to process the "stanford (no-download)" and "cdl" rights statuses. Attempting to use these statuses leads to this error:

```
2022-01-07 02:39:11 UTC RegisterDruidsJob: {"access"=>"stanford", "download"=>"none"} isn't one of in #/components/schemas/Access
2022-01-07 02:39:11 UTC RegisterDruidsJob: {"access"=>"cdl-stanford", "download"=>"none"} isn't one of in #/components/schemas/Access
```

Both of these rights statuses work when used from the registration form.

**User Impact**
It's not possible to bulk register items with these rights statuses. We have one project trying to register with "stanford (no-download) at the moment.

**To Reproduce**
1. Go to Bulk actions --> New bulk action
2. Choose "Register druids"
3. Upload a CSV with the following data:
```
APO,Collection,Rights,Initial Workflow,Content Type,Project Name,Source ID,Label,Tags,Tags
druid:sc012gz0974,,stanford-nd,registrationWF,map,bulk registration,bulk:abcd3800,stanford no-download,csv : test,
druid:sc012gz0974,,cdl-stanford-nd,registrationWF,document,bulk registration,bulk:abcd6151,cdl,csv : test,additional : tag
```
4. See errors in the logs

**Expected behavior**
All rights statuses on the registration form menu should also be supported in bulk registration.

**Additional context**
It's possible this hasn't worked since we updated Cocina to incorporate the CDL (digital borrowing) status in fall 2020. Our most common uses of bulk registration use other rights settings.

When we first implemented bulk registration, I tested all of the rights statuses that existed at the time (no CDL) and those worked, so I know stanford (no-download) has worked before (in August 2020). I don't routinely test bulk registration after changes elsewhere have been made so it's possible no one has checked this until now.

aaron-collier:
## Why was this change made?

Fixes #2944 

The bulk registration job had its own code for setting the access hash that did not match the registration form. This causes a validation failure by not including the required controlledDigitalLending related values when required.

A future refactor may rename/update the CocinaDroAccess method for clarity.

### DIFF
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -52,7 +52,7 @@ class RegistrationCsvConverter
     structural = {}
     structural[:isMemberOf] = [row['Collection']] if row['Collection']
     model_params[:structural] = structural
-    model_params[:access] = access(row['Rights']) if row.fetch('Rights') != 'default'
+    model_params[:access] = CocinaDroAccess.from_form_value(row['Rights']).value_or(nil)
     model_params[:administrative][:partOfProject] = row['Project Name'] if row['Project Name'].present?
 
     tags = []
@@ -84,28 +84,6 @@ class RegistrationCsvConverter
       Cocina::Models::Vocab.book
     else
       Cocina::Models::Vocab.object
-    end
-  end
-
-  # @param [String] the rights representation from the form
-  # @return [Hash<Symbol,String>] a hash representing the Access subschema of the Cocina model
-  def access(rights)
-    if rights.end_with?('-nd') || %w[dark citation-only].include?(rights)
-      {
-        access: rights.delete_suffix('-nd'),
-        download: 'none'
-      }
-    elsif rights.start_with?('loc:')
-      {
-        access: 'location-based',
-        readLocation: rights.delete_prefix('loc:'),
-        download: 'location-based'
-      }
-    else
-      {
-        access: rights,
-        download: rights
-      }
     end
   end
 end


app/services/registration_csv_converter.rb
```diff
@@ -52,7 +52,7 @@ def convert_row(row)
     structural = {}
     structural[:isMemberOf] = [row['Collection']] if row['Collection']
     model_params[:structural] = structural
-    model_params[:access] = access(row['Rights']) if row.fetch('Rights') != 'default'
+    model_params[:access] = CocinaDroAccess.from_form_value(row['Rights']).value! if row.fetch('Rights') != 'default'
```
jcoyne, line 55: Could you do this instead?
```suggestion
    model_params[:access] = CocinaDroAccess.from_form_value(row['Rights']).value_or(nil)
```

app/services/registration_csv_converter.rb
```diff
@@ -52,7 +52,7 @@ def convert_row(row)
     structural = {}
     structural[:isMemberOf] = [row['Collection']] if row['Collection']
     model_params[:structural] = structural
-    model_params[:access] = access(row['Rights']) if row.fetch('Rights') != 'default'
+    model_params[:access] = CocinaDroAccess.from_form_value(row['Rights']).value_or(nil) if row.fetch('Rights') != 'default'
```
jcoyne, line 55: I don't think `if row.fetch('Rights') != 'default'` is needed anymore, is it?

jcoyne (approved):

PR #2974, opened 2022-01-14
Expand and collapse cocina in place

### LINKED ISSUE #2618: add "View in new window" Cocina link to get full JSON for cocina
Since we now have collapsed Cocina by default in the record view, it can be a pain to get the full cocina.  If we add a link to this left navvy area:

image

that opens a new window with something as simple as the raw JSON, that could be a help to us, the developers and to Andrew and others for troubleshooting.

mjgiarlo:
Fixes #2618 (again)

## Why was this change made?

The previous impl allowed users to expand and collapse individual nodes, and as a user, I expect to see widgets for collapse-all and expand-all. So I added it. Note that I chatted with @andrewjbtw and he wants the "View Cocina in New Window" button to remain even with this in place as a means for easy copy/paste of large cocina structures.

Peek 2022-01-13 13-48

Locally. Clicked the buttons and watched the JS console.

None

### DIFF
--- a/app/views/catalog/_cocina_default.html.erb
+++ b/app/views/catalog/_cocina_default.html.erb
@@ -6,10 +6,15 @@
   </h2>
   <div id="document-cocina-section" class="accordion-collapse collapse" aria-labelledby="document-cocina-heading">
     <div class="accordion-body" data-controller="json-renderer">
+      <div>
+        <a class="btn btn-secondary " data-action="json-renderer#expand">Expand</a>
+        <a class="btn btn-secondary " data-action="json-renderer#collapse">Collapse</a>
+      </div>
 
       <script type="application/json" data-json-renderer-target="cocina">
         <%= @cocina.to_json.html_safe %>
       </script>
+
       <div class="detail" data-json-renderer-target="section"></div>
     </div>
   </div>
--- a/app/javascript/controllers/json_renderer.js
+++ b/app/javascript/controllers/json_renderer.js
@@ -5,7 +5,15 @@ export default class extends Controller {
   static targets = ['cocina', 'section']
 
   connect() {
-    const cocina = JSON.parse(this.cocinaTarget.innerText)
-    this.sectionTarget.appendChild(renderjson.set_show_to_level(1)(cocina))
+    this.cocina = JSON.parse(this.cocinaTarget.innerText)
+    this.collapse()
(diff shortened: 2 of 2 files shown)


app/views/catalog/_cocina_default.html.erb
```diff
@@ -6,10 +6,15 @@
   </h2>
   <div id="document-cocina-section" class="accordion-collapse collapse" aria-labelledby="document-cocina-heading">
     <div class="accordion-body" data-controller="json-renderer">
+      <div class="argo-show-btn-group">
```
jcoyne, line 9: Is this class necessary?

app/views/catalog/_cocina_default.html.erb
```diff
@@ -6,10 +6,15 @@
...
+        <a class="btn btn-primary " data-action="json-renderer#expand">Expand</a>
```
jcoyne, line 10: Could these be `btn-secondary` or `btn-outline-primary`?  I don't really think of collapse expand as primary functions.

app/views/catalog/_cocina_default.html.erb
mjgiarlo: @jcoyne Necessary? No. It made the buttons layout a little nicer to my eyes, though.

app/views/catalog/_cocina_default.html.erb
mjgiarlo: Sure. That's why I asked you about it yesterday. 😆 I'll tweak 'em.

mjgiarlo: Done!

app/views/catalog/_cocina_default.html.erb
jcoyne: If we don't have any styles or actions affiliated with this class, then I think it causes unnecessary cognitive load and it should be removed.

mjgiarlo: @jcoyne Should I remove the div entirely then or use an empty div?

jcoyne: I think an empty div is fine if you want to group those.

mjgiarlo: @jcoyne OK, done

PR #2988, opened 2022-01-18
Add a form for updating copyright

### LINKED ISSUE #2864: Edit "copyright" field inline
Once https://github.com/sul-dlss/argo/pull/2844 is merged, we want to edit the copyright statement inline. 

See https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/445123823
https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/444556663

jcoyne:
## Why was this change made?

Fixes #2864

Screen Shot 2022-01-18 at 4 22 16 PM

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -9,6 +9,7 @@ class ItemsController < ApplicationController
     purge_object
     source_id
     tags_bulk
+    update
   ]
 
   before_action :enforce_versioning, only: %i[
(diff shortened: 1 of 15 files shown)


app/controllers/items_controller.rb
```diff
@@ -223,6 +224,20 @@ def collection_ui
...
+  def edit_copyright
+    @change_set = ItemChangeSet.new(@cocina)
+  end
+
+  # save the copyright form
+  def update
+    change_set = ItemChangeSet.new(@cocina)
+    change_set.validate(copyright: params[:item][:copyright])
+    change_set.save
+    reindex
```
justinlittman, line 237: Should reindex be performed in change_set.save? Seems that others replicating this pattern might easily forget to include.

justinlittman: Is there going to be a race condition between the reindex and the redirect?

justinlittman: Why is the reindex triggered here instead of DSA?

justinlittman (approved):

PR #2992, opened 2022-01-19
Support setting copyright on collection

jcoyne:

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -208,7 +208,7 @@ class ItemsController < ApplicationController
 
     authorize! :manage_governing_apo, @cocina, params[:new_apo_id]
 
-    change_set = @cocina.is_a?(Cocina::Models::Collection) ? CollectionChangeSet.new(@cocina) : ItemChangeSet.new(@cocina)
+    change_set = build_change_set
     change_set.validate(admin_policy_id: params[:new_apo_id])
     change_set.save
     reindex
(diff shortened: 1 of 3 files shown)


app/controllers/items_controller.rb
```diff
@@ -231,7 +231,7 @@ def edit_copyright
 
   # save the copyright form
   def update
-    change_set = ItemChangeSet.new(@cocina)
+    change_set = @cocina.is_a?(Cocina::Models::Collection) ? CollectionChangeSet.new(@cocina) : ItemChangeSet.new(@cocina)
```
ndushay, line 234: random thought, not for this PR:  should we have `cocina-obj.collection?` `.dro?` and `.apo?` convenience methods provided in cocina-models?

spec/requests/item_update_spec.rb
```diff
@@ -18,36 +18,71 @@
...
+        expect(object_client).to have_received(:update)
+          .with(params: updated_model)
+        expect(Argo::Indexer).to have_received(:reindex_pid_remotely)
+        expect(response.code).to eq('303')
+      end
     end
 
-    it 'sets the new collection' do
-      patch "/items/#{pid}", params: { item: { copyright: 'in public domain' } }
+    context 'with an collection' do
```
ndushay, line 55: `a` not `an`

ndushay (approved):
looks good.  One quibble you can take or ignore and a random question.

app/controllers/items_controller.rb
jcoyne: I believe they already exist.

jcoyne: https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/checkable.rb

PR #2993, opened 2022-01-19
Don't show fields not applicable to an admin policy on the overview

jcoyne:


ndushay:
oh no - test failure!

PR #2996, opened 2022-01-19
Add a form to edit barcode inline

### LINKED ISSUE #2865: Edit "barcode" field inline
Once https://github.com/sul-dlss/argo/pull/2844 is merged, we want to edit the barcode inline. 

See https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/445123823
https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/444556663

aaron-collier:
## Why was this change made?

Fixes #2865 

Before:
Screen Shot 2022-01-20 at 7 13 29 AM

Expanded:
Screen Shot 2022-01-20 at 7 13 51 AM

After Save:
Screen Shot 2022-01-20 at 7 13 57 AM

Unit tests

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -224,6 +224,11 @@ class ItemsController < ApplicationController
     end
   end
 
+  # Draw form for barcode
+  def edit_barcode
+    @change_set = ItemChangeSet.new(@cocina)
+  end
+
   # Draw form for copyright
   def edit_copyright
     @change_set = ItemChangeSet.new(@cocina)
@@ -242,7 +247,7 @@ class ItemsController < ApplicationController
   # save the copyright form
   def update
     change_set = build_change_set
-    attributes = params.require(:item).permit(:copyright, :use_statement, :license)
+    attributes = params.require(:item).permit(:barcode, :copyright, :use_statement, :license)
     change_set.validate(**attributes)
     change_set.save
     reindex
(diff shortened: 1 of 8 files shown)


app/controllers/items_controller.rb
```diff
@@ -229,10 +229,16 @@ def edit_copyright
...
+  # Draw form for barcode
+  def edit_barcode
+    @change_set = ItemChangeSet.new(@cocina)
+  end
+
   # save the copyright form
   def update
     change_set = ItemChangeSet.new(@cocina)
     change_set.validate(copyright: params[:item][:copyright])
+    change_set.validate(barcode: params[:item][:barcode])
```
jcoyne, line 241: There is no need for a second call to validate, just pass in all the arguments at once

aaron-collier: Just saw this in your license PR, I'll update and/or rebase depending on which goes in first.

app/models/collection_change_set.rb
```diff
@@ -3,6 +3,7 @@
 # Represents a set of changes to a collection
 class CollectionChangeSet < ApplicationChangeSet
   property :admin_policy_id, virtual: true
+  property :barcode, virtual: true
```
jcoyne, line 6: This should be removed since Collections don't have barcode.

app/services/collection_change_set_persister.rb
```diff
@@ -26,7 +26,7 @@ def update
 
   attr_reader :model, :change_set
 
-  delegate :admin_policy_id, :license, :copyright, :use_statement, :catkey, :changed?, to: :change_set
+  delegate :admin_policy_id, :license, :barcode, :copyright, :use_statement, :catkey, :changed?, to: :change_set
```
jcoyne, line 29: This should be removed since collections don't have barcode

PR #3002, opened 2022-01-20
Upload structural CSV

### LINKED ISSUE #2887: Upload CSV that updates structural metadata
See https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/444556663/comments
There is an "Upload CSV" link in the "Content section".  When this is pushed and the user uploads a CSV file, the structure of the object should be updated.

The format of this CSV file will be specified by @andrewjbtw  as well as a list of validations.

@astridu will provide a mockup of what happens when this button is pushed an how to present any validation errors that occur.

jcoyne:
## Why was this change made?
Fixes #2887

### DIFF
--- a/app/components/contents_component.html.erb
+++ b/app/components/contents_component.html.erb
@@ -5,13 +5,27 @@
     </button>
   </h2>
   <div id="document-contents-section" class="accordion-collapse collapse show" aria-labelledby="document-contents-heading">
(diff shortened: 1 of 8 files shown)


justinlittman (approved):
What happens when user submits Excel instead of CSV?

jcoyne:
The upload blocks everything except CSV

PR #3013, opened 2022-01-27
Update code comment related to APO access defaults

mjgiarlo:
Connects to #3168 and #3250

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -196,7 +196,7 @@ class ItemsController < ApplicationController
     render status: :bad_request, plain: 'Invalid new rights setting.'
   end
 
-  # set the rightsMetadata to the AdminPolicies' defaultObjectRights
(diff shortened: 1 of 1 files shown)


app/controllers/items_controller.rb
```diff
@@ -196,7 +196,7 @@ def set_rights
     render status: :bad_request, plain: 'Invalid new rights setting.'
   end
 
-  # set the rightsMetadata to the AdminPolicies' defaultObjectRights
+  # set the object's access to its' admin policy's defaultAccess
```
justinlittman, line 199: "its" not "its'"

justinlittman (changes requested):

app/controllers/items_controller.rb
mjgiarlo: @justinlittman Good eye! Fixed in a follow-up drive-by commit. If you approve, can you squash-merge?

justinlittman (approved):

PR #3036, opened 2022-02-01
Allow the cancel button to re-render the show partial

### LINKED ISSUE #3024: (new Argo interface) "cancel" button in the inline editing should only cancel the edit on the field where it was clicked
If you have multiple in-line edit fields open, the "cancel" button will cancel all open edit boxes, not just the one that the button is attached to.
Screen Shot 2022-01-28 at 12 20 57 AM

The button should only cancel the specific edit being canceled and leave other open fields alone.

The "Save" button successfully saves only the one field being saved even if multiple edit fields are open.

jcoyne:
## Why was this change made?
Fixes #3024

### DIFF
--- a/app/presenters/argo_show_presenter.rb
+++ b/app/presenters/argo_show_presenter.rb
@@ -9,5 +9,15 @@ class ArgoShowPresenter < Blacklight::ShowPresenter
     CitationPresenter.new(@document).render
   end
 
-  attr_accessor :techmd, :cocina
+  def change_set
+    cocina.collection? ? CollectionChangeSet.new(cocina) : ItemChangeSet.new(cocina)
+  end
+
+  def id
+    cocina.externalIdentifier
+  end
+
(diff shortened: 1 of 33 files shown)


justinlittman (approved):

app/presenters/argo_show_presenter.rb
```diff
@@ -9,5 +9,15 @@ def heading
     CitationPresenter.new(@document).render
   end
 
-  attr_accessor :techmd, :cocina
+  def change_set
+    cocina.collection? ? CollectionChangeSet.new(cocina) : ItemChangeSet.new(cocina)
```
mjgiarlo, line 13: Is the ArgoShowPresenter not used for APOs? Or might this be using the ItemChangeSet for APOs? https://app.honeybadger.io/projects/49894/faults/83811520

jcoyne: Yeah, that seems like a possibility.

PR #3039, opened 2022-02-01
Allow the license to be cleared out

### LINKED ISSUE #3020: (new Argo interface): show "none" on the license menu
The new Argo interface shows a license menu for all items and collections. This menu should include the "none" option, which means an item isn't assigned a license at all. Currently, that option is not listed on the menu:

Screen Shot 2022-01-27 at 11 03 51 PM

Without the "none" option, it isn't possible to remove a license from an item.

jcoyne:
## Why was this change made?

Fixes #3020

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -283,10 +283,14 @@ class ItemsController < ApplicationController
   def update
     change_set = ItemChangeSet.new(@cocina)
     attributes = params.require(:item).permit(:barcode, :copyright, :use_statement, :license)
-    change_set.validate(**attributes)
-    change_set.save # may raise Dor::Services::Client::BadRequestError
-    reindex
-    redirect_to solr_document_path(params[:id]), status: :see_other
+    if change_set.validate(**attributes)
+      change_set.save # may raise Dor::Services::Client::BadRequestError
+      reindex
+      redirect_to solr_document_path(params[:id]), status: :see_other
+    else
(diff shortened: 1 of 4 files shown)


justinlittman (approved):

app/controllers/items_controller.rb
```diff
@@ -283,10 +283,14 @@ def show_license
...
-    change_set.save # may raise Dor::Services::Client::BadRequestError
-    reindex
-    redirect_to solr_document_path(params[:id]), status: :see_other
+    if change_set.validate(**attributes)
+      change_set.save # may raise Dor::Services::Client::BadRequestError
+      reindex
+      redirect_to solr_document_path(params[:id]), status: :see_other
+    else
+      logger.error "Errors: #{change_set.errors.full_messages.join}"
+      render 'error', locals: { message: message }
```
mjgiarlo, line 292: These lines were added in this PR and are uncovered which may be hiding a bug: where is `message` defined? In this case, the 🔴 coverage among CI checks might have been telling us something worth taking a look at. @jcoyne @justinlittman

jcoyne: I actually fixed this in https://github.com/sul-dlss/argo/pull/3167/files#diff-22da4de4953a8a5748f5a8f4425eb896283d008abd9525a4a33c9fd76f35f40bR294  Maybe we should pull that to it's own pr.

PR #3046, opened 2022-02-02
Display the file level access

### LINKED ISSUE #3033: (new Argo UI) show the file-level rights in the new UI contents section
I had not realized we were going to be able to show the new "Contents" display before the next phase of UI changes. Since we do have the new display, we should add the file-level rights, per Astrid's design (see the contents section here: https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/461215225 ):
Screen Shot 2022-01-31 at 11 22 35 AM

jcoyne:
## Why was this change made?

Fixes #3033
Tested on QA:

Screen Shot 2022-02-02 at 9 07 18 AM

### DIFF
--- a/app/components/contents/file_component.html.erb
+++ b/app/components/contents/file_component.html.erb
@@ -11,4 +11,6 @@
   <td class="ps-4"><%= tag :span, class: 'bi-check' if publish %></td>
   <td class="ps-4"><%= tag :span, class: 'bi-check' if shelve %></td>
   <td class="ps-4"><%= tag :span, class: 'bi-check' if sdrPreserve %></td>
+  <td><%= view_access %></td>
+  <td><%= download_access %></td>
 </tr>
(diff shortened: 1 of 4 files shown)


mjgiarlo (commented):
This change is gonna be super useful as we wind our way through the weeds of file-level rights. I wonder if the view should also include one or both of the other file-level properties to give a fuller view:

* `readLocation` (especially important when `access.access == 'location-based'`)
* `controlledDigitalLending`

Edit: I guess this is an @andrewjbtw question, since the issue this PR closes only specified View and Download. Also happy for this change to be made in a separate follow-up PR, if desired.

jcoyne:
@mjgiarlo yeah, let's make that a follow on PR as I don't think we have designs for it yet.

mjgiarlo (approved):

PR #3056, opened 2022-02-02
show workflow errors on item page

### LINKED ISSUE #3027: (new Argo UI) display workflow errors in the UI
When there's a workflow error, this should be displayed above the Overview/Details grid and below the buttons:

Screen Shot 2022-01-28 at 10 44 11 AM

(This has been added to the wireframe here: https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/461215225 )

peetucket:
## Why was this change made?

Fixes #3027 - show workflow errors in alert dialog box on item/collection/apo pages below buttons and above details, all other alerts/errors/notices are last as is in same location as before

Screen Shot 2022-02-02 at 3 59 42 PM

Localhost browser, existing tests

### DIFF
--- a/app/components/show/admin_policy_component.html.erb
+++ b/app/components/show/admin_policy_component.html.erb
@@ -7,7 +7,7 @@
   </div>
 </div>
 
-<%= render partial: 'shared/flash_msg', layout: 'shared/flash_messages' %>
+<%= render Show::WorkflowErrorComponent.new(document: document) %>
 
 <div class="row mt-5">
   <div class="col-md-6">
--- a/app/components/show/admin_policy_component.html.erb
+++ b/app/components/show/admin_policy_component.html.erb
@@ -7,6 +7,8 @@
   </div>
 </div>
 
+<%= render partial: 'shared/flash_msg', layout: 'shared/flash_messages' %>
+
 <div class="row mt-5">
   <div class="col-md-6">
     <%= render OverviewComponent.new(presenter: presenter) %>
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -43,8 +43,8 @@ class CatalogController < ApplicationController
     config.add_index_field 'identifier_tesim',                label: 'IDs', helper_method: :value_for_identifier_tesim
     config.add_index_field SolrDocument::FIELD_RELEASED_TO,   label: 'Released to'
     config.add_index_field 'status_ssi',                      label: 'Status'
-    config.add_index_field 'wf_error_ssim',                   label: 'Error', helper_method: :value_for_wf_error
-    config.add_index_field 'rights_descriptions_ssim',        label: 'Access Rights'
+    config.add_index_field SolrDocument::FIELD_WORKFLOW_ERRORS, label: 'Error', helper_method: :value_for_wf_error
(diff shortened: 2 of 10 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -207,6 +207,10 @@ def show
 
     @cocina = maybe_load_cocina(params[:id])
     flash[:alert] = 'Warning: this object cannot currently be represented in the Cocina model.' if @cocina.instance_of?(NilModel)
+    if @document[SolrDocument::FIELD_WORKFLOW_ERRORS]
+      flash[:error] =
+        "Error: #{helpers.value_for_wf_error(document: @document, field: SolrDocument::FIELD_WORKFLOW_ERRORS)}"
+    end
```
peetucket, line 214: if Solr document contains workflow errors, use the existing helper method to add them to the error flash

app/models/solr_document.rb
```diff
@@ -37,6 +37,7 @@ class SolrDocument # rubocop:disable Metrics/ClassLength
   FIELD_TAGS                      = 'tag_ssim'
   FIELD_SOURCE_ID                 = 'source_id_ssim'
   FIELD_BARCODE_ID                = 'barcode_id_ssim'
+  FIELD_WORKFLOW_ERRORS           = 'wf_error_ssim'
```
peetucket, line 41: this is the field that contains Workflow errors in solr

app/views/layouts/blacklight/base.html.erb
```diff
@@ -28,7 +28,7 @@
   <main id="main-container" class="<%= container_classes %>" role="main" aria-label="<%= t('blacklight.main.aria.main_container') %>">
     <%= content_for(:container_header) %>
 
-    <%= render partial: 'shared/flash_msg', layout: 'shared/flash_messages' %>
+    <%= render partial: 'shared/flash_msg', layout: 'shared/flash_messages' unless params[:id] && current_page?(solr_document_path)%>
```
peetucket, line 32: there may be a better way to do this...basically, we don't want to render the flash messages in the layout if we are on detail page, as that is handled in the components (so they can be rendered in a different location instead)

jcoyne (changes requested):

app/views/layouts/blacklight/base.html.erb
```diff
@@ -28,7 +28,7 @@
...
```
jcoyne, line 31: please revert this.  This box is not for all the messages, only the message about the workflow error state.

app/components/show/admin_policy_component.html.erb
```diff
@@ -7,6 +7,8 @@
   </div>
 </div>
 
+<%= render partial: 'shared/flash_msg', layout: 'shared/flash_messages' %>
```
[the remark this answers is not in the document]
jcoyne: that is:
```ruby
  field_config =  blacklight_config.show_fields_for(:show).fetch(SolrDocument::FIELD_WORKFLOW_ERRORS)
  Blacklight::FieldPresenter.new(self, @solr_document, field_config).render
 ```

peetucket:
Ok, switched from flash to a new workflow error view component ... render component in the correct place in the detail pages.  All flash messages and locations left untouched.

jcoyne (approved):

PR #3058, opened 2022-02-03
allow paging to work when using rails console in docker

peetucket:
## Why was this change made?

The default pager (`less`) is not available in the docker container which will throw errors if you do this: ` docker compose run workflow rails c` to get to the rails console and then have a lot of output.  This disables the less pager for the workflow container.

localhost


edsu:
I think "more" is probably available? Or is all paging undesirable?

peetucket:
> @peetucket I think `more` is probably available? Or is all paging undesirable?

you're right, just tried it and adjusted this PR

edsu (approved):
Looks good, thanks for checking if `more` was available.

PR #3064, opened 2022-02-04
Use the new mods update DSA endpoint for saving MODS

jcoyne:


justinlittman (approved):

andrewjbtw:
I did a very basic update but it looks like it worked: https://argo-stage.stanford.edu/view/druid:bj416cv1812

PR #3067, opened 2022-02-04
Allow barcode to be cleared out

jcoyne:


andrewjbtw:
I just tested this again and got a weird error after saving
Screen Shot 2022-02-08 at 12 55 25 PM

I'm pretty sure I was able to clear a barcode last week so I'm not sure if this is related to the current stage deployments rather than the code.

jcoyne:
@andrewjbtw this PR is not deployed anywhere, so it is not testable

aaron-collier (approved):

andrewjbtw:
Ah, ok. Then I'll ignore the error for now.

app/services/collection_change_set_persister.rb
```diff
@@ -16,7 +16,7 @@ def initialize(model, change_set)
 
   def update
     updated = model
-    updated = update_identification(updated) if changed?(:source_id) || changed?(:catkey) || changed?(:barcode)
+    updated = update_identification(updated) if changed?(:source_id) || changed?(:catkey)
```
jcoyne, line 19: collections don't have barcode

jcoyne:
Currently our openapi.yaml  isn't going to support `barcode: null` see https://github.com/OAI/OpenAPI-Specification/issues/1368 for a discussion of how nullable doesn't combine well with anyOf.

PR #3069, opened 2022-02-04
hide upload CSV button if the object is locked

### LINKED ISSUE #3062: (new Argo UI) don't allow structural metadata CSV upload when item is locked
When an item is locked, users should not be able to update structural metadata via the CSV. The structural metadata CSV download should remain available for locked objects, as it is now.

If structural updates get made without running the accessionWF afterwards, they can cause the embed viewer and the state of the files on Stacks to go out of sync. This has happened in the past when someone changed the filename in the contentMetadata without changing the actual file name on Stacks, resulting in a broken download link. By making sure users unlock the object before editing, they will then trigger the accessionWF when closing the object.

peetucket:
## Why was this change made?

Fixes #3062 - hide upload csv button if object is locked (does not allow modifications)

- Updated component test
- New structure controller test

### DIFF
--- a/app/components/contents_component.rb
+++ b/app/components/contents_component.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 class ContentsComponent < ApplicationComponent
-  def initialize(cocina:, document:, state_service:)
-    @cocina = cocina
-    @document = document
-    @state_service = state_service
+  def initialize(presenter:)
+    @document = presenter.document
+    @cocina = presenter.cocina
+    @state_service = presenter.state_service
   end
 
   def render?
--- a/app/components/contents_component.rb
+++ b/app/components/contents_component.rb
@@ -1,12 +1,15 @@
 # frozen_string_literal: true
 
 class ContentsComponent < ApplicationComponent
-  def initialize(cocina:, document:)
+  def initialize(cocina:, document:, state_service:)
     @cocina = cocina
     @document = document
+    @state_service = state_service
   end
 
   def render?
     @cocina.respond_to?(:structural)
   end
+
+  delegate :allows_modification?, to: :@state_service
 end
--- a/app/controllers/structures_controller.rb
+++ b/app/controllers/structures_controller.rb
@@ -29,7 +29,7 @@ class StructuresController < ApplicationController
         redirect_to solr_document_path(@cocina.externalIdentifier), flash: { error: status.failure.join('\n') }
       end
     else
-      redirect_to solr_document_path(@cocina.externalIdentifier), status: :not_acceptable, flash: { error: 'Updates not allowed on this object.' }
+      redirect_to solr_document_path(@cocina.externalIdentifier), flash: { error: 'Updates not allowed on this object.' }
     end
   end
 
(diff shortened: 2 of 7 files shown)


jcoyne (changes requested):
This should also add some enforcement on the controller: https://github.com/sul-dlss/argo/blob/main/app/controllers/structures_controller.rb#L19

app/controllers/structures_controller.rb
```diff
@@ -18,13 +18,18 @@ def show
...
+      if status.success?
+        object_client.update(params: @cocina.new(structural: status.value!))
+        redirect_to solr_document_path(@cocina.externalIdentifier), status: :see_other, notice: 'Structural metadata updated'
+      else
+        redirect_to solr_document_path(@cocina.externalIdentifier), flash: { error: status.failure.join('\n') }
+      end
     else
-      redirect_to solr_document_path(@cocina.externalIdentifier), flash: { error: status.failure.join('\n') }
+      redirect_to solr_document_path(@cocina.externalIdentifier), flash: { error: 'Updates not allowed on this object.' }
     end
```
peetucket, line 33: Is this what you were thinking to ensure we don't update if locked?

jcoyne: Yes, it just needs a test now.

app/components/contents_component.rb
```diff
@@ -1,12 +1,15 @@
 # frozen_string_literal: true
 
 class ContentsComponent < ApplicationComponent
-  def initialize(cocina:, document:)
+  def initialize(cocina:, document:, state_service:)
```
jcoyne, line 4: Since the presenter holds `cocina`, `document` and `state_service`, how about changing this to accept one argument, the presenter?

jcoyne (changes requested):

app/controllers/structures_controller.rb
```diff
@@ -18,13 +18,18 @@ def show
...
+
+      if status.success?
+        object_client.update(params: @cocina.new(structural: status.value!))
+        redirect_to solr_document_path(@cocina.externalIdentifier), status: :see_other, notice: 'Structural metadata updated'
+      else
+        redirect_to solr_document_path(@cocina.externalIdentifier), flash: { error: status.failure.join('\n') }
+      end
     else
-      redirect_to solr_document_path(@cocina.externalIdentifier), flash: { error: status.failure.join('\n') }
+      redirect_to solr_document_path(@cocina.externalIdentifier), status: :not_acceptable, flash: { error: 'Updates not allowed on this object.' }
```
jcoyne, line 32: I don't think you can redirect with a 406 status.

jcoyne (changes requested):

app/components/contents_component.rb
peetucket: done

app/controllers/structures_controller.rb
peetucket: yeah, i wasn't sure either, though the tests seemed to work this way... i put it back to a redirect and do other verification in the test to ensure we hit the correct code path

app/controllers/structures_controller.rb
peetucket: Added

jcoyne (approved):

PR #3072, opened 2022-02-07
Extract bulk styles from argo.scss

jcoyne:
## Why was this change made?

This helps provide context for where these styles apply in the application

### DIFF
--- a/app/assets/stylesheets/bulk-actions.scss
+++ b/app/assets/stylesheets/bulk-actions.scss
@@ -1,0 +1,10 @@
+.new_bulk_action {
+  .tab-pane {
+    @extend .mt-3;
+    @extend .mb-3;
+  }
+
+  .help-block {
+    font-weight: bold;
+  }
+}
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/bulk-actions.scss
```diff
@@ -0,0 +1,10 @@
+.new_bulk_action {
+  .tab-pane {
+    @extend .mt-3;
+    @extend .mb-3;
+  }
+
+  .help-block {
+    font-weight: bold;
+  }
+}
```
mjgiarlo, line 10: And I assume everywhere these are used they're inside the `.new_bulk_action` context? 👍🏻

mjgiarlo (approved):

app/assets/stylesheets/bulk-actions.scss
jcoyne: I believe that is the case

PR #3074, opened 2022-02-07
Allow editing and display of multiple catkeys

### LINKED ISSUE #3011: catkey entry form should support submission of multiple catkeys
The catkey entry form in Argo only supports entering one catkey at a time. We need to support entry of multiple catkeys.

I had not realized until recent conversations that multiple catkeys-per-druid are officially allowed, but we are mapping them here https://github.com/sul-dlss/dor-services-app/issues/3221 The only way to enter multiple catkeys at the moment is to edit the XML in identityMetadata to make the field repeat. Supporting multiple catkeys in the form and in the bulk catkey update is necessary to get us off of XML.

Since catkeys are always integers, I propose we support multiple catkey entry by separating each one with a comma on the same line. If we need to a more complex entry form, we'll need a new design.

jcoyne:
## Why was this change made?

Fixes #3011

Screen Shot 2022-02-07 at 3 03 23 PM

Screen Shot 2022-02-07 at 3 03 34 PM

Screen Shot 2022-02-07 at 3 03 49 PM

### DIFF
--- a/app/components/show/catkey_component.rb
+++ b/app/components/show/catkey_component.rb
@@ -2,16 +2,16 @@
 
 module Show
   class CatkeyComponent < ApplicationComponent
-    def initialize(document:, state_service:)
-      @document = document
+    def initialize(change_set:, state_service:)
+      @change_set = change_set
       @state_service = state_service
     end
 
     def catkey
-      catkey_id || 'None assigned'
+      @change_set.catkey.presence || 'None assigned'
     end
 
     delegate :allows_modification?, to: :@state_service
-    delegate :id, :catkey_id, to: :@document
+    delegate :id, to: :@change_set
   end
 end
--- a/app/models/concerns/catkey.rb
+++ b/app/models/concerns/catkey.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Catkey
+  def self.deserialize(model)
+    Array(model.identification.catalogLinks).filter_map { |link| link.catalogRecordId if link.catalog == 'symphony' }.join(', ')
+  end
+
+  def self.serialize(catkey)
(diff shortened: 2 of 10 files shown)


app/components/show/catkey_component.rb
```diff
@@ -2,16 +2,16 @@
...
     end
 
     def catkey
-      catkey_id || 'None assigned'
+      @change_set.catkey.presence || 'None assigned'
     end
 
     delegate :allows_modification?, to: :@state_service
-    delegate :id, :catkey_id, to: :@document
+    delegate :id, to: :@change_set
```
mjgiarlo, line 15: Not a big deal, but I wonder when (and when not) we want to delegate, specifically wondering about `:catkey`.

jcoyne: In this case we don't delegate, because we display a default value when the model has no catkeys.

app/models/concerns/catkey.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Catkey
+  def self.deserialize(model)
+    Array(model.identification.catalogLinks).select { |link| link.catalog == 'symphony' }.map(&:catalogRecordId).join(', ')
```
mjgiarlo, line 5: Not a giant deal, but I wonder if it's worth cou can replace the double iteration here with `#filter_map`?

mjgiarlo (approved):
Have a minor question about iterator choices, but approving

app/models/concerns/catkey.rb
jcoyne: I'm old skool!

jcoyne: Change made.

mjgiarlo (approved):

PR #3083, opened 2022-02-08
Track previous catkeys

### LINKED ISSUE #3057: Move set previous catkey logic to argo
There is currently logic in the [catkey convenience method](https://github.com/sul-dlss/dor-services/blob/c2c7eeb84cfae6e84a7b636f2e08159838cfe699/lib/dor/datastreams/identity_metadata_ds.rb#L124-L140) that adds a previous catkey when a new catkey is set.

This logic should be applied when a catkey is changed.

Currently this logic is applied at the datastream class level in dor-services gem.  The logic needs to be moved to argo at the location where we update catkeys and needs to apply the change to the cocina model.

jcoyne:
## Why was this change made?

Fixes #3057


app/models/concerns/catkey.rb
```diff
@@ -1,11 +1,50 @@
 # frozen_string_literal: true
 
-module Catkey
+class Catkey
```
mjgiarlo, line 3: Should this move out of `app/models/concerns/` now that it's a class?

mjgiarlo (approved):
Could consider moving `Catkey` but otherwise LGTM. Approving.

app/models/concerns/catkey.rb
jcoyne: Good idea. I moved it.

app/services/catkey.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class Catkey
+  SYMPHONY = 'symphony'
+  PREVIOUS_CATKEY = 'previous_catkey'
```
justinlittman, line 5: This should be `previous symphony`.

app/services/catkey.rb
```diff
@@ -0,0 +1,50 @@
...
+  def initialize(model)
+    @model = model
+  end
+
+  def deserialize
+    symphony_links.join(', ')
+  end
+
+  # If there was already a catkey in the record, store that in the "previous" spot (assuming there is no change)
+  # @return [Array<Hash>] a list of catalog links
```
justinlittman, line 24: Can you document the supported format in Rdoc?

app/services/catkey.rb
```diff
@@ -0,0 +1,50 @@
...
+  def serialize(catkey)
+    new_catkeys = split(catkey)
+    removed_links = symphony_links - new_catkeys
+    links = (previous_links + removed_links).map { |record_id| { catalog: 'previous_catkey', catalogRecordId: record_id } }
```
justinlittman, line 28: This should be `previous symphony`.

justinlittman (changes requested):

app/services/catkey.rb
jcoyne: @justinlittman the format of the catkey parameter?

app/services/catkey.rb
jcoyne: @justinlittman isn't that the behavior of the current system?

justinlittman (approved):

PR #3089, opened 2022-02-09
Keep tmp/pids in git for Argo to start

edsu:
## Why was this change made?

If you try to `docker-compose up -d` without a `tmp/pids` argo_web_1
will exit with:

```
/usr/local/bundle/gems/puma-5.6.1/lib/puma/launcher.rb:248:in `write': No such file or directory @ rb_sysopen - tmp/pids/server.pid (Errno::ENOENT)
```

This commit keeps `tmp/pids/` by adding `tmp/pids/.gitkeep`. The `tmp/.gitignore` was modified
to allow this but still ignore everything else.

Testing locally with `docker compose up -d`

None.

### DIFF
--- a/tmp/.gitignore
+++ b/tmp/.gitignore
@@ -1,4 +1,10 @@
 # Ignore everything in this directory
-*
+/*
+
 # Except this file
-!.gitignore
+!/.gitignore
+
+# And the pids directory needed for Argo to start
+!/pids
+/pids/*
+!/pids/.gitkeep


ndushay (approved):

tmp/.gitignore
```diff
@@ -1,4 +1,10 @@
 # Ignore everything in this directory
-*
+/*
```
ndushay, line 2: is the beginning slash the root directory from this app?  it's been a while ...

edsu: Good eye, yes, the docs I was reading used it. 

https://git-scm.com/docs/gitignore

If you look at the last example?

PR #3092, opened 2022-02-09
Use the new virtual object structure

mjgiarlo:
Connects to sul-dlss/dor-services-app#3416

## Why was this change made?

To reflect public API changes in DSA and avoid breaking virtual object creation.

CI. Will be tested via integration test too.

None


app/services/virtual_objects_creator.rb
```diff
@@ -29,7 +29,7 @@ def create
     Honeybadger.notify("WARN: Argo virtual object job errors from #{background_result_url}: #{job_output[:errors].inspect}")
 
     job_output[:errors].map do |error|
-      "Problem children for #{error.keys.first}: #{error.values.flatten.to_sentence}"
```
jcoyne, line 32: I'll miss you "problem child" 😢

jcoyne (approved):

PR #3095, opened 2022-02-10
Remove resource part of item link

### LINKED ISSUE #3088: "content" section for virtual objects shows double "druid:" prefix
The content section for a virtual object shows the constituent druids that make up the virtual object. The links are supposed to take you to each constituent but a druid: prefix is showing twice, causing the links to be broken:

Screen Shot 2022-02-09 at 10 20 28 AM

I'm not sure whether this is supposed to be `druid:` prefixed at all or just show the druid. It probably doesn't make a difference as long as the link goes to the right page.

edsu:
## Why was this change made?

Cocina now only has the druid for the structural members, so we can only display those. This change removes the "resource" portion and only displays an "item" link.

I viewed the contentMetadata for [druid:bf746ns6287](https://argo-stage.stanford.edu/view/druid:bf746ns6287) in my dev environment and the changes worked.

I also ran:

    bundle exec rspec spec/components/contents/external_file_component_spec.rb

@andrewjbtw tested in QA as well.

The full suite of tests unfortunately do not run reliably in my dev environment.

Closes #3088

### DIFF
--- a/app/components/contents/external_file_component.html.erb
+++ b/app/components/contents/external_file_component.html.erb
@@ -4,9 +4,7 @@
       <span class="label">
         External File
       </span>
-      <%= filename %>
-      (from item '<%= link_to druid, solr_document_path(druid) %>',
-       resource '<%= resource_id %>')
+      (from item '<%= link_to druid, solr_document_path(druid) %>')
     </li>
   </ul>
 </li>
--- a/app/components/contents/external_file_component.rb
+++ b/app/components/contents/external_file_component.rb
@@ -4,15 +4,12 @@ module Contents
   # Displays external files for a virtual object
   # e.g. https://argo.stanford.edu/view/druid:tm280sk2404
   class ExternalFileComponent < ViewComponent::Base
-    def initialize(external_file:)
-      @resource_id, @filename = external_file.split('/')
+    with_collection_parameter :druid
+
+    def initialize(druid:)
+      @druid = druid
     end
 
-    attr_reader :resource_id, :filename
-
-    def druid
-      druid, = resource_id.split('_')
-      "druid:#{druid}"
-    end
+    attr_reader :druid
   end
 end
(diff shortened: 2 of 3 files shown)


app/components/contents/external_file_component.html.erb
```diff
@@ -5,8 +5,7 @@
         External File
       </span>
       <%= filename %>
```
jcoyne, line 7: I think filename can be removed too

jcoyne (changes requested):
Please update the spec

app/components/contents/external_file_component.html.erb
edsu: @jcoyne I updated the spec, removed filename and added attr_reader for druid. I guess I should see how this works in stage before pulling out of draft?

app/components/contents/external_file_component.rb
```diff
@@ -4,15 +4,10 @@ module Contents
   # Displays external files for a virtual object
   # e.g. https://argo.stanford.edu/view/druid:tm280sk2404
   class ExternalFileComponent < ViewComponent::Base
-    def initialize(external_file:)
-      @resource_id, @filename = external_file.split('/')
+    def initialize(druid)
```
jcoyne, line 7: the convention for view components is to use kwargs (see https://viewcomponent.org/). You'll also need to update the caller to provide this kwarg rather than external_file:

jcoyne (changes requested):

app/components/contents/external_file_component.rb
edsu: Ok, thanks for the huddle -- I just pushed up the `with_collection_parameter` change.

mjgiarlo (approved):
LGTM. Might be worth a quick sniff-test by @andrewjbtw in a deployed environment, too, before we merge.

jcoyne (approved):

mjgiarlo:
@edsu Still LGTM. Feel free to merge unless you think it needs any additional testing.

edsu:
@andrewjbtw took a look in QA and gave it a 👍

PR #3106, opened 2022-02-12
refactoring state into StateService; unify lock/pencil icon logic

### LINKED ISSUE #3093: (new Argo UI) pencil icons should not appear while an object is "in accessioning"
When locking (closing) an object, the lock icon moves to the locked state immediately, but the pencil icons remain visible in the overview/details grid. The pencil icons should disappear at the same time that the lock gets locked, as it shouldn't be possible to edit items when they've been locked.

I think what's happening is that the pencils are not shown when the object has reached "Accessioned" but they are showing when the object is "in accessioning". I noticed this when recording my demo where I close an object to show the lock functionality.

Screen Shot 2022-02-09 at 3 28 40 PM

peetucket:
## Why was this change made?

Fixes #3093 - make the pencil icon and the lock icon use similar logic

- Refactor and move logic related to determining state of an object into a single location (the `StateService`) and get it out of the controller.
- Use the same logic for editing as is used to show lock/unlock icon

- Updated existing tests where necessary (new mocking required in places)
- Added new tests to `StateService` for new methods
- Tested on QA


README.md
```diff
@@ -161,8 +161,8 @@ This starts css/js bundling and the development server
 
 To begin registering items in the Argo UI, there will need to be at least one agreement object and one APO object in the index. To create and index one of each of these objects, run the following command:
 
-```shell
-$ docker compose exec web bin/rails db:seed
+```
+docker compose exec web bin/rails db:seed
 ```
```
peetucket, line 167: i'd like to be able to copy and paste this command more quickly into the terminal...

app/controllers/workflow_service_controller.rb
```diff
@@ -3,73 +3,26 @@
...
   # Draw the lock/unlock button
   def lock
-    version = Dor::Services::Client.object(params[:id]).version.current
-    opened = active_lifecycle('opened', druid: params[:id], version: version)
-    submitted = active_lifecycle('submitted', druid: params[:id], version: version)
-    template = find_template(opened, submitted)
-
-    render template, locals: { id: params[:id] }
+    render StateService.new(params[:id], version: @cocina.version).object_state.to_s, locals: { id: params[:id] }
   end
```
peetucket, line 13: All logic is now moved to the `StateService`

app/controllers/workflow_service_controller.rb
```diff
@@ -3,73 +3,26 @@
...
-  def active_assembly_wf?
-    workflow_client.workflow_status(druid: params[:id], workflow: 'assemblyWF', process: 'accessioning-initiate') == 'waiting'
-  end
+  def load_cocina
+    raise 'missing druid' unless params[:id]
 
-  def workflow_client
-    @workflow_client ||= WorkflowClientFactory.build
+    @cocina = maybe_load_cocina(params[:id])
   end
```
peetucket, line 27: All of these methods are now in the `StateService`.  We will fetch the version  (which is passed to the `StateService` since it is needed there) from the cocina model as we do in other controllers.

jcoyne (changes requested):

jcoyne (approved):

app/services/state_service.rb
```diff
@@ -1,25 +1,72 @@
...
+  # @return [Boolean]
+  def published?
+    get_lifecycle('published') ? true : false
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def object_state
+    # This item is closeable, display a working unlock button
+    return :unlock if !active_assembly_wf? && opened? && !submitted?
```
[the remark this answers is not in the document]
peetucket: Added the dry types to the top of the class file, but used symbols instead of strings as was the case before.  Added comment about the views.

app/services/state_service.rb
```diff
@@ -1,25 +1,72 @@
...
+  # rubocop:enable Metrics/PerceivedComplexity
+  # rubocop:enable Metrics/CyclomaticComplexity
+
   private
 
   attr_reader :pid, :version
 
-  def client
-    WorkflowClientFactory.build
+  def get_lifecycle(task)
```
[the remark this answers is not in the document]
peetucket: done

PR #3108, opened 2022-02-14
fix the build

peetucket:
## Why was this change made?

Fix bug and test broken by https://github.com/sul-dlss/argo/commit/3be8ef295c680c53d458cfec2fc9f68384e58d95

### DIFF
--- a/app/controllers/datastreams_controller.rb
+++ b/app/controllers/datastreams_controller.rb
@@ -48,7 +48,7 @@ class DatastreamsController < ApplicationController
     end
 
     respond_to do |format|
-      format.any { redirect_to solr_document_path(params[:item_id]), notice: msg, flash: { error: truncate_flash(error_msg) } }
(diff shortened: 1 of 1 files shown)


app/controllers/datastreams_controller.rb
```diff
@@ -48,7 +48,7 @@ def update
     end
 
     respond_to do |format|
-      format.any { redirect_to solr_document_path(params[:item_id]), notice: msg, flash: { error: truncate_flash(error_msg) } }
+      format.any { redirect_to solr_document_path(params[:item_id]), notice: msg, flash: { error: error_msg&.truncate(254) } }
     end
```
peetucket, line 52: rails has a built in method for truncation which we can use instead of writing our own method (but we need to guard against error_msg being nil in this case, which is what was causing the original bug)

mjgiarlo (approved):

PR #3110, opened 2022-02-15
Spiff up virtual object UI experience

mjgiarlo:
## Why was this change made?

Improve the virtual object experience for users by:

1. Adding a facet to find all virtual objects; and
2. Rendering a pill on the show page to let users know they're looking at a virtual object

### Facet UI

Yes, there's only one virtual object in QA. Happy to take suggestions for how to label these better. @andrewjbtw. Both the collapsible heading *and* the link could both say `Virtual Objects` for instance? 🤷🏻‍♂️ 

Screenshot from 2022-02-14 16-25-27

### Show Page

Screenshot from 2022-02-14 16-24-51

CI, QA

None

### DIFF
--- a/app/assets/stylesheets/show.scss
+++ b/app/assets/stylesheets/show.scss
@@ -8,7 +8,7 @@
   }
 
   .object-type {
-    width: 11rem;
+    width: 15rem;
     height: 2.4rem;
 
     &.object-type-apo {
@@ -19,6 +19,9 @@
     }
     &.object-type-item {
       background-color: $illuminating-light;
+    }
+    &.object-type-virtual-object {
+      background-color: $bay-light;
     }
   }
 
--- a/app/components/document_title_component.html.erb
+++ b/app/components/document_title_component.html.erb
@@ -5,7 +5,7 @@
     <%= title -%>
   <% end %>
 
-  <div class="object-type object-type-<%= object_type %> fw-bold text-center text-uppercase rounded-pill fs-4 mt-2 float-end">
-    <%= object_type %>
+  <div class="object-type object-type-<%= object_type_class %> fw-bold text-center text-uppercase rounded-pill fs-4 mt-2 float-end">
+    <%= object_type_label %>
   </div>
 </header>
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -114,6 +114,12 @@ class CatalogController < ApplicationController
     # common method since search results and reports all do the same configuration
     add_common_date_facet_fields_to_config! config
 
+    config.add_facet_field SolrDocument::FIELD_CONSTITUENTS, label: 'Virtual Objects', home: false,
+                                                             component: true,
+                                                             query: {
+                                                               has_constituents: { label: 'Virtual Objects', fq: "#{SolrDocument::FIELD_CONSTITUENTS}:*" }
+                                                             }
+
     # This will help us find records that need to be fixed before we can move to cocina.
     config.add_facet_field 'data_quality_ssim', label: 'Data Quality', home: false, component: true
 
--- a/spec/models/solr_document_spec.rb
+++ b/spec/models/solr_document_spec.rb
@@ -58,6 +58,42 @@ RSpec.describe SolrDocument, type: :model do
     end
   end
 
+  describe '#constituents' do
+    context 'when field present' do
+      let(:constituents) { ['druid:item1', 'druid:item2'] }
+      let(:document_attributes) do
+        {
+          SolrDocument::FIELD_CONSTITUENTS => constituents,
+          SolrDocument::FIELD_OBJECT_TYPE => ['item']
+        }
+      end
+
+      it 'returns the constituents' do
+        expect(document.constituents).to eq(constituents)
+      end
+
+      it 'knows it is a virtual object' do
+        expect(document).to be_virtual_object
+      end
+    end
+
+    context 'when field not present' do
+      let(:document_attributes) do
+        {
+          SolrDocument::FIELD_OBJECT_TYPE => ['item']
+        }
+      end
+
+      it 'returns nil' do
+        expect(document.constituents).to be_empty
+      end
+
+      it 'knows it is not a virtual object' do
+        expect(document).not_to be_virtual_object
+      end
+    end
+  end
+
   describe '#title' do
     subject(:title) { document.title }
 
--- a/app/assets/stylesheets/sul_variables.scss
+++ b/app/assets/stylesheets/sul_variables.scss
@@ -9,4 +9,5 @@ $sonic-silver: #767674; // "60% black" in Stanford brand identity
 $illuminating-light: #ffe781;
 $sky-light: #67afd2;
 $fog: #dad7cb;
+$bay-light: #8ab8a7;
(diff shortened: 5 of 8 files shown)


app/assets/stylesheets/show.scss
```diff
@@ -8,7 +8,7 @@
   }
 
   .object-type {
-    width: 11rem;
+    width: 15rem;
```
mjgiarlo, line 11: Pill must be wider to render `VIRTUAL OBJECT`

app/assets/stylesheets/show.scss
```diff
@@ -20,6 +20,9 @@
     &.object-type-item {
       background-color: $illuminating-light;
     }
+    &.object-type-virtual-object {
+      background-color: $bay-light;
```
mjgiarlo, line 24: Color chosen because it's in the brand guide and it's visually distinct from the other object type colors. Defer to @andrewjbtw and @astridu as to whether that's the best color to use here.

app/components/document_title_component.html.erb
```diff
@@ -5,7 +5,7 @@
     <%= title -%>
   <% end %>
 
-  <div class="object-type object-type-<%= object_type %> fw-bold text-center text-uppercase rounded-pill fs-4 mt-2 float-end">
-    <%= object_type %>
+  <div class="object-type object-type-<%= object_type_class %> fw-bold text-center text-uppercase rounded-pill fs-4 mt-2 float-end">
```
mjgiarlo, line 8: Now that we have an object type with a space in it (`virtual object`), we need a wee routine for deriving the class from the label.

app/controllers/catalog_controller.rb
```diff
@@ -114,6 +114,12 @@ class CatalogController < ApplicationController
     # common method since search results and reports all do the same configuration
     add_common_date_facet_fields_to_config! config
 
+    config.add_facet_field SolrDocument::FIELD_CONSTITUENTS, label: 'Virtual Objects', home: false,
+                                                             component: true,
+                                                             query: {
+                                                               has_constituents: { label: 'Has Constituents', fq: "#{SolrDocument::FIELD_CONSTITUENTS}:*" }
```
mjgiarlo, line 120: I'm not sure if `Has Constituents` is the best label here. Defer to @andrewjbtw

app/controllers/catalog_controller.rb
```diff
@@ -114,6 +114,12 @@ class CatalogController < ApplicationController
     # common method since search results and reports all do the same configuration
     add_common_date_facet_fields_to_config! config
 
+    config.add_facet_field SolrDocument::FIELD_CONSTITUENTS, label: 'Virtual Objects', home: false,
```
mjgiarlo, line 117: I'm not sure if `Virtual Objects` is the best label here. Defer to @andrewjbtw

astridu:
Can we make it a white box with black border? Thank you!

mjgiarlo:
@astridu 💬 
> Can we make it a white box with black border? Thank you!

For sure. Thanks!

Currently, the other pills (for items, agreements, APOs, collections) do not have borders. Maybe we need a design issue around all five that we can apply at the same time in a later PR? Tagging @andrewjbtw to assess & prioritize.

app/assets/stylesheets/show.scss
andrewjbtw: I was not expecting to go all the way to a new page type with a pill. I thought we were just making a new facet.

mjgiarlo: This commit adds both the facet and the show page pill. It doesn't need to, though, if we're not ready or if this isn't the design we want.

andrewjbtw: I'm not opposed to going down this path, I just haven't thought through the implications. It does mean we would be able to take out the "download" links in the Contents section, which don't work for virtual objects since their content is metadata that links to content.

jcoyne:
@mjgiarlo can you rebase this?

app/controllers/catalog_controller.rb
andrewjbtw: Maybe this is a theoretical question, but is it possible for something to tick the "I'm a virtual object" box and not have constituents? If so, having two values "has constituents" and "empty" could help with QA to identify empty objects. But if taking apart a virtual object removes it from the facet then I'd prefer the simplicity of "Virtual Object" - even if that ends up being redundant.

app/controllers/catalog_controller.rb
andrewjbtw: This is probably getting ahead of things but if we were to start indexing other special case objects, I could see using a more generic title for the facet. The other special case I can think of are digital serials, which are items that share a title and catkey but have their own druids, like issues of the same magazine.

In that scenario you could imagine a facet header with some name to be determined and then under it you'd see:

- Virtual Object
- Digital Serial

but that might just be making more work for ourselves at this stage.

app/controllers/catalog_controller.rb
mjgiarlo: I don't think that is possible here. This is a list of things in solr that have constituents. So there's a label for the facet header (currently `Virtual Objects`) and there's a label for the query (currently `Has Constituents`). Do you want one or both of these to change to `Virtual Object`?

app/controllers/catalog_controller.rb
mjgiarlo: These labels are super easy to change with very little in the way of consequences, so whatever works for both labels right now is fine by me!

app/controllers/catalog_controller.rb
andrewjbtw: Ok, let's just go with "Virtual Objects" in both and not overthink it. I can already see a huge benefit in having the facet.

mjgiarlo: @andrewjbtw OK, changed! Thank you.

mjgiarlo:
@jcoyne @andrewjbtw OK, made the labeling change. We can consider this unheld whenever y'all want.

jcoyne:
@mjgiarlo I've marked is as "HOLD" until Andrew has tested it on stage and he wants it to go to prod.

mjgiarlo:
Unholding per conversation with @andrewjbtw

spec/models/solr_document_spec.rb
```diff
@@ -26,35 +26,39 @@
     end
   end
 
-  describe '#publishable?' do
```
jcoyne, line 29: Is there a reason you removed this test?

jcoyne (changes requested):

spec/models/solr_document_spec.rb
mjgiarlo: @jcoyne rebase mishap. I'll fix it in the morning. Good 👁️

mjgiarlo: Restored.

mjgiarlo:
@jcoyne I fixed the rebase goof. Have another look-see?

mjgiarlo:
I assume the build is failing because cocina updates are in progress: the DSA container has already updated and #3157 is pending. Not sure it's worth holding this PR up because of that. 🤷🏻‍♂️

jcoyne (approved):

PR #3122, opened 2022-02-16
Require description when closing a version

edsu:
## Why was this change made?

Let the browser ensure that a description has been provided when closing a version.

Using my dev environment I tested closing version before and after the change, with and without a description.

### DIFF
--- a/app/views/versions/_close_ui.html.erb
+++ b/app/views/versions/_close_ui.html.erb
@@ -18,7 +18,7 @@
     </div>
     <div class='mb-3'>
       <label for="description">Version description</label><br>
-      <textarea id="description" name="description" class='form-control'><%= @description %></textarea>
+      <textarea id="description" name="description" class="form-control" required="true"><%= @description %></textarea>
     </div>
     <button type='submit' class='btn btn-primary'>
       Close Version


app/views/versions/_close_ui.html.erb
```diff
@@ -18,7 +18,7 @@
     </div>
     <div class='mb-3'>
       <label for="description">Version description</label><br>
-      <textarea id="description" name="description" class='form-control'><%= @description %></textarea>
+      <textarea id="description" name="description" class="form-control" required="true"><%= @description %></textarea>
```
mjgiarlo, line 21: I maybe should know this but I'll defer to @andrewjbtw: are there any instances where we want not to require version descriptions?

andrewjbtw: I would reformulate this as "what should we do if no description is provided"? In the Argo interface, we should always require a description. 

I'm highly suspicious that there's an automated workflow, perhaps GIS-related based on the patterns I've seen in druids that lack version names, that does not provide a description. But that's probably out of scope here. I think if a user is prompted for a description they must provide one.

mjgiarlo (approved):
LGTM, and I doubt a test suite change is needed to cover basic HTML functionality.

app/views/versions/_close_ui.html.erb
mjgiarlo: @andrewjbtw thx!

PR #3124, opened 2022-02-17
add back default access rights display for APOs

### LINKED ISSUE #3121: APO "default object rights" should use the same "access rights" value that items display
On the APO show page, the access rights value should display the same value that an item would display, given those access rights. Currently, the value showing in the APO seems to be the Cocina "access" value.

Screen Shot 2022-02-16 at 1 07 10 PM

For consistency, the value on the APO page and the value on the item show page should follow the same display rules. The goal of showing this value (previously only visible in the XML or the Edit form) is to show people exactly what items that apply these defaults will look like. 

Later in this workcycle we will be updating the rights menu to handle the more complex Cocina rights model. At that point, I'd like to make sure we update the APO page and the item page at the same time.

peetucket:
## Why was this change made?

Fixes #3121 - show the default access rights for APOs in a consistent way to that shown for items (by showing what is in the newly indexed solr documents for APOs)

See https://github.com/sul-dlss/dor_indexing_app/pull/754 for new indexing work this depends on

~~Note: HOLD until the new indexing code above is deployed, and APOs re-indexed for this to show correctly (note: it should not break if Argo is deployed before this occurs, but the display of default object rights will be blank until then).~~

Localhost browser

### DIFF
--- a/app/components/show/apo/default_object_rights_component.html.erb
+++ b/app/components/show/apo/default_object_rights_component.html.erb
@@ -1,6 +1,12 @@
 <table class="table table-sm mb-5 caption-header">
     <caption>Default object rights</caption>
     <tbody>
+      <tr>
+        <th class="col-3" scope="row">Access rights</th>
+        <td>
+            <%= access_rights %>
+        </td>
+      </tr>
 
       <tr>
         <th class="col-3" scope="row">Copyright</th>
--- a/app/components/show/apo/default_object_rights_component.rb
+++ b/app/components/show/apo/default_object_rights_component.rb
@@ -9,6 +9,10 @@ module Show
         @default_access = @presenter.cocina.administrative.defaultAccess
       end
 
+      def access_rights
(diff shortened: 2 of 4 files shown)


jcoyne:
In order to move this forward, we need to deploy dor_indexing_app to all environments and start the bulk action to reindex all the APOs.  Drop a note on this ticket when that has been completed.

peetucket:
Deployed dor-indexing-app to all environments, re-indexed APOs in all environments, and tested with this branch on QA.  It looks ok to me.

jcoyne (approved):

jcoyne (changes requested):
Please add a test.

peetucket:
Suggestions on where to add? Can't find a pattern of where other display fields like this are tested.

jcoyne:
@peetucket add it to the component test.  Set the value in the SolrDocument and check that it's drawn by the component.

jcoyne (approved):

PR #3137, opened 2022-02-21
bundle and npm updates to release branch

ndushay:
[DRAFT to ensure @jcoyne or someone can weigh in and say this is what should be done;  also to check this with integration tests after production dependency updates are done.]

lots of little dependency updates are easier than fewer big dependency updates

test suite.


justinlittman:
Aren't we done with the release branch?

ndushay:
I don't know if we wanted to control the release of additional changes to argo that are forthcoming ...

justinlittman:
If so, then I think we would want to create a new release branch rather than re-use existing.

PR #3145, opened 2022-02-23
Don't display MODS or Dublin Core button for adminPolicy

jcoyne:

This button shows the public MODS, which is a preview of what is published to PURL.  Since adminPolicies and agreements can't be published, don't show this preview.

Ref https://github.com/sul-dlss/dor-services-app/pull/3578#discussion_r812526754


app/components/show/external_links_component.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+module Show
+  # This draws links to external services for this object. Used on the sidebar of the item show page.
+  class ExternalLinksComponent < ViewComponent::Base
```
jcoyne, line 5: This was just moved into this namespace, thus "everything changed" in the diff.

app/components/show/external_links_component.rb
```diff
@@ -0,0 +1,71 @@
...
+    # @param [SolrDocument] document
+    def initialize(document:)
+      @document = document
+    end
+
+    delegate :publishable?, to: :@document
```
jcoyne, line 11: This is the only line I changed other than the namespace.

ndushay (commented):
a) oh no - failing lint/tests

b) I'm a bit muddled from slack, but ... is this the solution Andrew wanted?

justinlittman (approved):
@ndushay based on my reading of the slack conversation, this seems like the correct solution.

ndushay:
I thought that @andrewjbtw wanted to be able to use the MODS button for adminPolicy, since it has mods ... did that get resolved in a conversation somewhere that I missed?

jcoyne:
I saw Andrew say, "In that case, we can remove the buttons from APOs and agreements. I would still like APOs to handle descriptive metadata internally through the bulk metadata upload function."  

Maybe I missed something?  Is there something else we should consider?

ndushay:
I missed Andrew's comment in the other PR.  I just saw it.

PR #3150, opened 2022-02-23
[HOLD] Edit project tag from the show page

### LINKED ISSUE #3103: Show the pencil icon next to the "Project" row
The pencil icon is not showing in the "Project" row in the details table on the Argo show page. The "Project" field is essentially a tag and editable using the Tags interface, but for consistency we should show a pencil next to it as well.

Note that there is a Cocina complication around the Project tag, which is in the Cocina model. Non-project tags are not in the Cocina model. I don't think that should affect the pencil icon but there is a separate discussion about Project tags that we should return to. I thought we had a ticket about the Project tag (in DSA or Cocina) that but I'm having trouble finding it.

jcoyne:
~~Blocked by https://github.com/sul-dlss/dor-services-app/issues/3581~~

Fixes #3103

### DIFF
--- a/app/components/show/agreement/details_component.html.erb
+++ b/app/components/show/agreement/details_component.html.erb
@@ -18,7 +18,7 @@
       <tr>
         <th class="col-3" scope="row">Project</th>
         <td>
-          <%= render Show::ProjectTagComponent.new(document: @presenter.document) %>
+          <%= render Show::ProjectTagComponent.new(change_set: @presenter.change_set) %>
         </td>
       </tr>
 
--- a/app/components/show/collection/details_component.html.erb
+++ b/app/components/show/collection/details_component.html.erb
@@ -11,7 +11,7 @@
       <tr>
         <th class="col-3" scope="row">Project</th>
         <td>
-          <%= render Show::ProjectTagComponent.new(document: @presenter.document) %>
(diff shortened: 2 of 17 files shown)


justinlittman (approved):

andrewjbtw:
I tested this on stage with a collection and an item. I'm able to add a Project tag via the Project tag in-line edit but I can't delete it. When I save a blank cell the tag disappears from the Project row but remains in the "tags" box and then returns to the Project row when I reindex.

Screen Shot 2022-02-25 at 3 09 09 PM

If I remove the Project tag from the tags modal, then it really does get deleted.

jcoyne:
@justinlittman I think this problem is because partOfProject is not nillable: https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L177-L180

justinlittman:
https://github.com/sul-dlss/cocina-models/issues/321 for making partOfProject nillable.

andrewjbtw:
Should this be closed now?

jcoyne:
@andrewjbtw I guess this is a question for you (or @astridu).  Do the users want to conceptualize "Project(s)" as a tag or do we want to make "Project(s)" a separate property. I think it's confusing that we're treating it both ways at present.

andrewjbtw:
Sorry, I meant to close the issue not the PR. But I think the PR can be closed too, I'm just not sure if I'm supposed to do that.

PR #3154, opened 2022-02-24
Agreement page tweaks

### LINKED ISSUE #3112: Remove "Manage release" button and "released to" cell from Agreement page
I missed this when I looked at Agreement pages earlier. Since agreements are internal documents they'll never go out to SW so they shouldn't display the button or release status.

Note: this is not a blocker for releasing the new UI.

ndushay:

Closes #3112 -- there was cruft in the Agreement page.

### After

Note "Manage release" button is gone, and "Released to" is no longer in Details table.

image

### Before

image

As JCoyne points out, it is difficult to test for the absence of something.  These changes only affect the Argo display on the Agreeement change, and proof is in what the eye beholds?  (Per images above)

### DIFF
--- a/app/components/show/agreement/details_component.html.erb
+++ b/app/components/show/agreement/details_component.html.erb
@@ -37,13 +37,6 @@
       </tr>
 
       <tr>
-        <th class="col-3" scope="row">Released to</th>
-        <td>
-            <%= released_to %>
-        </td>
-      </tr>
-
-      <tr>
         <th class="col-3" scope="row">Preservation size</th>
         <td>
             <%= number_to_human_size(preservation_size) %>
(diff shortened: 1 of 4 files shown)


app/components/show/agreement/details_component.html.erb
```diff
@@ -36,13 +36,6 @@
         </td>
       </tr>
 
-      <tr>
-        <th class="col-3" scope="row">Released to</th>
-        <td>
-            <%= released_to %>
```
jcoyne, line 42: Can you remove the `released_to` method from Agreement::DetailsComponent now that it is not used?

jcoyne (changes requested):

app/components/show/agreement/details_component.html.erb
ndushay: yes!

ndushay: done.  Ready for re-review, @jcoyne

jcoyne (approved):

Issue #2967, opened 2022-01-13
Move object-level access to multi-select design


jcoyne:
![Screen Shot 2022-01-13 at 4.26.36 PM.png](https://images.zenhubusercontent.com/5d827302218c340001a2e38f/7ebfdddd-7bca-4614-8fbe-f4c1c92c9515)

"Download" can't offer a higher level of access than the "access field".  Can't have download: world and access: stanford.

andrewjbtw:
It should be possible to reach all possible object-level Cocina rights settings with the following two-menu design:

| view | download |
| -- | --
| world | world
| stanford | stanford
| cdl | none
| location: ars | location: ars
| location: art | location: art
| location: hoover | location: hoover
| location: m&m | location: m&m
| location: music | location: music
| location: spec | location: spec
| citation-only |  
| dark |  

Moving to this design would enable users to enter all possible object-level rights without editing XML. In the current one-menu design, certain rights can only be set via XML editing. Examples of these cases include:

- adding no-download settings to location-based rights
- combinations like "stanford, world (no-download)", where Stanford users can download but everyone else can only view

This rights menu should take the place of the current rights menu in all places where it appears:

- item show pages
- bulk rights update
- APO default object rights

Note that this menu should only apply to object-level rights. File-level rights have different requirements because certain settings do not apply to the file level, such as "citation".

andrewjbtw:
To see how these combinations map to the full cocina access values, see the [SDR rights spreadsheet](https://docs.google.com/spreadsheets/d/1BesuZ3iqouFMaur79A61Uq4ewsHQc4I0w1fRBQoYugk/edit#gid=998877965). I've added two columns to the beginning of the spreadsheet to show the Argo menu values for each valid rights setting.

We do need to prevent people from selecting invalid combinations. For example, you cannot make the download less restricted than the view. For example:
```
download: world
view: stanford
```
should be prohibited. I think the model will enforce this in the validation but the UI either needs to prevent invalid selections from being made by adjusting the menu choices interactively or it needs to show a message when an attempt to make an invalid combination is saved.

andrewjbtw:
I've filed individual tickets for each of the places where the access rights menu appears.

andrewjbtw:
Update: we're trialing a three-selection menu owing to the difficult of enforcing only the valid combinations with the two-menu design.

andrewjbtw:
[...]

PR #3155, opened 2022-02-25
object show: list all indexed rights for objects with complex rights combinations

### LINKED ISSUE #3146: "Access rights" on the show page should show full values for complex rights combinations
The "Access rights" display on Argo show pages is only showing the first value of a complex rights setting. It should show the full value.

Example: https://argo.stanford.edu/view/druid:bd336ff4952

The object-level rights for this item are:

- access: world
- download: stanford

and one file within the object is set to 

  - access: world
  - download: world

The remaining files have the same rights as the object-level rights. 

This is a valid combination of rights used for a set of tapes from a particular collection. This combination of rights comes out in the indexing as:

"stanford, world (no-download), and world (file)"

This multi-valued list shows up properly in the facet listing and the search result display: https://argo.stanford.edu/catalog?utf8=%E2%9C%93&q=bd336ff4952

Screen Shot 2022-02-23 at 12 16 44 AM

But on the show page, only the first value is displayed for the access rights:

Screen Shot 2022-02-23 at 12 16 26 AM

The access rights should show the full set of values for all complex rights settings. These values should match the ones computed for the search result display. As far as I can tell, this issue applies to all mutli-valued rights settings, not just ones as complex as this one. "Citation and world (file)" displays only as "citation", for example.

jmartin-sul:

fixes #3146

i wasn't sure if there was a better or more idiomatic way to do this, happy to change if so.

- [x] unit tests
- [x] manually on laptop instance

in search results (desired display behavior for reference):
Screen Shot 2022-02-24 at 11 44 15 PM

on show page before fix:
Screen Shot 2022-02-24 at 11 45 04 PM

on show page after fix:
Screen Shot 2022-02-24 at 11 45 32 PM

n/a

- [ ] integration test

### DIFF
--- a/app/components/show/access_rights_component.html.erb
+++ b/app/components/show/access_rights_component.html.erb
@@ -1,4 +1,4 @@
-<%= access_rights %>
+<%= access_rights.to_sentence %>
 <% if allows_modification? && !admin_policy? %>
   <%= link_to rights_item_path(id: id),
               aria: { label: 'Set rights' },
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -66,7 +66,7 @@ class SolrDocument # rubocop:disable Metrics/ClassLength
   attribute :publisher, Blacklight::Types::String, FIELD_PUBLISHER
   attribute :mods_created_date, Blacklight::Types::String, FIELD_MODS_CREATED_DATE
   attribute :status, Blacklight::Types::String, FIELD_STATUS
-  attribute :access_rights, Blacklight::Types::String, FIELD_ACCESS_RIGHTS
+  attribute :access_rights, Blacklight::Types::Array, FIELD_ACCESS_RIGHTS
   attribute :default_access_rights, Blacklight::Types::String, FIELD_DEFAULT_ACCESS_RIGHTS
   attribute :copyright, Blacklight::Types::String, FIELD_COPYRIGHT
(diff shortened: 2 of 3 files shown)


jcoyne (commented):
@andrewjbtw @astrid if we intend to switch to an in-line editor that shows one value, won't this display of multiple values be confusing?

app/components/show/access_rights_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= access_rights %>
+<%= access_rights.join(', ') %>
```
jcoyne, line 1: how about using `.to_sentence`? That's what the index page is using.

jmartin-sul: fixed

jmartin-sul:
rebased on fixed up `main`, build is passing again

app/components/show/access_rights_component.html.erb
jmartin-sul: Screen Shot 2022-02-25 at 9 49 28 AM

andrewjbtw:
This is a short-term fix to make the one-line display look like the previous display. I've found myself not realizing I was looking at an object with more complex rights when only the first value shows in Argo.

Ultimately, I want to move to a two-line display that uses two dropdown menus that will enable people to set the full range of Cocina rights. See https://github.com/sul-dlss/argo/issues/2967

jcoyne (approved):

PR #3157, opened 2022-02-25

justinlittman:


mjgiarlo:
I'm working on getting this PR 💚

app/services/rights_labeler.rb
```diff
@@ -2,48 +2,35 @@
...
-    elsif machine_read_access_node && machine_read_access_node.search('./location[text()="music"]').size.positive?
-      'loc:music'
-    elsif machine_discover_access_node && machine_discover_access_node.search('./world').size.positive?
-      # if it's not stanford restricted, world readable, or location restricted, but it is world discoverable, it's "citation only"
-      'citation-only'
-    elsif machine_discover_access_node && machine_discover_access_node.search('./none').size.positive?
-      # if it's not even discoverable, it's "dark"
-      'dark'
+    if default_access&.controlledDigitalLending
+      'cdl-stanford-nd'
```
mjgiarlo, line 21: This is NEW CODE that I added because I thought we were missing a valid case, but if this is not needed, I am happy to remove it.

app/services/rights_labeler.rb
```diff
@@ -2,48 +2,35 @@
...
+  # @param [Cocina::Models::AdminPolicyDefaultAccess] default_access admin policy default access instance
+  def initialize(default_access)
+    @default_access = default_access
   end
 
   # @return [String] a label expressing the rights of the object
   def label
-    machine_read_access_node = rights_xml.search('//rightsMetadata/access[@type="read"]/machine').first
-    machine_discover_access_node = rights_xml.search('//rightsMetadata/access[@type="discover"]/machine').first
+    return '' if default_access.nil? || default_access.access.nil?
```
mjgiarlo, line 18: We have at least one test that breaks if we don't have a case for returning an empty label, and I don't want to change this behavior in this PR.

app/services/rights_labeler.rb
mjgiarlo: Calling this out for @justinlittman @jcoyne @andrewjbtw

jcoyne (approved):

PR #3158, opened 2022-02-25
Bulk export of structural metadata as CSV

### LINKED ISSUE #3130: Create bulk action to download structural metadata from multiple objects in CSV form
This should be nested under the "CSV" section of the bulk actions. The CSV should have the same structure as the single-object structural metadata CSV, with the addition of two columns:

- druid
- content type

This will make it possible to modify structural metadata (including content type) for any arbitrary set of item druids and help with bulk remediation. This issue is paired with an issue for uploading a CSV as a bulk action.

**Additional information**

If any druids on the provided list are not items then the bulk action should report this to the log and skip over the druids while generating the CSV. If that's not possible, then the action should report to the log and stop.

jcoyne:

Fixes #3130

## How was this change tested? �
Screen Shot 2022-02-25 at 11 43 14 AM

### DIFF
--- a/app/components/bulk_actions_form_component.html.erb
+++ b/app/components/bulk_actions_form_component.html.erb
@@ -126,6 +126,12 @@
           <%= render 'bulk_actions/forms/import_tags_form', f: f %>
         </div>
 
+        <div role='tabpanel' class='tab-pane' id='ExportStructuralJob'>
+          <span class='help-block'>
+            Export structural metadata as CSV (comma-separated values) for druids specified below
+          </span>
+        </div>
+
         <div role='tabpanel' class='tab-pane' id='RegisterDruidsJob'>
           <div class="mb-3">
             <span class='help-block'>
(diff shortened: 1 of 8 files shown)


justinlittman (changes requested):
This could use a test of the contents of the CSV.

jcoyne:
@justinlittman isn't that already tested by https://github.com/sul-dlss/argo/blob/main/spec%2Fservices%2Fstructure_serializer_spec.rb?

justinlittman:
Yes, I mistakingly thought that StructureSerializer was new.

justinlittman (approved):

andrewjbtw:
This looks good for items that have structural metadata. 

For items where the CSV doesn't have anything to report, I'd like the log to report that there's no structure to report. These items are:

- registered
- accessioned but with no content
- virtual objects (not really compatible with the CSV, and I don't know that we can do anything about that)
- non-items

jcoyne:
@andrewjbtw I'm unclear about the log. Should it be one line that says "The item #{druid} has no structure" or did you want it divided up into those different classes.   As far as the cocina model is concerned, there is no reason a virtual object could not contain a file of it's own.  We also don't differentiate registered objects or other objects with no content.  In fact h2 registers and provides content at the same time.

andrewjbtw:
Just one line that says "The item #{druid} has no structural metadata to export" is fine. 

Right now the log says: "Exporting structural metadata for #{druid}" whether or not any lines are added to the CSV for that druid. The log will let people know why some of their druids aren't in the CSV if their druid list includes anything without content and they don't realize it.

PR #3167, opened 2022-02-25
Port setting source_id to a bulk action

### LINKED ISSUE #2029: Port "Set source id" bulk update to bulk action
**Existing update method**

This bulk update takes a list of druids and a tab-separated list of druids+source_ids as input. It then inserts the source_ids into the druids according to the rows on the TSV.

**When porting to bulk action**

Rename this to "Change source id" and take a CSV as input. The CSV should have these headers:

`druid,source_id`

jcoyne:
Fixes #2029

To be tested on stage by Andrew.

### DIFF
--- a/app/components/bulk_actions_form_component.html.erb
+++ b/app/components/bulk_actions_form_component.html.erb
@@ -85,6 +85,13 @@
           <%= render 'bulk_actions/forms/set_catkeys_and_barcodes_csv_form', f: f %>
         </div>
 
+        <div role="tabpanel" class="tab-pane" id="SetSourceIdsCsvJob">
+          <span class="help-block">
+            Adds or updates source IDs associated with objects.
+          </span>
+          <%= render 'bulk_actions/forms/set_source_ids_csv_form', f: f %>
+        </div>
+
         <div role='tabpanel' class='tab-pane' id='PrepareJob'>
           <span class='help-block'>
             Open items not yet open for versioning.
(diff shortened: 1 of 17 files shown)


jmartin-sul (approved):

andrewjbtw:
I tested this with objects in the various processing states:

- registered: works fine
- opened: works fine
- in accessioning: generates error as expected because item can't be opened
- accessioned: generates an error and fails

For accessioned items, it looks like the bulk action opens the items but then fails to update them. I attempted to update this set:
https://argo-stage.stanford.edu/catalog?f%5Bproject_tag_ssim%5D%5B%5D=throwaway

Each item has a version name of:
```
Source ID updated to .
```
The log shows this message for each druid:
```
2022-03-05 00:22:40 UTC Beginning set source_id for druid:df553hg0173
2022-03-05 00:22:40 UTC Adding source ID of bulk:action-83c0e607-d45b-4909-b740-10ed8330488c
2022-03-05 00:22:42 UTC Unexpected error setting source_id for druid:df553hg0173: Internal Server Error: 500 ({"status":500,"error":"Internal Server Error"})
```

We seem to be moving towards having the bulk actions automatically open accessioned items and apply the change, so what this should do is open, update the version name, and apply the new source ID.

andrewjbtw:
One more note, the instructions list the second header as `catkey_id` but this should say `source_id`. The CSV processing is accepting `source_id` as the header so I'm guessing the `catkey_id` is coming from a copy and paste for the instructions.

jcoyne:
@andrewjbtw I see this error https://app.honeybadger.io/projects/50568/faults/84382624 corresponding with the problem you found on Friday. That looks like it's a virtual object: https://argo-stage.stanford.edu/view/druid:zx135yy6705  I'm wondering if you saw that error on anything that was not a virtual object or a constituent?  I think the bug you encountered may be due to the virtual object handling in DSA and nothing about this change.

andrewjbtw:
I would need to test again. It looks like the only Accessioned items I tested on were in a virtual object. They successfully open and close individually, so it didn't occur to me that a bulk action would behave differently.

jcoyne:
@andrewjbtw this is ready to be retested when you have time.

andrewjbtw:
Looks good to me now, tested on all the item statuses.

PR #3182, opened 2022-03-02
add apply APO defaults button for items and collections

### LINKED ISSUE #3041: Add show page button for applying APO defaults
In testing, I realized that we never had a blue button for applying APO defaults on an item or collection show page. You have to paste the druid in to the bulk update interface. While going to the new layout, we should add this option to the Argo item and collection pages.

**Details**

**Object type**: this button should appear only on Collection and Item pages

**Position**: this button should appear on the "Actions" menu, after the "Purge" button

**State**: this button should only be active when the object is unlocked

**Label**: "Apply APO defaults"

peetucket:

Fixes #3041 - Add an Apply APO Defaults button that is shown for items and collections

~~HOLD Pending testing on QA~~
~~Note: Testing currently blocked by https://github.com/sul-dlss/dor-services-app/issues/3648 and https://github.com/sul-dlss/argo/issues/3185~~

Localhost browser, and updated component and request specs

Integration tests passed on QA

### DIFF
--- a/app/components/action_button.rb
+++ b/app/components/action_button.rb
@@ -1,14 +1,14 @@
 # frozen_string_literal: true
 
-# Draws a blue button in the side menu
+# Draws a red button on the item detail page
 class ActionButton < ApplicationComponent
   # @param [Hash] properties the button properties
-  def initialize(label:, url:, method: nil, confirm: nil, new_page: nil, check_url: nil, disabled: nil)
+  def initialize(label:, url:, method: nil, confirm: nil, open_modal: false, check_url: nil, disabled: nil)
     @label = label
     @url = url
     @method = method
     @confirm = confirm
-    @new_page = new_page
+    @open_modal = open_modal
     @check_url = check_url
     @disabled = disabled
   end
@@ -22,7 +22,7 @@ class ActionButton < ApplicationComponent
       if confirm
         # :confirm trumps :blacklight_modal, because :blacklight_modal would negate :confirm by firing the ajax request regardless of the user's decision
         data[:turbo_confirm] = confirm
-      elsif !new_page
+      elsif open_modal
         data[:action] = 'click->button#open'
       end
       data[:button_check_url_value] = check_url if check_url
@@ -31,5 +31,5 @@ class ActionButton < ApplicationComponent
     end
   end
 
-  attr_reader :label, :confirm, :new_page, :check_url, :url, :disabled, :method
+  attr_reader :label, :confirm, :open_modal, :check_url, :url, :disabled, :method
 end
--- a/app/components/action_button.rb
+++ b/app/components/action_button.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Draws a red button on the item detail page
+# Draws button on the item detail page
 class ActionButton < ApplicationComponent
   # @param [Hash] properties the button properties
   def initialize(label:, url:, method: nil, confirm: nil, open_modal: false, check_url: nil, disabled: nil)
--- a/app/components/show/controls_component.rb
+++ b/app/components/show/controls_component.rb
@@ -40,29 +40,41 @@ module Show
       doc.catkey_id.present?
     end
 
-    delegate :admin_policy?, :agreement?, :item?, :embargoed?, to: :doc
+    delegate :admin_policy?, :agreement?, :item?, :collection?, :embargoed?, to: :doc
 
     private
 
     def manage_release
-      render ActionButton.new(url: item_manage_release_path(pid), label: 'Manage release')
+      render ActionButton.new(url: item_manage_release_path(pid), label: 'Manage release', open_modal: true)
+    end
+
+    def apply_apo_defaults
+      return unless item? || collection?
+
+      render ActionButton.new(
+        url: apply_apo_defaults_item_path(id: pid),
+        method: 'post',
+        label: 'Apply APO defaults',
+        disabled: !state_service.allows_modification?
+      )
     end
 
     def create_embargo
       return if !item? || embargoed?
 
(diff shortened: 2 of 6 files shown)


app/components/action_button.rb
```diff
@@ -1,14 +1,14 @@
 # frozen_string_literal: true
 
-# Draws a blue button in the side menu
+# Draws a red button on the item detail page
 class ActionButton < ApplicationComponent
   # @param [Hash] properties the button properties
-  def initialize(label:, url:, method: nil, confirm: nil, new_page: nil, check_url: nil, disabled: nil)
+  def initialize(label:, url:, method: nil, confirm: nil, open_modal: false, check_url: nil, disabled: nil)
     @label = label
```
peetucket, line 7: I got tripped up by this "new_page" attribute, which basically means "don't create a modal window" when clicked.  It you don't add a "new_page", then anything other than a GET method will not work, which I found confusing.  I switched the parameter to "open_modal" instead and adjusted the buttons to reflect this.

app/components/show/controls_component.rb
```diff
@@ -40,29 +40,41 @@ def symphony_record?
...
     private
 
     def manage_release
-      render ActionButton.new(url: item_manage_release_path(pid), label: 'Manage release')
+      render ActionButton.new(url: item_manage_release_path(pid), label: 'Manage release', open_modal: true)
+    end
+
+    def apply_apo_defaults
+      return unless item? || collection?
+
```
peetucket, line 53: I added a second guard against this button being shown for APOs (you'll note that the view template itself actually doesn't render this button at all for APOs)

app/controllers/items_controller.rb
```diff
@@ -220,7 +220,13 @@ def set_rights
...
     reindex
-    render status: :ok, plain: 'APO defaults applied.'
+    respond_to do |format|
+      if params[:bulk]
+        format.html { render status: :ok, plain: 'APO defaults applied.' }
+      else
+        format.any { redirect_to solr_document_path(params[:id]), notice: 'APO defaults applied!' }
+      end
+    end
   rescue Dor::Services::Client::UnexpectedResponse => e
```
peetucket, line 230: This method is currently only used by the bulk action -- this lets us re-purpose it for the new button.

app/components/action_button.rb
```diff
@@ -1,14 +1,14 @@
 # frozen_string_literal: true
 
-# Draws a blue button in the side menu
+# Draws a red button on the item detail page
```
mjgiarlo, line 3: Haha! `s/blue/red/` 😆 

Given we will be changing these buttons' colors over time, I imagine, I wonder if we can drop the color altogether.

app/components/action_button.rb
```diff
@@ -1,14 +1,14 @@
...
 class ActionButton < ApplicationComponent
   # @param [Hash] properties the button properties
-  def initialize(label:, url:, method: nil, confirm: nil, new_page: nil, check_url: nil, disabled: nil)
+  def initialize(label:, url:, method: nil, confirm: nil, open_modal: false, check_url: nil, disabled: nil)
```
jcoyne, line 6: didn't this just flip the default?  If they specified nothing before, it would open a modal.  Now if you specify nothing it doesn't open a modal.  This seems like a change that could inadvertently cause problems.  Can you move it to a separate PR?

jcoyne (changes requested):

app/components/action_button.rb
```diff
@@ -22,7 +22,7 @@ def button_data
       if confirm
         # :confirm trumps :blacklight_modal, because :blacklight_modal would negate :confirm by firing the ajax request regardless of the user's decision
         data[:turbo_confirm] = confirm
-      elsif !new_page
+      elsif open_modal
```
mjgiarlo, line 25: I'm not saying this is a problem, but I'm observing that the default behavior before (since `new_page` defaulted to `nil`) was to trigger this condition and now the default behavior is not to trigger this condition. Intended? Not a problem if you touched all the places that call this code and do not want to trigger this condition.

app/components/action_button.rb
peetucket: well yes it flips it, but makes the intent clearer to me (I wasted an hour trying to figure out why my POST request wasn't working for the new button)

could move to a new PR, but will then have to wait for this one to merge to move forward

mjgiarlo: I think it's fine as long as you touched all the callers to ensure the right behavior?

app/components/action_button.rb
peetucket: There weren't that many spots that used this (all in the component itself) and I changed them.  i just want the attribute to match the intent ("new_page" currently means no modal AND only GET requests).

in my mind, showing a modal better implies a get request, so that name seems clearer to me

mjgiarlo: WFM

app/components/action_button.rb
peetucket: Removed color

app/components/action_button.rb
peetucket: I  fixed all  of the callers (they were all in the `ControlsComponent`, which also had other changes as part of this PR anyway)

mjgiarlo (approved):

app/components/action_button.rb
jcoyne: What about [app/views/items/_button_link_list.html.erb](https://github.com/sul-dlss/argo/blob/f728b0922975739ede99bfcb86d7cf82d611d209/app/views/items/_button_link_list.html.erb) ?

jcoyne: Okay 👍

jcoyne (approved):

PR #3188, opened 2022-03-02
Add OSS licenses

mjgiarlo:

Fixes #2408

This commit adds the OSS licenses from H2 to Argo, so that Argo now supports the same set of licenses as H2 (+ CC3). It also arranges the licenses into an order specified by the product owner.

Also, bumps coverage 7% by moving coverage block from rails helper to spec helper

CI, and may test in a deployed environment


ndushay (approved):
approving without reviewing this PR, as it duplicates the one i merged by accident.

PR #3194, opened 2022-03-03
disallow setting or editing embargo for locked objects

### LINKED ISSUE #3132: Creating or updating an embargo from Argo should require versioning
We are currently not requiring users to unlock an object in order to create or edit an embargo. As a result, changes to embargoes are not automatically propagated to the public XML, so they won't take effect in the Access systems.

We should treat embargoes like other actions that require versioning:

- Create embargo button should be inactive when an object is locked
- Embargo pencil icon should appear only when an object is unlocked

I suspect the embargo blue button didn't require versioning but we didn't notice because it never changed color.

peetucket:

Fixes #3132 - disallow setting or editing an embargo when an object is in a state where it cannot be modified (i.e. version needs to be opened)

Localhost and updated tests 

### DIFF
--- a/app/components/show/controls_component.rb
+++ b/app/components/show/controls_component.rb
@@ -62,7 +62,7 @@ module Show
     def create_embargo
       return if !item? || embargoed?
 
-      render ActionButton.new(url: new_item_embargo_path(pid), label: 'Create embargo', open_modal: true)
+      render ActionButton.new(url: new_item_embargo_path(pid), label: 'Create embargo', open_modal: true, disabled: !state_service.allows_modification?)
     end
 
     def edit_apo
--- a/app/components/show_embargo_component.rb
+++ b/app/components/show_embargo_component.rb
@@ -1,17 +1,21 @@
 # frozen_string_literal: true
 
 class ShowEmbargoComponent < ApplicationComponent
-  def initialize(solr_document:)
-    @solr_document = solr_document
+  def initialize(presenter:)
+    @solr_document = presenter.document
+    @state_service = presenter.state_service
   end
 
   delegate :id, :embargoed?, :embargo_release_date, to: :@solr_document
+  delegate :allows_modification?, to: :@state_service
 
   def render?
     embargoed? && embargo_release_date.present?
   end
 
   def edit_embargo
+    return unless allows_modification?
+
     link_to edit_item_embargo_path(id),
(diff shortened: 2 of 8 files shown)


app/components/show_embargo_component.rb
```diff
@@ -1,17 +1,21 @@
...
+  delegate :allows_modification?, to: :@state_service
 
   def render?
     embargoed? && embargo_release_date.present?
   end
 
   def edit_embargo
+    return unless allows_modification?
+
     link_to edit_item_embargo_path(id),
```
peetucket, line 19: this removes the pencil icon for editing an embargo when the object is locked

app/components/show/controls_component.rb
```diff
@@ -62,7 +62,7 @@ def apply_apo_defaults
     def create_embargo
       return if !item? || embargoed?
 
-      render ActionButton.new(url: new_item_embargo_path(pid), label: 'Create embargo', open_modal: true)
+      render ActionButton.new(url: new_item_embargo_path(pid), label: 'Create embargo', open_modal: true, disabled: !state_service.allows_modification?)
     end
```
peetucket, line 66: this disables the button to add an embargo when the object is locked

spec/components/show/controls_component_spec.rb
```diff
@@ -90,6 +116,7 @@
   describe '#registered_only?' do
     subject { component.send(:registered_only?) }
 
+    let(:allows_modification) { true }
     let(:item_id) { 'druid:kv840xx0000' }
 
     context 'when registered' do
```
peetucket, line 122: note: this spec looks like is has more changes that it actually does ... basically I created new context blocks for the DRO so that we can verify the embargo buttons are enabled or disabled as needed based on the locked/unlocked status ... the indenting caused a big diff

spec/components/show_embargo_component_spec.rb
```diff
@@ -39,6 +53,7 @@
   end
 
   context 'not embargoed with release date' do
+    let(:allows_modification) { true }
     let(:document) do
       SolrDocument.new(
         id: 1,
```
peetucket, line 59: same deal as above ... new context blocks so we can test locked/unlocked object and edit icon status ... indenting causes a big diff

jcoyne (approved):

PR #3197, opened 2022-03-04

justinlittman:


lwrubel (approved):
Approved by automated merge script

app/services/admin_policy_change_set_persister.rb
```diff
@@ -25,7 +25,11 @@ def model_for_registration
       label: title,
       version: 1,
       type: Cocina::Models::Vocab.admin_policy,
-      administrative: { hasAdminPolicy: SolrDocument::UBER_APO_ID, hasAgreement: agreement_object_id }
+      administrative: {
+        hasAdminPolicy: SolrDocument::UBER_APO_ID,
+        hasAgreement: agreement_object_id,
+        defaultAccess: { access: 'world', download: 'world' }
```
jcoyne, line 31: I think this is a dangerous default. Have you cleared this with @andrewjbtw ?

andrewjbtw: This is something where I don't understand exactly what the code means. Are the values entered in this context as examples or would they actually get applied in some contexts?

As Justin says, "world,world" is a risky default because if someone forgot to fill in the rights we shouldn't assume what they are. The safest default is "dark,none" since it won't put anything online and the depositor can come back and change it.

justinlittman: The current default (i.e., before this PR) for the form is "world".

image

If you would like it to be something different @andrewjbtw, please ticket.

justinlittman: I just tested on stage. I think "the world is OK" scenario is true.

PR #3199, opened 2022-03-04
Show registration workflows on APO page

### LINKED ISSUE #3100: Show the registration workflows in the Overview table on the APO page
https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/446060026

apo-workflows

lwrubel:
Fixes #3100 to display APO registration workflow(s) in the Overview component. 

Added specs. Deployed to QA and confirmed that one, multiple, or 'None' workflows appear. To be reviewed by @andrewjbtw. 

Screen Shot 2022-03-04 at 4 02 42 PM

Screen Shot 2022-03-07 at 10 28 34 AM

### DIFF
--- a/app/components/show/apo/overview_component.rb
+++ b/app/components/show/apo/overview_component.rb
@@ -7,6 +7,11 @@ module Show
       def initialize(presenter:)
         @presenter = presenter
         @solr_document = presenter.document
+        @registration_workflow = presenter.cocina.administrative.registrationWorkflow
+      end
+
+      def registration_workflow
+        @registration_workflow.present? ? @registration_workflow.join(', ') : 'None'
       end
 
       delegate :id, :status, to: :@solr_document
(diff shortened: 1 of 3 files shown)


app/components/show/apo/overview_component.rb
```diff
@@ -7,6 +7,11 @@ class OverviewComponent < ApplicationComponent
       def initialize(presenter:)
         @presenter = presenter
         @solr_document = presenter.document
+        @registration_workflow = @presenter.cocina.administrative.registrationWorkflow
```
mjgiarlo, line 10: Very minor: could this line use `presenter` (the kwarg) instead of `@presenter` (the instance variable)? There's no functional difference, given the code on line 8... my thinking is that the kwarg might be preferable on line 10 because that's how we're referring to the presenter on the prior line. (Again... very minor!)

lwrubel: Makes sense to me, and it works, of course. Pushing that update.

jcoyne (approved):

PR #3204, opened 2022-03-07
Bulk export of descriptive metadata

### LINKED ISSUE #3009: To support editing Cocina, the bulk descriptive metadata download should output in the form of a spreadsheet
The bulk action to download descriptive metadata currently outputs a set of MODS XML files. As we are moving away from MODS and XML, this download should be updated to output in the format of the spreadsheet that users submit when updating metadata.

Additionally, outputting the metadata in the spreadsheet format will save multiple steps in the metadata update workflow. Currently, bulk edits of metadata involve outputting MODS and then transforming that into a spreadsheet using these tools: https://stanford-library-metadata.herokuapp.com/

@arcadiafalcone will supply the form the spreadsheet download should take.

jcoyne:
Fixes #3009


jmartin-sul (commented):
one question about `Set` usage, and i'm not sure i 100% grok the recursive squish and flatten of `DescriptionExport`.  but other than the `Set` question, LGTM, since the squishing/flattening mostly makes sense to me, and the tests do a fine job of illustrating the expectations around the behavior there.

jcoyne:
@jmartin-sul the recursive call is because the descriptive schema is recursive. I don't love this, but that's the best we can do with this model. I'll take a look at the order of the headers, good call on that.

jmartin-sul:
> @jmartin-sul the recursive call is because the descriptive schema is recursive. I don't love this, but that's the best we can do with this model.

oh yeah, the recursiveness itself makes sense for that reason, i think i'm just not quite familiar enough with the schema to feel as comfortable as i'd like evaluating the particulars of the logic.  but it mostly makes sense, and i think the tests give me enough confidence to be 👍 on it.

jmartin-sul (approved):

jmartin-sul:
huh, looks like there's a stray outdated cocina-models reference in there (incomplete rebase?) resulting in `uninitialized constant Cocina::Models::Vocab` failing one of the new tests, but otherwise LGTM, thanks for the `Set` tweak!

PR #3207, opened 2022-03-08
Adds maintenance page.

### LINKED ISSUE #3180: Redirect users to a maintenance page during maintenance periods
Similar to H2, redirect users to a maintenance page when Argo should be down for maintenance. 

There are enough Argo users that it is difficult to ensure that no one uses the system during a maintenance period. Occasionally, modifications to druids made during maintenance periods can cause downstream issues. It's safer if we can make the system inaccessible altogether during maintenance.

justinlittman:
closes #3180

So that Argo can be put in maintenance mode.

QA

image

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -115,6 +115,8 @@ GEM
       capistrano-bundler (>= 1.4)
     capistrano-bundler (2.0.1)
       capistrano (~> 3.1)
+    capistrano-maintenance (1.2.1)
+      capistrano (>= 3.0)
     capistrano-one_time_key (0.1.0)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -345,10 +347,6 @@ GEM
     nokogiri (1.13.3)
       mini_portile2 (~> 2.8.0)
       racc (~> 1.4)
-    nokogiri (1.13.3-x86_64-darwin)
```
jcoyne, line 348: Did you mean to remove this?

jcoyne (approved):

PR #3210, opened 2022-03-08
Rights bulk action

### LINKED ISSUE #2030: Port "Set object rights" bulk update to bulk action
**Existing update method**

This bulk update takes a list of druids and presents the user with the rights menu.

**When porting to bulk action**

The bulk action should offer the same interaction: a list of druids + the rights menu. The rights chosen from the menu will then be applied to the druids on the list.

edsu:

Argo users would like to be able to update object rights as an asynchronous Bulk Action similar to how they do with synchronous Bulk Updates.

Closes #2030

I tested in my dev environment and on QA. See below for the more thorough testing that @andrewjbtw did. And some initial problems that were fixed.

### DIFF
--- a/app/jobs/set_rights_job.rb
+++ b/app/jobs/set_rights_job.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+##
+# Job to update rights for objects
+class SetRightsJob < GenericJob
+  ##
+  # A job that allows a user to update the rights for a list of druids
+  # @param [Integer] bulk_action_id GlobalID for a BulkAction object
+  # @param [Hash] params additional parameters that an Argo job may need
+  def perform(bulk_action_id, params)
+    super
+
+    @new_rights = params[:set_rights][:rights]
+
+    with_bulk_action_log do |log|
+      log.puts("#{Time.current} Starting SetRightsJob for BulkAction #{bulk_action_id}")
+      raise StandardError, 'Must provide rights' if @new_rights.blank?
+
+      update_druid_count
+
+      pids.each do |druid|
+        log.puts("#{Time.current} #{self.class}: Attempting #{druid} (bulk_action.id=#{bulk_action_id})")
+        set_rights(druid, log)
+      end
+      log.puts("#{Time.current} Finished SetRightsJob for BulkAction #{bulk_action_id}")
+    end
+  end
+
+  private
+
+  def set_rights(druid, log)
+    object_client = Dor::Services::Client.object(druid)
+    cocina_object = object_client.find
+    return unless verify_access(cocina_object, log)
+
+    # use dor services client to update the access
+    begin
+      state_service = StateService.new(druid, version: cocina_object.version)
+      raise StandardError, 'Object cannot be modified in its current state.' unless state_service.allows_modification?
+
+      form_type = cocina_object.collection? ? CollectionRightsForm : DroRightsForm
+      form = form_type.new(cocina_object)
+      form.validate(rights: @new_rights)
+      form.save
+
+      log.puts("#{Time.current} #{self.class}: Successfully updated rights of #{druid} (bulk_action.id=#{bulk_action.id})")
+      bulk_action.increment(:druid_count_success).save
+    rescue StandardError => e
+      log.puts("#{Time.current} SetRights failed #{e.class} #{e.message}")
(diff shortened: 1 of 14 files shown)


andrewjbtw:
When changing the object rights, the changes should be applied to the file rights as well. I ran the following sequence of changes on stage and the file rights stayed the same:

Starting point: 
```
_object-level_
view: world
download: world 

_file-level_
view: world
download: world
```

After changing to citation only:
```
_object-level_
view: citation
download: none 

_file-level_
view: world
download: world
```

After changing to stanford (no-download):
```
_object-level_
view: stanford
download: none 

_file-level_
view: world
download: world
```

Note that the expected file-rights for "citation" are:
```
view: dark
download: none
```

The set of objects I tested on: https://argo-stage.stanford.edu/catalog?f%5Bproject_tag_ssim%5D%5B%5D=throwaway&page=1

andrewjbtw:
Essentially, the bulk rights job should treat each druid the same way that the individual rights change treats them.

app/jobs/set_rights_job.rb
```diff
@@ -0,0 +1,64 @@
...
+    object_client = Dor::Services::Client.object(druid)
+    cocina_object = object_client.find
+    return unless verify_access(cocina_object, log)
+
+    # use dor services client to update the access
+    begin
+      state_service = StateService.new(druid, version: cocina_object.version)
+      raise StandardError, 'Object cannot be modified in its current state.' unless state_service.allows_modification?
+
+      object_client.update(params: cocina_object.new(cocina_update_attributes))
```
jcoyne, line 41: Have you considered using this code here:
```
form_type = cocina_object.collection? ? CollectionRightsForm : DroRightsForm
form = form_type.new(cocina_object)
form.validate(params)
form.save
```

edsu: That looks like it will handle a lot of the details of the update, thanks for the pointer! I'm still trying to figure out what `params` should be there...

jcoyne: I thought that `form.validate(rights: @new_rights)` would be the thing.  Did you follow it up with `form.save` ?

app/jobs/set_rights_job.rb
```diff
@@ -0,0 +1,64 @@
...
+      Argo::Indexer.reindex_pid_remotely(cocina_object.externalIdentifier)
```
jcoyne, line 42: Please remove this line, since this is an async job, updating right away is not important. The async job will take care of it.

app/jobs/set_rights_job.rb
edsu: Ah, it did work. I just wasn't passing in the variable the way I thought I was.

edsu: @jcoyne I'd like to adjust the tests to ensure that the rights down in the [down in the structural] section gets updated too. Is there a good example of how to do that in our rspec?

app/jobs/set_rights_job.rb
edsu: I think I removed this with my latest commit?

jcoyne: yep

app/jobs/set_rights_job.rb
edsu: I was wondering if the [a_cocina_object_with_access](https://github.com/sul-dlss/argo/blob/main/spec/support/cocina_matchers.rb#L45-L60) should check access in `structural` too? My tests were passing when those weren't getting updated.

But maybe that could be a separate task? This PR was a bear to rebase because almost every file I touched had a conflict, and it wasn't trivial for me to resolve them.

edsu: @andrewjbtw I've rebased and pushed some changes to address the file-rights update (thanks for the help @jcoyne). I've tested in my development environment and on QA to make sure it works there. Would you like to take another look before merging?

andrewjbtw:
This looks good to me. I tested the following statuses:

- opened
- registered
- accessioned
- in accessioning

I also sent objects through the transitions from:

- world
- citation
- location
- stanford no-download
- back to world
- dark (which also changes publish/shelve settings appropriately)

Everything is working as expected, so I think this is ready to go.

jcoyne (approved):

PR #3217, opened 2022-03-08
Edit access using the cocina fields

### LINKED ISSUE #2967: Move object-level access to multi-select design

jcoyne:
Fixes #2967

To be tested by Andrew on stage


aaron-collier (approved):

andrewjbtw:
I went through many if not all of the options on stage. I like the clarity of the multi-menu approach, especially how it reduces the need to mentally convert from what the menus say to how the model represents the data.

I've organized my comments more or less by component.

**View menu**

I like how the View menu selections are controlling whether the other menus actually appear. I think if we add CDL to the view menu, we can account for all of the possible view options. Until we hear otherwise, we're required to make CDL come out to "none" for download, so selecting CDL by itself could be handled the way Citation and Dark are handled where no other menu appears when it's selected.

**Download menu**

For the Download menu, am I correct that once a menu is displayed it's always going to show the same options? 

Currently, if you choose "location-based" for View, you still see the "World" and "Stanford" options in Download, even though those choices aren't valid. If there's no way to prevent people from selecting those, then we can work around that with the error messaging. But if we can prevent selections of "World" or "Stanford" download with a location-based view, then people wouldn't be able to trigger an invalid rights choice. I think this may be the only non-valid rights combination that people can select using the current design.

**Location menu**

This is works when adding a location but I'm running into errors trying to remove a location. It looks like once the location field is set and saved, it will not be removed if you open the menu again and choose a non-location set of rights. For example, I couldn't go from location-based to "World" or "Citation" or even "Dark". Even when the location menu disappears the location-based value seems to be kept behind the scenes because I see an error message when I try to save.

**Save/Cancel buttons**

There's no cancel button, just a save. For the save button, would it be possible to make it active only when the user has chosen a valid rights setting?

**Error messaging**

I like having the error message in the access rights box itself. Would it be possible to show the pencil again when an error mssage is displayed? It looks like you have to reload the page to get the pencil back.

For the message itself, would it be possible shorten the message to say something like:
```
The selected rights of "View: #{view}, Download: #{download}, Location: #{location} is not valid."
```

**File-level rights**

As with the previous rights menu, this menu should apply rights down the file-level, even if that means overriding special file-level rights that might be in place. It looked like the file-level rights were not changing while I cycled through the choices.

jcoyne:
See https://app.honeybadger.io/projects/49894/faults/84627334

andrewjbtw:
Results of the second round of testing:

Working now:

- Cancel button is there
- File-level access is being set
- "World" and "Stanford" are not selectable when location-based view is selected
- I'm able to go through the most rights transitions (from one setting to another)

Remaining issues:

- Once a location-based setting is chosen, it can be difficult to take it away

I haven't been able to take this item out of location-based access: https://argo-stage.stanford.edu/view/druid:bb259kq2628

I can change the download to None or I can change the View to Stanford or World, but I can't take location out entirely. I think the file access isn't getting the location removed since I do seem to be able to change location-based registered items that don't have files. I also see error messages like this on the item with files:
```
Error building Cocina: {"view"=>"world", "download"=>"world", "location"=>"spec", "controlledDigitalLending"=>false} isn't one of in #/components/schemas/FileAccess 
```

- I'm not able to select Citation-only or Dark if the transition is from a setting where the download is not already "None". I can go from "World, None" to Citation or Dark but I can't go from "World, World" to Citation.

andrewjbtw:
Looks good to me now after the latest deployment. One thing to consider for planning is whether to implement the new menu everywhere first before the next Argo deploy or to bring it in incrementally. I'm ok with making this change first and then updating the remaining menus.

Issue #3181, opened 2022-03-02
Support removing all content from an item via upload CSV


andrewjbtw:
To fully replace XML editing, it must be possible to remove all content from a druid. The simplest way to do this would be to make the structural metadata CSV support uploading a file that consists of just the header row.

The use cases for removing all content are the decommissioning and expunging procedures. Decommissioning requires removing all files from the current version of an item, leaving them only in preservation storage. Expunging requires following the decommissioning process, followed by removing the files from the Moab.

The current procedure for decommissioning involves either deleting all resources from the contentMetadata or posting an empty contentMetadata to Fedora (if removing a batch). It is possible to accession an empty folder from Preassembly as a workaround but it would be simpler to support removing all content using the CSV within Argo.

justinlittman:
Would it be easier to have a "remove all content", "decommission", or "expunge" button?

andrewjbtw:
It would be extremely helpful to have buttons for decommission and expunge but both of those involve much more than the contentMetadata, so there would still be a need for a separate "remove all content" option. I ticketed decommission last year but backlogged it when I worked out how to decommission without editing XML: https://github.com/sul-dlss/argo/issues/2393

For just removing content, I'm not sure having a button would be much easier from a user perspective than being able to upload a blank CSV. We could reuse the existing interfaces with the CSV.

PR #3218, opened 2022-03-09
ensure blank structural CSVs are allowed in argo

peetucket:

Part of #3181 - it turns out that you can upload a blank structural CSV and it sets the changeset to an empty array, which seems fine.  So this not working must be an issue in DSA.  Just adding a test to confirm the call to DSA is made correctly in Argo.

Added a new test

### DIFF
--- a/app/services/structure_updater.rb
+++ b/app/services/structure_updater.rb
@@ -4,6 +4,7 @@
 # So an upload CSV that lists a file not already present should be rejected.
 # At present, the CSV update only supports changing settings,
 # changing the order of resources, or removing content.
+# You can remove all content by leaving the file blank (except header).
 class StructureUpdater
   include Dry::Monads[:result]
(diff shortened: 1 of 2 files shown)


app/services/structure_updater.rb
```diff
@@ -21,7 +22,7 @@ def initialize(model, csv)
   attr_reader :model, :csv, :errors
 
   # @return [Bool] true if there are no problems
-  def validate
+  def valid?
     @errors = []
```
peetucket, line 26: small refactor - this method name better matches Rails standards

jcoyne:
Andrew, how do we remove all files from a bulk CSV? If we can do it with one object, will people expect the same for bulk?

peetucket:
Not needed - will use blank pre-assembly folders instead.

PR #3222, opened 2022-03-10
bulk actions: use search params instead of search history to get druids

### LINKED ISSUE #2792: Bulk action "populate from previous search" button sometimes returns unexpected results
**Describe the bug**

Reported from the argo-feedback mailing list:

>I've been encountering this bug with Argo lately, and I wondered if it's a known issue:
>
>I'll search a set of druids by catkey and click the white Bulk Action button to start a new bulk action.  I'll select my job type (usually either downloading descMD or opening/closing versions) and click "Populate with previous search", hoping to get the druid from the catkey I just searched.  However, it frequently populates this field with druids from whatever my "second to last" search was.  For example, this morning I looked up the Conrad Collection APO tag, and later for an unrelated task, I searched catkey 11756419 to download the descMD for those items.  When I clicked "populate with previous search" in the Bulk Action screen, it populated the field with the Conrad Collection druids.

**Additional background**

Argo has a feature that allows users to start a bulk action by clicking on the "Bulk Action" button near the top of a search results page:

Screen Shot 2021-08-03 at 4 26 55 PM

Clicking on this button takes the user to a page that lists all of their bulk actions. From that page, they can click on a link to start a new bulk action. On the new bulk action page, they then have the choice to click on a button to populate the input box with the list of druids from their previous search:

Screen Shot 2021-08-03 at 4 26 09 PM

Usually, this button will return the list of druids from the previous search. But under some circumstances the button returns a list of druids matching one of the user's older searches, not the most recent one. This happens even when clicking the exact sequence above to arrive at the bulk action directly from the most recent search result: search results --> bulk action button --> new bulk action --> populate with previous search

**User Impact**
If a user is given the wrong druid list, they may apply updates to the wrong set of druids. It can be difficult to tell from a list of druids that you are not seeing the correct druids.

**To Reproduce**

I've reproduced this but it's been difficult to pin down exactly what set of steps leads to it. Here is what I've done, all within a single browser tab starting with a new browser session:

1. Go to Argo
2. Run a search or click on a facet
3. Go to a new bulk action and check the druid list
4. Repeat this process
5. Eventually see that the druids being returned from the search are not the correct ones

One possibility is that the "populate from previous search" stops working once the number of searches has exceeded a certain amount, with additional searches not being recorded as the most recent one. The most recent time I reproduced this issue, I kept getting the same druid list even after running multiple additional searches.

To reduce the possibility of multiple browser tabs being the cause of the issue, I tested doing all of my Argo activity in a single browser tab and was able to reproduce the error under these conditions.

**Expected behavior**
When launching a bulk action from a list of search results, the bulk action should be able to populate the druid list from that set of search results.

It is possible to launch a bulk action independently (i.e. not from a results list) by launching it from the menu on the top red bar. I don't think we can easily determine what the "most recent" search was in this situation, so I wouldn't necessarily expect a bulk action launched independently to return the "correct" druid list. If we can have it return the most recent search (based on timestamp?), that would be great but not essential. I would expect people to run a search before using it to generate a bulk action druid list.

**Additional context**
I've been using Firefox and the user who reported this is on Chrome (both of us on MacOSes), so it seems like it's not a browser-specific issue.

Note also that the "bulk update" equivalent action to populate from a previous search seems to work correctly all the time. But I think the bulk updates may use a different method of generating the druid list.

peetucket:

Another attempt to fix #2792 

This does a few things:

- for bulk actions, it uses the search parameters in the query to determine if a previous search is available to get druids from (the same mechanism currently used for bulk update) instead of using the blacklight search history from cookie storage
- if no search parameters are available, it hides the "Populate with previous search" button on the New Bulk Action page
- if no search parameters are available, it hides the "Get druids from search " button on the Bulk Update page
- it stops showing the previous search from history in small text next to that button (as it is no longer relevant)

The net result is that unless you use the "Bulk Update" or "Bulk Action" buttons above the search results, you will not get those "get druid list" buttons (i.e. if you use the links in the top navigation to get to bulk update or bulk action, you will have to paste druids in).

This should fix any issues with search history getting confused between tabs, browser windows, etc.

Localhost brower, and updated tests

### DIFF
--- a/app/components/bulk_actions_form_component.html.erb
+++ b/app/components/bulk_actions_form_component.html.erb
@@ -168,7 +168,7 @@
     </div>
 
     <div data-bulk-actions-target="commonFields" >
-     <% unless search_params.blank? %>
+     <% if search_state.has_constraints? %>
         <button class='btn btn-primary' data-action="click->bulk-actions#populateDruids">
           Populate with previous search
         </button>
--- a/app/components/bulk_actions_form_component.html.erb
+++ b/app/components/bulk_actions_form_component.html.erb
@@ -62,7 +62,7 @@
             Deletes unpublished DOR objects
           </span>
         </div>
-        
+
         <div role='tabpanel' class='tab-pane' id='AddWorkflowJob'>
           <span class='help-block'>
             Starts a workflow for individual objects.
@@ -168,11 +168,10 @@
     </div>
 
     <div data-bulk-actions-target="commonFields" >
-      <% if @last_search %>
+     <% unless search_params.blank? %>
         <button class='btn btn-primary' data-action="click->bulk-actions#populateDruids">
           Populate with previous search
         </button>
-        <small><%= @last_search.query_params.except('controller', 'action') %></small>
       <% end %>
       <div class='mb-3'>
         <label>Druids to perform bulk action on</label>
--- a/app/components/bulk_actions_form_component.rb
+++ b/app/components/bulk_actions_form_component.rb
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 class BulkActionsFormComponent < ApplicationComponent
-  def initialize(form:, last_search:)
+  attr_reader :search_params
+
+  def initialize(form:, search_params:)
     @form = form
-    @last_search = last_search
+    @search_params = search_params
   end
 
   ##
@@ -13,9 +15,9 @@ class BulkActionsFormComponent < ApplicationComponent
   # @param [Blacklight::Search, nil]
   # @return [Hash]
   def search_of_pids
-    return '' if @last_search.blank?
+    return '' if search_params.blank?
 
-    @last_search.query_params.merge('pids_only' => true)
+    search_params.merge('pids_only' => true)
   end
 
   def action_types
--- a/app/components/bulk_actions_form_component.rb
+++ b/app/components/bulk_actions_form_component.rb
@@ -1,11 +1,8 @@
 # frozen_string_literal: true
 
 class BulkActionsFormComponent < ApplicationComponent
-  attr_reader :search_params
-
-  def initialize(form:, search_params:)
+  def initialize(form:)
     @form = form
-    @search_params = search_params
   end
 
   ##
@@ -15,9 +12,9 @@ class BulkActionsFormComponent < ApplicationComponent
   # @param [Blacklight::Search, nil]
   # @return [Hash]
   def search_of_pids
-    return '' if search_params.blank?
+    return '' unless search_state.has_constraints?
 
-    search_params.merge('pids_only' => true)
+    search_state.params_for_search('pids_only' => true)
   end
 
   def action_types
--- a/app/controllers/bulk_actions_controller.rb
+++ b/app/controllers/bulk_actions_controller.rb
@@ -13,7 +13,6 @@ class BulkActionsController < ApplicationController
   # GET /bulk_actions/new
   def new
     @form = BulkActionForm.new(BulkAction.new, groups: current_user.groups)
-    @last_search = session[:search].present? ? searches_from_history.find(session[:search]['id']) : nil
   end
 
   # POST /bulk_actions
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,10 +5,6 @@ module ApplicationHelper
     'Argo'
   end
 
-  def search_params
-    params.except(:controller, :action).permit!
-  end
-
   def location_to_send_search_form
     return report_path if report_view?
     return report_workflow_grid_path if workflow_grid_view?
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,6 +5,10 @@ module ApplicationHelper
     'Argo'
   end
 
+  def search_params
+    params.except(:controller, :action).permit!
+  end
+
   def location_to_send_search_form
     return report_path if report_view?
     return report_workflow_grid_path if workflow_grid_view?
--- a/app/views/bulk_actions/index.html.erb
+++ b/app/views/bulk_actions/index.html.erb
@@ -3,7 +3,7 @@
   <h1>Bulk Actions</h1>
   <div class="row">
     <div class="col-md-10">
-      <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-primary' %>
+      <%= link_to 'New Bulk Action', send(:new_bulk_action_path, search_params), class: 'btn btn-primary' %>
     </div>
     <div class="col-md-2">
       <%= link_back_to_catalog label: 'Back to search', class: 'btn btn-default' %>
--- a/app/views/bulk_actions/index.html.erb
+++ b/app/views/bulk_actions/index.html.erb
@@ -3,7 +3,7 @@
   <h1>Bulk Actions</h1>
   <div class="row">
     <div class="col-md-10">
-      <%= link_to 'New Bulk Action', send(:new_bulk_action_path, search_params), class: 'btn btn-primary' %>
+      <%= link_to 'New Bulk Action', new_bulk_action_path(search_state.to_h), class: 'btn btn-primary' %>
     </div>
     <div class="col-md-2">
       <%= link_back_to_catalog label: 'Back to search', class: 'btn btn-default' %>
--- a/app/views/catalog/_bulk_action_button.html.erb
+++ b/app/views/catalog/_bulk_action_button.html.erb
@@ -1,3 +1,3 @@
 <div class='btn-group'>
-  <%= link_to 'Bulk Actions', send(:bulk_actions_path, search_params), class: 'btn btn-outline-secondary', type: 'button' %>
+  <%= link_to 'Bulk Actions', bulk_actions_path(search_state.to_h), class: 'btn btn-outline-secondary', type: 'button' %>
 </div>
--- a/app/views/catalog/_bulk_action_button.html.erb
+++ b/app/views/catalog/_bulk_action_button.html.erb
@@ -1,3 +1,3 @@
 <div class='btn-group'>
-  <%= link_to 'Bulk Actions', bulk_actions_path, class: 'btn btn-outline-secondary', type: 'button' %>
+  <%= link_to 'Bulk Actions', send(:bulk_actions_path, search_params), class: 'btn btn-outline-secondary', type: 'button' %>
 </div>
--- a/app/views/catalog/_bulk_update_view_button.html.erb
+++ b/app/views/catalog/_bulk_update_view_button.html.erb
@@ -1,3 +1,3 @@
 <div id='bulk-update-button' class='btn-group'>
-  <%= link_to 'Bulk Update', send(:report_bulk_url, search_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
+  <%= link_to 'Bulk Update', report_bulk_url(search_state.to_h), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
 </div>
--- a/app/views/catalog/_bulk_update_view_button.html.erb
+++ b/app/views/catalog/_bulk_update_view_button.html.erb
@@ -1,4 +1,3 @@
 <div id='bulk-update-button' class='btn-group'>
-  <% cleaned_params = params.except(:controller, :action).permit! %>
-  <%= link_to 'Bulk Update', send(:report_bulk_url, cleaned_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
+  <%= link_to 'Bulk Update', send(:report_bulk_url, search_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
 </div>
--- a/app/views/report/bulk.html.erb
+++ b/app/views/report/bulk.html.erb
@@ -43,7 +43,7 @@ ul li{
 
 <h1>Bulk update operations</h1>
 
-<% unless search_params.blank? %>
+<% if search_state.has_constraints? %>
   <button class="btn btn-primary" id="get_druids" name="get_druids">Get druids from search</button>
 <% end %>
 <button class="btn btn-primary" id="paste-druids-button">Paste a druid list</button>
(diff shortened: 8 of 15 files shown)


app/components/bulk_actions_form_component.html.erb
```diff
@@ -62,7 +62,7 @@
             Deletes unpublished DOR objects
           </span>
         </div>
-        
+
         <div role='tabpanel' class='tab-pane' id='AddWorkflowJob'>
```
peetucket, line 66: white space cleanup

app/components/bulk_actions_form_component.html.erb
```diff
@@ -168,11 +168,10 @@
     </div>
 
     <div data-bulk-actions-target="commonFields" >
-      <% if @last_search %>
+     <% unless search_params.blank? %>
         <button class='btn btn-primary' data-action="click->bulk-actions#populateDruids">
           Populate with previous search
         </button>
-        <small><%= @last_search.query_params.except('controller', 'action') %></small>
       <% end %>
```
peetucket, line 175: don't need to show this to the user anymore

app/components/bulk_actions_form_component.html.erb
```diff
@@ -168,11 +168,10 @@
     </div>
 
     <div data-bulk-actions-target="commonFields" >
-      <% if @last_search %>
+     <% unless search_params.blank? %>
         <button class='btn btn-primary' data-action="click->bulk-actions#populateDruids">
```
peetucket, line 172: hide this button if we don't have any incoming search parameters in the query string

app/controllers/bulk_actions_controller.rb
```diff
@@ -13,7 +13,6 @@ def index
   # GET /bulk_actions/new
   def new
     @form = BulkActionForm.new(BulkAction.new, groups: current_user.groups)
-    @last_search = session[:search].present? ? searches_from_history.find(session[:search]['id']) : nil
   end
```
peetucket, line 16: don't need blacklight session search history anymore

app/helpers/application_helper.rb
```diff
@@ -5,6 +5,10 @@ def application_name
     'Argo'
   end
 
+  def search_params
+    params.except(:controller, :action).permit!
+  end
+
   def location_to_send_search_form
```
peetucket, line 12: move this to a helper since we use it in two places now (though actually it will go away for bulk update soon...)

app/views/bulk_actions/index.html.erb
```diff
@@ -3,7 +3,7 @@
   <h1>Bulk Actions</h1>
   <div class="row">
     <div class="col-md-10">
-      <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-primary' %>
+      <%= link_to 'New Bulk Action', send(:new_bulk_action_path, search_params), class: 'btn btn-primary' %>
     </div>
```
peetucket, line 7: link needs to pass on any incoming search params

app/views/catalog/_bulk_action_button.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class='btn-group'>
-  <%= link_to 'Bulk Actions', bulk_actions_path, class: 'btn btn-outline-secondary', type: 'button' %>
+  <%= link_to 'Bulk Actions', send(:bulk_actions_path, search_params), class: 'btn btn-outline-secondary', type: 'button' %>
 </div>
```
peetucket, line 3: pass this search params to bulk actions page when coming from the search results page

app/views/catalog/_bulk_update_view_button.html.erb
```diff
@@ -1,4 +1,3 @@
 <div id='bulk-update-button' class='btn-group'>
-  <% cleaned_params = params.except(:controller, :action).permit! %>
-  <%= link_to 'Bulk Update', send(:report_bulk_url, cleaned_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
+  <%= link_to 'Bulk Update', send(:report_bulk_url, search_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
 </div>
```
peetucket, line 3: use our new helper method instead

app/views/report/bulk.html.erb
```diff
@@ -43,7 +43,9 @@ ul li{
 
 <h1>Bulk update operations</h1>
 
-<button class="btn btn-primary" id="get_druids" name="get_druids">Get druids from search</button>
+<% unless search_params.blank? %>
+  <button class="btn btn-primary" id="get_druids" name="get_druids">Get druids from search</button>
+<% end %>
 <button class="btn btn-primary" id="paste-druids-button">Paste a druid list</button>
```
peetucket, line 49: hide this button in bulk updates if there are no incoming search params

spec/controllers/bulk_actions_controller_spec.rb
```diff
@@ -31,23 +31,6 @@
...
-      it 'with last session[:search]' do
-        Search.create
-        request.session[:search] = { 'id' => 1 }
-        all_searches = Search.all
-        expect(all_searches).to receive(:find).with(1)
-        expect(subject).to receive(:searches_from_history).and_return all_searches
-        get :new
-      end
-    end
-
```
peetucket, line 50: no need to test this anymore since we aren't using search history, and we already test the views elsewhere

app/components/bulk_actions_form_component.html.erb
```diff
@@ -168,11 +168,10 @@
     </div>
 
     <div data-bulk-actions-target="commonFields" >
-      <% if @last_search %>
+     <% unless search_params.blank? %>
```
jcoyne, line 171: Can this be if...present?

app/helpers/application_helper.rb
```diff
@@ -5,6 +5,10 @@ def application_name
     'Argo'
   end
 
+  def search_params
```
jcoyne, line 8: Can this be on the component rather than in the helpers? If not, can it be in the BulkActionHelper?

app/views/bulk_actions/index.html.erb
```diff
@@ -3,7 +3,7 @@
   <h1>Bulk Actions</h1>
   <div class="row">
     <div class="col-md-10">
-      <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-primary' %>
+      <%= link_to 'New Bulk Action', send(:new_bulk_action_path, search_params), class: 'btn btn-primary' %>
```
jcoyne, line 6: Why send?

app/views/catalog/_bulk_action_button.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class='btn-group'>
-  <%= link_to 'Bulk Actions', bulk_actions_path, class: 'btn btn-outline-secondary', type: 'button' %>
+  <%= link_to 'Bulk Actions', send(:bulk_actions_path, search_params), class: 'btn btn-outline-secondary', type: 'button' %>
```
jcoyne, line 2: Why send?

app/views/catalog/_bulk_update_view_button.html.erb
```diff
@@ -1,4 +1,3 @@
 <div id='bulk-update-button' class='btn-group'>
-  <% cleaned_params = params.except(:controller, :action).permit! %>
-  <%= link_to 'Bulk Update', send(:report_bulk_url, cleaned_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
+  <%= link_to 'Bulk Update', send(:report_bulk_url, search_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
```
jcoyne, line 2: I know this isn't new, but do you know why they used send? Can it be removed?

app/helpers/application_helper.rb
jcoyne: Have you considered using the blacklight method "search_state" instead?

spec/features/report_view_spec.rb
```diff
@@ -41,9 +41,16 @@
 
     let(:workflow_client) { instance_double(Dor::Workflow::Client, workflow_templates: []) }
 
-    it 'returns a page with the expected elements' do
+    it 'returns a page with no get druids from previous search link' do
```
jcoyne, line 44: This sentence doesn't quite parse right.

spec/features/report_view_spec.rb
```diff
@@ -41,9 +41,16 @@
...
       visit '/report/bulk'
       expect(page).to have_content('Bulk update operations')
+      expect(page).not_to have_css('.btn-primary', text: 'Get druids from search') # no previous search params, so no get druids button
+      expect(page).to have_css('.btn-primary', text: 'Paste a druid list')
+    end
+
+    it 'returns a page with a get druids from previous search link' do
```
jcoyne, line 51: This description needs work too

jcoyne (changes requested):

spec/features/report_view_spec.rb
peetucket: Will adjust descriptions to be less awkward

app/helpers/application_helper.rb
peetucket: Didn't know that exited.  I moved this helper to the component and am using blacklight's search_state instead (in the component and in other places directly).

app/components/bulk_actions_form_component.rb
```diff
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
 class BulkActionsFormComponent < ApplicationComponent
-  def initialize(form:, last_search:)
+  def initialize(form:)
     @form = form
-    @last_search = last_search
+  end
```
peetucket, line 6: we don't need the last_search passed in anymore, we can instead use the blacklight search state (which parses from the querystring)

app/views/report/bulk.html.erb
```diff
@@ -43,7 +43,7 @@ ul li{
 
 <h1>Bulk update operations</h1>
 
-<% unless search_params.blank? %>
+<% if search_state.params_for_search.present? %>
```
jcoyne, line 46:
```suggestion
<% if search_state.has_constraints? %>
```

app/components/bulk_actions_form_component.html.erb
```diff
@@ -168,7 +168,7 @@
     </div>
 
     <div data-bulk-actions-target="commonFields" >
-     <% unless search_params.blank? %>
+     <% if search_params.present? %>
```
jcoyne, line 171: can we use `helpers.search_state.has_constraints?` here?

app/components/bulk_actions_form_component.rb
```diff
@@ -1,11 +1,12 @@
...
 class BulkActionsFormComponent < ApplicationComponent
-  attr_reader :search_params
-
-  def initialize(form:, search_params:)
+  def initialize(form:)
     @form = form
-    @search_params = search_params
+  end
+
+  def search_params
```
jcoyne, line 8: And get rid of this?

jcoyne (changes requested):

app/views/bulk_actions/index.html.erb
```diff
@@ -3,7 +3,7 @@
   <h1>Bulk Actions</h1>
   <div class="row">
     <div class="col-md-10">
-      <%= link_to 'New Bulk Action', new_bulk_action_path, class: 'btn btn-primary' %>
+      <%= link_to 'New Bulk Action', new_bulk_action_path(search_state.params_for_search), class: 'btn btn-primary' %>
```
jcoyne, line 6:
```suggestion
      <%= link_to 'New Bulk Action', new_bulk_action_path(search_state.to_h), class: 'btn btn-primary' %>
```

app/views/catalog/_bulk_update_view_button.html.erb
```diff
@@ -1,4 +1,3 @@
 <div id='bulk-update-button' class='btn-group'>
-  <% cleaned_params = params.except(:controller, :action).permit! %>
-  <%= link_to 'Bulk Update', send(:report_bulk_url, cleaned_params), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
+  <%= link_to 'Bulk Update', report_bulk_url(search_state.params_for_search), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
```
jcoyne, line 2:
```suggestion
  <%= link_to 'Bulk Update', report_bulk_url(search_state.to_h), class: 'btn btn-outline-secondary', type: 'button', id: 'bulk_update_view' %>
```

jcoyne (changes requested):

app/views/catalog/_bulk_update_view_button.html.erb
peetucket: done

app/components/bulk_actions_form_component.rb
```diff
@@ -5,20 +5,16 @@ def initialize(form:)
...
   # Blacklight::Search
   # This is used by the "Populate with previous search" feature of bulk actions
   # @param [Blacklight::Search, nil]
   # @return [Hash]
   def search_of_pids
-    return '' if search_params.blank?
+    return '' unless search_state.has_constraints?
 
-    search_params.merge('pids_only' => true)
+    search_state.params_for_search.merge('pids_only' => true)
```
jcoyne, line 17:
```suggestion
    search_state.params_for_search('pids_only' => true)
```

spec/components/bulk_actions_form_component_spec.rb
```diff
@@ -6,12 +6,17 @@
...
+  let(:search_state) { Blacklight::SearchState.new(search_params, blacklight_config) }
 
-  before { allow(controller).to receive(:current_user).and_return(build(:user)) }
+  before do
+    allow(controller).to receive(:current_user).and_return(build(:user))
+    allow(subject).to receive(:search_state).and_return(search_state)
+  end
 
   describe '#search_of_pids' do
     context "when last search isn't present" do
```
jcoyne, line 18: I don't think this is a meaningful test anymore (it's the same as the next one)

jcoyne (changes requested):

spec/components/bulk_actions_form_component_spec.rb
peetucket: removed

app/components/bulk_actions_form_component.rb
peetucket: changed

jcoyne (approved):

PR #3225, opened 2022-03-10
Remove bulk tags

### LINKED ISSUE #3224: Remove the bulk update for tags
Last year, or maybe even the year before, we implemented a set of bulk actions to update tags via exporting and then importing a CSV. At the time, we left the old bulk update for tags in place.

Now that we're porting everything to bulk actions we should remove the tags bulk update so we can remove all bulk updates.

jcoyne:
Fixes #3224


jmartin-sul (approved):

jmartin-sul:
needs a rebase, but otherwise lgtm

PR #3229, opened 2022-03-11
Refactor bulk jobs

### LINKED ISSUE #3165: Refactor BulkActionForm#csv_as_string
Is there a way to make a list of the types that take CSV as input and run the method based of that?

This would do away with this ugly conditional:
https://github.com/sul-dlss/argo/blob/main/app/forms/bulk_action_form.rb#L36-L40

### LINKED ISSUE #3166: Consider refactoring bulk action to use turbo-frames to load each form
This allows us to have 1 form per type of bulk action, rather than 1 big form, where parts of it are hidden.  This might make us have more classes, but each should be simpler. 

This means we would create a new controller for each form type:
```
/bulk_actions/manage_release
/bulk_actions/reindex
/bulk_actions/republish
...
```

Each of those would have a `new` method that draws the form and a `create` method that creates the bulk action and starts the appropriate job. 

This means the `BulkActions#create`method can be removed. https://github.com/sul-dlss/argo/blob/main/app/controllers/bulk_actions_controller.rb#L20-L30

jcoyne:
This removes the BulkActionForm and BulkActionFormPersister. Each type of bulk job is given its own controller responsible for showing its form and for creating the job.  The goal is that the logic is less complex.

Fixes #3166 
Fixes #3165


config/routes.rb
```diff
@@ -6,10 +6,43 @@
...
-  resources :bulk_actions, except: %i[edit show update] do
+  resources :bulk_actions, only: %i[index new destroy] do
     member do
       get :file
     end
+    scope module: 'bulk_actions' do
+      collection do
+        # In the order they appear in the dropdown
+        resource :apply_apo_defaults_job, only: %i[new create]
+        resource :manage_release_job, only: %i[new create]
```
mjgiarlo, line 17: I believe the order may have changed since you create this PR? The order we want is in https://github.com/sul-dlss/argo/pull/3229/files#diff-2e44031d3ab23dbf7a0da286d7b5f23043e6e55fffc309a178a33a58325176a1R17

mjgiarlo: But I'm not sure it matter what order the routes are in... 🤷🏻

mjgiarlo (commented):
We don't normally go out of our way to user-test refactorings but given the scale of the changes here, that might be warranted. If already tested, consider this approved. If not, I might like to get PO testing sign-off before we merge this.

config/routes.rb
jcoyne: Thanks, I've updated them to match.

mjgiarlo (approved):

jcoyne:
Maybe, but that's really something that should be a new ticket and should get Andrew and Astrid's buy in.

andrewjbtw:
I'll file issues in comments and separate each one into its own comment.

The "populate with previous search" button should only appear if the user clicks through from a previous search. I'm seeing it at all time when going to "New bulk actions" from this page: https://argo-stage.stanford.edu/bulk_actions

andrewjbtw:
The "manage release" action isn't working. I tried to release https://argo-stage.stanford.edu/view/druid:bd329tx5888 and that send sidekiq into a retry loop. 
https://argo-stage.stanford.edu/queues/retries
https://app.honeybadger.io/projects/49894/faults/84611231

andrewjbtw:
The open version bulk action is opening a version but not including the version description or tag in the versionMetadata. I can't tell what's being written to the database but since the version display shows a blank, I'm guessing it's not writing anything there.

Screen Shot 2022-03-22 at 1 25 09 PM

andrewjbtw:
The Update Governing APO bulk action is reporting success but not actually changing the governing APO. I tried on an accessioned and a registered item.

To reproduce:

Registered item: bs651wm4021
APO I tried to select:  2021-04-02 (https://argo-stage.stanford.edu/view/druid:kc818fc2996)
APO stayed as the "berger-testing" APO.

andrewjbtw:
The "Edit license and rights statements" bulk action is reporting success but no update is being made. I attempted to update the statements on https://argo-stage.stanford.edu/view/druid:bd329tx5888

andrewjbtw:
The form version of Edit Catkeys and Barcodes does not appear to be doing anything.

I got the following log when trying to update a single druid with a catkey of "111"
```
2022-03-22 22:00:45 UTC Starting SetCatkeysAndBarcodesJob for BulkAction 1508
2022-03-22 22:00:45 UTC Finished SetCatkeysAndBarcodesJob for BulkAction 1508
```

What I entered:
Screen Shot 2022-03-22 at 3 02 55 PM

The bulk action page reported back 1 druid total with 0 success or failed. 
Screen Shot 2022-03-22 at 3 04 19 PM

andrewjbtw:
Set collection is not working. Even when I choose a new collection, the result is that the listed druids are being removed from all collections. I assume it removes the existing collection relationship before adding the new one, so maybe it's getting partway through the process before failing to add.

I used these settings:

Screen Shot 2022-03-22 at 3 17 28 PM

Both of the druids on that list had existing collection relationships. After the bulk action ran, neither druid was in any collection anymore. The log reported success:
```
2022-03-22 22:17:29 UTC Starting SetCollectionJob for BulkAction 1517
2022-03-22 22:17:29 UTC SetCollectionJob: Starting update for druid:bs651wm4021 (bulk_action.id=1517)
2022-03-22 22:17:30 UTC SetCollectionJob: Successfully updated druid:bs651wm4021 (bulk_action.id=1517)
2022-03-22 22:17:30 UTC SetCollectionJob: Finished update for druid:bs651wm4021 (bulk_action.id=1517)
2022-03-22 22:17:30 UTC SetCollectionJob: Starting update for druid:bd329tx5888 (bulk_action.id=1517)
2022-03-22 22:17:32 UTC SetCollectionJob: Successfully updated druid:bd329tx5888 (bulk_action.id=1517)
2022-03-22 22:17:32 UTC SetCollectionJob: Finished update for druid:bd329tx5888 (bulk_action.id=1517)
2022-03-22 22:17:32 UTC Finished SetCollectionJob for BulkAction 1517
```

andrewjbtw:
Selecting the "Change source_id" option isn't pulling up the correct form. Looks like it's staying on whichever option was selected at the time the "Change source_id" is chosen.

I took this video of what happens when I navigate directly to the bulk actions from the home page and then choose "Change source_id":

https://user-images.githubusercontent.com/8975583/159590161-2979b711-c0fd-42f4-bbcc-dfa298a007d4.mov

andrewjbtw:
I've tested everything but create virtual objects so far. Tests have been mostly with one or two druids that I've already opened for versioning since the automatically opened versions don't seem to be getting descriptions.

Aside from the ones I've comment on above, these look like they're working as expected:

- reindex
- republish
- purge
- add workflow
- close
- apply APO defaults
- set object rights
- refresh MODS
- set content type
- edit catkeys and barcodes from CSV
- export tags
- import tags
- export structural metadata
- manage embargo
- register druids
- download MODS
- checksum reports

andrewjbtw:
I've tested all the actions that I reported above as not working and all of them work now. I haven't tested virtual objects, but I don't think that's necessary at this point.

The one remaining issue is that the "populate with previous search" remains visible at all times and it looks like it might be pulling in the first 10 druids of a search for all druids if you click on it when you've opened the page without coming from a previous query.

PR #3230, opened 2022-03-11
Remove the vestiges of the Bulk Update

jcoyne:

All bulk updates are now bulk actions


ndushay (approved):
looks good;  wondering if there should be a replacement "when request fails" test but no strong feelings.

jcoyne:
There probably should be, but that's not code I'm touching in this PR. That "success" test was just promoted up out of a context, I didn't add it.

PR #3236, opened 2022-03-14
Upgrade to Rails 7

jcoyne:


config/environments/production.rb
```diff
@@ -74,21 +68,15 @@
...
-  # Send deprecation notices to registered listeners.
-  config.active_support.deprecation = :notify
-
-  # Log disallowed deprecations.
-  config.active_support.disallowed_deprecation = :log
-
-  # Tell Active Support which deprecation messages to disallow.
-  config.active_support.disallowed_deprecation_warnings = []
+  # Don't log any deprecations.
+  config.active_support.report_deprecations = false
```
jmartin-sul, line 72: does this mean we don't get deprecation warnings at all in the logs/alerts anymore?

jmartin-sul (approved):
one small question, but overall 👍

config/environments/production.rb
jmartin-sul: ah but i guess we'll still see warnings when running tests locally?  seems like it'd be nice to get deprecation warnings somewhere.

jmartin-sul: (but also didn't seem worth holding up merge over that question)

PR #3266, opened 2022-03-16
Remove unused code

jcoyne:
This code is not used.


justinlittman:
Does this have extra commits?

aaron-collier (approved):

aaron-collier:
Merging. I've seen small coverage drops like this when removing code.

PR #3307, opened 2022-03-22
Refactor jobs to have a common pattern for jobs that load from cocina

### LINKED ISSUE #3161: Extract a `with_druids` method from bulk jobs
This method would be used like this:

```ruby
  with_druids(params[:druids], log: log) do |druid, success, failure|
    result = some_operation_on(druid)
    if result == 'failed'
      failure.call('it just did not work for this one')
    else
      success.call('this one is good')
    end
  end
```

This can encapsulate the log writing  and success/failure incrementing so that the individual jobs don't have to.
This should automatically call `update_druid_count`

This method can also rescue any error within the yield and log it. Then the jobs won't need to.

jcoyne:
Fixes #3161

### DIFF
--- a/app/jobs/export_structural_job.rb
+++ b/app/jobs/export_structural_job.rb
@@ -3,21 +3,18 @@
 ##
 # A job that exports structural metadata to CSV for one or more objects
 # @param [Integer] bulk_action_id GlobalID for a BulkAction object
-# @param [Hash] _params additional parameters that an Argo job may need
+# @param [Hash] params additional parameters that an Argo job may need
 class ExportStructuralJob < GenericJob
-  def perform(bulk_action_id, _params)
+  def perform(bulk_action_id, params)
     super
 
-    with_bulk_action_log do |log_buffer|
-      update_druid_count
-
-      CSV.open(csv_download_path, 'w', headers: true) do |csv|
-        csv << ['druid', *StructureSerializer::HEADERS]
-        druids.each do |druid|
-          rows_for_file = item_to_rows(druid, log_buffer)
-          rows_for_file.each do |row|
-            csv << [druid.delete_prefix('druid:'), *row]
-          end
+    CSV.open(csv_download_path, 'w', headers: true) do |csv|
+      csv << ['druid', *StructureSerializer::HEADERS]
+      with_items(params[:druids], name: 'Export structural metadata') do |cocina_object, success, failure|
+        rows_for_file = item_to_rows(cocina_object, success, failure)
+        druid = cocina_object.externalIdentifier.delete_prefix('druid:')
+        rows_for_file.each do |row|
+          csv << [druid, *row]
         end
       end
     end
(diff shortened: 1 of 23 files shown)


app/jobs/export_structural_job.rb
```diff
@@ -26,32 +22,24 @@ def perform(bulk_action_id, _params)
...
       result << row
     end
-    bulk_action.increment(:druid_count_success).save
+    success.call('Exported structural metadata')
     result
   rescue StandardError => e
-    log_buffer.puts("#{Time.current} #{self.class}: Unexpected error exporting structural metadata for #{druid}: #{e}")
-    bulk_action.increment(:druid_count_fail).save
+    failure.call("Unexpected error exporting structural metadata: #{e.message}")
+    Honeybadger.notify(e)
```
justinlittman, line 39: Does this need a context? Does it need to be rescued at all or is it rescued by `with_items`?

app/jobs/generic_job.rb
```diff
@@ -56,6 +56,28 @@ def with_bulk_action_log
     end
   end
 
+  def with_items(druids, name:)
```
justinlittman, line 59: Can you document how this method should be used? In particular, the success and failure lambdas.

app/jobs/set_collection_job.rb
```diff
@@ -14,43 +14,26 @@ def perform(bulk_action_id, params)
...
-    log.puts("#{Time.current} SetCollectionJob: Unexpected error for #{current_druid} (bulk_action.id=#{bulk_action.id}): #{e} #{e.backtrace}")
-    bulk_action.increment(:druid_count_fail).save
   end
 
-  def check_can_set_collection!(cocina, state_service)
-    raise "#{cocina.externalIdentifier} is not open for modification" unless state_service.allows_modification?
+  private
 
+  def check_can_set_collection!(cocina)
     new_collection_ids.each do |new_collection_id|
```
justinlittman, line 37: It would be good to be consistent about whether this should be raised and not caught or handled by invoking `failure.call`.

app/jobs/set_rights_job.rb
```diff
@@ -10,50 +10,22 @@ class SetRightsJob < GenericJob
...
     super
 
-    @new_rights = params[:set_rights][:rights]
+    new_rights = params[:set_rights][:rights]
+    raise 'Must provide rights' if new_rights.blank?
 
-    with_bulk_action_log do |log|
-      raise StandardError, 'Must provide rights' if @new_rights.blank?
+    with_items(params[:pids], name: 'Set rights') do |cocina_object, success, failure|
+      # byebug
```
justinlittman, line 17: cruft

justinlittman (changes requested):
My primary concern is the inconsistency between raising errors and invoking `failure.call`.

app/jobs/set_collection_job.rb
justinlittman: Merged.

justinlittman (approved):

PR #3313, opened 2022-03-22
Strip druid: prefix in structural metadata CSV

### LINKED ISSUE #3292: Do not include the "druid:" prefix in the structural metadata CSV
I recently had the opportunity/need to use a structural CSV export with the preassembly file manifest. This made it far easier to reaccession OCR'd files than ever before, which is great.

The one issue I ran into is that users (including me) often use the druid in folder names, and the ":" in the prefix can cause issues with the filesystems they copy from. If we don't use the "druid:" prefix in the structural CSV, then the druid column can be reused in a preassembly manifest without needing to do a find and replace to take the prefix out.

edsu:

This is a small change to strip the "druid:" prefix from druids in the structural metadata bulk action export. The existing test has been updated to parse the CSV and make sure the druids are correct.

Adjusted unit test to look at druid column. Ran a report with 7 items on QA.

Closes #3292

### DIFF
--- a/spec/jobs/export_structural_job_spec.rb
+++ b/spec/jobs/export_structural_job_spec.rb
@@ -424,8 +424,13 @@ RSpec.describe ExportStructuralJob, type: :job do
         expect(log_buffer.string).to include "Exporting structural metadata for #{druid1}"
         expect(log_buffer.string).to include "Exporting structural metadata for #{druid2}"
         expect(File).to exist(csv_path)
-        File.open(csv_path, 'r') do |file|
-          expect(file.readlines.size).to eq 9 # one row for each file plus the headers.
+      end
+
+      it 'is valid csv and druids are prefixless' do
+        output = CSV.read(csv_path, headers: true)
+        expect(output.size).to eq 8
+        output.each do |row|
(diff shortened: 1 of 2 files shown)


spec/jobs/export_structural_job_spec.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'csv'
```
justinlittman, line 3: Maybe this should this go in `rails_helper`? (I'm not entirely clear on this convention.)

justinlittman (approved):
Approve pending consideration of comment. Can you check with another dev on the proper convention?

spec/jobs/export_structural_job_spec.rb
jcoyne: I'm surprised it's needed at all, it's required in the app code.

edsu: The test failed without it.

edsu: @jcoyne you were right, the require doesn't seem needed -- it's weird because I thought I saw it fail before.  I've removed the require.

PR #3317, opened 2022-03-22
allow the search state params to pass through bulk actions

peetucket:

Now that we use query string parameters to define the previous search used for the "populate druids" button in Bulk Actions (see #3222), the previous search is lost once you submit a bulk action job.  This simply passes the existing search params on through the form submission used to create a new bulk action and then onto the redirect to create a new job.  This allows the search request to be preserved if you then create a new bulk action immediately and want to apply them to the same list of druids.

Localhost browser.  Though I did not that it seems pass along more params than just the search related ones, and I'm not sure why.  It does work though.

### DIFF
--- a/app/components/bulk_actions_form_component.html.erb
+++ b/app/components/bulk_actions_form_component.html.erb
@@ -2,6 +2,7 @@
   <h1>New Bulk Action</h1>
 
   <%= form_for(@form,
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
Please wait until #3229 is merged

peetucket:
Closed in favor of #3331

PR #3318, opened 2022-03-22
rename bulk edit to bulk action

### LINKED ISSUE #3315: Replace "Bulk Edits" with "Bulk Actions" in the Argo top menu link
We only have Bulk Actions now, so we should use that language everywhere. I think the Argo top menu bar is the only place left that says something else.

peetucket:

Fixes #3315 - rename bulk edit to bulk action

Once merged, please proceed directly to sul-dlss/infrastructure-integration-test#384 without passing go or collecting $200

Existing tests

### DIFF
--- a/spec/features/bulk_desc_metadata_download_spec.rb
+++ b/spec/features/bulk_desc_metadata_download_spec.rb
@@ -11,12 +11,25 @@ RSpec.describe 'Bulk Descriptive Metadata Download', js: true do
 
   it 'New page has a populate druids button and div with last search' do
     visit search_catalog_path q: 'stanford'
-    click_link 'Bulk Actions'
+    within '.search-widgets' do
+      click_link 'Bulk Actions'
+    end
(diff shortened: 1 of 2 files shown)


spec/features/bulk_desc_metadata_download_spec.rb
```diff
@@ -11,14 +11,27 @@
 
   it 'New page has a populate druids button and div with last search' do
     visit search_catalog_path q: 'stanford'
-    click_link 'Bulk Actions'
+    within '.search-widgets' do
+      click_link 'Bulk Actions'
+    end
     expect(page).to have_css 'h1', text: 'Bulk Actions'
```
peetucket, line 17: this link becomes ambiguous now, and we need to ensure we click the button above search results and not in the top nav

jcoyne (approved):

peetucket:
Note: I think the new test I added in this PR will fail until the bug in #3324 is fixed.

PR #3320, opened 2022-03-23
Add a way to use factories to create Cocina objects

jcoyne:

To reduce the amount of boilerplate in tests.

### DIFF
--- a/spec/components/show/item/overview_component_spec.rb
+++ b/spec/components/show/item/overview_component_spec.rb
@@ -7,20 +7,7 @@ RSpec.describe Show::Item::OverviewComponent, type: :component do
   let(:presenter) { instance_double(ArgoShowPresenter, document: doc, cocina: cocina, change_set: change_set, state_service: state_service) }
   let(:change_set) { ItemChangeSet.new(cocina) }
   let(:cocina) do
-    Cocina::Models::DRO.new(externalIdentifier: 'druid:bc234fg5678',
-                            type: Cocina::Models::ObjectType.document,
-                            label: 'my dro',
-                            version: 1,
-                            description: {
-                              title: [{ value: 'my dro' }],
-                              purl: 'https://purl.stanford.edu/bc234fg5678'
-                            },
-                            access: {},
-                            identification: {},
-                            structural: {},
-                            administrative: {
-                              hasAdminPolicy: 'druid:hv992ry2431'
-                            })
+    build_for_repository(:item)
   end
   let(:rendered) { render_inline(component) }
   let(:allows_modification) { true }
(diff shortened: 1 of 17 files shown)


spec/factories/apos.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :apo, class: 'Cocina::Models::RequestAdminPolicy' do
+  factory :persisted_apo, class: 'Cocina::Models::RequestAdminPolicy' do
```
justinlittman, line 4: Why is this `persisted_apo`? The name choice doesn't make sense to me.

spec/factories/collections.rb
```diff
@@ -25,4 +25,34 @@
...
+      )
+    end
+
+    admin_policy_id { 'druid:hv992ry2431' }
+    external_identifier { 'druid:bc123df4568' }
+
+    label { 'test object' }
+    type { Cocina::Models::ObjectType.collection }
+    title { 'my collection' }
+    purl { 'https://purl.stanford.edu/bc123df4568' }
```
justinlittman, line 56: Can the purl be derived from the `external_identifier`?

spec/factories/items.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :item, class: 'Cocina::Models::RequestDRO' do
+  factory :persisted_item, class: 'Cocina::Models::RequestDRO' do
```
justinlittman, line 4: For naming consistency / clarity, can we use `dro` instead of `item`?

spec/factories/apos.rb
justinlittman: Oh wait, this is actually saved to the datastore. The "request" threw me. Can you please add comments to make the distinction between persisted and not persisted clear?

spec/components/show/item/overview_component_spec.rb
```diff
@@ -7,20 +7,7 @@
...
-                              title: [{ value: 'my dro' }],
-                              purl: 'https://purl.stanford.edu/bc234fg5678'
-                            },
-                            access: {},
-                            identification: {},
-                            structural: {},
-                            administrative: {
-                              hasAdminPolicy: 'druid:hv992ry2431'
-                            })
+    build_for_repository(:item)
```
justinlittman, line 10: What is the distinction between `build_for_repository` and `create_for_repository`? Also, the "_for_repository" feels a bit odd. "_cocina" instead?

spec/factories/items.rb
justinlittman: Oh, I take that back. This does following naming conventions https://github.com/sul-dlss/cocina-models#usage-conventions

spec/support/build_strategy_repository_pattern.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class BuildStategyForRepositoryPattern
```
justinlittman, line 3: Typo in class name.

Class documentation?

justinlittman (commented):
Mostly naming suggestions / requests. Will it make sense to add to cocina models so that can be easily used in other repos? Or just copy to other repos?

jcoyne (commented):
The create_ puts it in DSA, the build_ just builds one in memory

justinlittman:
How does that create_/build_ relate to persisted_?

jcoyne:
@justinlittman `create` and `build` are the default functions from FactoryBot.  So this keeps that convention for naming.

spec/components/show/item/overview_component_spec.rb
jcoyne: We've had `create_for_repository` in this repo for 2 years. I think it's pretty good. It was copied from Hyrax: https://github.com/samvera/hyrax/blob/573f1ce2f9eea19b32bf9905b1725db63d5be730/spec/support/create_for_repository.rb

PR #3323, opened 2022-03-23
[SPIKE] Create local models so we can avoid cocina edits

jcoyne:


spec/jobs/set_source_ids_csv_job_spec.rb
```diff
@@ -17,46 +17,13 @@
...
-                           'label' => 'My Item3',
-                           'version' => 3,
-                           'type' => Cocina::Models::ObjectType.collection,
-                           'externalIdentifier' => druids[2],
-                           'access' => {},
-                           'administrative' => { hasAdminPolicy: 'druid:cg532dg5405' },
-                           'description' => { title: [{ value: 'Stored title' }], purl: 'https://purl.stanford.edu/dd111ff2222' },
-                           identification: { sourceId: 'sul:1234' }
-                         })
+    build(:item, id: druids[2])
```
jcoyne, line 26: This should be collection

ndushay:
@jcoyne is this still valid, or is it superseded by PR #3453

PR #3325, opened 2022-03-24
Display image props

### LINKED ISSUE #2925: Display image resource properties in content section
Add "Height" and "Width" (see #3170 for a recent example touching this code)

Depends on #2891

Mockup: https://projects.invisionapp.com/share/ZR1072MLCMEW#/screens/444558684

lwrubel:
Resolves #2925. Image height and width display for image files in the Content resources section.

Added and ran unit tests, deployed to QA. 

When the fileset has a type of image, the Height and Width columns display:
Screen Shot 2022-03-24 at 11 51 51 AM

When the image files do not have presentation metadata, the columns appear but values are blank. 
Screen Shot 2022-03-24 at 11 50 58 AM

When the fileset is not for an image, the columns do not appear.
Screen Shot 2022-03-24 at 11 53 33 AM

### DIFF
--- a/spec/components/contents/file_component_spec.rb
+++ b/spec/components/contents/file_component_spec.rb
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Contents::FileComponent, type: :component do
-  let(:component) { described_class.new(file: file, object_id: 'druid:kb487gt5106', viewable: true) }
+  let(:component) { described_class.new(file: file, object_id: 'druid:kb487gt5106', viewable: true, image: true) }
   let(:rendered) { render_inline(component) }
   let(:file) do
     instance_double(Cocina::Models::File,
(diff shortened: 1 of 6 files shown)


spec/components/contents/file_component_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Contents::FileComponent, type: :component do
-  let(:component) { described_class.new(file: file, object_id: 'druid:kb487gt5106', viewable: true) }
+  let(:component) { described_class.new(file: file, object_id: 'druid:kb487gt5106', viewable: true, image: true) }
```
jcoyne, line 6: Please add one context where image is true and one context where image is false.

spec/components/contents/resource_component_spec.rb
```diff
@@ -3,13 +3,56 @@
...
+                               publish: true,
+                               sdrPreserve: true,
+                               shelve: true
+                             },
+                             presentation: presentation)
+  end
+  let(:type) { 'https://cocina.sul.stanford.edu/models/resources/image' }
+  let(:presentation) { { height: 11_839, width: 19_380 } }
+
+  it 'renders the component' do
```
jcoyne, line 37: Can you wrap this in "with an image" context?

jcoyne (changes requested):

spec/components/contents/file_component_spec.rb
lwrubel: Right, I see how that would be good. Just added another spec to cover image: false. Let me know if you were thinking of something further.

spec/components/contents/resource_component_spec.rb
lwrubel: Added that (and did similar on `file_component_spec.rb`)

PR #3330, opened 2022-03-25
Bump sdr-client to 0.74.0

mjgiarlo:

To stay up-to-date with sdr-client releases.

CI + was successfully integration tested in stage


edsu (approved):
Onward and upward!

PR #3339, opened 2022-03-28
Download descriptive CSV from the show page

### LINKED ISSUE #3015: Add "Download CSV" of descriptive metadata for single-objects that outputs a spreadsheet
For single objects in Argo, add an option (in the red buttons) to download a spreadsheet containing the descriptive metadata for that object. The goal is to give people an editable file that can then be uploaded directly to the object. The upload CSV will be ticketed separately.

This is intended to replace direct XML editing and to be more user-friendly than editing the Cocina JSON.

jcoyne:
Fixes #3015

### DIFF
--- a/app/assets/stylesheets/show.scss
+++ b/app/assets/stylesheets/show.scss
@@ -27,9 +27,12 @@
 
   .argo-show-btn-group {
     .btn {
-      width: $button-width;
+      min-width: $button-width;
       margin-right: .5rem;
       margin-bottom: .5rem;
+      &.small {
+        @extend .small;
+      }
     }
   }
 
--- a/app/components/show/controls_component.html.erb
+++ b/app/components/show/controls_component.html.erb
@@ -13,10 +13,10 @@
       <%= manage_release %>
     <% end %>
     <div class="dropdown d-inline-block">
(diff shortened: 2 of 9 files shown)


justinlittman (approved):
Approve pending consideration of question.

arcadiafalcone:
The source ID should appear in the second column (after druid) with header "sourceId."

arcadiafalcone:
The header ordinal numbers should start with 1 instead of 0.

arcadiafalcone:
Yes, any changes I mention here should also apply to the bulk download.

mjgiarlo (approved):

arcadiafalcone:
Dates in structured values (date ranges) are not appearing in the output (ex. cc262xq5907). May be related to first record in spreadsheet having simple date only (testing with download of data in this collection: https://argo-stage.stanford.edu/view/druid:bz491mc6190).

arcadiafalcone:
(Bulk download) Group similar columns together in order - contributor2 columns should immediately follow contributor1 columns (ex. jn362bp1522 in sample set).

arcadiafalcone:
Sure, will do.

PR #3340, opened 2022-03-28
Ordinal keys should begin at 1 rather than zero

jcoyne:
See https://github.com/sul-dlss/argo/pull/3339#issuecomment-1080833153


jmartin-sul (approved):

ndushay (approved):
Looks good

PR #3345, opened 2022-03-28
Let with_index handle setting the initial value

mjgiarlo:

Straight-up refactor so that the methods we're calling do our heavy lifting.

CI


jcoyne (approved):

jcoyne:
Thanks @mjgiarlo !

PR #3346, opened 2022-03-28
Remove temporary PO prompt from production deploys

mjgiarlo:

@andrewjbtw says we can unhold this after some descriptive metadata changes drop. Not until then.

CI


jmartin-sul (approved):

mjgiarlo:
Connects to #3383

PR #3347, opened 2022-03-28
Create a bulk structural metadata upload

### LINKED ISSUE #3131: Create bulk action to upload structural metadata from multiple objects in CSV form
This should be nested under the "CSV" section of the bulk actions. The CSV should have the same structure as the single-object structural metadata CSV, with the addition of two columns:

- druid
- content type

This will make it possible to modify structural metadata (including content type) for any arbitrary set of item druids and help with bulk remediation. One big use case for this will be making items searchable that contain OCR but are not searchable because they were accessioned before we had full text search. 

This issue is paired with an issue for downloading a bulk action structural CSV so that users do not need to create one from scratch. See #3130 

**Additional information**

If any druids on the provided list are not items then the bulk action should report this to the log and skip over those druids while processing the CSV. 

If possible, we should check that the files in the upload CSV already exist in the current structural metadata for that druid. It is not possible to add or rename files using the structural metadata, so errors in the CSV will result in accessioning errors once the item is closed. However, it is permissible to remove files by removing them from the CSV.

jcoyne:
Fixes #3131


app/jobs/application_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ApplicationJob < ActiveJob::Base
```
edsu, line 3: I was curious what this new empty class is for given that there is already GenericJob.

app/jobs/import_structural_job.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class ImportStructuralJob < GenericJob
```
edsu, line 3: This looks so much better after the refactor!

edsu (approved):

app/jobs/application_job.rb
jcoyne: ApplicationJob is a modern rails convention. If it is missing it is generated when you do `bin/rails generate job MyNewJobName`

ndushay (approved):

PR #3361, opened 2022-03-30
Load technical metadata lazily

### LINKED ISSUE #3756: Store filenames for CSV uploads in the bulk action logs
Request from @arcadiafalcone for descriptive metadata spreadsheet uploads: would it be possible to put the filename for the input CSV into the bulk action log?

The replayable spreadsheet stores the filename for the uploaded spreadsheet in a log, so this would bring parity with that functionality. I asked about this when we were implementing the bulk actions and I remember there was some difficulty with keeping the filename but I can't remember the explanation. 

**Additional context**

If we can work this out for descriptive uploads then it would be useful to apply it to any bulk action CSV import.

jcoyne:

Fixes #3756

Avoid having a call to the technical metadata service block rendering of the show page.

locally

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -222,7 +222,13 @@ class CatalogController < ApplicationController
                                                     versions: object_client.version.inventory)
 
     @datastreams = object_client.metadata.datastreams
-    @techmd = TechmdService.techmd_for(druid: params[:id])
+
+    # If you have this token, it indicates you have view_metadata access to the object
+    @verified_token_with_expiration = Argo.verifier.generate(
+      { key: params[:id] },
+      expires_in: 1.hour,
+      purpose: :view_token
+    )
 
     respond_to do |format|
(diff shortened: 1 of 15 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -222,7 +222,13 @@ def show
                                                     versions: object_client.version.inventory)
 
     @datastreams = object_client.metadata.datastreams
-    @techmd = TechmdService.techmd_for(druid: params[:id])
+
+    # If you have this token, it indicates you have view_metadata access to the object
+    @verified_token_with_expiration = Argo.verifier.generate(
+      { key: params[:id] },
+      expires_in: 5.minutes,
```
justinlittman, line 229: I'd make this longer, like 30/60 minutes?

justinlittman (approved):

app/controllers/catalog_controller.rb
jcoyne: Good idea. I made it an hour.

PR #3371, opened 2022-03-31
Fix mapping for sort_field

### LINKED ISSUE #3368: [dor-services-app/stage] Cocina::RoundtripValidationError: Roundtripping of descriptive metadata unsuccessful. Expected {"title":[{"structuredValue":[{"structuredValue":[],"parallelValue":[],"groupedValue":[],"value":"FFIEC HMDA raw data 2003","type":"main title","identifier":[],"note":[],"appliesTo":[]},{"structuredValue":[],"parallelValue":[],"groupedValue":[],"value":"part 17","type":"part number","identifier":[],"note":[],"appliesTo":[]},{"structuredValue":[],"parallelValue":[],"groupedValue":[],"val...

jcoyne:
Fixes #3368
The prior mapping was causing a roundtrip validation error in DSA

Added an integration test.

### DIFF
--- a/spec/forms/serials_form_spec.rb
+++ b/spec/forms/serials_form_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe SerialsForm do
                            'access' => {},
                            'administrative' => { hasAdminPolicy: 'druid:cg532dg5405' },
                            'structural' => {},
-                           'identification' => {}
+                           identification: { sourceId: 'sul:123' }
                          })
   end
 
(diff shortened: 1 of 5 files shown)


spec/forms/serials_form_spec.rb
```diff
@@ -30,13 +30,13 @@
...
                 { value: '7', type: 'part number' },
                 { value: 'samurai', type: 'part name' }
-              ],
-              note: [
-                { value: '1990', type: 'date/sequential designation' }
               ]
             }
           ],
-          'purl' => "https://purl.stanford.edu/#{druid.delete_prefix('druid:')}"
+          note: [
```
ndushay, line 36: is it intended that you changed the level at which the note is present in the cocina object?

ndushay (changes requested):
I'm not sure this is going in the correct direction.  See comment I put on #3368.  Please confer with Arcadia, and please can we have a better description of the problem?  how this relates to a sort field is beyond me.

spec/forms/serials_form_spec.rb
jcoyne: Yes, that's the change.

jcoyne: Screen Shot 2022-03-31 at 2 24 34 PM

ndushay (approved):

PR #3374, opened 2022-03-31
Create a factory for a cocina object

jcoyne:
Per discussion on #3323 it was agreed to first make factories for cocina models. This is a minimal implementation of a factory that is useful in a test.


spec/components/show/item/overview_component_spec.rb
```diff
@@ -7,20 +7,7 @@
...
-                              title: [{ value: 'my dro' }],
-                              purl: 'https://purl.stanford.edu/bc234fg5678'
-                            },
-                            access: {},
-                            identification: { sourceId: 'sul:1234' },
-                            structural: {},
-                            administrative: {
-                              hasAdminPolicy: 'druid:hv992ry2431'
-                            })
+    Cocina::Models::Factories.build(:item)
```
justinlittman, line 10: For what it is worth, I'd suggest going with `dro` instead of `item`. It avoids the mental overhead of having to remember that an item is a DRO.

jmartin-sul (approved):

Issue #3364, opened 2022-03-30
Clarify label for descriptive metadata upload/download


arcadiafalcone:
"Upload CSV" in the manage description menu may cause confusion as users could try to upload a replayable spreadsheet instead.

astridu:
![image](https://user-images.githubusercontent.com/42216982/161173628-1ececb91-2769-4fff-8e51-a80bde879df5.png)

jcoyne:
When we say "cocina" we're typically talking about the JSON data model. I'm not sure these new labels clearly indicates that it is a spreadsheet that maps to cocina. This new phrasing may cause confusion in the future.

arcadiafalcone:
Does "spreadsheet" fit after "Cocina" in the label? We'll be referring to it in general as the "Cocina spreadsheet," to distinguish it from the "replayable spreadsheet."

andrewjbtw:
Works for me

astridu:
yo tambien

PR #3377, opened 2022-04-04
Import descriptive CSV

### LINKED ISSUE #3016: Add "Upload descMetadata" option for single-objects that inputs a spreadsheet
For single objects in Argo, add an option (in the red buttons) to upload a spreadsheet containing the descriptive metadata for that object. The download spreadsheet option has been ticketed as https://github.com/sul-dlss/argo/issues/3015 

This is intended to replace direct XML editing and to be more user-friendly than editing the Cocina JSON.

jcoyne:

Fixes #3016


app/services/description_import.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def nest_hashes(value, *keys)
+    return value if keys.empty?
+
+    key = keys.shift
+    val = keys.empty? ? value : nest_hashes(value, *keys)
+    key.is_a?(Integer) ? [].tap { |arr| arr[key] = val } : { key => val }
+  end
+
+  def visit(what, rest, value, path = [])
```
lwrubel, line 43: I'm unclear on what `what` and `rest` would be. Are those variable naming conventions we use in SDR?

jcoyne: In this case `what` is a tree like data structure.  It's one of the nodes in the Cocina descriptive.  `rest` is the "path" we have to follow to get to the location in the tree.

app/controllers/descriptives_controller.rb
```diff
@@ -17,8 +39,13 @@ def show
...
-  def load_cocina
-    @cocina = Dor::Services::Client.object(params[:item_id]).find
+  def load_and_authorize_cocina
+    @cocina = dor_client.find
+    authorize! :manage_item, @cocina
+  end
+
+  def dor_client
+    Dor::Services::Client.object(params[:item_id])
   end
```
peetucket, line 49: if these methods are going to be repeated in other controllers (like here: https://github.com/sul-dlss/argo/blob/main/app/controllers/structures_controller.rb#L32-L39), thoughts on moving to the application controller to avoid duplication?

app/services/description_import.rb
```diff
@@ -0,0 +1,68 @@
...
+    params = @description.to_h
+
+    row = @csv.first
+    # The source_id is only there for the user to reference and should be ignored for data processing
+    headers = row.headers.excluding('source_id')
+
+    headers.each do |address|
+      split_address = address.scan(/[[:alpha:]]+|[[:digit:]]+/)
+                             .map { |item| /\d+/.match?(item) ? item.to_i - 1 : item.to_sym }
+      visit(params, split_address, row[address])
```
peetucket, line 25: I'm not sure what this bit of code is for, what is split_address ?

app/controllers/descriptives_controller.rb
jcoyne: item_id isn't always the parameter to use. I find that having this in the class makes it easier to reason about.  This class isn't very long and there isn't much chance of getting this call wrong.

app/services/description_import.rb
lwrubel: I'm not sure I would understand that from these names--aybe a comment would help, if there aren't other names for these variables that would convey that.

jcoyne: I'm not great with names, do you have suggestions that would make it clearer?

jcoyne: I've added yard doc to document the method parameters.

app/components/show/controls_component.html.erb
```diff
@@ -43,6 +43,9 @@
                                      data: { turbo_method: 'post' } %></li>
         <% end %>
         <li><%= link_to 'Download CSV', item_descriptive_path(doc, format: :csv), class: 'dropdown-item' %></li>
+        <li><%= link_to 'Upload CSV', new_item_descriptive_path(doc),
+                                      class: 'dropdown-item',
```
peetucket, line 47: based on #3364 (which I just happen to pick up), do you want to just rename both options while you are here in this PR: "Download Cocina spreadsheet" and "Upload Cocina spreadsheet"

jcoyne: #3364 should be marked as blocked and should not be worked on yet.

app/services/description_import.rb
lwrubel: Thanks, that helps. I was going to suggest `loc` or `location` for `rest` and `tree` for `what`.

app/controllers/descriptives_controller.rb
jcoyne: I've addressed this duplication in a more systematic way via https://github.com/sul-dlss/argo/pull/3406

app/services/description_import.rb
peetucket: perhaps adding this a code comment will help a future developer (i.e. show example input from csv and resulting parsed output)

PR #3382, opened 2022-04-05
[refactor] Use the existing methods for ensuring an item is modifiable

jcoyne:
So that we handle this state consistently.

### DIFF
--- a/app/controllers/content_types_controller.rb
+++ b/app/controllers/content_types_controller.rb
@@ -11,15 +11,15 @@ class ContentTypesController < ApplicationController
 
   # set the content type in the content metadata
   def update
-    authorize! :manage_item, @cocina_object
+    authorize! :manage_item, @cocina
 
     # if this object has been submitted and doesnt have an open version, they cannot change it.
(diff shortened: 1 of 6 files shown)


app/controllers/content_types_controller.rb
```diff
@@ -11,15 +11,15 @@ def show
 
   # set the content type in the content metadata
   def update
-    authorize! :manage_item, @cocina_object
+    authorize! :manage_item, @cocina
```
jmartin-sul, line 14: what was setting up `@cocina_object` before?  i can see how `@cocina` gets set up, just wondering how the old instance variable was working (was curious since i didn't notice its setup being removed in this PR).

jmartin-sul (approved):
~~one "how did it work before type question?" but LGTM~~ nm, bad read on my part, LGTM, no questions, merged

app/controllers/content_types_controller.rb
jmartin-sul: oh nm, bad read on my part

Issue #3211, opened 2022-03-08
Update content type menu to handle reading order for images


andrewjbtw:
While mapping structural metadata, we learned that the "image" content type supports reading order. Previously, we thought it was only supported for books. This means that we need to update the content type menu to handle reading order for both books and images. Right now, you can set reading order for books from the menu but you have to edit XML to set it for images.

Since reading order applies only to book or image, the ideal interaction would be:

- User selects book or image, new menu for reading order appears on the page
- User selects any other content type, reading order menu either does not appear or it disappears

I'll work with Astrid on a design.

andrewjbtw:
Adding Astrid's design and more details about the desired interaction:

- When the user selects "book" or "image", show an additional menu to select the reading order (see below for positioning)
- Show only one choice for "book" now - no need to have "book (rtl)" and "book (ltr)" anymore
- Do not show this menu or set a reading order for any content type other than book or image

<img width="964" alt="reading-order" src="https://user-images.githubusercontent.com/8975583/159968938-d07d86f5-ff21-4480-b840-49c5c38ba47f.png">

This ticket is for the show page menu only. Additional tickets will be made to update:

- bulk action content type menu (which is the same structure as this menu)
- registration form (a different page layout)
- preassembly content type selection (possibly not in scope for this workcycle)

andrewjbtw:
I think the help text needs updating but I'll file that separately because it's probably going to be a full re-write.

justinlittman:
@andrewjbtw This change can easily be made for all locations in Argo by replacing "image" with "image (rtl)" and "image (ltr)". (They are all fed from the same list of content types so the code changes are minimal.)

Would you like to proceed with that approach or the current design (which requires quite a bit more code changes and additional follow-up tickets)?

justinlittman:
The decision was to proceed with the current design.

PR #3388, opened 2022-04-05
Update content type menu to handle reading order for images

jcoyne:

Fixes https://github.com/sul-dlss/argo/issues/3211

Automated test suite

### DIFF
--- a/app/javascript/controllers/content_type_controller.js
+++ b/app/javascript/controllers/content_type_controller.js
@@ -1,0 +1,34 @@
+import { Controller } from 'stimulus'
+
+export default class extends Controller {
+    static targets = [ "direction", "directionRow", "contentType"]
+
+    connect() {
+      this.render()
+    }
+
+    render() {
+      switch (this.currentType()) {
+        case 'https://cocina.sul.stanford.edu/models/image':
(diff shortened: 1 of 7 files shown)


app/javascript/controllers/content_type_controller.js
```diff
@@ -0,0 +1,34 @@
...
+    static targets = [ "direction", "directionRow", "contentType"]
+
+    connect() {
+      this.render()
+    }
+
+    render() {
+      switch (this.currentType()) {
+        case 'https://cocina.sul.stanford.edu/models/image':
+        case 'https://cocina.sul.stanford.edu/models/book':
```
peetucket, line 13: image types can have a reading direction set too?

peetucket (approved):
Just a question about reading direction for images

app/javascript/controllers/content_type_controller.js
jcoyne: Yes, that's exactly the change requested in the ticket.

andrewjbtw: I was surprised but we found it in existing data and the viewer was parsing it correctly.

PR #3393, opened 2022-04-06
Avoid public_send as using it with untrusted input could lead to a DOS

jcoyne:
Avoid a DOS using a malicious upload.

Test suite.

### DIFF
--- a/app/services/structure_updater.rb
+++ b/app/services/structure_updater.rb
@@ -26,7 +26,9 @@ class StructureUpdater
     # 1. Ensure all files in the csv are present in the existing object
     csv.each.with_index(2) do |row, index|
       errors << "On row #{index} found #{row['filename']}, which appears to be a new file" unless existing_files_by_filename.key?(row['filename'])
-      errors << "On row #{index} found \"#{row['resource_type']}\", which is not a valid resource type" unless Cocina::Models::FileSetType.respond_to?(row['resource_type'])
+      unless Cocina::Models::FileSetType.properties.key?(row['resource_type'].to_sym)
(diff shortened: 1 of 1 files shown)


app/services/structure_updater.rb
```diff
@@ -26,7 +26,9 @@ def validate
     # 1. Ensure all files in the csv are present in the existing object
     csv.each.with_index(2) do |row, index|
       errors << "On row #{index} found #{row['filename']}, which appears to be a new file" unless existing_files_by_filename.key?(row['filename'])
-      errors << "On row #{index} found \"#{row['resource_type']}\", which is not a valid resource type" unless Cocina::Models::FileSetType.respond_to?(row['resource_type'])
+      unless Cocina::Models::FileSetType.properties.key?(row['resource_type'].to_sym)
```
mjgiarlo, line 29: Will this handle 3D ok?

jcoyne: Yes

mjgiarlo: @jcoyne Sorry, my question is really more like "what values are using entering for 3D objects?" If "three_dimensional," that used to work and would now be broken. If "3d," that is broken now and will be fixed by this PR. Not something I'd block the PR over, but I am curious.

jcoyne: I don't know what values would be entered. I assumed they would use `"3d"`

mjgiarlo: @jcoyne w00t, then it's both a security enhancement and a bugfix!

PR #3398, opened 2022-04-06
Add a shorthand way of creating cocina-models fixtures

jcoyne:

Less typing. Easier to read.


lib/cocina/models/factories.rb
```diff
@@ -3,6 +3,12 @@
 module Cocina
   module Models
     class Factories
+      module Methods
+        def build(...)
+          Factories.build(...)
```
mjgiarlo, line 8: Nice >= 2.7 idiom!

mjgiarlo (approved):

PR #3406, opened 2022-04-07
[refactor] add a repository to encapsulate the find/store function 

jcoyne:
… of the DSA endpoint

I think this makes the code easier to read and it looks a lot more like a "regular" Rails app

### DIFF
--- a/app/controllers/apo_controller.rb
+++ b/app/controllers/apo_controller.rb
@@ -3,7 +3,7 @@
 class ApoController < ApplicationController
   include Blacklight::FacetsHelperBehavior # for facet_configuration_for_field
 
-  before_action :create_obj, except: %i[
+  before_action :load_resource, except: %i[
     new
     create
     spreadsheet_template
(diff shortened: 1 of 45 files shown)


app/controllers/apo_controller.rb
```diff
@@ -123,9 +123,7 @@ def solr_conn
     @solr_conn ||= blacklight_config.repository_class.new(blacklight_config).connection
   end
 
-  def create_obj
-    raise 'missing druid' unless params[:id]
-
-    @cocina = maybe_load_cocina(params[:id])
+  def load_resource
+    @cocina = Repository.find(params[:id])
   end
```
peetucket, line 128: this makes me happy

peetucket (approved):
I like all these changes, makes the code much cleaner.  Only question is the name `#store` instead of `#save` (save seems more Rails like).

jcoyne:
@peetucket I considered `.save`, but that seems odd because in rails you tell the object to save itself (e.g. `item.save`) and in this case you are saying "store the thing I'm passing to you"

PR #3412, opened 2022-04-07
Collapse the datastream section by default

jcoyne:

This helps to encourage users to look elsewhere for the information they need. This will assist the transition to datastreams going away.

This came up in a conversation I had with Andrew.

Locally


edsu:
Good catch on the test -- I did a quick grep through the integration tests and didn't see anything?

PR #3414, opened 2022-04-07
add rake tasks for seed data and update readme

peetucket:

- Restructure the README to be clearer on how to do local development.
- Add a script to start the bin/rails server with appropriate env variables.
- Add rake tasks for seeding solr for local development and for creating items.

Localhost

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,22 @@ bundle install
 
 Note that `bundle install` may complain if MySQL isn't installed.  You can either comment out the `mysql2` inclusion in `Gemfile` and come back to it later (you can develop using `sqlite3`), or you can install MySQL.
 
+Install foreman for local development (foreman is not supposed to be in the Gemfile, See this [wiki article](https://github.com/ddollar/foreman/wiki/Don't-Bundle-Foreman)):
+
+```bash
+bundle install foreman
+```
+
+## Local Development TLDR
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -70,15 +70,22 @@ CI runs a series of steps;  this the sequence to do it locally, along with some
     bin/rake
     ```
```
peetucket, line 71: Note: The diffs view makes it hard to see what changed, since some stuff just got moved around.  May be easier to just view the whole file to see what it looks like now.

jcoyne (approved):

README.md
```diff
@@ -90,74 +113,90 @@ If you want to run background jobs, which are necessary for spreadsheet bulk upl
...
+bin/dev
+```
 
-It can be useful when debugging to see the local rails server output in realtime and pause with 'byebug'.  You can do
-this while running in the app in the web container.  First stop any existing web container (if running already):
+Most of the time (unless you already have data), you will want seed data and a single item.  Note tha all servers must be running first, including web, and this will clear solr:
 
 ```
-docker compose stop web
+rake argo:seed_data
```
jcoyne, line 134: You probably want `bin/rake` here so it's in the bundle context

peetucket: fixed

PR #3415, opened 2022-04-07
[SPIKE] bin/dev and docker

edsu:

This PR modifies the docker-compose web service to use bin/dev as an entrypoint. This means that JS and CSS assets will get compiled automatically when the files change. Hopefully it should make getting set up to develop Argo a little bit easier.

Local development environment.

The docker configuration isn't used outside of development.


jcoyne (approved):
Looks good to me. Thank you for taking this on.

PR #3418, opened 2022-04-07
ensure uploaded csv does not change preserve from no to yes

### LINKED ISSUE #3400: Structural metadata CSV upload should not allow "preserve=no" to be changed to "preserve=yes"
The structural metadata upload CSV only supports working with files that already exist in Stacks or Preservation, and the only way to add new files to Preservation is to run accessioning from one of the accessioning applications (Preassembly, Goobi, H2, etc.).

When uploading structural metadata in Argo, we should not accept a change from "preserve=no" to "preserve=yes". What would happen if someone entered this change is that the accessionWF would look for the file in `/dor/workspace` and try to send it to Stacks. But since the file wouldn't be in `/dor/workspace` without having been staged there first, the workflow would halt with an error.

**Related**

For files that are already on Stacks, one way around this would be to support copying the existing file from Stacks into the `/dor/workspace` so that it can be sent to preservation. I've ticketed that here: https://github.com/sul-dlss/common-accessioning/issues/853 Unless/until we support that, it would be better to prevent this change to the preserve setting entirely.

peetucket:

Fixes #3400 - validate uploaded structural metadata in CSV to ensure the user does not change an existing file's preserve value from no to yes

Added new test

### DIFF
--- a/spec/services/structure_updater_spec.rb
+++ b/spec/services/structure_updater_spec.rb
@@ -204,9 +204,9 @@ RSpec.describe StructureUpdater do
       <<~CSV
         resource_label,resource_type,sequence,filename,file_label,publish,shelve,preserve,rights_access,rights_download,mimetype,role
         Image 1,image,1,bb045jk9908_0001.tiff,bb045jk9908_0001.tiff,yes,yes,yes,stanford,none,image/one,
-        Image 1,image,1,bb045jk9908_0001.jp2,bb045jk9908_0001.jp2,yes,yes,yes,world,world,image/two,transcription
+        Image 1,image,1,bb045jk9908_0001.jp2,bb045jk9908_0001.jp2,yes,yes,no,world,world,image/two,transcription
         Image 2,image,2,bb045jk9908_0002.tiff,bb045jk9908_0002.tiff,yes,yes,yes,stanford,none,image/three,
-        Image 2,image,2,bb045jk9908_0002.jp2,bb045jk9908_0002.jp2,yes,yes,yes,world,world,image/four,
+        Image 2,image,2,bb045jk9908_0002.jp2,bb045jk9908_0002.jp2,yes,yes,no,world,world,image/four,
       CSV
     end
 
@@ -217,9 +217,9 @@ RSpec.describe StructureUpdater do
       new_file_routing = new_files.map { |file| [file.administrative.publish, file.administrative.shelve, file.administrative.sdrPreserve] }
       expect(new_file_routing).to eq [
         [true, true, true],
+        [true, true, false],
         [true, true, true],
-        [true, true, true],
-        [true, true, true]
+        [true, true, false]
       ]
 
       new_file_mime = new_files.map(&:hasMimeType)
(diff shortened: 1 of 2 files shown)


spec/services/structure_updater_spec.rb
```diff
@@ -204,9 +204,9 @@
       <<~CSV
         resource_label,resource_type,sequence,filename,file_label,publish,shelve,preserve,rights_access,rights_download,mimetype,role
         Image 1,image,1,bb045jk9908_0001.tiff,bb045jk9908_0001.tiff,yes,yes,yes,stanford,none,image/one,
-        Image 1,image,1,bb045jk9908_0001.jp2,bb045jk9908_0001.jp2,yes,yes,yes,world,world,image/two,transcription
+        Image 1,image,1,bb045jk9908_0001.jp2,bb045jk9908_0001.jp2,yes,yes,no,world,world,image/two,transcription
         Image 2,image,2,bb045jk9908_0002.tiff,bb045jk9908_0002.tiff,yes,yes,yes,stanford,none,image/three,
-        Image 2,image,2,bb045jk9908_0002.jp2,bb045jk9908_0002.jp2,yes,yes,yes,world,world,image/four,
+        Image 2,image,2,bb045jk9908_0002.jp2,bb045jk9908_0002.jp2,yes,yes,no,world,world,image/four,
       CSV
```
peetucket, line 210: note: our existing test data would have violated this new constraint, so i changed it so it wouldn't

spec/services/structure_updater_spec.rb
```diff
@@ -240,9 +240,9 @@
       <<~CSV
         resource_label,resource_type,sequence,filename,file_label,publish,shelve,preserve,rights_access,rights_download,mimetype,role
         Picture 1,object,1,bb045jk9908_0001.tiff,bb045jk9908_0001.tiff,yes,yes,yes,stanford,none,image/tiff,
-        Picture 1,object,1,bb045jk9908_0001.jp2,bb045jk9908_0001.jp2,yes,yes,yes,world,world,image/jp2,
+        Picture 1,object,1,bb045jk9908_0001.jp2,bb045jk9908_0001.jp2,yes,yes,no,world,world,image/jp2,
         Picture 2,page,2,bb045jk9908_0002.tiff,bb045jk9908_0002.tiff,yes,yes,yes,stanford,none,image/tiff,
-        Picture 2,page,2,bb045jk9908_0002.jp2,bb045jk9908_0002.jp2,yes,yes,yes,world,world,image/jp2,
+        Picture 2,page,2,bb045jk9908_0002.jp2,bb045jk9908_0002.jp2,yes,yes,no,world,world,image/jp2,
       CSV
     end
```
peetucket, line 247: same comment as above, if not changed, we would get a validation error with this data

spec/services/structure_updater_spec.rb
```diff
@@ -281,4 +281,22 @@
...
+
+  context 'with switching preservation from no to yes for a file in the csv' do
+    let(:csv) do
+      <<~CSV
+        resource_label,resource_type,sequence,filename,file_label,publish,shelve,preserve,rights_access,rights_download,mimetype,role
+        Picture 1,object,1,bb045jk9908_0001.tiff,bb045jk9908_0001.tiff,yes,yes,yes,stanford,none,image/tiff,
+        Picture 1,object,1,bb045jk9908_0001.jp2,bb045jk9908_0001.jp2,yes,yes,no,world,world,image/jp2,
+        Picture 2,page,2,bb045jk9908_0002.tiff,bb045jk9908_0002.tiff,yes,yes,yes,stanford,none,image/tiff,
+        Picture 2,page,2,bb045jk9908_0002.jp2,bb045jk9908_0002.jp2,yes,yes,yes,world,world,image/jp2,
+      CSV
```
peetucket, line 293: this file starts with preserve == no in the original cocina

jcoyne (approved):

PR #3426, opened 2022-04-08
Remove pid file when starting up rails dev server

edsu:

If you start the containers with `docker compose up` and then stop them
with ctrl-c or a `docker compose stop` the tmp/pids/server.pid in the
working directory (shared with the docker container) is not deleted.
This means when you try to start it back up again you get an error about
the server already running.

I suspect there's a better way to do this, but this small change will remove the
server.pid before starting up `rails server`.

Starting and stopping my local Argo Docker development environment.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,8 @@ services:
       SETTINGS__WORKFLOW_URL: http://workflow:3000
       SETTINGS__TECH_MD_SERVICE__URL: http://techmd:3000
       SOLR_URL: http://solr:8983/solr/argo
+      # Docker doesn't allow rails server to remove PID on shutdown
+      PID_FILE: /dev/null
     depends_on:
       - dor-indexing-app
     # To allow yarn --watch from Procfile.dev to continue after stdin is closed


Procfile.dev
```diff
@@ -1,3 +1,3 @@
-web: bin/rails server -b 0.0.0.0 -p 3000
+web: rm -f tmp/pids/server.pid & bin/rails server -b 0.0.0.0 -p 3000
```
jcoyne, line 1: Can you move this to the dockerfile instead?   This pid check is useful locally.

jcoyne: I think you can do `     PIDFILE: /dev/null` in the environment variable section of docker-compose.yml

edsu: I did try to put it there at first, but it didn't work as:

    RUN rm -f /app/tmp/pids/server.pid

I guess because that just runs once when the container is being built! Maybe if I put it into the entry point it will take though? I'll give that a try.

edsu: `PIDFILE: /dev/null` worked nicely!

jcoyne (approved):

PR #3427, opened 2022-04-08
Add a bulk job for descriptive metadata import

### LINKED ISSUE #3190: Create new bulk descriptive spreadsheet upload that works directly with Cocina
The bulk action to upload descriptive metadata currently converts data from a spreadsheet into MODS, or takes MODS XML directly. This MODS is then inserted into the descriptive metadata for each item in the batch. As we are moving away from MODS and XML, we should create a new spreadsheet upload that works directly with Cocina and does not go through a MODS conversion on input.

@arcadiafalcone will supply the form the spreadsheet download should take.

This spreadsheet upload will go in the Bulk actions menu under "Modify data via CSV". We will eventually retire the current system where users have to go to specific APOs to make MODS uploads.

jcoyne:

Fixes #3190

### DIFF
--- a/app/models/bulk_action.rb
+++ b/app/models/bulk_action.rb
@@ -3,37 +3,7 @@
 # Stores data about an asynchonous background job
 class BulkAction < ApplicationRecord
   belongs_to :user
-  validates :action_type,
-            inclusion: {
-              in: %w[GenericJob
-                     AddWorkflowJob
-                     ApplyApoDefaultsJob
-                     DescmetadataDownloadJob
-                     DescriptiveMetadataExportJob
-                     ReleaseObjectJob
-                     RemoteIndexingJob
-                     PurgeJob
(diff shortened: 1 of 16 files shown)


app/models/bulk_action.rb
```diff
@@ -3,37 +3,7 @@
...
-                     ImportStructuralJob
-                     RegisterDruidsJob
-                     SetLicenseAndRightsStatementsJob
-                     SetSourceIdsCsvJob
-                     SetContentTypeJob
-                     ManageEmbargoesJob
-                     SetCollectionJob
-                     SetRightsJob]
-            }
+  validates :action_type, inclusion: { in: GenericJob.descendants.map(&:to_s) }
```
aaron-collier, line 6: 👍

aaron-collier (approved):

PR #3429, opened 2022-04-08
add rights_location, rename rights_access, add validation for location

### LINKED ISSUE #3327: Include access location in the structural CSV (import and export) -- and rename "rights_access" to "rights_view"
Updating the ticket for two requests:

1. add "rights_location" to the structural CSV

We need to support including a location in the structural metadata CSV. Currently you can choose "location-based" but there's no column for the actual location.

I had an idea about supporting this with just two columns but now that we're going to three selections for the access rights menu it will be more consistent to have three columns.

2. rename "rights_access" to "rights_view" since we've renamed that Cocina field.

peetucket:

Fixes #3327 - add a new `rights_location` column to structural spreadsheet and rename `rights_access` column to `rights_view`

Note: I also added an additional validation to ensure the new `rights_location` column has a value if either `rights_view` or `rights_download` contains a value of `location-based`.  

Question: is there an easy way to verify the location specified is an allowed value in the enum here: https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L318-L324 

Updated existing tests to show location can be set, and to account for new column and renamed column, and added a new test for location validation.

### DIFF
--- a/app/services/structure_updater.rb
+++ b/app/services/structure_updater.rb
@@ -23,20 +23,34 @@ class StructureUpdater
   # @return [Bool] true if there are no problems
   def validate
     @errors = []
-    # Ensure all files in the csv are present in the existing object and ensure no existing files change preserve from no to yes
     csv.each.with_index(2) do |row, index|
-      if !existing_files_by_filename.key?(row['filename'])
-        errors << "On row #{index} found #{row['filename']}, which appears to be a new file"
-      elsif !existing_files_by_filename[row['filename']].administrative.sdrPreserve && row['preserve'] == 'yes'
-        errors << "On row #{index} found #{row['filename']}, which changed preserve from no to yes, which is not supported"
-      end
-      # ensure all supplied resource types are valid
-      unless Cocina::Models::FileSetType.properties.key?(row['resource_type'].to_sym)
-        errors << "On row #{index} found \"#{row['resource_type']}\", which is not a valid resource type"
-      end
+      errors << "On row #{index} found #{row['filename']}, which appears to be a new file" unless existing_file?(row)
+      errors << "On row #{index} found #{row['filename']}, which changed preserve from no to yes, which is not supported" if invalid_preservation_change?(row)
+      errors << "On row #{index} found #{row['filename']}, which set view or download rights to location-based but did not specify a location" if invalid_location_rights?(row)
+      errors << "On row #{index} found \"#{row['resource_type']}\", which is not a valid resource type" if invalid_resource_type?(row)
     end
     csv.rewind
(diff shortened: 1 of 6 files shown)


app/services/structure_updater.rb
```diff
@@ -23,22 +23,36 @@ def initialize(model, csv)
...
+  def invalid_location_rights?(row)
+    existing_file?(row) && [row['rights_view'], row['rights_download']].include?('location-based') && row['rights_location'].blank?
+  end
+
+  # Ensure all supplied resource types are valid
+  def invalid_resource_type?(row)
+    !Cocina::Models::FileSetType.properties.key?(row['resource_type'].to_sym)
+  end
+
   def existing_files_by_filename
```
peetucket, line 56: some refactoring of validation .... Rubucop was unhappy with the complexity of the single method when I added a new condition

app/services/structure_updater.rb
```diff
@@ -23,22 +23,36 @@ def initialize(model, csv)
...
-      if !existing_files_by_filename.key?(row['filename'])
-        errors << "On row #{index} found #{row['filename']}, which appears to be a new file"
-      elsif !existing_files_by_filename[row['filename']].administrative.sdrPreserve && row['preserve'] == 'yes'
-        errors << "On row #{index} found #{row['filename']}, which changed preserve from no to yes, which is not supported"
-      end
-      # ensure all supplied resource types are valid
-      unless Cocina::Models::FileSetType.properties.key?(row['resource_type'].to_sym)
-        errors << "On row #{index} found \"#{row['resource_type']}\", which is not a valid resource type"
-      end
+      errors << "On row #{index} found #{row['filename']}, which appears to be a new file" unless existing_file?(row)
```
jcoyne, line 27: Is it possible to do `next` if the file doesn't exist? Then you don't need to repeat this check in all the other validations?

jcoyne (changes requested):

app/services/structure_updater.rb
peetucket: I had the same thought and did it at first, but then you wouldn't get follow on errors for a missing file like the test expectation shows: https://github.com/sul-dlss/argo/blob/main/spec/services/structure_updater_spec.rb#L280   But perhaps that's not important and we can just adjust the test.

jcoyne: What if we move the invalid_resource_type check up to the top?

peetucket: Yup, that works.  Adjusted the code and test (which now reverses the order of the error message).

jcoyne (approved):

PR #3430, opened 2022-04-08
Add Generate Tracking sheets bulk action

### LINKED ISSUE #3416: Create new "Generate tracking sheets" bulk action
Note: This is related to #3203 . I'll explain how below the description of the feature.

This request is for a new bulk action that will:

- Take a list of druids as the only input
- Generate a tracking sheet PDF for those druids, downloadable in the way other reports are downloadable
- Name of bulk action: "Generate tracking sheets"
- Location: under the "Reports" section of the bulk actions menu

**What are tracking sheets?**

Tracking sheets are PDFs that can be generated from the registration form by following these steps:

- Register druids using the form
- While the list of new druids is still visible, click on the "tracking sheets" button
tracking-sheets

The tracking sheets contain barcodes and essential pieces of information (source id, tags, etc.) about each druid, one druid per page. The digitization labs print these out and use them to track physical items through the digitization process.

**Why a tracking sheets bulk action?**
 
The only way to generate tracking sheets right now is to use the registration form. If you need to generate tracking sheets for druids that have already been registered, you can 

- paste a list of druids into the "druid" column of the registration form and then 
- click the tracking sheets button (as if you just registered those druids)

This is a weird use of the registration form, since druids are supposed to be auto-generated and you shouldn't supply any data in the druid column. We could prevent people from supplying druids on the registration form if we had an alternative way to get to the tracking sheets. This is the connection to #3203 - the form makes it seem like you could pre-supply a druid for registration, but that doesn't actually work.

aaron-collier:

Fixes #3416 

Updated unit test

### DIFF
--- a/app/jobs/tracking_sheet_report_job.rb
+++ b/app/jobs/tracking_sheet_report_job.rb
@@ -29,7 +29,7 @@ class TrackingSheetReportJob < GenericJob
         error_message = "#{Time.current} TrackingSheetReportJob creation failed #{e.class} #{e.message}"
         log.puts(error_message) # this one goes to the user via the bulk action log
         logger.error(error_message) # this is for later archaeological digs
-        Honeybadger.context(bulk_action_id: bulk_action_id, params: params)
+        Honeybadger.context(bulk_action_id:, params:)
         Honeybadger.notify(error_message) # this is so the devs see it ASAP
       end
     end
@@ -49,7 +49,7 @@ class TrackingSheetReportJob < GenericJob
 
     doc = nil
     begin
-      doc = find_or_create_in_solr_by_id(druid)
+      doc = find_by_id(druid)
     rescue StandardError
       pdf.text "DRUID #{druid} not found in index", size: 15, style: :bold, align: :center
       return pdf
@@ -126,12 +126,8 @@ class TrackingSheetReportJob < GenericJob
   # @return [Hash] doc from Solr or to_solr
   # @note To the extent we use Solr input filters or copyField(s), the Solr version will differ from the to_solr hash.
   # @note That difference shouldn't be important for the few known fields we use here.
-  def find_or_create_in_solr_by_id(druid)
+  def find_by_id(druid)
     namespaced_druid = Druid.new(druid).with_namespace
-    doc = SearchService.query(%(id:"#{namespaced_druid}"), rows: 1)['response']['docs'].first
-    return doc unless doc.nil?
-
-    Argo::Indexer.reindex_druid_remotely(namespaced_druid)
     SearchService.query(%(id:"#{namespaced_druid}"), rows: 1)['response']['docs'].first
   end
 
--- a/app/jobs/tracking_sheet_report_job.rb
+++ b/app/jobs/tracking_sheet_report_job.rb
@@ -2,7 +2,6 @@
 
 ##
 # Job to create downloadable tracking sheets
-# rubocop:disable Metrics/ClassLength
 class TrackingSheetReportJob < GenericJob
   ##
   # A job that...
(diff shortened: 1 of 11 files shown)


app/jobs/tracking_sheet_report_job.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    druids = params[:druids]
+
+    with_bulk_action_log do |log|
+      update_druid_count
+      begin
+        pdf = Prawn::Document.new(page_size: [5.5.in, 8.5.in])
+        pdf.font('Courier')
+        druids.each_with_index do |druid, i|
+          find_or_create_in_solr_by_id(druid)
```
jcoyne, line 23: why do we do this here if we do it again in generate_tracking_sheet?

spec/jobs/tracking_sheet_report_job_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    allow(Ability).to receive(:new).and_return(ability)
+  end
+
+  after do
+    FileUtils.rm_rf(output_directory) if Dir.exist?(output_directory)
+  end
+
+  describe '#perform_now' do
+    context 'with authorization' do
+      let(:ability) { instance_double(Ability, can?: true) }
```
jcoyne, line 30: I don't see your job checking authorization anywhere, but you're testing it here?

aaron-collier: Copy pasta, will remove.

app/jobs/tracking_sheet_report_job.rb
aaron-collier: I was refactoring to remove some duplication and missed this. Will fix.

app/jobs/tracking_sheet_report_job.rb
```diff
@@ -0,0 +1,142 @@
...
+  # rubocop:enable Metrics/PerceivedComplexity
+
+  # @param [String] druid unqualified DRUID identifier
+  # @return [Hash] doc from Solr or to_solr
+  # @note To the extent we use Solr input filters or copyField(s), the Solr version will differ from the to_solr hash.
+  # @note That difference shouldn't be important for the few known fields we use here.
+  def find_or_create_in_solr_by_id(druid)
+    namespaced_druid = Druid.new(druid).with_namespace
+    doc = SearchService.query(%(id:"#{namespaced_druid}"), rows: 1)['response']['docs'].first
+    return doc unless doc.nil?
```
jcoyne, line 132: I don't understand this use case. Can you explain it to me?  Why wouldn't it be found in solr? Why is this the right place to fix it?

jcoyne (changes requested):

app/jobs/tracking_sheet_report_job.rb
aaron-collier: This was a straight copy of the existing code used in the registration form, so I can't answer that. it is however 6 yo code (https://github.com/sul-dlss/argo/commit/ee0a7be508fde4c28d26dbb2e333f435420ab6b4)... if we'd like to refactor that, I'm happy to do so.

jcoyne: I think that code makes sense in registration, when the object may have not been created, but I don't think we need it here.

app/jobs/tracking_sheet_report_job.rb
```diff
@@ -0,0 +1,138 @@
...
+  # @param [hash] params additional parameters that an Argo job may need
+  # @option params [Array] :druids required list of druids
+  def perform(bulk_action_id, params)
+    super
+
+    druids = params[:druids]
+
+    with_bulk_action_log do |log|
+      update_druid_count
+      begin
```
jcoyne, line 18: Is there a reason you aren't calling the existing `TrackSheet` code here?

jcoyne (changes requested):

app/jobs/tracking_sheet_report_job.rb
aaron-collier: There was. I don't recall why. I'll rework to use the existing and see if it raises any issues.

aaron-collier: @jcoyne fixed. I had a mistaken assumption about how the files were written early so copied the code... That was unnecessary, so the existing `TrackSheet` model works fine. Good catch.

jcoyne (approved):

PR #3434, opened 2022-04-11
[refactor] use load_and_authorize_resource consistently

jcoyne:

We were mostly using the load_and_authorize_resource, so use that consistently across all controllers
```
$ git grep load_and_authorize_cocina |wc -l
       5
$ git grep load_and_authorize_resource |wc -l
      10
```

test suite

### DIFF
--- a/app/controllers/descriptives_controller.rb
+++ b/app/controllers/descriptives_controller.rb
@@ -2,7 +2,7 @@
 
 # Download the CSV descriptive metadata
 class DescriptivesController < ApplicationController
-  before_action :load_and_authorize_cocina
+  before_action :load_and_authorize_resource
 
   # Display the form for uploading the descriptive metadata spreadsheet
   def new; end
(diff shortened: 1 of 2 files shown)


app/controllers/descriptives_controller.rb
```diff
@@ -42,7 +42,7 @@ def convert_metadata_fail(failure)
     render :new, status: :unprocessable_entity
   end
 
-  def load_and_authorize_cocina
+  def load_and_authorize_resource
```
justinlittman, line 45: Is there a reason this isn't in ApplicationController? Then there is no need to worry about naming conventions.

mjgiarlo: I have the same question.

jcoyne: That's where I was going... but in small steps.

justinlittman (approved):

PR #3446, opened 2022-04-11
add support for Excel files to desc spreadsheet upload

### LINKED ISSUE #3363: Support Excel and ODS uploads for descriptive metadata
For parity with current service. Downloads can remain CSV-only.

peetucket:

Fixes #3363 - allow the user to upload ODS, XLS, XLSX as well as CSV when uploading descriptive metadata (only).  CSV is still required for all other uploads.  

Use the roo gem to open the spreadsheet, then ask it to covert to CSV ... this allows us to parse and use the CSV just as before without other code changes.

TODO: test (both in code and in browser)

### DIFF
--- a/app/jobs/descriptive_metadata_import_job.rb
+++ b/app/jobs/descriptive_metadata_import_job.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'roo'
+
 class DescriptiveMetadataImportJob < GenericJob
   queue_as :default
 
@@ -14,7 +16,14 @@ class DescriptiveMetadataImportJob < GenericJob
   # @option params [String] :use_barcodes_option option to update the barcodes
   def perform(bulk_action_id, params)
     super
-    csv = CSV.parse(params[:csv_file], headers: true)
+    # uploaded_file is the file provided by the user, it can be a CSV, ODS or Excel file (xls or xlsx)
+    uploaded_file = params[:uploaded_file]
+
+    raise 'Unsupported upload file type' unless %w[.csv .ods .xls .xlsx].include? File.extname(uploaded_file)
+
+    spreadsheet = Roo::Spreadsheet.open(uploaded_file) # open the spreadsheet
+
+    csv = CSV.parse(spreadsheet.to_csv, headers: true) # convert data to CSV and then parse by the CSV library as done in other parts of Argo
     with_csv_items(csv, name: 'Import descriptive metadata') do |cocina_object, csv_row, success, failure|
       next failure.call('Not authorized') unless ability.can?(:manage_item, cocina_object)
 
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,9 @@ gem 'honeybadger', '~> 4.1'
 gem 'lograge'
 gem 'sidekiq', '~> 6.0'
 
+gem 'roo', '~> 2.9.0' # work with newer Excel files and other types (xlsx, ods, csv)
+gem 'roo-xls' # needed to work with legacy Excel files (xls)
+
 gem 'newrelic_rpm'
 gem 'nokogiri', '~> 1.12'
 # Prawn is used to create "tracksheets"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -448,6 +448,13 @@ GEM
       railties (>= 5.0)
     retries (0.0.5)
     rexml (3.2.5)
+    roo (2.9.0)
+      nokogiri (~> 1)
+      rubyzip (>= 1.3.0, < 3.0.0)
+    roo-xls (1.2.0)
+      nokogiri
+      roo (>= 2.0.0, < 3)
+      spreadsheet (> 0.9.0)
     rsolr (2.5.0)
       builder (>= 2.1.2)
       faraday (>= 0.9, < 3, != 2.0.0)
@@ -487,6 +494,7 @@ GEM
       rubocop (>= 1.7.0, < 2.0)
     rubocop-rspec (2.9.0)
       rubocop (~> 1.19)
+    ruby-ole (1.2.12.2)
     ruby-prof (1.4.3)
     ruby-progressbar (1.11.0)
     ruby2_keywords (0.0.5)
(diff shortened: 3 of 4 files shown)


app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -14,7 +16,14 @@ class DescriptiveMetadataImportJob < GenericJob
   # @option params [String] :use_barcodes_option option to update the barcodes
   def perform(bulk_action_id, params)
     super
-    csv = CSV.parse(params[:csv_file], headers: true)
+    # uploaded_file is the file provided by the user, it can be a CSV, ODS or Excel file (xls or xlsx)
+    uploaded_file = params[:uploaded_file]
+
+    raise 'Unsupported upload file type' unless %w[.csv .ods .xls .xlsx].include? File.extname(uploaded_file)
+
+    spreadsheet = Roo::Spreadsheet.open(uploaded_file) # open the spreadsheet
```
justinlittman, line 24: I think roo should be in CsvUploadNormalizer.

peetucket: Ahh, I missed this class.  Though doing it here will then allow Excel files to be uploaded everywhere (and not just in the Desc Metadata job as specced in the ticket).  Though I supposed this won't matter much, since the user should be able to continue using CSV and letting roo deal with it?

justinlittman: Exactly.

app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'roo'
```
jcoyne, line 3: This isn't necessary since you've declared roo in the Gemfile.

app/jobs/descriptive_metadata_import_job.rb
jcoyne: Another issue is that we have argo on two different machines. You're implementation here would require that all the nodes have access to the same upload area. I'm not sure we can make that assumption.

peetucket: But using the `CsvUploadNormalized` class to just open with roo and convert to CSV immediately gets around this right?  Because that class is dealing with the file that the user just uploaded ?

app/services/csv_upload_normalizer.rb
```diff
@@ -11,10 +11,14 @@ def initialize(path, druid_headers: %w[Druid druid])
...
+  # Reads uploaded file (could be csv or Excel) and normalizes to CSV.
   # This includes handling BOM and druids without prefixes.
   def read
-    table = CSV.open(path, 'rb:bom|utf-8', headers: true).read
+    raise 'Unsupported upload file type' unless %w[.csv .ods .xls .xlsx].include? File.extname(path)
+
+    spreadsheet = Roo::Spreadsheet.open(path, { csv_options: { encoding: 'bom|utf-8' } }) # open the spreadsheet
+    table = CSV.parse(spreadsheet.to_csv, headers: true) # convert spreadsheet to CSV and then parse by the CSV library
+
     table.each do |row|
```
peetucket, line 22: this is the core change, just have Roo open the spreadsheet, using the CSV encoding (if you happen to open a CSV file to start), then ask roo to covert to CSV and parse it as normal

the desc metadata import job has it's form changed to accept other file types (and note this in the message to the user), the other forms do not even though they could now, but should continue to work exactly as before since CSV is an accepted type

peetucket:
Ok, made some changes to move the roo parsing code to the `CsvUploadNormalizer` class

app/services/csv_upload_normalizer.rb
```diff
@@ -11,10 +11,14 @@ def initialize(path, druid_headers: %w[Druid druid])
...
   end
 
-  # Reads and normalizes CSV.
+  # Reads uploaded file (could be csv or Excel) and normalizes to CSV.
   # This includes handling BOM and druids without prefixes.
   def read
-    table = CSV.open(path, 'rb:bom|utf-8', headers: true).read
+    raise 'Unsupported upload file type' unless %w[.csv .ods .xls .xlsx].include? File.extname(path)
+
+    spreadsheet = Roo::Spreadsheet.open(path, { csv_options: { encoding: 'bom|utf-8' } }) # open the spreadsheet
```
jcoyne, line 19: are the encoding options still necessary? Does roo take care of this automatically?

peetucket: I believe that it does for Excel files, but if you start with a CSV it may not handle automatically.  The roo documentation lists this option for input encoded CSV files, so it seems reasonable to include it. We could verify on stage by deploying and seeing (if Andrew has different encoded files).

justinlittman: You can create a CSV with a BOM by saving in Excel as "CSV UTF-8" (the default CSV option in Mac Excel).

peetucket: Actually I just noticed there is an existing test for regular vs UTF-8 CSVs (with both kinds in the fixture files), see https://github.com/sul-dlss/argo/pull/3446/files#diff-1b4024a38dec55440d3468e90ef2fa1be8ba1b1bd4a11d31059f481efc496cfbL14-L27

Both tests pass in this PR, but the UTF-8 CSV test fails in this PR if I remove the encoding options from the `Roo::Spreadsheet.open` on line 19.  So at least the tests suggest this encoding option is needed and is working as expected.

peetucket:
@andrewjbtw and @arcadiafalcone : based on the tests, I'm fairly certain that Excel files and UTF-8 encoded CSV files will work for desc metadata uploads in this PR, but if you'd like to verify with some actual spreadsheets you have, we can deploy this to QA or stage and you can check there.

Also note this change would allow us to accept Excel fairly easily in other jobs that currently only take CSVs, though this PR only makes the UI changes for desc metadata uploads.

jcoyne (approved):

peetucket:
> @andrewjbtw and @arcadiafalcone : based on the tests, I'm fairly certain that Excel files and UTF-8 encoded CSV files will work for desc metadata uploads in this PR, but if you'd like to verify with some actual spreadsheets you have, we can deploy this to QA or stage and you can check there.

Never mind, it was merged. We can still test on qa/stage if we deploy main now.

PR #3453, opened 2022-04-12
Create a set of local models

jcoyne:

These models are routable and mutable so the work more like rails defaults unlike the dry-struct models. 
This has been extracted from #3323


ndushay:
is this PR ready to go or is it on hold pending discussion?

jcoyne:
I thought this was part of what was discussed.

ndushay:
@jcoyne can you move this forward by getting appropriate review and/or buy in?  Bring it up in standup if nothing else?

PR #3455, opened 2022-04-12
Load using cancancan

jcoyne:

### DIFF
--- a/app/controllers/catkeys_controller.rb
+++ b/app/controllers/catkeys_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class CatkeysController < ApplicationController
-  before_action :load_and_authorize_resource
+  load_and_authorize_resource :cocina, parent: false, class: 'Repository', id_param: 'item_id'
 
   rescue_from Dor::Services::Client::UnexpectedResponse do |exception|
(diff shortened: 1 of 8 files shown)


ndushay:
um ... why _was_ this change made?  There's no attached ticket, and the title isn't helping me figure that out.

jcoyne:
This lets the gem do the work. We can remove boiler plate code.

app/controllers/catkeys_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class CatkeysController < ApplicationController
-  before_action :load_and_authorize_resource
+  load_and_authorize_resource :cocina, parent: false, class: 'Repository', id_param: 'item_id'
```
justinlittman, line 4: Can you explain this?

jcoyne: It is a method in the Cancancan gem that we use for authorization: https://github.com/CanCanCommunity/cancancan/blob/develop/docs/controller_helpers.md#authorize_resource-load_resource-load_and_authorize_resource

justinlittman: Thanks. Ruby magic.

justinlittman (approved):

PR #3465, opened 2022-04-12

sul-devops-team:


spec/forms/serials_form_spec.rb
```diff
@@ -182,7 +181,6 @@
                                        title: [
                                          {
                                            structuredValue: [
-                                             { type: 'odd thing', value: '99' },
```
jcoyne, line 185: I think it's useful to test that "It doesn't overwrite existing stuff".  Is there a valid type we could use here?

jcoyne (changes requested):

spec/forms/serials_form_spec.rb
justinlittman: Yeah, this change was a mistake. Fixed.

jcoyne (approved):

PR #3471, opened 2022-04-13
bulk action creation: remove bulk action form values from params passed along location URL in redirect when bulk action is created.

### LINKED ISSUE #3413: Bulk actions for large batches hit error in browser
**Describe the bug**
When I submit a large batch for a bulk action, I see an error message in the browser. The bulk action actually does start running, but in stage I see a page like this after pressing submit:

Screen Shot 2022-04-07 at 8 55 52 AM

In production, I may not see an error at all. Instead, my browser continues to show the "new bulk action" form even after the action has been submitted.

I haven't seen a honeybadger error when this issue happens. I'm also not sure how large a batch has to be to trigger it. I've generally seen it for lists of druids larger than 1000.

**User Impact**
It could lead someone to submit a bulk action twice, not realizing that it had already started running.

**To Reproduce**
1. Go to bulk actions
2. Submit a list of 1000+ druids. It doesn't seem to matter which bulk action runs as long as it's not a CSV one.
3. See either the error message or see that the browser stays on the bulk action form.

**Expected behavior**
Submitting a bulk action should take the browser back to the list of bulk actions.

jmartin-sul:

closes #3413 

long druid lists were causing response handling upstream from argo to throw a 500 error... because of the possibly enormous redirect `Location` header?

e.g. the apache error logs (`argo_error_ssl.log`) would report `Premature end of script headers: bulk_actions, referer` when the user received a 500 in this case (despite the app logs showing a successful 303 response to the `POST`).  note that HTTP 303 responses should have a location.  perhaps the location header in the 303 was not parseable by apache once a certain size was exceeded?

this only occurred with long druid lists.  thanks to @tallenaz yesterday for talking through where we might insert some debug logging, and @edsu today for suggesting a look at the 303 response the user gets when there's a long-ish druid list but the redirect after the bulk job succeeds in getting a 303 to the user.

i think the long druid list was actually just a holdover from the bulk job creation form, not part of the blacklight search params?

i wasn't sure whether filtering `druids_only` from the redirect URL's param list was the right thing to do.

tested manually on QA with a search that resulted in 964 druids.

needs a new unit test? 🤷 

`main`, with 964 druid search result, reproducing the issue:
Screen Shot 2022-04-12 at 5 47 03 PM

`main` with a short druid list and a successful response (the total count jumped up to 10 a moment later on page reload after the job had a chance to initialize -- note druid list in `Location`):
Screen Shot 2022-04-12 at 5 54 25 PM

this branch with a 964 druid search result and a successful 303 back to the user (druid list filtered from `Location` URL):
Screen Shot 2022-04-12 at 5 50 12 PM

### DIFF
--- a/app/controllers/concerns/creates_bulk_actions.rb
+++ b/app/controllers/concerns/creates_bulk_actions.rb
@@ -15,7 +15,11 @@ module CreatesBulkActions
     if bulk_action.save
       bulk_action.enqueue_job(job_params)
 
-      redirect_to bulk_actions_path(Blacklight::Parameters.sanitize(search_state.to_h.except(:authenticity_token))), status: :see_other, notice: success_message
+      # strip the CSRF token, and the parameters that happened to be in the bulk job creation form
+      # this can be removed when this is resolved: https://github.com/projectblacklight/blacklight/issues/2683
+      search_state_subset = search_state.to_h.except(:authenticity_token, :druids, :druids_only, :description)
+      path_params = Blacklight::Parameters.sanitize(search_state_subset)
+      redirect_to bulk_actions_path(path_params), status: :see_other, notice: success_message
     else
       render :new, status: :unprocessable_entity
     end


app/controllers/concerns/creates_bulk_actions.rb
```diff
@@ -15,7 +15,10 @@ def create
     if bulk_action.save
       bulk_action.enqueue_job(job_params)
 
-      redirect_to bulk_actions_path(Blacklight::Parameters.sanitize(search_state.to_h.except(:authenticity_token))), status: :see_other, notice: success_message
+      # strip the CSRF token, and the parameters that happened to be in the bulk job creation form
+      search_state_subset = search_state.to_h.except(:authenticity_token, :druids, :druids_only, :description)
```
jcoyne, line 19: It seems to me this is a bug in blacklight. search_state.to_h should not return parameters that are not part of the search state.

jmartin-sul: i think i agree, and i did try a couple of other methods that seemed more promising based on the name, but didn't work out (e.g. `Blacklight::SearchState#params_for_search` still included `druids`).

maybe we should leave this workaround in place with a comment till the issue is fixed upstream? 🤷

edsu: It [looks like](https://github.com/projectblacklight/blacklight/blob/main/lib/blacklight/search_state.rb#L26-L46) `Blacklight::SearchState` does do some normalization, but I didn't see that it removed things the params provided?

edsu: I'm inclined to merge this as is, but I defer to @jcoyne if it needs further work. One question I had for @jmartin-sul  was why preserving the parameters is needed given it's going back to the `/bulk_actions` -- are there links on that page that need the params preserved?

jcoyne: Yeah, that is my question too.  Why is it preserving these parameters when it goes back to `/bulk_actions`?

jcoyne: It's preserving the parameters because it needs to know what the last search was, in case you again want to click "populate with last search"

edsu:
I added a comment pointing to the Blacklight issue noting that we can remove this stripping in Argo once it is resolved there.

app/controllers/concerns/creates_bulk_actions.rb
edsu: I added a comment pointing to the Blacklight issue noting that we can remove this stripping in Argo once it is resolved there.

edsu (approved):
I think this looks good to solve the immediate problem and we can revisit when the upstream Blacklight issue gets resolved.

PR #3473, opened 2022-04-13
Fix structure upload.

### LINKED ISSUE #3451: Can't change file-level rights using the import structural CSV without a location
**Describe the bug**
When I try to change the file-level rights for a druid with the import structural CSV, I get an error message like:

```
Cocina::Models::ValidationError: {"view"=>"stanford", "download"=>"none", "location"=>nil} isn't one of in #/components/schemas/FileAccess
```

This happens whether or not I include a `rights_location` column in the CSV. It does not happen if I import a CSV where the rights do not change. 

**User Impact**
It's not possible to change the rights on specific files without the structural CSV.

**To Reproduce**
1. Go to a druid like https://argo-stage.stanford.edu/view/druid:db064kh7115
2. Download the structural CSV
3. Make changes to the CSV that involve changes to rights
4. Import the CSV
5. See the error

**Expected behavior**
Locations are only required when view or download is "location-based". If the rights don't include a location, it should be ok if the location cell is blank. 

**Additional context**
I get the same outcome for single item and bulk action changes.

justinlittman:
closes #3451

Valid cocina wasn't being created.

Unit

### DIFF
--- a/app/services/structure_updater.rb
+++ b/app/services/structure_updater.rb
@@ -73,10 +73,14 @@ class StructureUpdater
         shelve: row['shelve'] == 'yes',
         sdrPreserve: row['preserve'] == 'yes'
       ),
-      access: existing_file.access.new(
-        view: row['rights_view'],
-        download: row['rights_download'],
-        location: row['rights_location']
+      access: existing_file.access.class.new(
+        {
+          view: row['rights_view'],
+          download: row['rights_download'],
+          location: row['rights_location'],
+          # stanford/none required controlledDigitalLending set to false. All others should omit.
+          controlledDigitalLending: row['rights_view'] == 'stanford' && row['rights_download'] == 'none' ? false : nil
+        }.compact
       )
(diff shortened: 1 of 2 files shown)


app/services/structure_updater.rb
```diff
@@ -73,10 +73,14 @@ def update_file(existing_file, row)
...
-        download: row['rights_download'],
-        location: row['rights_location']
+      access: existing_file.access.class.new(
+        {
+          view: row['rights_view'],
+          download: row['rights_download'],
+          location: row['rights_location'],
+          # stanford/none required controlledDigitalLending set to false. All others should omit.
+          controlledDigitalLending: row['rights_view'] == 'stanford' && row['rights_download'] == 'none' ? false : nil
+        }.compact
```
jcoyne, line 83: Won't this mean you can never remove a location once set?

jcoyne (approved):

PR #3480, opened 2022-04-13
Remove nil values from arrays when uploading descriptive metadata

### LINKED ISSUE #3470: Spreadsheet upload: Error message for contributor.name.parallelValue.structuredValue.value
`Line 5: #/components/schemas/Contributor does not allow null values for druid:nb406th1306 (2022-04-12 22:47:00 UTC)`
[spreadsheet_upload_test.xlsx](https://github.com/sul-dlss/argo/files/8476998/spreadsheet_upload_test.xlsx)

aaron-collier:

Closes #3470 

This addresses the issue where if a descriptive metadata upload spreadsheet as a record with blank values in `contributor1.*` for instance, but does have values in `contributor3.*` that creates parameters with nil values such as:

```
{
  :purl=>"https://purl.stanford.edu/jr825qh8124",
  :title=>[{:value=>"foo bar 2"}],
  :contributor=>[
    nil,
    nil,
    {:name=>[
      nil,
      nil, 
      {:parallelValue=>[
        nil,
        nil, 
        {:structuredValue=>[
          {:value=>"Parallel 1 part 1", :type=>"name"},
          {:value=>"1800-1900", :type=>"life dates"}
        ]},
        {:structuredValue=>[
          {:value=>"Parallel 2 part 1", :type=>"name"},
          {:value=>"marchioness", :type=>"term of address"}]}]}]}]
}
```

This recursively removes the nils from the parameters resulting in:

```
{
  :purl=>"https://purl.stanford.edu/jr825qh8124",
  :title=>[{:value=>"foo bar 2"}],
  :contributor=>[
    {:name=>[
      {:parallelValue=>[
        {:structuredValue=>[
          {:value=>"Parallel 1 part 1", :type=>"name"},
          {:value=>"1800-1900", :type=>"life dates"}
        ]},
        {:structuredValue=>[
          {:value=>"Parallel 2 part 1", :type=>"name"},
          {:value=>"marchioness", :type=>"term of address"}]}]}]}]
}
```

And avoiding `#/components/schemas/Contributor does not allow null values` errors.

### DIFF
--- a/app/services/description_import.rb
+++ b/app/services/description_import.rb
@@ -71,7 +71,10 @@ class DescriptionImport
 
   def compact_params(params)
     params.each do |key, value|
-      params[key].compact! if value.is_a?(Array)
+      next unless value.is_a?(Array)
+
+      params[key].compact!
+      params[key].map { |param| compact_params(param) }
     end
   end
 end
--- a/app/services/description_import.rb
+++ b/app/services/description_import.rb
@@ -24,7 +24,7 @@ class DescriptionImport
       visit(params, split_address, @csv_row[address]) if @csv_row[address]
     end
 
-    Success(Cocina::Models::Description.new(params))
+    Success(Cocina::Models::Description.new(compact_params(params)))
   rescue Cocina::Models::ValidationError => e
     Failure(e.message)
   end
@@ -66,6 +66,12 @@ class DescriptionImport
       (0...what.size).cover?(key)
     when Hash
       what.key?(key)
+    end
+  end
+
+  def compact_params(params)
+    params.each do |key, value|
+      params[key].compact! if value.is_a?(Array)
     end
   end
 end
--- a/spec/services/description_import_spec.rb
+++ b/spec/services/description_import_spec.rb
@@ -205,4 +205,42 @@ RSpec.describe DescriptionImport do
       expect(updated.value!).to eq expected
     end
   end
+
+  context 'with a csv that has varied nested values' do
+    # This case occurs when you do a bulk import of different object structures
+    let(:csv_data) do
+      <<~CSV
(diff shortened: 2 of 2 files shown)


app/services/description_import.rb
```diff
@@ -68,4 +68,10 @@ def key?(what, key)
       what.key?(key)
     end
   end
+
+  def compact_params(params)
+    params.each do |key, value|
```
jcoyne, line 73: Does this only work at the top level?  Is that good enough or do we need to ensure this happens at any level of the tree?

aaron-collier: Good catch. Not sure how likely that is, but this small change makes it recursive and I've updated the description above.

jcoyne (changes requested):
Please add a test for this.

aaron-collier:
@jcoyne test added.

app/services/description_import.rb
```diff
@@ -71,7 +71,10 @@ def key?(what, key)
 
   def compact_params(params)
     params.each do |key, value|
-      params[key].compact! if value.is_a?(Array)
+      next unless value.is_a?(Array)
```
jcoyne, line 74: Won't this kill the recursion? I'm not sure the test shows that this works on nested arrays that have nils.

jcoyne (changes requested):

app/services/description_import.rb
jcoyne: Can you test with the data you display in the PR?

aaron-collier: i can make that change. Just to point out, this does show the functionality. If I remove the compact_params method and run this test I get a "Cannot insert nil..." error - so I believe this is properly exercising the code, but I can make deeper test data, sure.

aaron-collier: @jcoyne ok, this is now testing with the data from the comment above.

jcoyne (approved):

PR #3483, opened 2022-04-13
Close the modal when the upload is successful

### LINKED ISSUE #3475: For descriptive file upload, display validation errors
1. For an item, download the descriptive spreadsheet: Manage description > Download cocina spreadsheet.
2. Open the spreadsheet and change the value for any field that includes "type" (e.g., `contributor1:name1:structuredValue1:type`) to an invalid value (e.g., "foo").
3. Save the spreadsheet.
4. Upload the descriptive spreadsheet: Manage description > Upload cocina spreadsheet.

Expected:
* Cocina::Models::ValidationError message is displayed to user.
* No HB alert

Actual result:
image

https://app.honeybadger.io/projects/49894/faults/84964293

jcoyne:
and catch validation errors

Fixes #3475

### DIFF
--- a/spec/features/upload_descriptive_metadata_spec.rb
+++ b/spec/features/upload_descriptive_metadata_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# This is an integration test (with DSA), because we need to show that the
+# data we submit from here doesn't cause a round trip to MODS error.
+RSpec.describe 'Descriptive metadata spreadsheet upload', js: true do
+  let(:user) { create(:user) }
+  # We need to set a catkey on this item, but we can't do it when it's created,
+  # because we don't have symphony running in our test environment. If you register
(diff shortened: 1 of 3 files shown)


spec/features/upload_descriptive_metadata_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# This is an integration test (with DSA), because we need to show that the
+# data we submit from here doesn't cause a round trip to MODS error.
```
edsu, line 6: Am I just missing it or is there not a test of what happens when there is a failure?

edsu (approved):

spec/features/upload_descriptive_metadata_spec.rb
jcoyne: This, being a feature spec, is a full stack spec. They are slow, so we only do them where necessary.  We have more granular specs in `spec/requests` that are not full stack and are faster.

PR #3484, opened 2022-04-13
Rescue validation error in DescriptiveMetadataImportJob to avoid alerting honeybadger

### LINKED ISSUE #3478: Do not HB alert when ValidationError on DescriptiveMetadataImportJob
Currently, if a Cocina::Models::ValidationError is raised during a DescriptiveMetadataImportJob it logs the error and HB alerts. It should not HB alert.

```
2022-04-13 19:39:43 UTC Starting DescriptiveMetadataImportJob for BulkAction 307
Line 2: Line 2: Import descriptive metadata failed Cocina::Models::ValidationError Unrecognized types in description: description.contributor1.name1.structuredValue1 for druid:bb408qn5061 (2022-04-13 19:39:43 UTC)
2022-04-13 19:39:43 UTC Finished DescriptiveMetadataImportJob for BulkAction 307
```

## Backtrace

    line 22 of [PROJECT_ROOT]/app/jobs/descriptive_metadata_import_job.rb: block (2 levels) in perform
    line 21 of [PROJECT_ROOT]/app/jobs/descriptive_metadata_import_job.rb: block in perform
    line 135 of [PROJECT_ROOT]/app/jobs/generic_job.rb: block (2 levels) in with_csv_items

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49894/faults/84964957)

aaron-collier:

Closes #3478 

This adds a pattern for filtering which jobs and error combinations will alert to honeybadger. 

### DIFF
--- a/app/jobs/descriptive_metadata_import_job.rb
+++ b/app/jobs/descriptive_metadata_import_job.rb
@@ -19,7 +19,14 @@ class DescriptiveMetadataImportJob < GenericJob
       next failure.call('Not authorized') unless ability.can?(:update, cocina_object)
 
       DescriptionImport.import(csv_row: csv_row).either(
-        ->(description) { Repository.store(cocina_object.new(description: description)) && success.call('Successfully updated') },
+        lambda { |description|
+          begin
+            Repository.store(cocina_object.new(description: description))
+            success.call('Successfully updated')
+          rescue Cocina::Models::ValidationError => e
+            failure.call(e.message)
+          end
+        },
         ->(error) { failure.call(error) }
       )
     end
--- a/spec/jobs/descriptive_metadata_import_job_spec.rb
+++ b/spec/jobs/descriptive_metadata_import_job_spec.rb
@@ -62,5 +62,29 @@ RSpec.describe DescriptiveMetadataImportJob, type: :job do
         expect(Repository).not_to have_received(:store)
       end
     end
+
+    context 'when validation fails' do
+      let(:csv_file) do
+        [
+          'druid,source_id,title1.structuredValue1.type,purl',
+          [item1.externalIdentifier, item1.identification.sourceId, 'new title 1', 'https://purl'].join(',')
+        ].join("\n")
+      end
+      let(:ability) { instance_double(Ability, can?: true) }
+
+      before do
+        allow(Ability).to receive(:new).and_return(ability)
+        allow(Honeybadger).to receive(:notify)
+        subject.perform(bulk_action.id, { csv_file: csv_file })
+      end
+
+      it 'updates the error count without alerting honeybadger' do
+        expect(bulk_action.druid_count_total).to eq 1
+        expect(bulk_action.druid_count_fail).to eq 1
+        expect(bulk_action.druid_count_success).to eq 0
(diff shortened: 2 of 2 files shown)


app/jobs/generic_job.rb
```diff
@@ -14,6 +14,9 @@
 class GenericJob < ApplicationJob
   # A somewhat easy to understand and informative time stamp format
   TIME_FORMAT = '%Y-%m-%d %H:%M%P'
+  DO_NOT_ALERT = {
+    'DescriptiveMetadataImportJob' => ['Cocina::Models::ValidationError']
```
jcoyne, line 18: How about rescuing this in the `DescriptiveMetadataImportJob` rather than add code to the base class just for the one subclass?

jcoyne (changes requested):

app/jobs/generic_job.rb
aaron-collier: @jcoyne,  I see that we already rescue in `DescriptiveImport` (https://github.com/sul-dlss/argo/blob/17cef1d40cde1f2c0521be6f0ceca65e863c9038/app/services/description_import.rb#L28) and that bubbles up into `GenericJob` where the alert is sent... 

I'm not clear on how to avoid the HB alert, which I think but could be wrong, is rescued in https://github.com/sul-dlss/argo/blob/main/app/jobs/generic_job.rb#L136 with the `failure.call` at https://github.com/sul-dlss/argo/blob/main/app/jobs/descriptive_metadata_import_job.rb#L23

Should I refactor the error path and remove the `failure.call` in `DescriptiveImportJob`? Or will that impact the failure count reporting?

Thanks.

jcoyne: There is no way for that rescued error to "bubble up".  It must have been from a different location.  Don't remove the failure.call, or you won't get proper counts.

jcoyne: It's the `cocina_object.new(description: description)` that is raising the error, not the DescriptionImport

aaron-collier: Done. Thanks.

app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -19,7 +19,14 @@ def perform(bulk_action_id, params)
       next failure.call('Not authorized') unless ability.can?(:update, cocina_object)
 
       DescriptionImport.import(csv_row: csv_row).either(
-        ->(description) { Repository.store(cocina_object.new(description: description)) && success.call('Successfully updated') },
+        lambda { |description|
+          begin
+            Repository.store(cocina_object.new(description: description))
+            success.call('Successfully updated')
+          rescue Cocina::Models::ValidationError => e
+            failure.call(e)
```
jcoyne, line 27: This should be `e.message`

jcoyne (changes requested):

app/jobs/descriptive_metadata_import_job.rb
aaron-collier: done.

aaron-collier:
@jcoyne done.

jcoyne (approved):

PR #3489, opened 2022-04-14
don't show 'Manage serials' option for non-item objects

### LINKED ISSUE #3460: Don't show the "manage serials" option on collection objects
Only items can be used with the digital serials model. Collection pages currently show the "manage serials" option under the "manage description" menu so that link should be removed.

ndushay:

"Manage serials" should not be an option in "Manage description" pulldown except for items with ckeys

collection objects with ckeys now show this:

image

Fixes #3460

### DIFF
--- a/spec/components/show/controls_component_spec.rb
+++ b/spec/components/show/controls_component_spec.rb
@@ -118,6 +118,45 @@ RSpec.describe Show::ControlsComponent, type: :component do
     end
   end
 
+  context 'when the object is a Collection the user can manage' do
+    let(:view_collection_id) { 'druid:kv840xx0000' }
+    let(:allows_modification) { true }
+
+    let(:doc) do
+      SolrDocument.new('id' => view_collection_id,
+                       'processing_status_text_ssi' => 'not registered',
+                       SolrDocument::FIELD_OBJECT_TYPE => 'collection',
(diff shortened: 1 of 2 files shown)


spec/components/show/controls_component_spec.rb
```diff
@@ -118,6 +118,45 @@
...
+      expect(page).to have_link 'Unpublish'
+      expect(page).to have_text 'Manage description'
+      expect(page).to have_link 'Purge', href: '/items/druid:kv840xx0000/purge'
+      expect(page).to have_link 'Download Cocina spreadsheet', href: '/items/druid:kv840xx0000/descriptive.csv'
+      expect(page).to have_link 'Upload Cocina spreadsheet', href: '/items/druid:kv840xx0000/descriptive/edit'
+
+      expect(rendered.css('a').size).to eq 9
+    end
+
+    context 'when the item has a catkey' do
```
jcoyne, line 148:
```suggestion
    context 'when the collection has a catkey' do
```

jcoyne (changes requested):

spec/components/show/controls_component_spec.rb
ndushay: thanks for catching this copypasta.  Corrected and ready for merge, captain!

jcoyne (approved):

PR #3492, opened 2022-04-14
Handle validation errors when updating structural

### LINKED ISSUE #3485: For structural file upload, display validation errors
To replicate, for an object with dark rights, change the file rights to world and re-upload.

Expected:
* Validation error is displayed to user.
* No HB alert.

Actual:
image

https://app.honeybadger.io/projects/49894/faults/84972823

jcoyne:

Fixes #3485

Test suite

### DIFF
--- a/app/controllers/descriptives_controller.rb
+++ b/app/controllers/descriptives_controller.rb
@@ -39,16 +39,15 @@ class DescriptivesController < ApplicationController
   end
 
   def convert_metdata_success(description:)
-    valid_model = begin
-      @cocina.new(description: description)
-    rescue Cocina::Models::ValidationError => e
-      return display_error(e.message) # rubocop:disable Lint/NoReturnInBeginEndBlocks
-    end
-
-    Repository.store(valid_model)
-    redirect_to solr_document_path(@cocina.externalIdentifier),
-                status: :see_other,
-                notice: 'Descriptive metadata has been updated.'
+    CocinaValidator.validate(@cocina, description: description).either(
+      lambda { |valid_model|
+        Repository.store(valid_model)
+        redirect_to solr_document_path(@cocina.externalIdentifier),
+                    status: :see_other,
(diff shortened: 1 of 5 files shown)


app/controllers/descriptives_controller.rb
```diff
@@ -39,16 +39,15 @@ def display_error(error)
...
-      @cocina.new(description: description)
-    rescue Cocina::Models::ValidationError => e
-      return display_error(e.message) # rubocop:disable Lint/NoReturnInBeginEndBlocks
-    end
-
-    Repository.store(valid_model)
-    redirect_to solr_document_path(@cocina.externalIdentifier),
-                status: :see_other,
-                notice: 'Descriptive metadata has been updated.'
+    CocinaValidator.validate(@cocina, description: description).either(
```
justinlittman, line 42: For what it is worth, I find this syntax obtuse compared against just evaluating the Success/Failure result.

app/controllers/structures_controller.rb
```diff
@@ -21,14 +21,18 @@ def show
   end
 
   def update
-    status = StructureUpdater.from_csv(@cocina, params[:csv].read)
-
-    if status.success?
```
justinlittman, line 26: This syntax is much clearer to me. (if/then vs lambdas).

justinlittman (approved):
Approve, but would prefer to see if/then over lambdas for clarity.

app/controllers/structures_controller.rb
jcoyne: I'm not attached. I'm going to merge to fix the error, but feel free to rejigger for readability.  FWIW, I this is the more idiomatic according to the dry-monad docs.

justinlittman: Yeah, but every time I've looked at the dry-monad docs I've been unable to wrap my head around the idiom.

PR #3494, opened 2022-04-14
Handle errors on saving to dor-services-app

jcoyne:
Extracted a method for validate_and_save.  This is often something we do at the same time.  This helps us always handle the necessary exceptions while simplifying the controllers.

Errors were not being handled properly, leading to an exception.

Test suite

### DIFF
--- a/app/controllers/descriptives_controller.rb
+++ b/app/controllers/descriptives_controller.rb
@@ -12,9 +12,12 @@ class DescriptivesController < ApplicationController
     csv = CSV.read(params[:data].tempfile, headers: true)
     validator = DescriptionValidator.new(csv)
     if validator.valid?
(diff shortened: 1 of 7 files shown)


app/controllers/descriptives_controller.rb
```diff
@@ -38,18 +38,11 @@ def display_error(error)
...
 
-  def convert_metdata_success(description:)
-    CocinaValidator.validate(@cocina, description: description).either(
-      lambda { |valid_model|
-        Repository.store(valid_model)
-        redirect_to solr_document_path(@cocina.externalIdentifier),
-                    status: :see_other,
-                    notice: 'Descriptive metadata has been updated.'
-      },
+  def convert_metadata_success(description:)
```
jcoyne, line 41: fixed a typo in this method name

justinlittman (approved):

PR #3497, opened 2022-04-14
[HOLD] Add a bulk action for Validating descriptive metadata uploads

### LINKED ISSUE #3373: Add bulk action for validating a Cocina descriptive metadata spreadsheet (without modifying any data)
Outcome of a conversation with @arcadiafalcone and @astridu regarding bulk descriptive metadata upload: we should make a separate bulk action for descriptive metadata validation. This would allow people to check their metadata and get a report back on the entire sheet of druids before processing begins. 

**Details**

Bulk action menu location: Under "Generate reports"

The label should be: "Validate Cocina descriptive metadata spreadsheet"

Interaction: Provide a "submit" box to input the file.

Formats to support: Excel, ODS, CSV (same formats as the Bulk MODS upload currently accepts)

Help text: 
>**Validate a Cocina descriptive metadata spreadsheet**
>
>*Validates a Cocina descriptive metadata spreadsheet and checks problems with the metadata. Results will be available as a downloadable report.*

Implementation details for the log/report: @arcadiafalcone will provide more details on this.

**Additional information**

Validation will be part of the spreadsheet import bulk action (#3190) but that bulk action will process the druids right away. Having a separate validation action should reduce the chances of running into issue during updates because people will be able to revise their spreadsheet until it's valid beforehand.

aaron-collier:

Closes #3373 

Adds a bulk action that takes a csv/spreadsheet equivalent to the descriptive metadata upload job without committing changes to the cocina objects.

New unit test.

### DIFF
--- a/app/jobs/validate_cocina_descriptive_job.rb
+++ b/app/jobs/validate_cocina_descriptive_job.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class ValidateCocinaDescriptiveJob < GenericJob
+  queue_as :default
+
+  ##
+  # A job that allows a user to verify that a spreadsheet for descriptive upload is valid
+  # @param [Integer] bulk_action_id GlobalID for a BulkAction object
+  # @param [Hash] params additional parameters that an Argo job may need
+  # @option params [String] :csv_file csv data to validate
+  def perform(bulk_action_id, params)
+    super
+
+    csv = CSV.parse(params[:csv_file], headers: true)
+    with_csv_items(csv, name: 'Validate Cocina descriptive metadata') do |cocina_object, csv_row, success, failure|
+      DescriptionImport.import(csv_row:)
+                       .bind { |description| CocinaValidator.validate(cocina_object, description:) }
+                       .either(
+                         ->(_validated) { success.call('Successfully validated') },
+                         ->(error) { failure.call(error) }
+                       )
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/jobs/validate_cocina_descriptive_job.rb
```diff
@@ -0,0 +1,29 @@
...
+  def perform(bulk_action_id, params)
+    super
+
+    csv = CSV.parse(params[:csv_file], headers: true)
+    with_csv_items(csv, name: 'Validate Cocina descriptive metadata') do |cocina_object, csv_row, success, failure|
+      DescriptionImport.import(csv_row: csv_row).either(
+        lambda { |description|
+          begin
+            cocina_object.new(description: description)
+            success.call('Successfully updated')
```
justinlittman, line 20: This should probably say "Successfully validated".

justinlittman (changes requested):

app/jobs/validate_cocina_descriptive_job.rb
```diff
@@ -0,0 +1,29 @@
...
+  # @option params [String] :csv_file csv data to validate
+  def perform(bulk_action_id, params)
+    super
+
+    csv = CSV.parse(params[:csv_file], headers: true)
+    with_csv_items(csv, name: 'Validate Cocina descriptive metadata') do |cocina_object, csv_row, success, failure|
+      DescriptionImport.import(csv_row: csv_row).either(
+        lambda { |description|
+          begin
+            cocina_object.new(description: description)
```
jcoyne, line 19: You can replace this with `CocinaValidator.valdate(cocina_object, description: description`.  That can be chained with `bind` to the previous call, similar to https://github.com/sul-dlss/argo/blob/main/app/jobs/descriptive_metadata_import_job.rb#L21-L26

jcoyne (changes requested):

app/jobs/validate_cocina_descriptive_job.rb
aaron-collier: Done.

app/jobs/validate_cocina_descriptive_job.rb
aaron-collier: Good catch. done.

jcoyne (approved):

justinlittman (approved):

PR #3499, opened 2022-04-14
Validate cell values for csv desc metadata spreadsheets

### LINKED ISSUE #3444: Descriptive upload validation: Validate cell values
For both individual and bulk upload:
* When cell value is #NA, #REF!, #VALUE?, or #NAME?, report "Value error: #{druid} has spreadsheet formula error in #{header}."
* When cell value is 0, report "Value error: #{druid} has 0 value in #{header}."

refs #3375

peetucket:

Fixes #3444 - validates cell values in spreadsheet for 0s and spreadsheet errors.

Added new tests

### DIFF
--- a/app/services/description_validator.rb
+++ b/app/services/description_validator.rb
@@ -14,11 +14,23 @@ class DescriptionValidator
   def valid?
     validate_duplicate_headers
     validate_title_headers
+    validate_cell_values
     if @bulk_job
       validate_druid_headers
       validate_druid_rows
     end
     errors.empty?
+  end
+
+  def validate_cell_values
+    @csv.each.with_index(2) do |row, i|
+      @headers.excluding('druid').each do |header|
+        location = row['druid'] || "row #{i}"
+        cell_value = row[header]&.strip
+        errors << "Value error: #{location} has 0 value in #{header}." if cell_value == '0'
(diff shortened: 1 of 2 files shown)


app/services/description_validator.rb
```diff
@@ -14,13 +14,24 @@ def initialize(csv, bulk_job: false)
...
       validate_druid_rows
     end
     errors.empty?
   end
 
+  def validate_cell_values
+    @csv.each.with_index(2) do |row, i|
+      @headers.excluding('druid').each do |header|
+        location = row['druid'] || "row #{i}"
+        errors << "Value error: #{location} has 0 value in #{header}." if row[header] == '0'
```
peetucket, line 29: the ticket tells us to add the druid to the error message, but single object spreadsheets may not have a druid column...in which case, we will reference the row instead

app/services/description_validator.rb
```diff
@@ -14,13 +14,24 @@ def initialize(csv, bulk_job: false)
...
+        errors << "Value error: #{location} has spreadsheet formula error in #{header}." if %w[#NA #REF! #VALUE? #NAME?].include? row[header]
+      end
```
peetucket, line 31: The ticket tells us to add the druid to the error message, but single object spreadsheets may not have a druid column, so we will use the row instead in that case.

jcoyne (approved):

Issue #3514, opened 2022-04-15
Redesign item registration form


andrewjbtw:
See Astrid's wireframe below. The motivation for the registration form redesign - beyond the fact that it desperately needs one - is to accommodate the new and more dynamic content type and access rights menus:

- The content type menu needs to expand to show reading order when that is relevant to the content type (book or image)
- The access rights menu needs to give access to view/download/location/CDL options

These two menus are separated out in the diagram as a way to make space to expand them. Depending on how we're able to implement this, we might need additional/modified labeling to indicate which menus belong together.

<img width="713" alt="registration-form" src="https://user-images.githubusercontent.com/8975583/163633983-b240f2fd-2624-4509-ad7f-c7bad3ee8176.png">

**If changes break an infrastructure integration tests, please provide fix for tests.**

jcoyne:
How does the display change when you push each of the buttons or change the selected fields?
What values should be in each of the select boxes. 
What sort of validations should be done on each field? When should the validation take place? How do we display invalid values?

justinlittman:
Notes from 4/18:
* Unless otherwise specified, current behavior should be replicated.
* "Rights" should be "View".
* "Downloadable by" should be "Download".
* Controlled digital lending should be added.
* Reading order should only be enabled when an appropriate content type is selected.
* Reading order should have the same values as Set Content type menu.
* Tracking sheets button should be disabled until after registration.

PR #3521, opened 2022-04-18
Update registration grid actions styles

jcoyne:
This removes the multiple select and the lock button per the designs.

Ref #3514

locally, test suite, on stage by Andrew
Screen Shot 2022-04-18 at 4 10 47 PM

### DIFF
--- a/app/assets/stylesheets/icons.scss
+++ b/app/assets/stylesheets/icons.scss
@@ -22,18 +22,6 @@
   }
 }
 
-.icon-lock {
-  &::before {
-    content: '🔒';
-  }
-}
-
-.icon-unlock {
-  &::before {
-    content: '🔓';
-  }
-}
-
 .icon-minus {
   &::before {
     content: '-';
--- a/app/assets/stylesheets/registration.scss
+++ b/app/assets/stylesheets/registration.scss
@@ -2,6 +2,8 @@
 @import 'registration/ui-override';
 @import 'jquery-ui-1.12.1.custom/jquery-ui.theme';
 
+// these style are used by the Grid/Text toggle
+
 .ui-controlgroup {
 	vertical-align: middle;
 	display: inline-block;
--- a/app/assets/stylesheets/registration/registration.scss
+++ b/app/assets/stylesheets/registration/registration.scss
@@ -1,19 +1,5 @@
-#data {
-  thead {
-    font-weight: bold;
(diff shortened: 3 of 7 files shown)


andrewjbtw:
I see I should have put this in the "redesign" ticket: we implemented auto-suggest for tags on the old form that we should support after redesign. It's been helpful for people wanting to reuse existing tags.

jcoyne:
@andrewjbtw are you saying that the autosuggest is no longer working?

andrewjbtw:
Oh, I keep forgetting to try different tags on stage vs. prod. Sorry. I'm seeing it now.

andrewjbtw:
Switching from grid view to text view is doing some strange things. When you choose the text view, the buttons remain visible in the background:

Screen Shot 2022-04-19 at 2 40 41 PM

In the text view, the farthest left column should be the "barcode" column. But it looks like the unlabeled column on the grid, which is there to signal if there are errors, is being treated as an enter-able field, which shifts the tab arrangement over one position. I pasted some data that I used in prod registration and the alignment when I switch back to grid view is off by one:
Screen Shot 2022-04-19 at 2 58 27 PM

That input should be filling in "catkey", "source id" and "label".

andrewjbtw:
Sample input data to go along with above: 

```
	6553131	sul:CSt-112-1360		6553131
	6553136	sul:CSt-112-1364		6553136
	6553344	sul:CSt-112-1368		6553344
	6553358	sul:CSt-112-1372		6553358
	6553379	sul:CSt-112-1376		6553379
	6553407	sul:CSt-112-1380		6553407
	6553465	sul:CSt-112-1384		6553465
	6553503	sul:CSt-112-1388		6553503
	6553587	sul:CSt-112-1392		6553587
	6553715	sul:CSt-112-1396		6553715
```

andrewjbtw:
Probably a question for Astrid: since we don't know how long the grid could be, maybe the buttons should go above it. I pasted in 300 druids and that pushed the interface way down the page.

jcoyne:
The current library, when switching to "text" depends on there being nothing at the bottom of the page.  So, even if the buttons weren't showing through, they would be inaccessible.

jcoyne:
@andrewjbtw Thank you for your feedback. Can you retest it now?

justinlittman (approved):

PR #3525, opened 2022-04-18
Upgrade registration form modal to bootstrap 5

jcoyne:
The registration form hasn't been displaying errors properly since the bootstrap 5 upgrade

locally

### DIFF
--- a/app/javascript/controllers/registration_controller.js
+++ b/app/javascript/controllers/registration_controller.js
@@ -168,11 +168,8 @@ export default class extends Controller {
           displayRequirements : function(text) {
             document.querySelector('#gridErrorModal .modal-body p').innerHTML = text;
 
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/registration_controller.js
```diff
@@ -37,6 +37,7 @@ export default class extends Controller {
     register() {
         var cells = $('#data').jqGrid('getGridParam','savedRow');
         if (cells.length > 0) {
+          console.log("still editin")
```
ndushay, line 40: whoops?

ndushay (approved):
I think you might have left a `console.log` in there by mistake

app/javascript/controllers/registration_controller.js
jcoyne: Yep.

PR #3533, opened 2022-04-19
Specify Ruby 3.1.x as required Ruby version

mjgiarlo:

To make it clearer what Ruby version an app requires.

CI


Gemfile
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+ruby '~> 3.1.1'
```
jcoyne, line 3: Will this cause the app to break when https://github.com/sul-dlss/puppet/pull/7883 is released?

mjgiarlo: It's one of the things I'm looking to test.

config/deploy/qa.rb
```diff
@@ -4,3 +4,11 @@
 server 'argo-qa-b.stanford.edu', user: 'lyberadmin', roles: %w[web db app worker]
 
 Capistrano::OneTimeKey.generate_one_time_key!
+
+before 'deploy:starting', :ruby_version_check do
+  app_ruby = Bundler::LockfileParser.new(Bundler.read_file('Gemfile.lock')).ruby_version.split.last.split('p').first
+  # default_system_ruby = `rvm list default string`.chomp.chomp.split('-').last
+  system_rubies = `rvm list rubies`.split.grep(/ruby/).map { |version| version.delete_prefix('ruby-') }
```
jcoyne, line 11: I'm not sure that looking for an installed ruby version is good enough. We need to know what passenger is using.  If the version in the Gemfile doesn't match what passenger has, I think it will break.

mjgiarlo: that's in the list of things @aaron-collier is looking at. I wouldn't look too closely at this PR until it's out of draft.

mjgiarlo:
For now, we have determined that investing time in making the passenger bits of puppet (specifically the version of ruby it is configured to use) more open to modification is not worth the potential benefit. Until that is tractable, this change is of limited value.

PR #3534, opened 2022-04-19
content type bulk job set reading order

### LINKED ISSUE #3527: Update bulk action "set content type" menu to include reading order
This is a follow up to the single item menu change completed in #3211 and should follow the same design ultimately implemented in https://github.com/sul-dlss/argo/pull/3388 

I thought I'd ticketed the bulk action menu but apparently I had not.

peetucket:

Fixes #3527 - update the set content type bulk action job UI so that it matches the single object content type update, which means showing a separate reading order drop down menu for Books and Images.  This involves re-using the existing JS controller actions for hiding/showing the reading-direction, re-using the drop down menu elements, and updating the job logic so that is uses the actual cocina types (which are now sent with the drop downs) instead of looking them up.

Updated the tests, and some localhost testing in browser (to ensure the form works as expected).


app/forms/content_type_form.rb
```diff
@@ -56,7 +56,7 @@ def cocina_update_attributes
     end
   end
 
-  # If the new content type is a book, we need to set the viewing direction attribute in the cocina model
+  # If the new content type is a book or image, we need to set the viewing direction attribute in the cocina model
   def member_orders
```
peetucket, line 60: make this comment consistent with the recent changes in this form

app/jobs/set_content_type_job.rb
```diff
@@ -5,10 +5,11 @@ class SetContentTypeJob < GenericJob
   def perform(bulk_action_id, params)
     super
 
-    # types are the label for content types, e.g. book (ltr)
+    # types are the label for content types, e.g. book
     @current_resource_type = params[:current_resource_type]
     @new_content_type = params[:new_content_type]
     @new_resource_type = params[:new_resource_type]
+    @viewing_direction = params[:viewing_direction]
```
peetucket, line 13: we have a new drop down menu in the bulk form to set viewing direction

app/jobs/set_content_type_job.rb
```diff
@@ -41,7 +42,7 @@ def set_content_type(cocina_object, success, failure)
 
   def cocina_update_attributes(cocina_object)
     {}.tap do |attributes|
-      attributes[:type] = Constants::CONTENT_TYPES[@new_content_type]
+      attributes[:type] = @new_content_type
```
peetucket, line 45: this and other similar changes are because the drop down menus now have the actual cocina types (e.g. `https://cocina.sul.stanford.edu/models/resources/page`) instead of the display value (e.g. `page`) -- so we don't look them up anymore but use them directly (as is done in https://github.com/sul-dlss/argo/blob/main/app/forms/content_type_form.rb)

app/jobs/set_content_type_job.rb
```diff
@@ -50,30 +51,29 @@ def cocina_update_attributes(cocina_object)
...
-                        else
-                          'right-to-left'
-                        end
-    [{ viewingDirection: viewing_direction }]
+    return [] unless may_have_direction? && @viewing_direction.present?
+
+    [{ viewingDirection: @viewing_direction }]
+  end
+
+  def may_have_direction?
```
peetucket, line 61: viewing direction set from the new drop down, and not looked up as before based on a content type value

jcoyne (approved):

PR #3537, opened 2022-04-19
Fix load_resource for the BulkJobsController for the index action

### LINKED ISSUE #3532: Users are getting a "forbidden" screen when trying to reach an Upload MODS page
Two users have reported this morning that they can no longer reach the "Upload MODS" page they previously had been able to reach. It could be doing a permissions check that previously wasn't being done.

Specifically, both users have reported the experience on different APOs:

- Go to an APO page like https://argo.stanford.edu/view/druid:xs835jp8197
- Click on the "Upload MODS" button
Screen Shot 2022-04-19 at 12 51 42 PM
- Get a forbidden message

**Additional context**

Other users have reported being able to see the Upload MODS button. I believe the difference between users who can see the "Upload MODS" and users who can't is that the users who can are in the "administrator" workgroup and the users who can't are not.

I think this change may have come with this week's dependency updates because one user definitely accessed Upload MODS last Friday.

**Expected behavior**

Any user who can see an APO page should also be able to see the Upload MODS interface linked from that page.

jcoyne:

Fixes #3532

Added tests such that we actually try the authorization logic.

### DIFF
--- a/app/controllers/bulk_jobs_controller.rb
+++ b/app/controllers/bulk_jobs_controller.rb
@@ -4,7 +4,7 @@
 class BulkJobsController < ApplicationController
   include Blacklight::Searchable
 
-  load_resource :cocina, parent: false, class: 'Repository', id_param: 'apo_id', only: :index
+  load_resource :cocina, class: 'Repository', id_param: 'apo_id', only: :index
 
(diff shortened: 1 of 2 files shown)


ndushay (approved):
I changed the title of the PR - I hope it's accurate.  The existing title made no sense to me.

PR #3539, opened 2022-04-19
close objects after bulk desc metadata import job

### LINKED ISSUE #3524: Automatically close object versions after running bulk descriptive metadata import
Automatically close objects after they've been successfully updated using bulk descriptive metadata import.

Per @arcadiafalcone , users often forget to close their objects after running bulk descriptive metadata import. Closing the object is necessary in order to trigger accessioning so that the purl and preservation are updated with the new metadata.

peetucket:

Fixes #3524 - close objects after running bulk desc metadata import job.  Also delete a tmp db file that was inadvertently added by #3534 and update gitignore to prevent from happening again.

Note: this will only close objects for which an update actually occurred.  It also assumes that any object updated is currently open, which now must be true (as we will open before applying update, if not already open). 

~~HOLD because:~~
~~Andrew tested and approved on May 13, but after rebasing from main I noticed one test started failing (likely due to this change: https://github.com/sul-dlss/argo/commit/686655e43c2f5149354fdcb56aece5638308e59c)~~

~~I think what is happening in that test is that the validation is failing, but the object is still getting closed, and so the final result is "success", which is probably not correct.  Not sure why it passed before.~~

Updated tests.

### DIFF
--- a/app/jobs/descriptive_metadata_import_job.rb
+++ b/app/jobs/descriptive_metadata_import_job.rb
@@ -22,9 +22,10 @@ class DescriptiveMetadataImportJob < GenericJob
       DescriptionImport.import(csv_row:)
                        .bind { |description| validate_changed(cocina_object, description) }
                        .bind { |description| open_version(cocina_object, description) }
-                       .bind { |description, new_cocina_object| CocinaValidator.validate_and_save(new_cocina_object, description:) }
-                       .bind { |new_cocina_object| close_version(new_cocina_object) }
-                       .either(
+                       .bind do |description, new_cocina_object|
+                         CocinaValidator.validate_and_save(new_cocina_object, description:)
+                         close_version(new_cocina_object)
+                       end.either(
                          ->(_updated) { success.call('Successfully updated') },
                          ->(messages) { failure.call(messages.to_sentence) }
                        )
--- a/app/jobs/descriptive_metadata_import_job.rb
+++ b/app/jobs/descriptive_metadata_import_job.rb
@@ -23,7 +23,7 @@ class DescriptiveMetadataImportJob < GenericJob
                        .bind { |description| validate_changed(cocina_object, description) }
                        .bind { |description| open_version(cocina_object, description) }
                        .bind { |description, new_cocina_object| CocinaValidator.validate_and_save(new_cocina_object, description:) }
-                       .bind { VersionService.close(identifier: cocina_object.externalIdentifier) }
+                       .bind { |new_cocina_object| close_version(new_cocina_object) }
                        .either(
                          ->(_updated) { success.call('Successfully updated') },
                          ->(messages) { failure.call(messages.to_sentence) }
@@ -43,6 +43,14 @@ class DescriptiveMetadataImportJob < GenericJob
     cocina_object = open_new_version_if_needed(cocina_object, 'Descriptive metadata upload')
 
     Success([description, cocina_object])
+  rescue RuntimeError => e
+    Failure([e.message])
+  end
+
+  def close_version(cocina_object)
+    VersionService.close(identifier: cocina_object.externalIdentifier)
+
+    Success()
   rescue RuntimeError => e
     Failure([e.message])
   end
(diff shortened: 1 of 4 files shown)


app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -23,6 +23,7 @@ def perform(bulk_action_id, params)
                        .bind { |description| validate_changed(cocina_object, description) }
                        .bind { |description| open_version(cocina_object, description) }
                        .bind { |description, new_cocina_object| CocinaValidator.validate_and_save(new_cocina_object, description:) }
+                       .bind { VersionService.close(identifier: cocina_object.externalIdentifier) }
```
jcoyne, line 26: Since `VersionService.close` doesn't return a promise, I don't think you can chain this result here, can you?  Can you write a test for `VersionService.close` raising `UnexpectedResponse`?  Does it call the failure lambda?

jcoyne (changes requested):

app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -23,6 +23,7 @@ def perform(bulk_action_id, params)
                        .bind { |description| validate_changed(cocina_object, description) }
                        .bind { |description| open_version(cocina_object, description) }
                        .bind { |description, new_cocina_object| CocinaValidator.validate_and_save(new_cocina_object, description:) }
+                       .bind { |new_cocina_object| close_version(new_cocina_object) }
                        .either(
```
peetucket, line 27: `new_cocina_object` does not appear to be getting returned from the `CocinaValidator.validate_and_save(new_cocina_object, description:)` call despite the fact that https://github.com/sul-dlss/argo/blob/main/app/services/cocina_validator.rb#L16 calls https://github.com/sul-dlss/argo/blob/main/app/models/repository.rb#L14-L17, which does appear to return the object

app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -45,4 +46,12 @@ def open_version(cocina_object, description)
   rescue RuntimeError => e
     Failure([e.message])
   end
+
+  def close_version(cocina_object)
+    VersionService.close(identifier: cocina_object.externalIdentifier)
```
peetucket, line 51: `cocina_object` comes in as nil here, causing the tests to fail

app/jobs/descriptive_metadata_import_job.rb
jcoyne: can you put those two items in the same block?

peetucket: Yes, tried that locally and it does work, only downside I can see is that it makes the calls less atomic (i.e. update can succeeed, close can fail).  Not sure that has any real world implications though.

peetucket: Just pushed a commit with this change.  Though would still like to understand why it wasn't working before.

peetucket: I'm guessing that this line in the code is returning a Success or Failure without any return: https://github.com/sul-dlss/argo/blob/main/app/services/cocina_validator.rb#L17

while this method (for example), is returning a Success with data structures returned which are then available to pass down the chain: https://github.com/sul-dlss/argo/blob/main/app/jobs/descriptive_metadata_import_job.rb#L44

So I guess I'm trying to understanding how `to_result` works

justinlittman (approved):

peetucket:
I think there was a legitimate test failure after rebasing caused if an object failed to validate and save, but then still got closed.  I have now separated out the validate/save and closing actions so that the method chain will be halted if there is a failure in `validate_and_save`.  Tests passing again (but we should re test this on stage with actual objects).

app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -45,4 +46,19 @@ def open_version(cocina_object, description)
...
     Failure([e.message])
   end
+
+  def validate_and_save(cocina_object, description)
+    result = CocinaValidator.validate_and_save(cocina_object, description:)
+    return Success(cocina_object) if result.success?
+
+    Failure(["validate_and_save failed for #{cocina_object.externalIdentifier}"])
+  end
+
```
peetucket, line 56: `CocinaValidator` does not return the object, which we need to pass onto `close_version` ... this new method wraps that action and allows us to pass the cocina object in the success result

PR #3541, opened 2022-04-19
mutate cocina object when versioning in mods metadata updates

### LINKED ISSUE #3538: Fix versioning in ApplyModsMetadata
Currently, ApplyModsMetadata optionally changes the version of an item (using the VersionService). However, it doesn't update the cocina object with the new version. This means that it will overwrite the changed version when performing an update.

Instead, it should mutate the cocina object with the new version. See `GenericJob.open_new_version`.

peetucket:

Fixes #3538 - mutate objects when versioning in mods metadata update jobs

Todo: update tests

In progress

### DIFF
--- a/app/services/apply_mods_metadata.rb
+++ b/app/services/apply_mods_metadata.rb
@@ -47,7 +47,7 @@ class ApplyModsMetadata
       return
     end
 
-    version_object
+    @updated_cocina = open_new_version_if_needed
     update_metadata
 
     log.puts("argo.bulk_metadata.bulk_log_job_save_success #{cocina.externalIdentifier}")
@@ -58,6 +58,7 @@ class ApplyModsMetadata
   private
 
   attr_reader :apo_druid, :mods, :existing_mods, :cocina, :original_filename, :ability, :log
+  attr_accessor :updated_cocina
 
   # Log the error
   def log_error!(exception)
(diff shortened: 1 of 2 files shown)


app/services/apply_mods_metadata.rb
```diff
@@ -103,11 +102,6 @@ def equivalent_xml?(node1, node2)
                               ignore_attr_values: ['version', 'xmlns', 'xmlns:xsi', 'schemaLocation'])
   end
 
-  # Returns true if the given object is accessioned, false otherwise.
-  def accessioned?
-    (6..8).cover?(status)
-  end
-
   # Checks whether or not a DOR object is in accessioning or not.
```
peetucket, line 105: unused method

spec/services/apply_mods_metadata_spec.rb
```diff
@@ -170,17 +170,4 @@
...
-        allow(action).to receive(:status).and_return(i)
-        if [6, 7, 8].include?(i)
-          expect(action.send(:accessioned?)).to be_truthy
-        else
-          expect(action.send(:accessioned?)).to be_falsy
-        end
-      end
-    end
-  end
 end
```
peetucket, line 173: unused method

app/services/apply_mods_metadata.rb
```diff
@@ -66,28 +67,26 @@ def log_error!(exception)
...
-      log.puts("argo.bulk_metadata.bulk_log_unable_to_version #{cocina.externalIdentifier}") # totally unexpected
-      return
-    end
-    commit_new_version
-  end
-
   def update_metadata
-    object_client = Dor::Services::Client.object(cocina.externalIdentifier)
+    object_client = Dor::Services::Client.object(updated_cocina.externalIdentifier)
     object_client.metadata.update_mods(mods)
```
peetucket, line 72: use the new object when updating metadata

jcoyne: We're not using the updated object, just it's identifier.

app/services/apply_mods_metadata.rb
```diff
@@ -47,7 +47,7 @@ def apply
       return
     end
 
-    version_object
+    @updated_cocina = open_new_version_if_needed
```
jcoyne, line 50: I don't understand why we need to store this value.

peetucket:
Not necessary.  See #3538

PR #3550, opened 2022-04-20
use object title from solr instead of label from solr

### LINKED ISSUE #3547: Use the descriptive title instead of the object label (i.e. the Cocina label) on tracking sheets
Tracking sheets are usually generated by registering a druid and then pressing the "Tracking sheets" button on the registration form. This generates a PDF containing these fields:

- druid
- object label
- project name
- tags
- source id

Following the example of https://github.com/sul-dlss/dor-services-app/pull/3791 we should use the descriptive title to generate a title for tracking sheets instead of using the label.

**Why**

As discussed elsewhere (for example, https://github.com/sul-dlss/dor-services-app/issues/3760) the "label' field is never updated after registration. All places we've found where "label" is still used are places where users expect to see the title. This leads to confusion when people update a title and don't see the change reflected in the data display.

In this case, a user updated some descriptive titles after registration and expected to see the updated title when they regenerated their tracking sheets, only to see the same label.

peetucket:

Fixes #3547: use the same field we use to show object titles in Argo in the tracking sheets (instead of the solr field having the Fedora label).

See for the Solr field name we are indexing to: https://github.com/sul-dlss/dor_indexing_app/blob/main/app/indexers/descriptive_metadata_indexer.rb#L19

and the method used to generate (using `Cocina::Models::Builders::TitleBuilder`):

https://github.com/sul-dlss/dor_indexing_app/blob/main/app/indexers/descriptive_metadata_indexer.rb#L54

And the field name defined in Argo that we will now use in the tracking sheet to get this title: https://github.com/sul-dlss/argo/blob/main/app/models/solr_document.rb#L24

It was not tested.

### DIFF
--- a/spec/models/track_sheet_spec.rb
+++ b/spec/models/track_sheet_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe TrackSheet do
 
     before do
       allow(SearchService).to receive(:query)
-        .with('id:"druid:xb482ww9999"', rows: 1)
+        .with("id:\"druid:#{druid}\"", rows: 1)
         .and_return(response)
     end
 
@@ -31,7 +31,7 @@ RSpec.describe TrackSheet do
         allow(Argo::Indexer).to receive(:reindex_druid_remotely)
 
         allow(SearchService).to receive(:query)
-          .with('id:"druid:xb482ww9999"', rows: 1)
+          .with("id:\"druid:#{druid}\"", rows: 1)
           .and_return(response, second_response)
       end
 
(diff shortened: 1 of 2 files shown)


justinlittman (changes requested):
Test?

peetucket:
> Test?

Sure, there weren't any for the method I changed, so I added some.  Which revealed that "catkey" is probably never being added to the sheet, which is now fixed in this PR (and tested).

spec/models/track_sheet_spec.rb
```diff
@@ -11,7 +11,7 @@
 
     before do
       allow(SearchService).to receive(:query)
-        .with('id:"druid:xb482ww9999"', rows: 1)
+        .with("id:\"druid:#{druid}\"", rows: 1)
         .and_return(response)
```
peetucket, line 15: DRY up the druid (set above in a let statement)

justinlittman (approved):
Thanks for the extra testing goodness!

PR #3552, opened 2022-04-21
disable/enable tracking sheet button based on registration

### LINKED ISSUE #3549: Disable tracking sheets button until after registration.
The "Tracking sheets" button is on the registration form.  If you push it before you register, then you get an error.

peetucket:

Fixes #3549 - the tracking sheet button starts out disabled, and then gets enabled once registration is complete (and gets re-enabled if the user resets the form).

~~UPDATE: it starts as disabled, but it is possible for it be re-enabled when it shouldn't be ... briefly looking into a better way of enabling based on there being all successful registrations in the grid~~

Localhost browser

### DIFF
--- a/app/javascript/controllers/registration_controller.js
+++ b/app/javascript/controllers/registration_controller.js
@@ -202,6 +202,7 @@ export default class extends Controller {
         $('#data').jqGrid('setColProp','barcode_id',{ editable: edit });
         $('#data').jqGrid('setColProp','druid',{ editable: edit }); //, formatter: edit ? null : druidFormatter });
         $('#data').jqGrid('setColProp','label',{ editable: edit });
+        document.querySelector("#tracking-sheet-btn").disabled = edit;
       }
 
       stopEditing(autoSave) {
--- a/app/javascript/controllers/registration_controller.js
+++ b/app/javascript/controllers/registration_controller.js
@@ -2,7 +2,7 @@ import { Controller } from "stimulus"
 import DorRegistration from '../registration/register'
 
 export default class extends Controller {
-  static targets = ["gridShowLabel", "textShowLabel"]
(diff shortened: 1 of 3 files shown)


app/javascript/registration/grid.js
```diff
@@ -3,7 +3,6 @@ export function gridContext() {
 
   var statusFormatter = function(val, opts, rowObj) {
     if (val in $t.statusIcons) {
-      console.error("hey")
       var result = '<span class="' + $t.statusIcons[val] + '" title="' +
```
peetucket, line 6: Despite it's friendliness, I don't think we need this.

app/javascript/controllers/registration_controller.js
```diff
@@ -202,6 +202,7 @@ export default class extends Controller {
         $('#data').jqGrid('setColProp','barcode_id',{ editable: edit });
         $('#data').jqGrid('setColProp','druid',{ editable: edit }); //, formatter: edit ? null : druidFormatter });
         $('#data').jqGrid('setColProp','label',{ editable: edit });
+        document.querySelector("#tracking-sheet-btn").disabled = edit;
       }
```
peetucket, line 206: hook into the existing method used to toggle other bits of the form to disable/enable tracking sheet button

app/views/registrations/show.html.erb
```diff
@@ -90,7 +90,7 @@
     <div class="float-end">
       <button class="btn btn-link" data-action="click->registration#resetDialog">Reset</button>
 
-      <button class="btn btn-outline-primary" data-action="click->registration#trackingSheet">
+      <button id="tracking-sheet-btn" disabled class="btn btn-outline-primary" data-action="click->registration#trackingSheet">
```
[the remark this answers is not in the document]
peetucket: using a stimulus target now

jcoyne (approved):

PR #3558, opened 2022-04-21
set reading order on registration page

### LINKED ISSUE #3548: Add new content-type/viewing direction menu to the registration form
registration-form

Use the UI from editing the content type from the show page.  The viewing direction selector should appear if the content type is book or image.

peetucket:

Fixes #3548 - set the reading order and content type on the registration page separately and in the same way they are done on the bulk action and edit content type pages.

IMPORTANT: Once approved and merged, please go to https://github.com/sul-dlss/infrastructure-integration-test/pull/415 and merge that PR, as it will fix the affected integration test.

Todo:
- ~~Should "None" be allowed as an option in the drop down when selecting Viewing Order (when the menu is shown, for books or images)?  "None" is currently an option in the bulk and individual content type update forms, which allows you to leave reading order out for books/images.  If left out, there will be no `viewingDirection` set in cocina.~~ Yes, leave none as an option.  Selecting it for image produces no explicit direction in the model, setting it for books sets as left-to-right in the model.
- ~~How does the "Process : Content Type : XXXX" tag get set and does this need to fixed to deal with reading order for books/images?  It seems to work, but I don't know how it is getting added to the object.~~  The `Process : Content Type : XXX` tag is automatically added by DSA during registration and does not need to be sent or set by Argo.

When image or book is selected:

Screen Shot 2022-04-21 at 3 15 01 PM

When other types are selected:

Screen Shot 2022-04-21 at 3 15 06 PM

- Existing tests were updated
- Localhost browser


app/forms/content_type_form.rb
```diff
@@ -18,7 +18,7 @@ class ContentTypeForm < ApplicationChangeSet
     'webarchive-seed' => Cocina::Models::ObjectType.webarchive_seed
   }.freeze
 
-  DIRECTIONS = %w[right-to-left left-to-right].freeze
+  DIRECTIONS = %w[left-to-right right-to-left].freeze
```
peetucket, line 22: Isn't "left-to-right" the more common option?  If so, shouldn't it be first in the drop down menu?

app/forms/registration_form.rb
```diff
@@ -50,38 +46,5 @@ def administrative_tags
...
-      Cocina::Models::ObjectType.object
-    end
-  end
-
-  # helper method to get just the content type tag
-  def content_type_tag
-    content_tag = params[:tag].find { |tag| tag.start_with?('Process : Content Type') }
-    content_tag.split(':').last.strip
-  end
 end
```
peetucket, line 50: we can rid of all these special code to parse content type and reading direction out of a content tag, and instead just pass them in as form parameters

app/helpers/registration_helper.rb
```diff
@@ -23,10 +23,4 @@ def apo_list(permission_keys)
       [Array(doc['sw_display_title_tesim']).first, doc['id'].to_s]
     end
   end
-
-  def valid_content_types
-    # the content types selectable in registration are capitalized but match what is available when setting content types
-    #  in the bulk and item detail page for setting content types (exception: '3d' needs to go '3D')
-    Constants::CONTENT_TYPES.keys.map { |content_type| content_type == '3d' ? '3D' : content_type.capitalize }
-  end
 end
```
peetucket, line 26: this method is not used anymore

app/javascript/registration/register.js
```diff
@@ -50,6 +58,8 @@ export default function DorRegistration(initOpts) {
         'admin_policy' : this.apoId(),
         'project' : this.projectName(),
         'workflow_id' : this.workflowId(),
+        'content_type': this.contentType(),
+        'viewing_direction': this.viewingDirection(),
         'label' : data.label || ':auto',
```
peetucket, line 63: the new parameters from our registration form

app/views/registrations/show.html.erb
```diff
@@ -29,11 +29,18 @@
...
+        <%= select_tag :content_type, options_for_select(ContentTypeForm::CONTENT_TYPES), 'data-rcparam': 'contentType',
+                    class: 'form-select', data: { content_type_target: 'contentType', action: 'content-type#render' }  %>
+      </div>
+    </div>
+    <div data-content-type-target="directionRow" class="row mb-3">
+      <label for="viewing_direction" class="col-sm-3 col-form-label">Viewing Direction</label>
+      <div class="col-sm-9">
+        <%= select_tag :viewing_direction, options_for_select(ContentTypeForm::DIRECTIONS), 'data-rcparam': 'viewingDirection',
+                    include_blank: 'none', class: 'form-select', data: { content_type_target: 'direction' } %>
       </div>
```
peetucket, line 41: replace the single content type drop down with two, using the same stimulus logic to hide/show viewing direction as needed

spec/helpers/registration_helper_spec.rb
```diff
@@ -34,15 +34,4 @@
...
-    it 'returns the expected content types in the expected order' do
-      expect(valid_content_types).to eq ['Book (ltr)', 'Book (rtl)', 'File', 'Image', 'Map', 'Media', '3D', 'Document', 'Geo', 'Webarchive-seed']
-    end
-  end
-
-  it 'returns nothing when permission_keys is empty' do
-    expect(SearchService).not_to receive(:query)
-    expect(apo_list([])).to eq []
-  end
 end
```
peetucket, line 37: this method is not used anymore

lib/constants.rb
```diff
@@ -115,19 +115,6 @@ module Constants
...
-    'image' => Cocina::Models::ObjectType.image,
-    'map' => Cocina::Models::ObjectType.map,
-    'media' => Cocina::Models::ObjectType.media,
-    '3d' => Cocina::Models::ObjectType.three_dimensional,
-    'document' => Cocina::Models::ObjectType.document,
-    'geo' => Cocina::Models::ObjectType.geo,
-    'webarchive-seed' => Cocina::Models::ObjectType.webarchive_seed
-  }.freeze
-
   RESOURCE_TYPES = {
```
peetucket, line 118: we shouldn't need this constant anymore

lib/constants.rb
```diff
@@ -116,8 +116,7 @@ module Constants
   ].freeze
 
   CONTENT_TYPES = {
-    'book (ltr)' => Cocina::Models::ObjectType.book,
-    'book (rtl)' => Cocina::Models::ObjectType.book,
+    'book' => Cocina::Models::ObjectType.book,
     'file' => Cocina::Models::ObjectType.object,
```
peetucket, line 120: we are now using the constant to create the `Process : Content Type` on the server side (it used to be generated by the javascript before, but now our html has the cocina types and not strings)

app/forms/registration_form.rb
```diff
@@ -33,55 +33,13 @@ def cocina_model
...
     model_params[:administrative][:partOfProject] = params[:project] if params[:project].present?
 
     Cocina::Models::RequestDRO.new(model_params)
   end
 
-  # All the tags from the form except the project and content type, which are handled specially
-  def administrative_tags
-    params[:tag].filter { |t| !t.start_with?('Process : Content Type') }
-  end
-
```
peetucket, line 52: there is no longer a `Process : Content Type` tag coming from the reg form since we don't need it to compute the content type, so need to strip it anymore

spec/features/item_registration_spec.rb
```diff
@@ -23,7 +23,8 @@
     visit registration_path
     select '[Internal System Objects]', from: 'Admin Policy' # "uber APO"
     select 'goobiWF', from: 'Initial Workflow'
-    select 'Book (ltr)', from: 'Content Type'
+    select 'book', from: 'Content Type'
+    select 'left-to-right', from: 'Viewing Direction'
```
peetucket, line 28: select the new options in the reg form

spec/features/item_registration_spec.rb
```diff
@@ -78,7 +79,9 @@
       'project' => 'special division : project #4',
       'rights' => 'default',
       'source_id' => 'source:id1',
-      'tag' => ['Process : Content Type : Book (ltr)', 'tag : test', "Registered By : #{user.sunetid}"],
+      'content_type' => 'https://cocina.sul.stanford.edu/models/book',
+      'viewing_direction' => 'left-to-right',
+      'tag' => ['tag : test', "Registered By : #{user.sunetid}"],
       'workflow_id' => 'goobiWF'
```
peetucket, line 85: verify the params come through ok (but without the Process : Content Type tag, which we don't need or want anymore)

app/javascript/registration/register.js
```diff
@@ -26,6 +26,14 @@ export default function DorRegistration(initOpts) {
       return document.querySelector('[data-rcparam="workflowId"]').value
     },
 
+    contentType : function() {
+      return document.querySelector('[data-rcparam="contentType"]').value
+    },
+
+    viewingDirection : function() {
+      return document.querySelector('[data-rcparam="viewingDirection"]').value
```
jcoyne, line 34: Will this always return a viewingDirection since it's not checking to see if the form element is disabled?

app/javascript/registration/register.js
```diff
@@ -26,6 +26,14 @@ export default function DorRegistration(initOpts) {
       return document.querySelector('[data-rcparam="workflowId"]').value
     },
 
+    contentType : function() {
+      return document.querySelector('[data-rcparam="contentType"]').value
```
jcoyne, line 30: Do we need to have a special rcparam? Can we just use the `[data-content-type-target='contentType']`?

jcoyne (changes requested):

app/forms/content_type_form.rb
peetucket: verified with Andrew, swapping the order

app/javascript/registration/register.js
peetucket: It will, but it's getting dropped downstream when creating the cocina model unless it's a book or image tyoe.  I'll change this function to return null unless the element is enabled though so we don't even send it.

app/javascript/registration/register.js
peetucket: yeah, we can ditch that (also on the viewing direction element) and just use the data-content-type-target

app/javascript/registration/register.js
```diff
@@ -26,6 +26,17 @@ export default function DorRegistration(initOpts) {
       return document.querySelector('[data-rcparam="workflowId"]').value
     },
 
+    contentType : function() {
+      return document.querySelector('[data-content-type-target="contentType"]').value
+    },
+
+    viewingDirection : function() {
+      var element = document.querySelector('[data-content-type-target="direction"]')
```
jcoyne, line 34: Please use `const` rather than `var`

app/javascript/registration/register.js
```diff
@@ -26,6 +26,17 @@ export default function DorRegistration(initOpts) {
...
+      if (element.disabled)
+        return null
```
jcoyne, line 36: null can be omitted here.

jcoyne (changes requested):
Just 2 small JS style changes. Then looks good.

app/javascript/registration/register.js
peetucket: changed

jcoyne (approved):

PR #3565, opened 2022-04-22
Don't truncate error messages so short

### LINKED ISSUE #3523: Add click-to-expand functionality to display long errors in the banner message
I'm not sure if we've always done this, but error messages in the Argo page banners are being truncated. This is preventing users from being able to see the information they need to fix the issue when the later part of the message contains that information. 

Example:

truncated-error

If users could click to see the whole message it would split the difference between the long messages that contain user-facing information and the long messages that seem to be more like stack traces.

jcoyne:
The cookie can hold up to 4k of data, so use more of that space

Fixes #3523

### DIFF
--- a/app/controllers/datastreams_controller.rb
+++ b/app/controllers/datastreams_controller.rb
@@ -40,11 +40,11 @@ class DatastreamsController < ApplicationController
       # if the content is not well-formed xml, inform the user rather than raising an exception
       error_msg = 'The datastream could not be saved due to malformed XML.'
     rescue Dor::Services::Client::UnexpectedResponse => e
-      error_msg = e.errors.first['detail']
+      error_msg = e.errors.first['detail']&.truncate(1024) # Truncate to avoid cookie overflow. Cookies can hold 4k.
     end
 
     respond_to do |format|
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

app/controllers/datastreams_controller.rb
```diff
@@ -40,11 +40,11 @@ def update
       # if the content is not well-formed xml, inform the user rather than raising an exception
       error_msg = 'The datastream could not be saved due to malformed XML.'
     rescue Dor::Services::Client::UnexpectedResponse => e
-      error_msg = e.errors.first['detail']
+      error_msg = e.errors.first['detail']&.truncate(1024) # Truncate to avoid cookie overflow. Cookies can hold 4k.
```
justinlittman, line 43: Should this be configuration?

jcoyne: I find think that would just make it harder to track down.

justinlittman: Yeah, but we're unlikely to remember that it needs to be changed in multiple locations. (Not a blocker, just a comment.)

jcoyne: @justinlittman  it might vary by location because there is a different preamble in each place.

PR #3572, opened 2022-04-25
[refactor] Use a turbo frame to implement the apo registration options 

jcoyne:

This sets the stage to implement the new rights menu for the registration page.

### DIFF
--- a/app/controllers/apo_controller.rb
+++ b/app/controllers/apo_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ApoController < ApplicationController
+class ApoController < ApplicationController # rubocop:disable Metrics/ClassLength
   include Blacklight::FacetsHelperBehavior # for facet_configuration_for_field
 
   load_and_authorize_resource :cocina, parent: false, class: 'Repository', only: %i[edit update]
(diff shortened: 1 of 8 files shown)


config/routes.rb
```diff
@@ -208,9 +209,9 @@
   resource :registration, only: :show do
     collection do
       get 'tracksheet'
-      get 'collection_list'
-      get 'workflow_list'
-      get 'rights_list'
+      # get 'collection_list'
```
justinlittman, line 212: Cruft

justinlittman (approved):
Approve pending cruft removal.

app/controllers/apo_controller.rb
```diff
@@ -19,6 +19,38 @@ def new
     render layout: 'one_column'
   end
 
+  # Draw the form controls for collection, rights and initial workflow
+  def registration_options
```
jcoyne, line 23: This method consists of code copied from the three methods previously part of RegistrationController https://github.com/sul-dlss/argo/pull/3572/files#diff-99d7bf3698ba31b7aedf0f330940825d7f298fe8a94391d8e08281c8c64078d1L17-L77

PR #3573, opened 2022-04-25
[HOLD] Use cocina properties in bulk action -> register druids

### LINKED ISSUE #3526: Update bulk action registration to use cocina properties
The bulk action registration form (available in the bulk actions menu under "Register Druids" in the CSV section) is still tied to how the registration form worked two years ago, meaning it does not include:

- the new Cocina access rights properties
- reading order for image and book content types
- barcode

It also uses spaces in the headers and I think while we're updating the headers we should also spell out abbreviations, go all lower-case, and use `_` instead of spaces.

New headers:

- administrative_policy_object (required)
- collection (optional)
- initial_workflow (required)
- content_type (required)
- reading_order (required if "content_type" is "book" or "image")
- source_id (required)
- catkey (optional)
- barcode (optional)
- label (required unless a catkey has been entered)
- rights_view (required)
- rights_download (required)
- rights_location (required if "view" or "download" uses "location-based")
- project_name (optional)
- tags (optional, may repeat)

For validation, I think we are currently doing row-by-row reporting to the log. I'd like to test this after the new headers are in place to see what's being returned when errors occur. The reporting I would like:

- a downloadable CSV of the druids successfully registered 
  - already implemented, should continue to function the same way (including the same columns)
- reporting to the log which rows failed
  - since failure on this job means that no druid will have been assigned yet, include the source id in the error message

aaron-collier:

Closes #3526 

Requires review/testing by @andrewjbtw 

New field list displayed:

Screen Shot 2022-04-26 at 1 22 42 PM

Updated unit test.

### DIFF
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -104,7 +104,7 @@ class RegistrationCsvConverter
       access[:view] = row.fetch('rights_view')
       access[:download] = row.fetch('rights_download')
       access[:location] = row.fetch('rights_location') if [access[:view], access[:download]].include?('location-based')
-      access[:controlledDigitalLending] = row['rights_controlledDigitalLending'].presence || 'false'
+      access[:controlledDigitalLending] = ActiveModel::Type::Boolean.new.cast(row['rights_controlledDigitalLending']) if row['rights_controlledDigitalLending'].present?
     end
   end
 end
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -43,7 +43,7 @@ class RegistrationCsvConverter
     catalog_links = row['catkey'] ? [{ catalog: 'symphony', catalogRecordId: row['catkey'], refresh: true }] : []
 
     model_params = {
-      type: dro_type(row.fetch('content_type')),
+      type: dro_type(row.fetch('content_type').downcase),
       version: 1,
       label: row['catkey'] ? row['label'] : row.fetch('label'),
       administrative: {
@@ -73,19 +73,19 @@ class RegistrationCsvConverter
 
   def dro_type(content_type)
     case content_type
-    when 'Image'
+    when 'image'
       Cocina::Models::ObjectType.image
-    when '3D'
+    when '3d'
       Cocina::Models::ObjectType.three_dimensional
-    when 'Map'
+    when 'map'
       Cocina::Models::ObjectType.map
-    when 'Media'
+    when 'media'
       Cocina::Models::ObjectType.media
-    when 'Document'
+    when 'document'
       Cocina::Models::ObjectType.document
-    when /^Manuscript/
+    when /^manuscript/
       Cocina::Models::ObjectType.manuscript
-    when 'Book (ltr)', 'Book (rtl)'
+    when 'book', 'book (ltr)', 'book (rtl)'
       Cocina::Models::ObjectType.book
     else
       Cocina::Models::ObjectType.object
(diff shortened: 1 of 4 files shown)


app/services/registration_csv_converter.rb
```diff
@@ -86,4 +90,27 @@ def dro_type(content_type)
...
+  def access(row)
+    {}.tap do |access|
+      access[:view] = row.fetch('rights_view')
+      access[:download] = row.fetch('rights_download')
+      access[:location] = row.fetch('rights_location') if [access[:view], access[:download]].include?('location-based')
+      access[:controlledDigitalLending] = controlled_digital_lending(access)
+    end
+  end
+
+  def controlled_digital_lending(access)
```
jcoyne, line 110: Should this method end with a `?`

jcoyne (changes requested):

app/services/registration_csv_converter.rb
```diff
@@ -86,4 +90,27 @@ def dro_type(content_type)
...
+      structural[:hasMemberOrders] = [{ viewingDirection: row['reading_order'] }] if row['reading_order'].present?
+    end
+  end
+
+  def access(row)
+    {}.tap do |access|
+      access[:view] = row.fetch('rights_view')
+      access[:download] = row.fetch('rights_download')
+      access[:location] = row.fetch('rights_location') if [access[:view], access[:download]].include?('location-based')
+      access[:controlledDigitalLending] = controlled_digital_lending?(access)
```
jcoyne, line 106: This doesn't seem right.  It's not automatically CDL if stanford/none.  But it should optionally allow CDL to be set if these preconditions are met.

jcoyne (changes requested):

app/services/registration_csv_converter.rb
aaron-collier: @jcoyne Suggestions on a field or how to handle that? I was basing this on https://github.com/sul-dlss/argo/blob/223ef77fa4f559387d4973d568132f5523968245/app/services/cocina_dro_access.rb which uses `rights == 'cdl-stanford-nd'` do set controlledDigitalLending to true, but I don't see anything going the other way.

aaron-collier: Perhaps @andrewjbtw can weigh in on how he may prefer to pass in CDL via the spreadsheet.

andrewjbtw: Could we cover this with an optional column in the CSV for `rights_controlledDigitalLending`? But I would like to avoid making people enter "false" when they don't want CDL. It's such an exceptional status that I'd like to set things up so people only indicate when it's "true".

That could mean:

- only require the column for CDL if one of the items on the spreadsheet will have "true"
- if the column is present and the cell for a particular row is blank, treat that as "false"
- if the column is present and the cell for a particular row has "true", set CDL
- if the column is present and all cells are blank, treat that as "false"

andrewjbtw: And to be explicit, the "view:stanford" and "download:none" must also be set to be valid for CDL.

aaron-collier: I think an optional column with `view:stanford` and `download:none` validation would be fine. :+1:

jcoyne (changes requested):
Please add a test where rights_controlledDigitalLending is set.

aaron-collier:
@jcoyne tests added.

jcoyne (approved):

jcoyne:
Whomever merges, please squash these commits.

ndushay:
whoops - another PR merged with [HOLD] in the title

PR #3579, opened 2022-04-26
Adjust css for the resource file table to truncate filenames

### LINKED ISSUE #3328: adjust contents display on the item show page to handle long filenames
Long filenames push the contents display on the Argo show page past the limits of the grid. Following the example of H2, we should truncate the visible filename in the contents display table so that the grid maintains a standard size. 

The filename should remain a link and users should be able to see the whole filename when they hover over the filename. Clicking on the link should continue to pop up a modal where the whole filename is visible.

**Screenshot**

This is an example made to illustrate the current problem. There are real prod filenames that push past the standard Argo width in similar ways.

extended-contents

**H2 example**

H2 has a hover feature that keeps filenames from breaking the grid:
Screen Shot 2022-04-25 at 10 29 37 AM

aaron-collier:

Closes #3328 

This updates the resource file table to truncate filenames. A bit of positional adjustment was required as well. This may truncate the filenames too much, which may require some additional design.

Screen Shot 2022-04-26 at 1 45 50 PM

### DIFF
--- a/app/assets/stylesheets/show.scss
+++ b/app/assets/stylesheets/show.scss
@@ -14,12 +14,15 @@
     &.object-type-apo {
       background-color: $sky-light;
     }
+
     &.object-type-collection {
       background-color: $fog;
     }
+
     &.object-type-item {
       background-color: $illuminating-light;
     }
+
     &.object-type-virtual-object {
       background-color: $bay-light;
     }
@@ -30,6 +33,7 @@
       min-width: $button-width;
       margin-right: .5rem;
       margin-bottom: .5rem;
+
       &.small {
         @extend .small;
       }
@@ -66,6 +70,7 @@
   .open-close {
     width: 2rem;
     float: left;
+
     .btn {
       font-size: x-large;
       color: $color-cardinal-red;
@@ -85,12 +90,14 @@
       background-color: $gray-200;
       padding: 1rem;
       margin-bottom: 1rem;
+
       .dl-horizontal {
         dt {
           clear: left;
           float: left;
           width: 7rem;
         }
+
         dd {
           margin-left: 8rem;
         }
@@ -100,7 +107,8 @@
         }
 
         dd:before,
-        dd:after, {
+        dd:after,
+          {
           display: table;
           content: " ";
         }
@@ -109,10 +117,23 @@
           clear: both;
         }
       }
+
       .resource-files {
         background-color: $body-bg;
-        margin-left: 3rem;
-        width: 90%
+        margin-left: .5rem;
+        margin-right: .5rem;
+
+        tr {
+          td {
+            max-width: 30vw;
+            text-overflow: ellipsis;
+            overflow: hidden;
+
+            a {
+              white-space: nowrap;
+            }
+          }
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/show.scss
```diff
@@ -109,10 +117,23 @@
...
       .resource-files {
         background-color: $body-bg;
-        margin-left: 3rem;
-        width: 90%
+        margin-left: .5rem;
+        margin-right: .5rem;
+
+        tr {
+          td {
+            max-width: 0;
```
jcoyne, line 128: What is this doing?

app/assets/stylesheets/show.scss
```diff
@@ -133,10 +154,10 @@
     @extend .header;
   }
 
-  .caption-header > caption {
+  .caption-header>caption {
```
jcoyne, line 157: I think this is harder to read without the spaces.

app/assets/stylesheets/show.scss
aaron-collier: Without this, the filename cell fills to allow the full text. I'm open to suggestion, I'd have to find where I got that example, but it solved that issue for me.

app/assets/stylesheets/show.scss
aaron-collier: This is my editor auto-fixing spaces. I'm happy to adjust.

app/assets/stylesheets/show.scss
jcoyne: Please document this.

jcoyne: I'm a little concerned we're depending on an indeterminate part of the CSS spec that might not be implemented the same in every browser both now and in the future.

aaron-collier: I'm open to suggestion if you have any ideas on a more idiomatic way to handle this.

aaron-collier: I added a comment to the css. Not sure if that is sufficient, but I'm not sure where/how else it would make sense to document this change.

jcoyne: How about `max-width: 30vw;`?

aaron-collier: That's looks pretty good actually. Since I'm no css expert, can you please let me know how you came to that?

jcoyne: 30% of the viewport width is `30vw`.  There's a new thing called container queries, which is really what we want, but the support is only experimental and only in a few browsers: https://caniuse.com/css-container-queries

ndushay (approved):
LGTM but i'm no css expert.

app/assets/stylesheets/show.scss
jcoyne, line 128: @aaron-collier you resolved this, but it's still 0?

jcoyne (changes requested):

app/assets/stylesheets/show.scss
aaron-collier: @jcoyne apologies, I didn't realize that change didn't push (due to rebasing I think). Fixed.

mjgiarlo (approved):
LGTM. Would be good for @jcoyne to review again, since he is currently blocking.

jcoyne (approved):

PR #3581, opened 2022-04-26
[HOLD] Style disabled primary buttons as greyed out

### LINKED ISSUE #3577: Update inactive style for buttons and links
Currently the default style for buttons and links that are disabled is a pinkish red. See example: 

It should be made to be a light gray.  If possible accomplish this with a bootstrap variable.

mjgiarlo:
On HOLD until Andrew takes a look in QA and approves.

Fixes #3577

This commit changes the appearance of all primary buttons (including outline-primary buttons) such that if they are disabled, whether via CSS class (for link-y buttons) or pseudoclass (for button-y buttons), they are greyed out, making it clear to users that they are not able to be pushed.

Screenshot from 2022-04-26 12-40-39

Screenshot from 2022-04-26 12-39-54

CI

### DIFF
--- a/app/assets/stylesheets/bootstrap-overrides.scss
+++ b/app/assets/stylesheets/bootstrap-overrides.scss
@@ -32,3 +32,10 @@
   color: $body-color;
   opacity: 1.0;
 }
+
+// Style disabled outline-primary buttons as greyed out using the `disabled` pseudo-class
+.btn-outline-primary:disabled {
+  border-color: $btn-link-disabled-color;
+  color: $btn-link-disabled-color;
+  opacity: $btn-disabled-opacity;
+}
--- a/app/assets/stylesheets/variables.scss
+++ b/app/assets/stylesheets/variables.scss
@@ -16,6 +16,9 @@ $silver: #c5c5c5;
 $primary: $color-cardinal-red;
 $success: $barley-corn;
 
+// Buttons
+$btn-link-disabled-color: $sonic-silver;
+
 // Links
 $link-color: $bright-blue;
 $link-hover-color: $blue;
@@ -32,5 +35,4 @@ $accordion-button-active-bg: $sonic-silver;
 $accordion-button-focus-box-shadow: 0 0 0.25rem 0.25rem rgba($sonic-silver, 0.25);
 
 // Table
-
 $table-border-color: $silver-sand;


mjgiarlo:
@jcoyne I welcome yr thoughts on whether this is sufficiently bootstrappy or not. I was hoping to be able to delete more of our custom SCSS in this PR but doing so did not have the desired effect. I'm all ears.

README.md
```diff
@@ -41,8 +41,7 @@ Brings up app at localhost:3000 with some test data:
 
 ```
 gem install foreman
-docker compose up -d
-docker compose stop web
+docker compose up -d sdr-api techmd
```
jcoyne, line 44: Please move to a separate PR for discussion

jcoyne (approved):
CSS looks good if you remove the README.md changes.

README.md
mjgiarlo: Dropped the commit

jcoyne:
@mjgiarlo the new version of bootstrap on the main branch is going to make this much easier to do: https://github.com/twbs/bootstrap/blob/acf6ea74a74328bcaa9f1c354f27e602cfbb8968/scss/mixins/_buttons.scss#L34-L36

andrewjbtw:
The gray on the SearchWorks button is so faint that I find it hard to read. Does the contrast fit within accessibility guidelines? 

Also, the red-filled buttons are now no longer gray when inactive. I'm not sure if that was intended. 

This is what I see in QA:
Screen Shot 2022-04-26 at 1 33 09 PM

Compare with Prod:
Screen Shot 2022-04-26 at 1 33 26 PM

mjgiarlo:
@andrewjbtw In short, I'm not sure. We have at least three different styles of buttons (that I know of) that are affected by this change, and the issue in question doesn't make clear which buttons should get what style. I suspect the issue this is fixing could be clearer about how the different buttons (Tracking sheets, SearchWorks, and e.g. Purge) should appear, and that'll make it easier for me to test all three have the specified styling.

andrewjbtw:
This must be stemming from the registration form redesign, but we hadn't discussed the buttons in other contexts.

@astridu, could you take a look at these style choices when you have a chance?

jcoyne:
@mjgiarlo I think we only wanted this ticket to impact `.btn-outline-primary`.

mjgiarlo:
@jcoyne 💬 
> @mjgiarlo I think we only wanted this ticket to impact `.btn-outline-primary`.

OK, that'd reduce the area of effect in this PR, so I'll do that in the morning. Thanks!

mjgiarlo:
@andrewjbtw @jcoyne This look better?

# BEFORE

## Registration

trackingsheets_before

## Show

showpage_before

# AFTER

## Registration

trackingsheets_after

## Show

showpage_after

jcoyne:
Looks great to me 👍

mjgiarlo:
@andrewjbtw 💬 
> It's definitely better, but is there a contrast checker we could run?

Yep: https://webaim.org/resources/contrastchecker/?fcolor=767674&bcolor=FFFFFF

This color combination has a contrast ratio of 4.55:1. This passes the WCAG AA standard but not the WCAG AAA standard. Which I guess we're OK with given this change applies a set of colors we're already using (e.g., the SearchWorks button) to the outline-primary buttons.

andrewjbtw:
Thanks! It seems like we're ok then.

PR #3582, opened 2022-04-26
[HOLD] Use the new access UI element on the registration page

### LINKED ISSUE #3212: Update registration form to use two-menu access rights design
The two-menu design is described in #2967

jcoyne:
Fixes #3212

Locally, test suite.

Screen Shot 2022-04-26 at 3 11 39 PM


andrewjbtw:
The menu interaction looks good but no matter which settings I try I get a "Bad request" error when I try to register.
Screen Shot 2022-04-26 at 4 22 56 PM

andrewjbtw:
I haven't been able to find any HB error corresponding to my registration attempts.

jcoyne:
@andrewjbtw I had a javascript error in the deploy.  It should be right now.

andrewjbtw:
I went through all the main combinations of rights (not the weird edge cases) and everything but two of the combinations worked as expected:

1. When I tried to register as CDL, I got "view: stanford" and "download: none" even when I had CDL as "yes"
2. I get a 400 error for citation-only. I couldn't find an HB error, but I tried to figure out what happened by looking at the browser console and I think the request being sent my be missing the "download" parameter when citation-only is chosen.

justinlittman (approved):

PR #3585, opened 2022-04-26
Render Cocina on show page without metadata

### LINKED ISSUE #3583: Don't display "created", "modified", "lock" in the Cocina section of the Argo show page
Screen Shot 2022-04-26 at 1 23 09 PM

mjgiarlo:

Fixes #3583

This commit adds a new option to the `CocinaHashPresenter` that allows it to render a Cocina structure without metadata, and in this case it means without the lock, created timestamp, and modified timestamp. This new option is then used on the catalog show page. As part of this work, convert the CocinaHashPresenter spec to use factories instead of hand-crafted Cocina model instances and add a flag to the factories to retain past test behavior.

CI

### DIFF
--- a/app/presenters/cocina_hash_presenter.rb
+++ b/app/presenters/cocina_hash_presenter.rb
@@ -9,8 +9,8 @@
 #    it's super easy to coerce a hash into other data structures, so it gives us
 #    flexibility.
 class CocinaHashPresenter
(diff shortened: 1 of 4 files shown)


jcoyne (approved):

ndushay (approved):
LGTM

PR #3589, opened 2022-04-27
Display location for location-based access.

### LINKED ISSUE #3574: show location column when a file has location-based rights
The Argo show page is not showing the actual location for files with location-based rights:
Screen Shot 2022-04-25 at 11 16 45 PM

We should show the location when there is one but not when there isn't.

lwrubel:

Resolves #3574. Shows location value rather than "Location-based" in structural View and Download columns.

Specs and integration tests. 

Screen Shot 2022-04-27 at 8 29 14 AM

### DIFF
--- a/app/components/contents/file_component.rb
+++ b/app/components/contents/file_component.rb
@@ -23,11 +23,19 @@ module Contents
     delegate :publish, :shelve, :sdrPreserve, to: :administrative
 
     def view_access
-      access.view.capitalize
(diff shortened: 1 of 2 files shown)


jcoyne (approved):

lwrubel:
@andrewjbtw is this one you'd like to review before merge?

andrewjbtw:
I think this looks good based on the screenshot.

PR #3605, opened 2022-04-28
Recognize parallel titles as valid titles.

### LINKED ISSUE #3600: Recognize parallel titles as valid titles when uploading the Cocina descriptive CSV (single and bulk)
**Describe the bug**
The Argo cocina descriptive spreadsheet upload appears to be rejecting CSVs with parallel titles. I'm trying to upload metadata with the following header row:

`source_id,title1.parallelValue1.structuredValue1.type,title1.parallelValue1.structuredValue1.value,title1.parallelValue1.structuredValue2.type,title1.parallelValue1.structuredValue2.value,title1.parallelValue2.value,title2.parallelValue1.type,title2.parallelValue1.value,title2.parallelValue2.type,title2.parallelValue2.value,note1.type,note1.value,purl`

This is a scaled down version of the descriptive spreadsheet you can obtain from downloading the descriptive metadata for https://argo.stanford.edu/view/gn988hy1229 I have not changed the title headers but I did remove many other columns to isolate the issue.

When I try to upload that spreadsheet I get the error "title column not found"
Screen Shot 2022-04-27 at 5 38 59 PM

**User Impact**
It's not possible to update metadata that uses parallel titles.
 
**To Reproduce**
1. Go to https://argo-stage.stanford.edu/view/druid:bc217ky7586
2. Download the cocina descriptive spreadsheet for that item
3. Upload the spreadsheet you just downloaded
4. See the error

**Expected behavior**
Parallel titles should be treated as valid.

**Additional context**
There may be other title structures that Argo needs to account for when determining validity.

justinlittman:
closes #3600

Allow parallel titles, which are valid for descriptive spreadsheets.

Unit

### DIFF
--- a/app/services/description_validator.rb
+++ b/app/services/description_validator.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 # Validate the descriptive metadata spreadsheet
+# rubocop:disable Metrics/ClassLength
 class DescriptionValidator
   def initialize(csv, bulk_job: false)
     @csv = csv
@@ -41,10 +42,7 @@ class DescriptionValidator
   end
 
   def validate_title_headers
-    if @headers.include?('title1.value') ||
-       (@headers.exclude?('title1.value') && @headers.include?('title1.structuredValue1.type') && @headers.include?('title1.structuredValue1.value'))
-      return
-    end
+    return if title_value_header? || title_structured_value_header? || title_parallel_value_header?
 
     errors << 'Title column not found.'
   end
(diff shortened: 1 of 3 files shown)


app/services/description_validator.rb
```diff
@@ -40,12 +40,13 @@ def validate_duplicate_headers
...
-      return
-    end
+    title_count = 0
+    title_count += 1 if title_value_header?
+    title_count += 1 if title_structured_value_header?
+    title_count += 1 if title_parallel_value_header?
 
-    errors << 'Title column not found.'
+    errors << 'Title column not found.' if title_count.zero?
+    errors << 'Multiple title columns found.' if title_count > 1
```
jcoyne, line 49: I believe multiple title columns is valid. Consider one row has a parallel title and one row has a structured title.

jcoyne (changes requested):

app/services/description_validator.rb
justinlittman: That makes sense. The previous code prevented it (notice the `exclude`) so I assumed that was intentional. Removing this check.

jcoyne (approved):

Issue #3528, opened 2022-04-19
Display "Agreement" name and link to the agreement on the APO show page


andrewjbtw:
One more addition to the APO show page that I missed in earlier ticketing: show the "Agreement" for that APO and link to it. The row should be in the "Details" pane, positioned under "# of items" and above "Tags".

<img width="668" alt="agreement-details" src="https://user-images.githubusercontent.com/8975583/163901709-64f40441-1761-44b6-9947-6c33f5381e58.png">

PR #3617, opened 2022-04-29
another attempt to fix flappy spec

peetucket:

Turbo links pull in the link name, which can sometimes be flappy.  This sleeps instead of waits.  1 second passed locally and failed in CI.  Trying for 3 seconds.

Existing tests

### DIFF
--- a/spec/features/apo_spec.rb
+++ b/spec/features/apo_spec.rb
@@ -55,10 +55,10 @@ RSpec.describe 'Create an apo', js: true do
     click_button 'Register APO'
     expect(page).to have_text 'created'
 
-    # Shows a link to the agreement
-    expect(page).to have_link 'Test Agreement',
-                              href: solr_document_path(agreement.externalIdentifier),
-                              wait: 30 # This link is rendered in a turbo-frame, so wait longer than the default
+    # Shows a link to the agreement. The agreement name is loaded by turbo-links, so pause to let it complete
+    # (capybara wait is flaky, see https://discuss.hotwired.dev/t/capybara-wait-for-ajax-replacement-for-turbo-stream-responses/2269/9)
+    sleep 3
+    expect(page).to have_link 'Test Agreement', href: solr_document_path(agreement.externalIdentifier)
 
     click_on 'Edit APO'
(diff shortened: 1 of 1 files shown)


spec/features/apo_spec.rb
```diff
@@ -56,9 +56,7 @@
     expect(page).to have_text 'created'
 
     # Shows a link to the agreement
-    expect(page).to have_link 'Test Agreement',
-                              href: solr_document_path(agreement.externalIdentifier),
-                              wait: 30 # This link is rendered in a turbo-frame, so wait longer than the default
+    expect(page).to have_link href: solr_document_path(agreement.externalIdentifier)
```
jcoyne, line 59: Has this previous one failed? Can you show me where?

jcoyne: Ah, this one: https://app.circleci.com/pipelines/github/sul-dlss/argo/4724/workflows/fa379774-4058-4999-884e-bffdc341ed41/jobs/10332 ?

jcoyne: I feel like this just completely removes the behavior I was testing for in the first place.

peetucket:
Yes, it does change the behavior of test, though it does confirm that the agreement is on the page and is associated with the APO, which is something (ie it verifies that the link is there, the text just hasn't been filled in by turbo).

jcoyne:
@peetucket I wrote the test to show that the name of the agreement was on the page. That was the ticket: https://github.com/sul-dlss/argo/issues/3528

jcoyne:
It seems like this is a problem other people have had too https://discuss.hotwired.dev/t/capybara-wait-for-ajax-replacement-for-turbo-stream-responses/2269/9

peetucket:
> It seems like this is a problem other people have had too https://discuss.hotwired.dev/t/capybara-wait-for-ajax-replacement-for-turbo-stream-responses/2269/9

So maybe a sleep in this case would be better than the wait?

jcoyne:
I think we could try that.

peetucket:
Sleep instead of wait...

jcoyne (approved):

PR #3622, opened 2022-05-02
Handle csv jobs missing druid column without repeating.

### LINKED ISSUE #3619: Do not retry a descriptive metadata bulk action that fails because of: RuntimeError: Column "druid" not found
Currently, if someone forgets to include a druid column on the descriptive CSV, the bulk action will get caught in an endless loop of retries with the error: `RuntimeError: Column "druid" not found`
https://app.honeybadger.io/projects/49894/faults/85168602

Since retrying cannot fix this error and it will apply to all rows, the job should fail and report the error back to the log so the user can see it.

justinlittman:
closes sul-dlss/dor-services-app#3619

Avoid unnecessary job retries.

Unit

### DIFF
--- a/app/jobs/generic_job.rb
+++ b/app/jobs/generic_job.rb
@@ -119,6 +119,8 @@ class GenericJob < ApplicationJob
   def with_csv_items(csv, name:, druid_column: 'druid')
     update_druid_count(count: csv.size)
     with_bulk_action_log do |log|
+      return unless check_druid_column(csv:, druid_column:, log:, bulk_action:)
+
       csv.each.with_index(2) do |csv_row, row_num|
         druid = csv_row.fetch(druid_column)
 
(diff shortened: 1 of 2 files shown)


app/jobs/generic_job.rb
```diff
@@ -134,11 +134,14 @@ def with_csv_items(csv, name:, druid_column: 'druid')
         cocina_object = Repository.find(druid)
         yield(cocina_object, csv_row, success, failure, row_num)
       rescue KeyError
-        raise "Column \"#{druid_column}\" not found"
+        raise
```
jcoyne, line 137: Can we just do `failure.call(...)` here? I think that would be easier to follow.

jcoyne (changes requested):

app/jobs/generic_job.rb
jcoyne: @justinlittman If this is for the whole sheet, can we validate it before making a background job (in the controller) and give instant feedback to the user?

justinlittman: Yes, but ticketed separately: https://github.com/sul-dlss/dor-services-app/issues/3971

Even if that is done, this is still necessary as backup as there is nothing that guarantees that this will be invoked with a csv with a druid column.

jcoyne: Can we just move the check to above the loop then?

justinlittman: Yes. Done.

jcoyne (approved):

PR #3633, opened 2022-05-02
Allow excel files to be uploaded for description update

### LINKED ISSUE #3620: Add support for Excel (.xlsx) to single item descriptive upload
We have support for Excel spreadsheets in the bulk descriptive upload but it does not appear that we support them for the single item upload. The single item upload prompts users to add CSV files and grays out all other file types. If a user figures out how to select an Excel file anyway, clicking on the "Update" button results in nothing happening.

For consistency, we should support Excel files for single-item uploads.

aaron-collier:

Closes #3620 by allowing `.xlsx` file extensions in the file open dialog and then uses `CsvUploadNormalizer` to parse either a CSV or Excel sheet into csv data.

### DIFF
--- a/spec/features/upload_descriptive_metadata_spec.rb
+++ b/spec/features/upload_descriptive_metadata_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe 'Descriptive metadata spreadsheet upload', js: true do
     "title1.value,purl\nmy title,https://purl.stanford.edu/#{Druid.new(item.externalIdentifier).without_namespace}\n"
   end
   let(:file) do
-    Tempfile.new('upload.csv').tap do |file|
+    Tempfile.new(%w[upload .csv]).tap do |file|
       file.write csv
       file.close
     end
--- a/spec/features/upload_descriptive_metadata_spec.rb
+++ b/spec/features/upload_descriptive_metadata_spec.rb
@@ -13,32 +13,65 @@ RSpec.describe 'Descriptive metadata spreadsheet upload', js: true do
     FactoryBot.create_for_repository(:persisted_item)
   end
 
-  let(:csv) do
-    "title1.value,purl\nmy title,https://purl.stanford.edu/#{Druid.new(item.externalIdentifier).without_namespace}\n"
-  end
-  let(:file) do
-    Tempfile.new(%w[upload .csv]).tap do |file|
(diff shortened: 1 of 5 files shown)


spec/features/upload_descriptive_metadata_spec.rb
```diff
@@ -17,7 +17,7 @@
     "title1.value,purl\nmy title,https://purl.stanford.edu/#{Druid.new(item.externalIdentifier).without_namespace}\n"
   end
   let(:file) do
-    Tempfile.new('upload.csv').tap do |file|
+    Tempfile.new(%w[upload .csv]).tap do |file|
```
aaron-collier, line 20: Note: using an array in this call honors the file extension, so now the tempfile will look like `upload2022-05-02.csv` instead of `upload.csv2022-05-02` which causes an unsupported filetype error on upload.

jcoyne (changes requested):
Can you please add a test that shows the DescriptivesController#upload can take an Excel file?

aaron-collier:
@jcoyne test added, ready for a re-review.

spec/features/upload_descriptive_metadata_spec.rb
```diff
@@ -13,32 +13,65 @@
...
   end
 
-  it 'uploads descriptive' do
-    click_button 'Manage description'
-    click_link 'Upload Cocina spreadsheet'
+  context 'with an excel file' do
+    let(:file) do
+      Tempfile.new(%w[upload .xlsx]).tap do |file|
+        # Open the temp file as an XLSX workbook
+        workbook = WriteXLSX.new(file.path)
```
jcoyne, line 51: Instead of adding a new dependency, can you just add a sample xls in `spec/fixtures/files` ?

jcoyne (changes requested):

spec/features/upload_descriptive_metadata_spec.rb
aaron-collier: No - it requires a matching purl, so we need to be able to write it at the time of the test the same way we do with the CSV.

jcoyne: Okay.

jcoyne (approved):

PR #3645, opened 2022-05-04
adjust test syntax for flaky test

peetucket:

Previous flaky test fix was potentially looking for nil text, which was sometimes not the case when turbo actually filled it in?  Just speculation.  This version again passes locally and in the CI branch consistently.

Tests


jcoyne (changes requested):
If it's the scrolling that is the problem, can you remove the sleep?

peetucket:
> If it's the scrolling that is the problem, can you remove the sleep?

sure, can try that ... though of course it might be both issues .... we shall see

jcoyne (approved):

PR #3653, opened 2022-05-04
Registration form redesign

### LINKED ISSUE #3610: Registration page : update to use form without a grid
Remove the existing grid and replace with form fields in a row, and the ability to add more rows (similar to what is done in H2 for adding authors).

jcoyne:

Fixes #3610

Locally and test suite

Screen Shot 2022-05-05 at 4 01 13 PM


andrewjbtw:
This looks good when all the data entered is valid. I ran into a couple of problems when I tried invalid data like duplicate source ids:

- The "Reset" link seems to be submitting data rather than resetting the form. I'm not sure it serves a function now in the new form system given that the trashcan icon interaction can clear the fields. If we do keep it, I think it should set everything to what you'd get if you hadn't entered any item information yet.
- It looks like the form saves data even if there's an error for one or more of the items. The result is that you get a druid for the items that don't have an error but the UI keeps you on the form:
Screen Shot 2022-05-10 at 5 11 35 PM
- When you hit an error and the UI keeps you on the form, the trash icons are no longer visible, so you can't remove any of the items that are listed

I'm not sure what the best action is for errors. I think either we should take people to the post-registration page and display what was and was not registered or we should prevent anything from being registered if any one item will fail registration. That would keep people on the form and give them a chance to correct errors, so I think that would be the better way to go.

jcoyne:
@andrewjbtw while we certainly can prevent "you entered a the same sourceId multiple times on the page", we still need to handle the error "you entered a sourceId that has already been used".   How do we show that sort of error for one of the records if another of the records would register fine?

andrewjbtw:
I met with Astrid earlier this week:

Her recommendation for handling errors in the UI:

- User fills out form
- Validate what can be validated on the form itself:
  - required fields are present
  - source id uses the correct format
  - no duplicate source ids within the set of items on the form itself
- User submits registration
- Some items succeed and some fail
- User stays on the form, but the form now shows
  - Items that failed are highlighted in red and remain editable
  - Items that succeeded remain on the form, but are grayed out so they cannot be edited
- User corrects or removes the items with errors
- User submits form again, sending data only for the items that didn't succeed before
- If remaining items succeeded, user gets the page that shows the list of what was registered

The reasoning behind this proposal:

- Gives users the opportunity to correct problems in place, within the same batch they were working on
- Avoids sending users to an error log which they then have to parse themselves before going back to fill out the form again
- If a user can't work out how to fix an error, they can remove that item from the form and allow the rest of the batch to finish

If it's not possible to handle errors within the single form interface, then the next option would be to send the user to the results page and show successes and failures in separate lists on the page. Or to provide a log for the errors.

ndushay:
@jcoyne extremely gentle reminder that this PR is still open due to error handling.  If you and Andrew share an understanding of what remains to be done, great.  If not, and if I can help in any way, I'm happy to do so.

jcoyne:
@ndushay those items have been addressed. This is the base for #3684 and it's been waiting for @andrewjbtw to approve.  We put it on stage last week, and Andrew said he would look this week.

andrewjbtw:
@jcoyne I found one more issue in testing. When submitting an invalid barcode in the form, I get sent to a 500 page instead of seeing an error message. I think an invalid barcode should be handled like an invalid source id and prevent form submission.

andrewjbtw:
The catkey behaves slightly differently:

- enter an invalid catkey format (not only integers), you get a 500 page
- enter a catkey that's not found, you get returned to the form with an error message

Being returned to the form is a problem because if there are multiple entries, the valid ones are submitted and get registered anyway. It may be better to fall back on the label as the title and register the item anyway if the catkey lookup fails. That would give the users the list of druids that registered and they could try to correct the catkey on the show page.

andrewjbtw:
Leaving a note here to say that based on discussion in standup, I'm going to re-write my comment above as follow-on tickets. That will allow this PR to be merged and make it so the follow-on tickets can go on the board in the current workcycle, plus allow us to get some user feedback on the new UI sooner. And relieve the need to keep rebasing this PR.

As far as I can tell, we are catching all validation issues in the CSV upload via the log file, while the form is catching all source id and label validation needs. That's the key functionality we need to go ahead.

ndushay:
It appears the code in this PR is included in PR #3684;   I'm making this a draft to avoid problematic merges

ndushay (approved):

PR #3654, opened 2022-05-04
fix bug where tags were not making it during registration

### LINKED ISSUE #3650: Tags are not being added to newly registered items
**Describe the bug**
Tags are not being assigned to newly registered items, with the exception of the project tag. If you enter any tags in the "Tags" boxes on the registration form, they are not carried over to the druids being registered.

**User Impact**
Users have to go back and add tags later.

**To Reproduce**
1. Go to the registration form
2. Fill it out, including tags in the "Tags" box
3. Register
4. See that the druids do not have the tags from the Tags box. The Project tag (in the Project box) is being included, but not the other tags.

**Expected behavior**
Tags should be assigned to the new druids.

**Additional context**
This seems to have started this morning.

peetucket:

Fixes #3650 - tags were not making it the object because the params being sent by javascript was "tag" and the params expected by the server were "tags" (started with changes made in https://github.com/sul-dlss/argo/pull/3637 )...I normalized to "tags" , which makes more sense.

Localhost browser.  Updated integration tests to catch this in the future: https://github.com/sul-dlss/infrastructure-integration-test/pull/426

### DIFF
--- a/app/javascript/registration/register.js
+++ b/app/javascript/registration/register.js
@@ -41,7 +41,7 @@ export default function DorRegistration(initOpts) {
       return document.getElementById('rights').value
     },
 
-    // Grab list of tags from the form and rejects blanks
+    // Grab list of tags from the form
     tags: function() {
         return Array.from(document.querySelectorAll('[data-controller="tag-validation"]')).map((elem) => elem.value)
     },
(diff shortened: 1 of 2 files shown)


app/javascript/registration/register.js
```diff
@@ -41,7 +41,7 @@ export default function DorRegistration(initOpts) {
       return document.getElementById('rights').value
     },
 
-    // Grab list of tags from the form and rejects blanks
+    // Grab list of tags from the form
```
peetucket, line 44: it's not rejecting blanks anymore!  they get rejected on the server now

jcoyne (changes requested):
Can you write a feature spec for this? We should be able to register an item with tags and show them on the page.

jcoyne (approved):

jcoyne:
Feature test to come in another PR, to follow.

PR #3660, opened 2022-05-05
Add back to search on show page.

### LINKED ISSUE #3657: Add "back to search results" to show pages
Add back the "back to search results" link that used to appear when you look at individual search results.

I'm not sure when the link disappeared from the top of the show pages, but it may even have been last year. I'll admit that I don't remember what it looked like, but it may have been near the "previous" and "next" links.

Screen Shot 2022-05-05 at 9 43 14 AM

I can tag this as "design needed" if necessary, but if there's a standard Blacklight way to do this that's consistent with the current design that's probably enough to meet the need.

justinlittman:
closes #3657

It used to be there.

image

### DIFF
--- a/app/assets/stylesheets/show.scss
+++ b/app/assets/stylesheets/show.scss
@@ -7,6 +7,18 @@
 
   }
 
+  .search-links {
+    border-bottom: $pagination-border-width solid $pagination-border-color;
+    margin-bottom: 1em;
+    padding-bottom: 1em;
+
+    .pagination-search-widgets {
+      border-bottom-width: 0px;
+      margin-bottom: 0em;
+      padding-bottom: 0em;
+    }  
+  }
+
   .object-type {
     width: 15rem;
     height: 2.4rem;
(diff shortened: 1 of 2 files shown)


app/views/catalog/_previous_next_doc.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="row search-links">
+  <div class="col-md-10">
+    <%= render(Blacklight::SearchContextComponent.new(search_context: @search_context, search_session: search_session)) %>
+  </div>
+  <div class="col-md-2">
+    <%= link_back_to_catalog label: 'Back to search', class: 'btn btn-default' %>
```
jcoyne, line 6: I would not put this in this partial, because the momentum has been to do away with overridable partials and switch to components.

justinlittman: But to invoke the new component don't you still need to override this partial?

jcoyne: Not if blacklight does away with this partial. I'd prefer we move this into the show page.

jcoyne (approved):

PR #3661, opened 2022-05-05
force reindex after desc CSV metadata upload

### LINKED ISSUE #3656: Refresh the show page after a descriptive metadata upload
After uploading a descriptive metadata CSV on an Argo show page, the item is reindex but the show page itself isn't refreshed. This makes it look like the data hasn't changed when it actually has.

To reproduce:

1. Go to an item show page
2. Download the Cocina descriptive spreadsheet via the "Manage Description" button
3. Change something that will be visible on the show page, like the title
4. Upload the spreadsheet
5. See the "success" message
6. See that the display hasn't changed
7. Refresh the page (but don't reindex)
8. See the change in display

peetucket:

Fixes #3656 - reindex will show new updated metadata

Note: I was unable to reproduce the described behavior of refreshing the page showing updated metadata, but re-indexing show it. The current code does a redirect, this forces a re-index right before the redirect.

Localhost


jcoyne (changes requested):

jcoyne (approved):
Please rebase, then it should pass tests

peetucket:
Rebased, tests passed, merging.

PR #3677, opened 2022-05-09
Updates to support blacklight 7.24

### LINKED ISSUE #3676: unpin blacklight
pinned to 7.24.x keep build passing, after 7.25.1 broke the build

see https://github.com/sul-dlss/argo/pull/3670/commits/21315db19516ab98e4f53fe33a7be9b7ee87d4b8 from https://github.com/sul-dlss/argo/pull/3670

jcoyne:

Blacklight 7.24 requires a configuration change.

Fixes #3676 

ci


jcoyne:
SearchState is a Blacklight concept of all the inputs that affect a search result. It can be serialized to query parameters, but there may be query parameters that are not represented in the search state.

PR #3681, opened 2022-05-10
adjust catkey form to use multiple rows

### LINKED ISSUE #3358: Update catkey form
See https://github.com/sul-dlss/cocina-models/pull/373

So that users can choose the "refresh" or "do not refresh" setting for any associated catkey.  See comments below with design.

peetucket:

Fixes #3358 - update catkey editing form so it allows for adding/removing catkeys and setting of refresh/no refresh.  Design in the ticket: https://github.com/sul-dlss/argo/issues/3358#issuecomment-1116329820

As implemented:

Screen Shot 2022-05-11 at 2 03 14 PM

Also: validate that there is no more than a single catkey set to refresh and that any entered catkey values are valid..

~~HOLD for PO testing~~

- Localhost browser
- Updated existing tests for new UI
- Added many more tests for all of the scenarios (deleting, updating, setting refresh)

### DIFF
--- a/app/controllers/catkeys_controller.rb
+++ b/app/controllers/catkeys_controller.rb
@@ -4,7 +4,7 @@ class CatkeysController < ApplicationController
   load_and_authorize_resource :cocina, parent: false, class: 'Repository', id_param: 'item_id'
 
   def edit
-    @form = catkey_form
+    @form = CatkeyForm.new(@cocina)
     @form.prepopulate!
     respond_to do |format|
       format.html { render layout: !request.xhr? }
@@ -14,29 +14,13 @@ class CatkeysController < ApplicationController
   def update
     return unless enforce_versioning
 
-    @form = catkey_form
-    if @form.validate(params[:catkey]) && @form.save(@cocina)
+    @form = CatkeyForm.new(@cocina)
+    if @form.validate(params[:catkey]) && @form.save
       Argo::Indexer.reindex_druid_remotely(@cocina.externalIdentifier)
       msg = "Catkeys for #{@cocina.externalIdentifier} have been updated!"
       redirect_to solr_document_path(@cocina.externalIdentifier, format: :html), notice: msg
     else
       render turbo_stream: turbo_stream.replace('modal-frame', partial: 'edit'), status: :unprocessable_entity
     end
-  end
-
-  private
-
-  def catkey_form
-    # fetch catkeys from object
-    object_catkeys = @cocina.identification.catalogLinks.filter_map { |catalog_link| catalog_link if catalog_link.catalog == Constants::SYMPHONY }
-
-    # form is initialized with catkeys in the object
-    catkeys = object_catkeys.map { |catkey| CatkeyForm::Row.new(value: catkey.catalogRecordId, refresh: catkey.refresh) }
-    CatkeyForm.new(
-      CatkeyForm::ModelProxy.new(
-        id: params[:item_id],
-        catkeys:
-      )
-    )
   end
 end
--- a/app/controllers/catkeys_controller.rb
+++ b/app/controllers/catkeys_controller.rb
@@ -3,15 +3,44 @@
 class CatkeysController < ApplicationController
   load_and_authorize_resource :cocina, parent: false, class: 'Repository', id_param: 'item_id'
 
-  rescue_from Dor::Services::Client::UnexpectedResponse do |exception|
-    md = /\((.*)\)/.match exception.message
-    detail = JSON.parse(md[1])['errors'].first['detail']
-    redirect_to solr_document_path(params[:item_id]),
-                flash: { error: "Unable to retrieve the cocina model: #{detail.truncate(200)}" }
+  ### classes that define a virtual catkey model object and data structure, used in form editing...persistence is in the cocina model
+  class ModelProxy
+    def initialize(id:, catkeys:)
+      @id = id # the object ID
+      @catkeys = catkeys # the array of catkey objects (defined in custom class below)
+    end
+
+    attr_reader :id, :catkeys
+
+    def to_param
+      @id
+    end
+
+    def persisted?
+      true
+    end
   end
 
+  class CatkeyRow
+    attr_accessor :value, :refresh, :id
+
+    def initialize(attrs = {})
+      @id = attrs[:value]
+      @value = attrs[:value]
+      @refresh = attrs[:refresh]
+    end
+
+    def persisted?
+      id.present?
+    end
+
+    # from https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/activerecord/lib/active_record/nested_attributes.rb#L382-L384
+    def _destroy; end
+  end
+  ###
+
   def edit
-    @change_set = CatkeyForm.new(@cocina)
+    @form = catkey_form
     respond_to do |format|
       format.html { render layout: !request.xhr? }
     end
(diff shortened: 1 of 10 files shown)


app/controllers/catkeys_controller.rb
```diff
@@ -3,15 +3,44 @@
 class CatkeysController < ApplicationController
   load_and_authorize_resource :cocina, parent: false, class: 'Repository', id_param: 'item_id'
 
-  rescue_from Dor::Services::Client::UnexpectedResponse do |exception|
-    md = /\((.*)\)/.match exception.message
-    detail = JSON.parse(md[1])['errors'].first['detail']
-    redirect_to solr_document_path(params[:item_id]),
-                flash: { error: "Unable to retrieve the cocina model: #{detail.truncate(200)}" }
+  ### classes that define a virtual catkey model object and data structure, used in form editing...persistence is in the cocina model
```
peetucket, line 6: no longer needed, this is being caught higher up the stack: https://github.com/sul-dlss/argo/blob/main/app/models/repository.rb#L26

app/controllers/catkeys_controller.rb
```diff
@@ -3,15 +3,44 @@
 class CatkeysController < ApplicationController
   load_and_authorize_resource :cocina, parent: false, class: 'Repository', id_param: 'item_id'
 
-  rescue_from Dor::Services::Client::UnexpectedResponse do |exception|
-    md = /\((.*)\)/.match exception.message
-    detail = JSON.parse(md[1])['errors'].first['detail']
-    redirect_to solr_document_path(params[:item_id]),
-                flash: { error: "Unable to retrieve the cocina model: #{detail.truncate(200)}" }
```
peetucket, line 10: No longer needed, being caught further up the stack: https://github.com/sul-dlss/argo/blob/main/app/models/repository.rb#L26

app/services/catkey.rb
```diff
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
 class Catkey
-  SYMPHONY = 'symphony'
-  PREVIOUS_CATKEY = 'previous symphony'
-
   def self.symphony_links(model)
```
peetucket, line 4: move these to the global constants so I can use them in the new code above

app/services/catkey.rb
```diff
@@ -21,17 +18,17 @@ def initialize(model)
...
-    find(SYMPHONY)
+    find(Constants::SYMPHONY)
   end
 
   def previous_links
-    find(PREVIOUS_CATKEY)
+    find(Constants::PREVIOUS_CATKEY)
   end
 
   def find(type)
```
peetucket, line 34: Note: this class is no longer used for individual catkey editing as changed in this PR, but it is still being used in the change set persister service classes, so I am leaving as is. 

see https://github.com/sul-dlss/argo/blob/main/app/services/collection_change_set_persister.rb#L61 and https://github.com/sul-dlss/argo/blob/main/app/services/item_change_set_persister.rb#L96

spec/requests/set_catkey_spec.rb
```diff
@@ -94,8 +97,8 @@
       context 'with a collection that has no existing catkeys' do
         let(:cocina_model) { build(:collection_with_metadata, id: druid, source_id: 'sul:1234') }
 
-        it 'updates the catkey, trimming whitespace' do
-          patch "/items/#{druid}/catkey", params: { catkey: { catkey: '   12345 ' } }
+        it 'updates the catkey' do
```
peetucket, line 100: no trimming of whitespace anymore, it will simply not validate anymore if you do this and refuse to save

ndushay (approved):
I think someone more reform fluent should review, but I'll approve.  Don't forget to remove "HOLD" in the PR title before merging.

app/controllers/catkeys_controller.rb
```diff
@@ -20,18 +49,37 @@ def edit
...
 
-  def update_params
-    params[CatkeyForm.model_name.param_key]
+  def catkey_form
+    # fetch catkeys from object
+    object_catkeys = @cocina.identification.catalogLinks.filter_map { |catalog_link| catalog_link if catalog_link.catalog == Constants::SYMPHONY }
+
+    # if there are no object catkeys, provide an initial blank row, else form is initialized with catkeys in the object
+    catkeys = if object_catkeys.size.zero?
+                [CatkeyRow.new(value: '', refresh: true)]
```
jcoyne, line 74: This should be done via prepopulate!

app/controllers/catkeys_controller.rb
```diff
@@ -3,15 +3,44 @@
...
+    def to_param
+      @id
+    end
+
+    def persisted?
+      true
+    end
   end
 
+  class CatkeyRow
```
jcoyne, line 24: Can you move this into CatkeyForm?

app/controllers/catkeys_controller.rb
```diff
@@ -3,15 +3,44 @@
 class CatkeysController < ApplicationController
   load_and_authorize_resource :cocina, parent: false, class: 'Repository', id_param: 'item_id'
 
-  rescue_from Dor::Services::Client::UnexpectedResponse do |exception|
-    md = /\((.*)\)/.match exception.message
-    detail = JSON.parse(md[1])['errors'].first['detail']
-    redirect_to solr_document_path(params[:item_id]),
-                flash: { error: "Unable to retrieve the cocina model: #{detail.truncate(200)}" }
+  ### classes that define a virtual catkey model object and data structure, used in form editing...persistence is in the cocina model
+  class ModelProxy
```
jcoyne, line 7: Can you move this into CatkeyForm?

app/views/catkeys/_row.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <%= form.text_field :value, class: "form-control" %>
+  </div>
+  <div class="col-sm-1">
+    <%= button_tag type: 'button', class: 'btn', aria: { label: 'Remove' },
+     data: { action: "click->nested-form#removeAssociation",
+             plain: true } do %>
+      <span class="bi bi-trash"></span>
+    <% end %>
+    <%= form.hidden_field :_destroy %>
+    <%= form.hidden_field :id %>
```
jcoyne, line 15: Why do we need these fields?  Catkey's don't have an id and it doesn't make sense to mark them for destruction, since we're just rewriting the whole set rather than deleting records.

jcoyne (changes requested):
I think we should find some time to work on this together.

app/views/catkeys/_edit.html.erb
```diff
@@ -1,10 +1,43 @@
...
+      <div class="col-sm-2"></div>
+      <div class="col-sm-10 col-xl-8">
+        <strong>Catkey</strong>
+      </div>
+      <div class="col-sm-1"></div>
+    </div>
+    <%= form_with model: @form,
+                      url: item_catkey_path(@form),
+                      method: :patch,
+                      data: { controller: 'nested-form', nested_form_selector_value: '.plain-container' } do |form| %>
```
jcoyne, line 14: I don't think this is an appropriate controller, because nested-form depends on the child resources actually being resources (like with an identifier) rather than parts of a bigger record.

app/controllers/catkeys_controller.rb
peetucket: Yes, will move (though we may want to consider doing the same thing in the `TagsController` for consistency.  I'll create a ticket for that.

app/controllers/catkeys_controller.rb
peetucket: yes, same comment as below for ModelProxy

app/controllers/catkeys_controller.rb
```diff
@@ -20,18 +49,37 @@ def edit
...
-    Argo::Indexer.reindex_druid_remotely(@cocina.externalIdentifier)
-
-    msg = "Catkey for #{@cocina.externalIdentifier} has been updated!"
-    redirect_to solr_document_path(@cocina.externalIdentifier), notice: msg
+    @form = catkey_form
+    respond_to do |format|
+      if @form.validate(params[:catkey]) && @form.save
+        Argo::Indexer.reindex_druid_remotely(@cocina.externalIdentifier)
+        msg = "Catkeys for #{@cocina.externalIdentifier} have been updated!"
+        format.html { redirect_to solr_document_path(@cocina.externalIdentifier, format: :html), notice: msg }
```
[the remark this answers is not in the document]
peetucket: This is called when you update the catkeys.  The logic for saving to the cocina model is in the CatkeyForm class.

app/controllers/catkeys_controller.rb
peetucket: moved the blank catkey row logic to a prepopulator on the form

app/controllers/catkeys_controller.rb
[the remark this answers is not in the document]
jcoyne: My question is more about, when would the format be "html" rather than "turbo_stream"?

app/forms/catkey_form.rb
```diff
@@ -1,22 +1,89 @@
...
+  def save
+    cocina_object = Repository.find(@model.id)
+
+    return false if catkeys.size.zero? # nothing submitted on the form
+
+    # fetch the existing previous catkey values from the cocina object
+    existing_previous_catkeys = Catkey.new(cocina_object).previous_links
+
+    # these are all of the existing catkey values the user wants to remove (i.e. for which they clicked the trash icon)
+    removed_catkeys = catkeys.filter_map { |catkey| catkey.value if catkey._destroy == '1' }
```
jcoyne, line 71: Can this destroy check be moved to the populator?

app/forms/catkey_form.rb
```diff
@@ -1,22 +1,89 @@
...
+    # must match the expected pattern
+    errors.add(:catkey, 'must be in an allowed format') if catkeys.count { |catkey| catkey.value.match(/^\d+(:\d+)*$/).nil? }.positive?
   end
 
-  # @raises [Dor::Services::Client::BadRequestError] when the server doesn't accept the request
-  # @raises [Cocina::Models::ValidationError] when given invalid Cocina values or structures
-  def save_model
-    return unless changed?(:catkey)
+  def save
+    cocina_object = Repository.find(@model.id)
```
jcoyne, line 63: This causes a second load, which should be unnecessary, since we already loaded this in the controller.

app/forms/catkey_form.rb
```diff
@@ -1,22 +1,89 @@
...
+  collection :catkeys, prepopulator: ->(*) { catkeys << CatkeyForm::Row.new(value: '', refresh: true) if catkeys.size.zero? },
+                       populator: lambda { |collection:, index:, **|
+                                    if item = collection[index] # rubocop:disable Lint/AssignmentInCondition
+                                      item
+                                    else
+                                      collection.insert(index, CatkeyForm::Row.new)
+                                    end
+                                  } do
+    property :value
+    property :refresh
```
jcoyne, line 13: can this use `Coercion` to be a boolean?

jcoyne (changes requested):

app/controllers/catkeys_controller.rb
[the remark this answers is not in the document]
peetucket: Got it.  Yes, I don't think the `respond_to` and formats are needed, it's a redirect in one case (valid) and a turbo response in the other (invalid).  Removed.

jcoyne (approved):

jcoyne:
Please do a squash merge on this.

PR #3682, opened 2022-05-10
change icon on tag form to trash icon

peetucket:

When working on https://github.com/sul-dlss/argo/pull/3681 (catkey editing), I noted that the design for removing catkey rows shows a trash icon, yet the icon for removing tags on a very similar modal form is currently a button.  This makes the tag editing form look like the design for catkeys: https://github.com/sul-dlss/argo/issues/3358#issuecomment-1116329820

**Current:**
Screen Shot 2022-05-10 at 3 08 19 PM

**Updated:**
Screen Shot 2022-05-10 at 3 08 07 PM

### DIFF
--- a/app/views/tags/_row.html.erb
+++ b/app/views/tags/_row.html.erb
@@ -1,15 +1,15 @@
 <div class="mb-3 row plain-container">
   <div class="col-sm-2">
-    <%= form.label :email, 'Tag', class: 'col-form-label' %>
+    <%= form.label :name, 'Tag', class: 'col-form-label' %>
   </div>
   <div class="col-sm-10 col-xl-8">
     <%= form.text_field :name, class: "form-control" %>
   </div>
   <div class="col-sm-1">
-    <%= button_tag type: 'button', class: 'btn btn-secondary', aria: { label: 'Remove' },
+    <%= button_tag type: 'button', class: 'btn', aria: { label: 'Remove' },
      data: { action: "click->nested-form#removeAssociation",
              plain: true } do %>
-      <span class='icon-remove-sign text-danger'></span>
+      <span class='bi bi-trash'></span>
     <% end %>
     <%= form.hidden_field :_destroy %>
     <%= form.hidden_field :id %>


app/views/tags/_row.html.erb
```diff
@@ -1,15 +1,15 @@
...
     <%= form.text_field :name, class: "form-control" %>
   </div>
   <div class="col-sm-1">
-    <%= button_tag type: 'button', class: 'btn btn-secondary', aria: { label: 'Remove' },
+    <%= button_tag type: 'button', class: 'btn', aria: { label: 'Remove' },
      data: { action: "click->nested-form#removeAssociation",
              plain: true } do %>
-      <span class='icon-remove-sign text-danger'></span>
+      <span class='bi bi-trash'></span>
     <% end %>
```
peetucket, line 13: change from button with an X to a trash can icon

app/views/tags/_row.html.erb
```diff
@@ -1,15 +1,15 @@
 <div class="mb-3 row plain-container">
   <div class="col-sm-2">
-    <%= form.label :email, 'Tag', class: 'col-form-label' %>
+    <%= form.label :tag, 'Tag', class: 'col-form-label' %>
   </div>
```
peetucket, line 4: typo?

app/views/tags/_row.html.erb
```diff
@@ -1,15 +1,15 @@
 <div class="mb-3 row plain-container">
   <div class="col-sm-2">
-    <%= form.label :email, 'Tag', class: 'col-form-label' %>
+    <%= form.label :tag, 'Tag', class: 'col-form-label' %>
```
jcoyne, line 3: Should this be `:name` as we don't have a field with a name of `:tag` on this page.

jcoyne (changes requested):

app/views/tags/_row.html.erb
peetucket: Fixed.

PR #3691, opened 2022-05-13
Set catkey to no refresh when updating with serial form.

### LINKED ISSUE #3687: When the digital serials form is submitted, mark the catkey as "do not refresh"
The main use case for the "Do not refresh" catkey setting are digital serials. These are items that have gone through the following steps:

- catkey entered
- metadata refreshed from Symphony
- serials metadata added in Argo
- catkey marked as "Do not refresh" (using the new form we're implementing right now in #3358 )

These items are then supposed to be marked "do not refresh" because a refresh from Symphony would overwrite the serials metadata added in Argo.

Given that we know that serials metadata should not have metadata automatically refreshed, we could eliminate the separate step of updating the catkey setting by automatically marking any catkey updated via the serials form with the "Do not refresh" attribute.

justinlittman:
closes #3687

Less work for user.

### DIFF
--- a/spec/forms/serials_form_spec.rb
+++ b/spec/forms/serials_form_spec.rb
@@ -321,5 +321,39 @@ RSpec.describe SerialsForm do
         expect(object_client).to have_received(:update).with(params: expected)
       end
     end
+
+    context 'when catalog links present' do
+      let(:cocina_item) { build(:dro_with_metadata, id: druid, catkeys: ['6671606']) }
+
+      let(:expected) do
+        build(:dro_with_metadata, id: druid).new(
+          description: {
+            title: [
+              {
+                structuredValue: [
+                  { value: 'factory DRO title', type: 'main title' }
+                ]
+              }
+            ],
+            purl: 'https://purl.stanford.edu/bc123df4567'
+          },
+          identification: {
+            catalogLinks: [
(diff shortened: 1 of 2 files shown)


justinlittman:
Andrew approved.

spec/forms/serials_form_spec.rb
```diff
@@ -321,5 +321,39 @@
...
+                structuredValue: [
+                  { value: 'factory DRO title', type: 'main title' }
+                ]
+              }
+            ],
+            purl: 'https://purl.stanford.edu/bc123df4567'
+          },
+          identification: {
+            catalogLinks: [
+              { catalog: 'symphony', refresh: false, catalogRecordId: '6671606' }
```
edsu, line 342: It's kind of too bad all the boilerplate is needed to just test `refresh: false`?

spec/forms/serials_form_spec.rb
```diff
@@ -321,5 +321,39 @@
         expect(object_client).to have_received(:update).with(params: expected)
       end
     end
+
+    context 'when catalog links' do
```
edsu, line 325: Maybe 'when catalog links present' reads better?

edsu (approved):

spec/forms/serials_form_spec.rb
justinlittman: Changing.

PR #3693, opened 2022-05-13
Removed editing a cell warning for registration form.

### LINKED ISSUE #3649: stop using the "editing a cell" warning that stops people from registering from the form
When users have a cursor active in a cell, they get this warning:

Screen Shot 2022-05-04 at 11 33 56 AM

I don't think it serves a purpose anymore and we should remove it. We're going to replace the form altogether soon but I've been getting feedback from users that this is interfering with their work because it's less clear how to "lock" cells now than it was before, when there was a "lock" button.

justinlittman:
closes #3649

Lock was causing problem for users.

### DIFF
--- a/app/javascript/controllers/registration_controller.js
+++ b/app/javascript/controllers/registration_controller.js
@@ -24,18 +24,11 @@ export default class extends Controller {
     }
 
     register() {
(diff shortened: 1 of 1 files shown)


edsu (approved):

justinlittman:
Andrew approved.

PR #3695, opened 2022-05-13
Set refresh to true for catkey when registering.

### LINKED ISSUE #3694: Items registered with catkeys are having the catkeys set to "do not refresh"
**Describe the bug**
When items are registered with a catkey, the catkey is being marked do not refresh. As a result no Symphony descriptive metadata is making it into the druid when the item is registered.

**User Impact**
Users do not get the metadata they need. It also presents a very confusing situation.

**To Reproduce**
1. Go to the item registration page
2. Enter a catkey when registering a druid
3. See that no Symphony metadata is ever imported.
4. See that the druid has the catkey marked as "refresh: false"

Screen Shot 2022-05-13 at 12 05 49 PM

**Expected behavior**
Druids should start with "refresh: true" when they are newly registered.

justinlittman:
closes #3694

Descriptive metadata isn't being refreshed upon registration.

### DIFF
--- a/app/forms/registration_form.rb
+++ b/app/forms/registration_form.rb
@@ -21,7 +21,7 @@ class RegistrationForm
     catalog_links = []
     if params[:other_id] != 'label:'
       catalog, record_id = params[:other_id].split(':')
-      catalog_links = [{ catalog:, catalogRecordId: record_id, refresh: false }]
+      catalog_links = [{ catalog:, catalogRecordId: record_id, refresh: true }]
     end
 
     model_params = {


app/forms/registration_form.rb
```diff
@@ -21,7 +21,7 @@ def cocina_model
     catalog_links = []
     if params[:other_id] != 'label:'
       catalog, record_id = params[:other_id].split(':')
-      catalog_links = [{ catalog:, catalogRecordId: record_id, refresh: false }]
+      catalog_links = [{ catalog:, catalogRecordId: record_id, refresh: true }]
     end
```
peetucket, line 25: looks fine to me

justinlittman:
Approved by Andrew.

peetucket (approved):

PR #3698, opened 2022-05-18
pin blacklight to a known working branch

### LINKED ISSUE #3697: Using the workflow grid to reset multiple errored steps is not working
**Describe the bug**
When I click on the "reset" link from the workflow grid, I see a blue progress bar appear near the top of the Argo window but the process never finishes and now workflow steps get reset.

A typical reset link from https://argo.stanford.edu/report/workflow_grid :

Screen Shot 2022-05-17 at 12 09 03 PM

I took a video of what happens when I click the link. It's subtle but you can see a blue line extend from left to right just above the Argo header:

https://user-images.githubusercontent.com/8975583/168892087-d0fe89b6-c266-4e4f-8eb8-b7f3a6df69fb.mp4

Eventually the bar crosses the screen and nothing else happens.

**User Impact**
It's not possible to use Argo to reset multiple steps at once. You either need to open each item's show page and go to the workflow modal to reset each step one at a time, or you have to go to the workflow server itself and use a rake task or the rails console to process a batch in a loop.

**To Reproduce**
1. Go to the workflow grid link above
2. Click on the reset link - probably safest to keep trying the "end-accession" step in my screenshot above
3. See the progress bar but no other effect.

**Expected behavior**
The steps corresponding to the reset link should all be reset back to a status of "waiting", which then leads them to run again. When this works properly, the workflow grid page is reloaded and a banner appears indicating that the steps were successfully reset.

**Additional context**
I'm not sure when this stopped working. I reset batches of steps on a weekly basis so it either stopped working this week or late last week. It has not worked at all this week.

peetucket:

Temporarily pin blacklight to a known working PR https://github.com/projectblacklight/blacklight/pull/2718 to fix #3697 

Once the blacklight PR https://github.com/projectblacklight/blacklight/pull/2718 is merged, we can unpin.

Localhost, test suite

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,7 +74,8 @@ gem 'zip_tricks', '5.3.1' # 5.3.1 is required as 5.4+ breaks the download all fe
 gem 'openapi_parser', '< 1.0'
 
 # Stanford related gems
-gem 'blacklight', '~> 7.25'
+# temporary pin of blacklight to fix https://github.com/sul-dlss/argo/issues/3697 .... remove when blacklight PR merged, 5/18/2022
+gem 'blacklight', '~> 7.25', git: 'https://github.com/projectblacklight/blacklight', branch: 'backport-unpermitted-params'
 gem 'blacklight-hierarchy', '~> 6.0'
 gem 'dor-services-client', '~> 12.0'
 gem 'dor-workflow-client', '~> 4.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,19 @@
+GIT
+  remote: https://github.com/projectblacklight/blacklight
+  revision: 9644976c01ad7ab4d785ca88b78607a60613cdfa
+  branch: backport-unpermitted-params
+  specs:
+    blacklight (7.25.1)
+      deprecation
+      globalid
+      hashdiff
+      i18n (>= 1.7.0)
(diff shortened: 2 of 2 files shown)


mjgiarlo (approved):

peetucket:
> How do we know this works and the current release doesn't? Can a test be written to demonstrate?

Based on console and browser testing using blacklight current release and this pinned branch.   Already spent about 8 hours on debugging the source of the problem and am a bit reluctant to write an argo test for something that is in blacklight.

It's also possible we need to send the page params to blacklight differently now to make it work with the current release, but that wasn't obvious to me.  If you have any insight, that would be helpful.

jcoyne:
If I had a simple test that showed the feature working successfully, I would be able to understand the problem and I could tell you.

peetucket:
Let's talk after standup -- I think testing this in argo would require actually exercising the blacklight service service and then verifying it's doing paging as it should, which I'm not convinced is warranted.   The test we have in Argo for this part of the code is mocking the blacklight response, which is why it works either way.

barmintor:
That PR is approved and merged: Can you test against https://github.com/projectblacklight/blacklight/tree/release-7.x ? It would expedite releasing a patch.

peetucket:
Thanks Ben.  A quick test on both localhost and our Argo QA server pointing to that release show it working ok.

PR #3700, opened 2022-05-19
do not update workflow back to waiting if not in error

peetucket:

Another attempt to deal with #3697

Guard against setting a workflow step back to waiting that is no longer in an error state.

Existing tests

### DIFF
--- a/app/controllers/report_controller.rb
+++ b/app/controllers/report_controller.rb
@@ -57,8 +57,13 @@ class ReportController < CatalogController
         druid:,
         workflow:,
         process: step,
+        current_status: 'error',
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):

peetucket:
Heads up: had to fix a couple test expectations with the new method call parameters

jcoyne:
spec/requests/reset_workflow_steps_spec.rb needs to be updated too

peetucket:
> spec/requests/reset_workflow_steps_spec.rb needs to be updated too

updated that file, re-running tests

PR #3711, opened 2022-05-25
Update access and file access simultaneously

### LINKED ISSUE #3709: Changing rights from "dark" to "location-based" generates a Cocina validation error
**Describe the bug**
Changing object rights from dark to a location-based right generates a Cocina validation error. It looks like the validation error that happens when you try to set rights to dark when files are published and/or shelved, but this shouldn't be happening because the items are already dark.

For example, when trying to change https://argo-stage.stanford.edu/view/druid:bk732ck6093 to "location: spec".
```
Error building Cocina: Not all files have dark access and/or are unshelved when object access is dark: ["CSt-110-0238.tif", "CSt-110-0238.jp2", "CSt-110-0239.tif", "CSt-110-0239.jp2", "CSt-110-0240.tif", "CSt-110-0240.jp...
```

**User Impact**
Users need to do a workaround to get rights from dark to location-based. It is possible to go to citation-only first and then to location-based, just not directly from dark to location-based.

**To Reproduce**
1. Go to https://argo-stage.stanford.edu/view/druid:bk732ck6093
2. Change the rights to "view:location-based, download:location-based, location:spec"
3. See error

**Expected behavior**
It should be possible to change to any valid rights value from dark.

jcoyne:

Otherwise combining dark access with published files can trigger a validation error.
Fixes #3709

Tested on stage and added a request test.

### DIFF
--- a/app/services/item_change_set_persister.rb
+++ b/app/services/item_change_set_persister.rb
@@ -18,8 +18,9 @@ class ItemChangeSetPersister
   def update
     updated_model = update_identification(model)
                     .then { |updated| updated_administrative(updated) }
+                    .then { |updated| updated_structural_collections(updated) }
                     .then { |updated| updated_access(updated) }
-                    .then { |updated| update_structural(updated) }
+
     Repository.store(updated_model)
   end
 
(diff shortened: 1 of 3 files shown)


app/services/item_change_set_persister.rb
```diff
@@ -64,6 +65,45 @@ def update_structural(updated)
     updated.new(structural: updated_structural)
   end
 
+  ### Access and structural have to be updated simultaneously or it may trigger
+  ### the validation about the file being published when access is dark.
```
ndushay, line 69: "validation ... error? ... about ... dark, and the object can't be updated"

ndushay (commented):
great job finding the solution!   A few suggestions inline for your consideration.

PR #3718, opened 2022-05-31
Remove problematic collections when returning registration options.

justinlittman:
Handle problematic data in stage without throwing an error.

See https://app.honeybadger.io/projects/49894/faults/85702827

Stage

### DIFF
--- a/app/controllers/apo_controller.rb
+++ b/app/controllers/apo_controller.rb
@@ -36,7 +36,7 @@ class ApoController < ApplicationController # rubocop:disable Metrics/ClassLengt
       end
 
       ["#{name.truncate(60, separator: /\s/)} (#{col_id.delete_prefix('druid:')})", col_id]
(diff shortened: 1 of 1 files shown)


app/controllers/apo_controller.rb
```diff
@@ -36,7 +36,7 @@ def registration_options
       end
 
       ["#{name.truncate(60, separator: /\s/)} (#{col_id.delete_prefix('druid:')})", col_id]
-    end.sort_by(&:first) # before returning the list, sort by collection name
+    end.compact.sort_by(&:first) # before returning the list, sort by collection name
```
jcoyne, line 39: Can you add a comment here about why this compact is necessary?  Otherwise someone in the future is likely to factor this out as they will see it as un-necessary.

jcoyne (changes requested):

jcoyne (approved):

PR #3724, opened 2022-06-01
Add facet configuration for released to searchworks/earthworks

### LINKED ISSUE #3721: Replace "Released To" facet with new "Released to Searchworks" and "Released to Earthworks" date facets
There is info about how these new facets should appear in the Epic ticket:  sul-dlss/dor_indexing_app#846 

Andrew should weigh in on where these should be in the facet list.

This is the facet that should be replaced:

image

And this is how we want the new facets to display:

image

We already have date facets similar to this, so you should be able to configure these new ones.  

If configuring for the sort of boolean value of "Ever" is difficult, feel free to split that out into a separate ticket with the TL.

Note that Solr's `facet.missing` must be used to indicate "not released" and there _may_ be a need to get fancy with the link so it has a label different from the Solr value itself for this.

jcoyne:

Part of sul-dlss/dor_indexing_app#846
Fixes #3721

Tested on stage.


ndushay:
I'm marking this for PO review ... particularly after there is some data indexed.

ndushay (approved):
@jcoyne merge at will.  @andrewjbtw  may need to notify users before this is deployed to prod - dunno.

PR #3727, opened 2022-06-02
Pass viewing direction to content type job.

### LINKED ISSUE #3726: Reading order is not being set when using the Bulk action content type menu
**Describe the bug**
When using the bulk action content type menu, the reading order is not being set for books or images. The `hasMemberOrders` is an empty array.

**User Impact**
Users can't bulk change reading order.

**To Reproduce**
1. Go to a druid like https://argo-stage.stanford.edu/view/druid:bs651wm4021
2. Use the bulk action "Set content type" function to set the reading order like so
Screen Shot 2022-06-01 at 2 53 49 PM

3. See that the `hasMemberOrders` is an empty array:
Screen Shot 2022-06-01 at 2 54 24 PM

**Expected behavior**
The reading order should be whatever was selected under "viewing direction" in the bulk action menu.

**Additional context**
Changing the reading order from the show page does work at the individual item level. Only the bulk action seems to have the problem.

justinlittman:
closes #3726

It was being ignored.

Andrew on stage.

### DIFF
--- a/app/controllers/bulk_actions/content_type_jobs_controller.rb
+++ b/app/controllers/bulk_actions/content_type_jobs_controller.rb
@@ -8,7 +8,8 @@ module BulkActions
     def job_params
       super.merge(current_resource_type: params[:current_resource_type],
                   new_content_type: params[:new_content_type],
-                  new_resource_type: params[:new_resource_type])
+                  new_resource_type: params[:new_resource_type],
+                  viewing_direction: params[:viewing_direction])
     end
   end
 end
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
Can you write a test for this change?

jcoyne (approved):

ndushay:
Is this still HOLD?  Does it need PO review?  Or should HOLD be removed from title and PR merged?

PR #3742, opened 2022-06-15
hardcode list of workflow options per issue #3741

### LINKED ISSUE #3741: limit the list of workflows that appears in the "Add workflow" dropdown menu
There's a dropdown menu on Argo show pages and in the Bulk actions list for "add workflow". Currently, this menu shows what appears to be every workflow the SDR has ever used. We should limit this menu to only the workflows that actually get used.

We don't have a whole lot of data on how this menu is used. Since we ported it to a Bulk action this spring, we do have a small number of logs to show recent usage. Based on that, some knowledge of the workflows, and my own usage, I think we can keep only the following workflows in the list:

- accessionWF
- gisAssemblyWF
- gisDeliveryWF
- gisDiscoveryWF
- goobiWF
- registrationWF
- wasCrawlDisseminationWF
- wasCrawlPreassemblyWF
- wasSeedPreassemblyWF

We may not need all of the GIS and WAS workflows but it's safer to leave them for now. GIS accessioning relies on "Add workflow" for some tasks and I've had to use it to remediate WAS druids in the past. That may stop being necessary after this workcycle.

ndushay:

Closes #3741 

### **AFTER**

image

image

### **BEFORE**

image

image

locally

### DIFF
--- a/app/helpers/apo_helper.rb
+++ b/app/helpers/apo_helper.rb
@@ -1,14 +1,21 @@
 # frozen_string_literal: true
 
 module ApoHelper
-  # Retrieve a list of workflow templates from  the workflow service and return
-  # an array suitable for select_tag options
+  # @return [Array<Array<String, String>] array suitable for select_tag options
   def workflow_options
-    Rails.cache.fetch 'workflow-templates-select-options' do
-      list = WorkflowClientFactory.build.workflow_templates
-      list.map do |name|
-        [name, name]
-      end
+    # per https://github.com/sul-dlss/argo/issues/3741, this should be hardcoded
+    %w[
+      accessionWF
+      gisAssemblyWF
+      gisDeliveryWF
+      gisDiscoveryWF
+      goobiWF
+      registrationWF
+      wasCrawlDisseminationWF
+      wasCrawlPreassemblyWF
+      wasSeedPreassemblyWF
+    ].map do |workflow|
+      [workflow, workflow]
     end
   end
 


ndushay:
@andrewjbtw do you want to approve this in situ, or are the screenshots of the workflow pulldowns above sufficient?

andrewjbtw:
Unless it would be a good idea for me to check that the options get picked up when selected, I think the screenshot is fine.

app/helpers/apo_helper.rb
```diff
@@ -1,14 +1,23 @@
...
+        ['gisDeliveryWF', 'gisDeliveryWF'],
+        ['gisDiscoveryWF', 'gisDiscoveryWF'],
+        ['goobiWF', 'goobiWF'],
+        ['registrationWF', 'registrationWF'],
+        ['wasCrawlDisseminationWF', 'wasCrawlDisseminationWF'],
+        ['wasCrawlPreassemblyWF', 'wasCrawlPreassemblyWF'],
+        ['wasSeedPreassemblyWF', 'wasSeedPreassemblyWF']
+        # rubocop:enable Style/WordArray
+      ]
     end
```
peetucket, line 21: small niggle, but to avoid having each name twice and risk having a typo, how about leave the existing `.map` block and just make `list = %w[accessionWF gisAssemblyWF etc etc]` in line 8?  

also, shouldn't need the Rails cache anymore since it's just a hardcoded array now

ndushay: good ideas;  didn't think of them.  will fix.

ndushay:
aaaaaaaand - fixed.  Thanks for the suggestions!

peetucket (approved):

PR #3743, opened 2022-06-15
Remove existing server.pid on startup when using docker

aaron-collier:

This is a small docker file change to add an entrypoint script that will remove an existing `tmp/pids/server.pid` on startup because when running with foreman it is not removed on shutdown. When attached (watching the docker output) and ending with `ctrl-c` it remains, as well as when detached and running either `docker compose stop` or `docker compose down`. The result is that subsequent starts of the web application currently require manually removing the `server.pid` file to avoid the web container stopping with the `pid file already exists` error. 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,4 +30,4 @@ RUN yarn install
 
 COPY . .
 
-CMD ["bin/dev"]
+CMD ["docker/entrypoint.sh"]
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+set -e
+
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -30,4 +30,4 @@ RUN yarn install
 
 COPY . .
 
-CMD ["bin/dev"]
+ENTRYPOINT ["docker/entrypoint.sh"]
```
jcoyne, line 33: Why entrypoint and not cmd? If the former, then we can't override this in docker-compose.yml

jcoyne (changes requested):

ndushay (approved):

jcoyne (approved):

PR #3752, opened 2022-06-22
Validate barcode on registration page

### LINKED ISSUE #3748: (new registration form) Validate barcode format during form entry on the new Argo registration form
**NOTE** This issue depends on the new registration form.

When a user enters a barcode on the new Argo registration form, validate that it matches an allowable pattern (defined in the cocina model). This should follow the example of how the form handles source id validation. Users should not be able to submit the form with a barcode that is not valid.

Currently, submitting an invalid barcode will take the user to a 500 error page.

Must be a PR against PR #3684 or must be done separately after PR #3684 is merged -- either way.

ndushay:
~~[HOLD - for PO review, but also so PR #3751 can go through first]~~

closes #3748 

Registration page validates barcode against allowed barcodes per cocina.

###  AFTER

image

### BEFORE

image

po review

### DIFF
--- a/app/forms/registration_form.rb
+++ b/app/forms/registration_form.rb
@@ -8,6 +8,9 @@ class RegistrationForm < Reform::Form
     end
   end
 
+  # standard, business, lane, catkey (matches allowed patterns in cocina-models)
+  VALID_BARCODE_REGEX = /\A(36105[0-9]{9}|2050[0-9]{7}|245[0-9]{8}|[0-9]+-[0-9]+)\z/
+
   include HasViewAccessWithCdl
 
   property :current_user, virtual: true
@@ -37,6 +40,7 @@ It's legal to have more than one colon in a hierarchy, but at least one colon is
     property :barcode, virtual: true
     validates :source_id, format: { with: /\A.+:.+\z/, message: 'ID is invalid' }
     validates :catkey, allow_blank: true, format: { with: /\A\d+(:\d+)*\z/, message: 'is invalid' }
+    validates :barcode, allow_blank: true, format: { with: VALID_BARCODE_REGEX, message: 'is invalid' }
   end
 
   def persisted?
--- a/app/views/registrations/_item_row.html.erb
+++ b/app/views/registrations/_item_row.html.erb
@@ -17,7 +17,7 @@
     <%= item_form.text_field :catkey, class: 'form-control',
       pattern: '^[0-9]+(:[0-9]+)*$',
       data: {
-        action: "change->registration#validateCatkey invalid->registration#displayValidation",        
+        action: "change->registration#validateCatkey invalid->registration#displayValidation",
     } %>
   </div>
 </div>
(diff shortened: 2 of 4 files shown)


justinlittman:
@ndushay As I just learned from my catkey ticket, this also needs to be done client-side.

ndushay:
> @ndushay As I just learned from my catkey ticket, this also needs to be done client-side.

@andrewjbtw do you want the barcodes to have lookups like sourceId and catkey to ensure they exist?

andrewjbtw:
I wasn't aware that we could do barcode lookups. I believe we've just been validating the pattern in the existing registration form and Argo entry forms. 

For purposes of this ticket, I think it's fine to skip lookups if we don't do them elsewhere. The source id and catkey lookups are needed because the registration process does those lookups once the data is submitted and fails if things don't work out. For barcodes, even if the barcode doesn't exist in Symphony registration will still succeed.

app/forms/registration_form.rb
```diff
@@ -37,6 +40,7 @@ def persisted?
     property :barcode, virtual: true
     validates :source_id, format: { with: /\A.+:.+\z/, message: 'ID is invalid' }
     validates :catkey, allow_blank: true, format: { with: /\A\d+(:\d+)*\z/, message: 'is invalid' }
+    validates :barcode, allow_blank: true, format: { with: VALID_BARCODE_REGEX, message: 'is invalid' }
```
justinlittman, line 43: This validation also needs to be client-side.

justinlittman (changes requested):

app/forms/registration_form.rb
ndushay: yup - got it.

ndushay:
@justinlittman - I fixed this PR and Andrew has signed off.

peetucket (approved):

app/views/registrations/_item_row.html.erb
```diff
@@ -34,6 +34,10 @@
 <div class="row mb-3">
   <%= item_form.label :barcode, class: 'col-sm-2 col-form-label' %>
   <div class="col-sm-10">
-    <%= item_form.text_field :barcode, class: 'form-control' %>
+    <%= item_form.text_field :barcode, class: 'form-control',
+    pattern: '^(36105[0-9]{9}|2050[0-9]{7}|245[0-9]{8}|[0-9]+-[0-9]+)',
+    data: {
+      action: "change->registration#validateBarcode invalid->registration#displayValidation",
```
justinlittman, line 40: There is no validateBarcode method on registration controller.

justinlittman (changes requested):

app/views/registrations/_item_row.html.erb
ndushay: right you are.

ndushay: there is now.  Though I'm not sure if there should be.

justinlittman (approved):

PR #3755, opened 2022-06-22
Support refresh in set catkeys CSV bulk action.

### LINKED ISSUE #3735: Update catkey CSV bulk action to allow catkeys to be marked as refreshable or non-refreshable
I thought I'd ticketed this in the last workcycle but apparently not. We need to include catkey refreshability in the update catkey bulk action. One approach we could take is to modify the CSV to add an optional Refresh column and then apply these rules:

1. Only the first catkey column (if the user enters multiple columns) can be designated "refresh"/"do not refresh"
2. Any additional catkey columns are automatically "do not refresh" (since only one catkey can be refreshable)
3. If the user submits a CSV with only "Druid,Catkey" as the headers, assume the catkeys are refreshable
4. If the user adds submits the headers "Druid,Catkey,Refresh" then use the "Refresh" column to set the refreshability

That's much more complicated than the current CSV but it does preserve the existing behavior for CSVs that only include "Druid,Catkey".

As for the "set catkeys via form" - I wonder if it's finally time to retire that and only keep the CSV. I think it's going to get harder to maintain both actions, plus adding refreshability as an option to the form-based action seems like a difficult challenge.

justinlittman:
closes #3735

Need to support setting refresh.

Unit, QA

### DIFF
--- a/app/views/bulk_actions/catkey_and_barcode_csv_jobs/_new.html.erb
+++ b/app/views/bulk_actions/catkey_and_barcode_csv_jobs/_new.html.erb
@@ -10,6 +10,7 @@
     <ul>
       <li>Druid (required)</li>
       <li>Catkey (optional). May be repeated if an item is associated with multiple catkeys.</li>
+      <li>Refresh (optional). Set to "true" to indicate descriptive metadata is refreshable from first catkey; set to "false" to indicate descriptive metadata is not refreshable.</li>
       <li>Barcode (optional)</li>
     </ul>
     </em></small></p>
--- a/spec/jobs/set_catkeys_and_barcodes_job_spec.rb
+++ b/spec/jobs/set_catkeys_and_barcodes_job_spec.rb
@@ -63,14 +63,14 @@ RSpec.describe SetCatkeysAndBarcodesJob do
 
       it 'attempts to update the catkey/barcode for each druid with correct corresponding catkey/barcode' do
         expect(subject).to receive(:with_bulk_action_log).and_yield(buffer)
-        expect(subject).to receive(:update_catkey_and_barcode).with(change_set1, { catkeys: catkeys[0].split(','), barcode: barcodes[0] }, buffer)
+        expect(subject).to receive(:update_catkey_and_barcode).with(change_set1, { catkeys: catkeys[0].split(','), barcode: barcodes[0], refresh: true }, buffer)
         expect(subject).not_to receive(:update_catkey_and_barcode).with(change_set2, { catkeys: [], barcode: nil }, buffer)
-        expect(subject).to receive(:update_catkey_and_barcode).with(change_set3, { catkeys: [catkeys[2]], barcode: barcodes[2] }, buffer)
+        expect(subject).to receive(:update_catkey_and_barcode).with(change_set3, { catkeys: [catkeys[2]], barcode: barcodes[2], refresh: true }, buffer)
         subject.perform(bulk_action_no_process_callback.id, params)
         expect(bulk_action_no_process_callback.druid_count_total).to eq druids.length
-        expect(change_set1).to have_received(:validate).with(barcode: barcodes[0], catkeys: %w[12345 66233])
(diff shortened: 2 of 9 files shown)


app/views/bulk_actions/catkey_and_barcode_csv_jobs/_new.html.erb
```diff
@@ -10,6 +10,7 @@
     <ul>
       <li>Druid (required)</li>
       <li>Catkey (optional). May be repeated if an item is associated with multiple catkeys.</li>
+      <li>Refresh (optional). Set to "true" to indicate first catkey is refreshable; set to "false" to indicate first catkey is not refreshable.</li>
```
ndushay, line 13: "... to indicate first catkey is refreshable; set to "false" to indicate *no* catkey is ~~not~~ refreshable" for your consideration.

Also, it's not the catkey that is refreshable, it's the descriptive metadata that is refreshable from the source represented by the catkey ...

"... to indicate descriptive metadata is refreshable from the first catkey ..." ?

spec/jobs/set_catkeys_and_barcodes_job_spec.rb
```diff
@@ -92,14 +92,14 @@
...
+        expect(subject).not_to receive(:update_catkey_and_barcode).with(change_set2, { catkeys: [], refresh: true }, buffer) # not changed
+        expect(subject).to receive(:update_catkey_and_barcode).with(change_set3, { catkeys: ['44444'], refresh: true }, buffer) # changed to a value
         subject.perform(bulk_action_no_process_callback.id, params)
         expect(bulk_action_no_process_callback.druid_count_total).to eq druids.length
-        expect(change_set1).to have_received(:validate).with(catkeys: []) # set to empty
-        expect(change_set2).to have_received(:validate).with(catkeys: []) # not changed
-        expect(change_set3).to have_received(:validate).with(catkeys: ['44444']) # changed to a value
+        expect(change_set1).to have_received(:validate).with(catkeys: [], refresh: true) # set to empty
+        expect(change_set2).to have_received(:validate).with(catkeys: [], refresh: true) # not changed
+        expect(change_set3).to have_received(:validate).with(catkeys: ['44444'], refresh: true) # changed to a value
```
ndushay, line 102: should there be a spec for `refresh: false`?

spec/services/item_change_set_persister_spec.rb
```diff
@@ -221,6 +221,28 @@
...
+              catalogLinks: [
+                { catalog: 'symphony', catalogRecordId: catkey_before, refresh: false }
+              ]
+            }
+          )
+        )
+      end
+    end
+
+    context 'when change set has changed refresh' do
```
ndushay, line 243: did you perhaps change the wrong description?

ndushay (changes requested):
a few quibbles, and one possible copypasta error in a spec

app/views/bulk_actions/catkey_and_barcode_csv_jobs/_new.html.erb
justinlittman: Updated.

spec/jobs/set_catkeys_and_barcodes_job_spec.rb
justinlittman: Refresh can't be set from the plain-old set_catkeys_and_barcodes_job invocation, which this is testing.

spec/services/item_change_set_persister_spec.rb
justinlittman: Indeed I did. Fixed.

ndushay (approved):

andrewjbtw:
Copying in my summary of testing from Slack. I think this is good to go:

I tested:

- remove all catkeys
- add catkeys “normally” - with no refresh
- set some refresh to false
- add multiple catkeys for a druid
- switch refresh false to refresh true

PR #3768, opened 2022-06-30
add druid column to all exported structural metadata

### LINKED ISSUE #3758: Include the druid in the structural metadata CSV when downloading from the show page
We currently don't include the druid in the structural metadata CSV when it's downloaded from an item show page (because it only applies to that one druid). We do include the druid when the structural metadata is exported via bulk action (which can include multiple druids).

For consistency, we should include the druid on the show page download. This will make it easier to combine multiple structural metadata CSVs if they were downloaded separately but the user wants to submit a bulk upload later.

peetucket:

Fixes #3758 - include the druid column for all structural metadata exports (even single objects).

This adds the druid column to the core export class, and then removes it from the bulk action job (since it will come automatically from the single object now).

~~HOLD to validate on QA/stage with real data~~ validated on qa, works fine

Updated/existing tests/localhost.  Verified on qa with actual data.

Note: existing spec for bulk action job is unchanged but passes, showing that it still works as expected (i.e. includes the druid column)

### DIFF
--- a/app/jobs/export_structural_job.rb
+++ b/app/jobs/export_structural_job.rb
@@ -9,13 +9,9 @@ class ExportStructuralJob < GenericJob
     super
 
     CSV.open(csv_download_path, 'w', headers: true) do |csv|
-      csv << ['druid', *StructureSerializer::HEADERS]
+      csv << StructureSerializer::HEADERS
       with_items(params[:druids], name: 'Export structural metadata') do |cocina_object, success, failure|
-        rows_for_file = item_to_rows(cocina_object, success, failure)
-        druid = Druid.new(cocina_object).without_namespace
-        rows_for_file.each do |row|
-          csv << [druid, *row]
-        end
+        item_to_rows(cocina_object, success, failure).each { |row| csv << row }
       end
     end
   end
@@ -30,7 +26,7 @@ class ExportStructuralJob < GenericJob
       return []
     end
 
-    StructureSerializer.new(item.structural).rows do |row|
+    StructureSerializer.new(item.externalIdentifier, item.structural).rows do |row|
       result << row
     end
     success.call('Exported structural metadata')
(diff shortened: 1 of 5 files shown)


app/jobs/export_structural_job.rb
```diff
@@ -9,13 +9,9 @@ def perform(bulk_action_id, params)
...
-      csv << ['druid', *StructureSerializer::HEADERS]
+      csv << StructureSerializer::HEADERS
       with_items(params[:druids], name: 'Export structural metadata') do |cocina_object, success, failure|
-        rows_for_file = item_to_rows(cocina_object, success, failure)
-        druid = Druid.new(cocina_object).without_namespace
-        rows_for_file.each do |row|
-          csv << [druid, *row]
-        end
+        item_to_rows(cocina_object, success, failure).each { |row| csv << row }
       end
```
peetucket, line 15: lines 14-18 above can be collapsed into one line, since now we don't need to add the druid specifically (it will come from the single object export)

jcoyne (changes requested):
Did you test that the exported structural metadata was able to be imported again?

ndushay:
@andrewjbtw do you want to test this?

peetucket:
> Did you test that the exported structural metadata was able to be imported again?

I tested the single object exported structural metadata could be changed and re-imported with the new druid column.  This matches with what I see in the code, where the druid columns will just be ignored at import time.

The bulk action format has not changed (it always included the druid column).

I did update `spec/services/structure_updater_spec.rb` though so that the new column is used in the test.

ndushay (approved):

peetucket:
Unless Andrew wants to verify this upon his return and discovers an issue, it should be good to go without further work.

jcoyne (approved):

ndushay:
(Andrew indicated in slack that this doesn't require his approval).

PR #3775, opened 2022-07-01
[WIP] Paste into the registration form

### LINKED ISSUE #3763: Enable copy and paste in the new registration form
A stub ticket for enabling a version of the copy and paste functionality present in the old registration form.

To be determined:

- handling errors, validation problems
- whether to report out to the form results page or the bulk action results page

jcoyne:

Fixes #3763


app/javascript/controllers/registration_items_controller.js
```diff
@@ -4,14 +4,49 @@ export default class extends Controller {
...
+    }
+
+    const rowsOnPage = this.element.querySelectorAll(this.selectorValue)
+
+    lines.forEach((line, index) => {
+      let sourceId, catkey, label, barcode;
+      [barcode, catkey, sourceId, label] = line.split(/\t/)
+  
+      const elements = rowsOnPage[index].querySelectorAll('input')
+      elements[0].value = barcode
```
justinlittman, line 40: This seems fragile.

app/views/registrations/_item_row.html.erb
```diff
@@ -1,26 +1,42 @@
...
 </div>
+
+<div class="row mb-3">
+  <%= item_form.label :source_id, 'Source ID', class: 'col-sm-2 col-form-label' %>
+  <div class="col-sm-10">
+    <%= item_form.text_field :source_id, class: 'form-control', required: true,
+                             pattern: '^.+:.+$',
+                             data: {
+                               action: "change->registration-item-row#validateSourceId invalid->registration#displayValidation",
+                               registration_target: 'sourceId',
```
justinlittman, line 32: This is no longer a target of registration?

justinlittman (changes requested):
Test?

app/views/registrations/_item_row.html.erb
jcoyne: no. It's now all done in `registration-item-row`

justinlittman (approved):

jcoyne:
@ndushay this PR is not up to date with the latest version of cocina-models, so the agreement creation fails.

andrewjbtw:
This PR is good to go, now that I've heard from users and the paste-in order will remain the same.

jcoyne:
@ndushay naw, that's just a flaky test.

PR #3778, opened 2022-07-01
Switch to using DSA to check if source id is unique.

justinlittman:
DSA is the canonical source of uniqueness.

Unit

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -43,10 +43,12 @@ class RegistrationsController < ApplicationController
   def source_id
     raise 'Malformed input' unless /\A.+:.+\z/.match?(params[:source_id])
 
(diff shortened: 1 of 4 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -43,10 +43,12 @@ def autocomplete
   def source_id
```
ndushay, line 43: if this is a boolean method, should it be `source_id_unique?`

ndushay (commented):
question for your consideration

app/controllers/registrations_controller.rb
jcoyne: This is not a boolean method, it's returning json.

jcoyne (approved):

PR #3779, opened 2022-07-01
Use bare druids in tracking sheet.

### LINKED ISSUE #3772: Registration form: tracking sheets should not include the druid prefix
Tracking sheets are now being produced with a druid prefix, which is throwing off workflows that involve scanning barcodes. Previously tracking sheets did not include the druid prefix.

This applies to:

- the tracking sheets generated off the registration form
- tracking sheets from the bulk action

**Examples**

An old tracking sheet without a prefix: 
[barcode-1.pdf](https://github.com/sul-dlss/argo/files/9023985/barcode-1.pdf)

Recent tracking sheet with a prefix:
[tracking_sheet-4.pdf](https://github.com/sul-dlss/argo/files/9023992/tracking_sheet-4.pdf)

justinlittman:
closes #3772

This was the previous behavior.

Unit


spec/requests/registration_spec.rb
```diff
@@ -28,7 +28,7 @@
     let(:doc) { instance_double(Prawn::Document, render: '') }
 
     it 'generates a tracking sheet with the right default name' do
-      get "/registration/tracksheet?druid=#{bare_druid}", headers: headers
+      get "/registration/tracksheet?druid=#{druid}", headers: headers
```
ndushay, line 31: I don't see how this is proving that it's a druid without a namespace - should it prove that?

ndushay (changes requested):
tests failing - tests related to this change, even.

ndushay (approved):

PR #3787, opened 2022-07-05
Add a back to registration form link

### LINKED ISSUE #3780: Registration form: add link to go back to form after registering
@astridu 's design now includes a "Back to form" link. This should take the user back to the form with the settings they just used already selected (including tags). It should not reload the form with the defaults they get when loading it for the first time.

image (2)

jcoyne:

Fixes #3780


mjgiarlo (approved):

andrewjbtw:
I tested this in stage today. All the settings previously selected were being retained when I clicked on the back-to-form link except for the rights. The rights were being set back to the APO defaults each time I went back. If possible, the back link should retain the rights from the previous registration, even if they're not the APO defaults.

andrewjbtw:
After the latest changes, the rights settings are being preserved. But I found one more issue that I'd probably missed in testing earlier: the "Initial workflow" is being reset to registrationWF when you go back to the form. 

One of the big use cases will be goobiWF users who will be using the back button to change only a tag or two between registrations.

justinlittman:
@mjgiarlo This has undergone some changes since you approved. Would you mind taking another look please?

mjgiarlo (approved):

mjgiarlo:
@justinlittman @jcoyne Approved. Does this need further testing? If not, :shipit:

PR #3789, opened 2022-07-06
Humanize barcode error message

### LINKED ISSUE #3744: Humanize error message for invalid barcodes
The current error message for entering an invalid barcode looks like:

`"invalid" isn't one of in #/components/schemas/Barcode`

The message should be more human-readable:

`"invalid" is not a valid barcode`

The barcode error message appears in the following contexts:

1. registration form (in the redesigned form)
2. bulk action registration (in the bulk action log)
3. on the Argo show page in the barcode field
4. in the bulk actions for updating catkeys/barcodes (appears in the log)

I don't know how much work is involved to get the same message out to all of those places. I think it's most important in the contexts where the user is entering data into a box: the registration form and the item show page.

jcoyne:
Fixes #3744

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -204,9 +204,7 @@ class ItemsController < ApplicationController
       reindex
       redirect_to solr_document_path(params[:id]), status: :see_other
     else
-      message = change_set.errors.full_messages.to_sentence
-      logger.error "Errors: #{message}"
-      render 'error', locals: { message: }
+      render 'error', locals: { message: change_set.errors.map(&:message).to_sentence }
     end
   end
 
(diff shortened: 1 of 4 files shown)


ndushay:
@andrewjbtw would still like to review this;   @jcoyne thanks for adding the label, but I guess John didn't notice the label and there was no "Hold" added to the beginning of the title.

jmartin-sul:
@ndushay 💬 
> @andrewjbtw would still like to review this; @jcoyne thanks for adding the label, but I guess John didn't notice the label and there was no "Hold" added to the beginning of the title.

d'oh, my fault, i didn't see the label, will try to keep an eye out for that in the future.

PR #3799, opened 2022-07-12
Remove the druid prefix from tracksheet barcode

### LINKED ISSUE #3795: Do not use druid prefix in tracking sheet generated by the Argo tracking sheet bulk action
Follow up to #3772 because https://github.com/sul-dlss/argo/pull/3779 only addressed the registration form tracking sheet bulk action:

Tracking sheets generated by the Argo bulk action are being produced with a druid prefix, which is throwing off workflows that involve scanning barcodes. Previously tracking sheets did not include the druid prefix.

**Examples**

An old tracking sheet without a prefix: 
[barcode-1.pdf](https://github.com/sul-dlss/argo/files/9023985/barcode-1.pdf)

Recent tracking sheet with a prefix:
[tracking_sheet-4.pdf](https://github.com/sul-dlss/argo/files/9023992/tracking_sheet-4.pdf)

aaron-collier:

Fixes #3795 

### DIFF
--- a/app/models/track_sheet.rb
+++ b/app/models/track_sheet.rb
@@ -31,6 +31,7 @@ class TrackSheet
   # @param [String] druid unqualified DRUID identifier
   # @param [Prawn::Document] pdf document being built (document is modified)
   # @return [Prawn::Document] the same document
+  # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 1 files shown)


ndushay (approved):
merge when 🟢

PR #3804, opened 2022-07-18
Guard against changing the selected content type in CSV registration

### LINKED ISSUE #3802: Media content type not sticking - draft
**Describe the bug**
When registering a media object using the spreadsheet upload method, the content type gets set to ‘file’ instead of ‘media,’ despite the drop down value being set correctly as 'media'. (Non-spreadsheet registrations are working as expected.)

**User Impact**
The incorrect content type value requires remediation by the user. It should just be set correctly from the start.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

aaron-collier:

Fixes #3802 

The `dro_types` method will always return `Cocina::Models::ObjectType.object` if the `content_type` is already a Cocina::Models type. This guards against changing it.

Locally, first selecting the content_type (in both cases I selected `media`):
Screen Shot 2022-07-18 at 3 07 24 PM

Before the change, a csv registration shows that it defaults to `Cocina::Models::ObjectType.object` (displayed as `file`):

Screen Shot 2022-07-18 at 3 07 57 PM

After this change the proper selection is displayed:
Screen Shot 2022-07-18 at 3 08 47 PM

### DIFF
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -4,6 +4,15 @@
 class RegistrationCsvConverter
   include Dry::Monads[:result]
 
+  CONTENT_TYPES = [Cocina::Models::ObjectType.image,
+                   Cocina::Models::ObjectType.three_dimensional,
+                   Cocina::Models::ObjectType.map,
+                   Cocina::Models::ObjectType.media,
+                   Cocina::Models::ObjectType.document,
+                   Cocina::Models::ObjectType.manuscript,
+                   Cocina::Models::ObjectType.book,
+                   Cocina::Models::ObjectType.object].freeze
+
   # @param [String] csv_string CSV string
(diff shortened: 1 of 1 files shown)


edsu (approved):

app/services/registration_csv_converter.rb
```diff
@@ -87,6 +95,9 @@ def tags(row)
   end
 
   def dro_type(content_type)
+    # for CSV registration, we already have the URI
```
jcoyne, line 98: @aaron-collier Why do we need the code below this then? Isn't this whole file for CSV registration?

aaron-collier: Oh you know, that's a good point @jcoyne - there was a bit of crossover in the forms, but not likely here. I'll do a follow up.

aaron-collier: Is it possible that content_type is included in the csv? I'll verify that first.

PR #3811, opened 2022-07-20
Handle invalid registration CSVs.

### LINKED ISSUE #3809: Improve error handling when a registration form CSV upload contains an invalid character encoding
**Describe the bug**
The registration form sends users to the library-status page if they upload a CSV with an invalid character encoding. A user tried to upload a CSV containing accent marks and the uploader failed by taking them to the status page.

As far as I can tell, re-saving the CSV in Excel using the "CSV UTF-8" option generated a CSV that could be processed, so the issue wasn't with the accent marks, it was with how the accent marks were encoded. 

If it's possible to handle the original CSV without needing to ask the user to re-save it, we should do that. But if it's not possible, Argo should show the user an error message.

**User Impact**
Without an error message in Argo, it's difficult to see what the problem is. The status page can lead to the conclusion that the system might be down or unable to register at all.

**To Reproduce**
1. Go to argo-stage registration
2. Use this CSV as input (you may need to change the source id if you've used it before)
[character-test.csv](https://github.com/sul-dlss/argo/files/9145362/character-test.csv)

3. See the library status page
4. See that HB did get an alert over character encoding: https://app.honeybadger.io/projects/49894/faults/86856157

**Expected behavior**
Users should see a message if the input CSV could not be processed.

justinlittman:
closes #3809

A notified user is better than a befuddled user.

Unit, local


justinlittman:
image

app/controllers/registrations_controller.rb
```diff
@@ -115,13 +113,21 @@ def csv_create
       # redirect_to bulk_actions_path(path_params), status: :see_other, notice: success_message
       redirect_to bulk_actions_path, status: :see_other, notice: 'Register druids job was successfully created.'
     else
-      @apo_list = AdminPolicyOptions.for(current_user)
-      @registration_form.prepopulate!
+      prepopulate
       render :show, status: :bad_request
     end
+  rescue CSV::MalformedCSVError => e
+    prepopulate
```
ndushay, line 120: did you mean to call it again here?  You already have the error.

ndushay (changes requested):
I assume there was no easy way to 

> "If it's possible to handle the original CSV without needing to ask the user to re-save it, we should do that. But if it's not possible, Argo should show the user an error message."

Like - can we force the encoding to utf-8 and see if it can prepopulate without error?

https://rubyapi.org/3.1/o/stringio#method-i-set_encoding -- CSV delegates some methods to StringIO ...

.rubocop.yml
```diff
@@ -39,6 +39,12 @@ Metrics/BlockLength:
     - 'app/services/forms_grouper.rb'
     - 'app/services/notes_grouper.rb'
 
+Metrics/ClassLength:
+  Exclude:
```
jcoyne, line 43: I'm not sure we should move these here, because this to me indicates that we never want to fix them (and we don't care how big they grow). I don't think that is the intent here. If I'm wrong about that, please move this to a separate PR and we can discuss there.

spec/features/item_registration_csv_spec.rb
```diff
@@ -82,4 +82,28 @@
       expect(page).to have_content 'Csv file missing headers: source_id.'
     end
   end
+
+  context 'invalid CSV' do
+    it 'reports error' do
```
jcoyne, line 87: Not a blocker, but this could have been a request test instead. Request tests are much faster and are capable of testing the same behavior that you are testing here.

jcoyne (changes requested):

.rubocop.yml
justinlittman: We have different philosophies about this, but not worth the effort to litigate. I backed this out, but re-running autogen resulted in a bunch of other changes.

spec/features/item_registration_csv_spec.rb
justinlittman: Spent a little time on this, but the complexity of getting the post params correct for this form wasn't worth it.

jcoyne (approved):

aaron-collier (approved):

aaron-collier:
I believe the question above from @ndushay has been answered, so will merge, any follow up work should be a seperate PR.

PR #3820, opened 2022-08-04
registration fix for access citation-only with csv

### LINKED ISSUE #3818: Registration is failing when attempting to set citation-only rights while using the CSV
**Describe the bug**
When using a CSV with the registration form, attempts to set citation-only rights are failing with this error:
`RegisterDruidsJob: {"view"=>"citation-only"} isn't one of in #/components/schemas/Access`

It looks like the problem is only the "view" setting is being passed through, without the "download" setting (which should be "none".

**User Impact**
Users can't use the CSV to register batches of druids with citation only rights.

**To Reproduce**
1. Go to the registration form
2. Choose citation-only rights on the menu
3. Upload a CSV
4. See error in bulk action logs

**Expected behavior**
The correct rights should be assigned.

**Additional context**
The form entry is working. Only the CSV upload seems to have this issue.

ndushay:

Registering items with a csv for access of citation-only was failing.  

Fixes #3818

I deployed it to stage and checked a csv that was failing before I deployed this fix.  It works now.

### DIFF
--- a/spec/services/registration_csv_converter_spec.rb
+++ b/spec/services/registration_csv_converter_spec.rb
@@ -95,4 +95,55 @@ RSpec.describe RegistrationCsvConverter do
       expect(results.first.value![:model]).to eq(expected_cocina.new(access: {}))
     end
   end
+
+  context 'when params have rights_view citation-only but no rights_download' do
+    let(:csv_string) do
+      <<~CSV
+        source_id,label
+        foo:123,My new object
+      CSV
+    end
+
+    let(:params) do
+      {
+        administrative_policy_object: 'druid:bc123df4567',
+        collection: 'druid:bk024qs1808',
+        initial_workflow: 'accessionWF',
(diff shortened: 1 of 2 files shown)


aaron-collier (approved):

spec/services/registration_csv_converter_spec.rb
```diff
@@ -95,4 +95,55 @@
...
+        initial_workflow: 'accessionWF',
+        content_type: Cocina::Models::ObjectType.book,
+        tags: ['csv : test', 'Project : two']
+      }
+    end
+
+    it 'returns result with access citation-only model' do
+      expect(results.size).to be 1
+      expect(results.first.success?).to be true
+      expect(results.first.value![:model]).to eq(expected_cocina.new(access: { 'view' => 'citation-only', 'download' => 'none' }))
```
jcoyne, line 146: I'd like to see the `expected_cocina` be the cocina we expect, rather than something we mutate to get what we expect.

PR #3829, opened 2022-08-22
Stop differentiating between Hydrus and non-Hydrus collections

### LINKED ISSUE #3827: Combine the Hydrus and nonHydrus collections facets into one
At long last, I think we're ready to remove the Hydrus collection facet and put all collections in the same facet.

This is part of work that needs to be done to fully remove the old Hydrus/non-Hydrus distinction. For background and context of that work, see https://docs.google.com/document/d/1g_kuGrLJcWWr1Nh-IfCj4FTxtf6D5sBn9iduHWZuTts/edit#

Adding a note to say that the APO facets will be addressed later -- this ticket is just for the collection facet.

jcoyne:

Fixes #3827

CI

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -57,21 +57,20 @@ class CatalogController < ApplicationController
     # exploded_tag_ssim indexes all tag prefixes (see IdentityMetadataDS#to_solr for a more exact
     # description), whereas tag_ssim only indexes whole tags.  we want to facet on exploded_tag_ssim
     # to get the hierarchy.
-    config.add_facet_field 'exploded_tag_ssim',                 label: 'Tag',                 limit: 9999,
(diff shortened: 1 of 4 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -57,21 +57,20 @@ class CatalogController < ApplicationController
     # exploded_tag_ssim indexes all tag prefixes (see IdentityMetadataDS#to_solr for a more exact
     # description), whereas tag_ssim only indexes whole tags.  we want to facet on exploded_tag_ssim
     # to get the hierarchy.
-    config.add_facet_field 'exploded_tag_ssim',                 label: 'Tag',                 limit: 9999,
```
jcoyne, line 60: This is mostly just spacing changes. You may want to "hide whitespace" when reviewing.

justinlittman (approved):

PR #3837, opened 2022-08-29
pin to redis 4

peetucket:

Redis 5 issues with resque

CI


jcoyne (changes requested):
Argo doesn't use resque

PR #3844, opened 2022-09-07
deploy.rb: capistrano first deploy works without sprockets

ndushay:

This fix made it possible to deploy the argo app to a newly built VM.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -44,3 +44,9 @@ set :honeybadger_env, fetch(:stage)
 
 # update shared_configs before restarting app
 before 'deploy:restart', 'shared_configs:update'
+
+# configure capistrano-rails to work with propshaft instead of sprockets
+# (we don't have public/assets/.sprockets-manifest* or public/assets/manifest*.*)
+set :assets_manifests, lambda {
+  [release_path.join('public', fetch(:assets_prefix), '.manifest.json')]
+}


config/deploy.rb
```diff
@@ -44,3 +44,7 @@
 
 # update shared_configs before restarting app
 before 'deploy:restart', 'shared_configs:update'
+
+# configure capistrano-rails to work without sprockets
```
jcoyne, line 48: This is only necessary for the first deploy right? How do we ensure it gets turned off after?

jcoyne (changes requested):
Could this be done by configuring 

```
    set :assets_manifests, -> {
        release_path.join("public", fetch(:assets_prefix), '.manifest.json')
    }
```

See https://github.com/capistrano/rails/issues/257

ndushay:
> set :assets_manifests, -> {
>         release_path.join("public", fetch(:assets_prefix), '.manifest.json')
>     }

I'll try it for the stage box rebuilds.

ndushay:
@jcoyne your suggestion didn't work.  :-(

ndushay:
@jcoyne and I paired and tweaked his suggestion.  It now works for the first deploy and for subsequent deploys.  Yay!

jcoyne (approved):

PR #3859, opened 2022-09-28
Filter inactive APOs from APO lists.

### LINKED ISSUE #3856: Use a tag to filter the list of APOs available during registration
Following the example in https://github.com/sul-dlss/argo/issues/3825 for collections, use a tag to keep specific APOs out of Argo dropdown menus.

The menus that list APOs in Argo are:

- the registration form
- the change APO modal on the show page (available via pencil icon)
- the change APO bulk action menu

Desired outcome: APOs that have been tagged `APO status : inactive` should not appear on these menus

Use case: preventing decommissioned APOs from being available for selection in Argo menus. Roughly 75% of all APOs in SDR are legacy Hydrus APOs. We do not want people to register new items to these APOs. A small number of other APOs have also been decommissioned.

justinlittman:
closes #3856

Remove APOs that we don't want user to select.

Unit

### DIFF
--- a/app/services/admin_policy_options.rb
+++ b/app/services/admin_policy_options.rb
@@ -13,7 +13,7 @@ class AdminPolicyOptions
       defType: 'lucene',
       rows: 99_999,
       fl: 'id,tag_ssim,sw_display_title_tesim',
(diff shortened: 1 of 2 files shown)


justinlittman:
@andrewjbtw approved.

ndushay (approved):

PR #3868, opened 2022-10-06
add validation before attempting desc metadata update to get better error messages

### LINKED ISSUE #3865: Show the same date validation log message in the Cocina descriptive upload bulk action that shows in the Cocina validation action
When uploading a Cocina descriptive metadata spreadsheet (via the bulk action, not MODS), if a druid contains an invalid date in a date range, the bulk action log reports only that the validation and save failed, like so:

`Line 4: validate_and_save failed for druid:gj203jm4155 for druid:gj203jm4155 (2022-10-03 21:11:37 UTC)`

But when running the same metadata spreadsheet through the Cocina validation bulk action, the log reports a more informative error:

`Line 2: ["Invalid date(s) for druid:bb546qn3753: [[\"86\", \"w3cdtf\"]]"] for druid:bb546qn3753 (2022-10-04 06:52:31 UTC)`

If possible, both bulk actions should show the same log message. Otherwise it can be difficult to debug when someone only sees the "failed" message.

**To reproduce**

1. In stage, go to the bulk actions page
2. Choose "upload descriptive metadata spreadsheet"
3. Upload this CSV, which has a date validation error for a stage druid: 
[descriptive-bb546qn3753.csv](https://github.com/sul-dlss/argo/files/9710425/descriptive-bb546qn3753.csv)
4. See the log file for the generic error.
5. Run the same spreadsheet through the validation action
6. See the log file for the more informative message

peetucket:

Fixes #3865 - validate each row of the spreadsheet explicitly first when running a desc metadata import job, before we apply the desc metadata update...this will catch any validation errors and provide the same nice error message as the validate only job does.

Localhost and QA

### DIFF
--- a/app/jobs/descriptive_metadata_import_job.rb
+++ b/app/jobs/descriptive_metadata_import_job.rb
@@ -17,6 +17,7 @@ class DescriptiveMetadataImportJob < GenericJob
       next failure.call('Not authorized') unless ability.can?(:update, cocina_object)
 
       DescriptionImport.import(csv_row:)
+                       .bind { |description| validate_input(cocina_object, description) }
                        .bind { |description| validate_changed(cocina_object, description) }
                        .bind { |description| open_version(cocina_object, description) }
                        .bind { |description, new_cocina_object| validate_and_save(new_cocina_object, description) }
@@ -29,6 +30,14 @@ class DescriptiveMetadataImportJob < GenericJob
   end
 
   private
+
+  # this validates input data from spreadsheet before any updates are applied to provide error messages to the user
+  def validate_input(cocina_object, description)
+    result = CocinaValidator.validate(cocina_object, description:)
+    return Success(description) if result.success?
+
+    Failure(["validation failed for #{cocina_object.externalIdentifier}: #{result.failure}"])
+  end
 
   def validate_changed(cocina_object, description)
     return Failure(['Description unchanged']) if cocina_object.description == description
--- a/app/jobs/descriptive_metadata_import_job.rb
+++ b/app/jobs/descriptive_metadata_import_job.rb
@@ -20,7 +20,7 @@ class DescriptiveMetadataImportJob < GenericJob
                        .bind { |description| validate_input(cocina_object, description) }
                        .bind { |description| validate_changed(cocina_object, description) }
                        .bind { |description| open_version(cocina_object, description) }
-                       .bind { |description, new_cocina_object| validate_and_save(new_cocina_object, description) }
+                       .bind { |description, new_cocina_object| save(new_cocina_object, description) }
                        .bind { |new_cocina_object| close_version(new_cocina_object) }
                        .either(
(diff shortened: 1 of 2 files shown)


mjgiarlo (commented):
Looks simple enough. Would you consider adding tests (one for success, one for failure)?

peetucket:
There are currently tests for both failure and success of this job.  I did consider amending the existing failure test at https://github.com/sul-dlss/argo/blob/main/spec/jobs/descriptive_metadata_import_job_spec.rb#L79-L102 (e.g. to verify it calls the `CocinaValidator.validate` method), but unclear as to what value is added.

peetucket:
> @peetucket I hear ya. So the intent here, is it to validate the cocina object **before a new version is opened**? If so, I think the test that may be warranted here is that if the supplied cocina object is invalid, a new version is not opened. 

Added an expectation for this.

 > Related: with validation moving earlier in the chain of operations, do we still need to second validation later in the validate_and_save step? Or can it save without validating again?

I don't think we need to validate again, though it won't do any harm.  I spent some trying to refactor to just do a `Repository.save` instead, but then had some failing tests that I didn't quite understand (perhaps due to my understanding of how Dry Monads work).  There is some performance benefit to not validating again, but it was not clear it worth the time investment right now.  Happy to pair though to see if I'm just missing something.

peetucket:
Ok, figured out to just do a save at the end directly.

app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -17,9 +17,10 @@ def perform(bulk_action_id, params)
       next failure.call('Not authorized') unless ability.can?(:update, cocina_object)
 
       DescriptionImport.import(csv_row:)
+                       .bind { |description| validate_input(cocina_object, description) }
                        .bind { |description| validate_changed(cocina_object, description) }
                        .bind { |description| open_version(cocina_object, description) }
-                       .bind { |description, new_cocina_object| validate_and_save(new_cocina_object, description) }
+                       .bind { |description, new_cocina_object| save(new_cocina_object, description) }
                        .bind { |new_cocina_object| close_version(new_cocina_object) }
```
peetucket, line 24: no need to do validation again at the end now

app/jobs/descriptive_metadata_import_job.rb
```diff
@@ -44,11 +53,12 @@ def open_version(cocina_object, description)
     Failure([e.message])
   end
 
-  def validate_and_save(cocina_object, description)
-    result = CocinaValidator.validate_and_save(cocina_object, description:)
-    return Success(cocina_object) if result.success?
```
peetucket, line 49: note: the `CocinaValidator.validate_and_save` is used elsewhere in the code, so we will leave it as is even though we don't need it anymore in this class

ndushay (approved):

mjgiarlo (approved):

PR #3870, opened 2022-10-13
[HOLD] Support hierarchical file structures.

justinlittman:
refs https://github.com/sul-dlss/dor-services-app/issues/4256

[HOLD] until we are sure we will deploy the full stack in support of file hierarchy


mjgiarlo:
@justinlittman What do you think we should do with this PR at this point?

PR #3877, opened 2022-11-01
Use standard as a linter on top of rubocop

mjgiarlo:
# Why was this change made? 🤔

The `standard` gem manages Rubocop upgrades and maintains strong opinions on the Ruby style guide so that we do not have to. This is a spike to explore what it would look like to replace Rubocop with standard and begin to evaluate which is a better fit for our team.

> [standard] aims to save uou (and others!) time in the same three ways:
>
> 1. **No configuration.** The easiest way to enforce consistent style in your project. Just drop it in.
> 1. **Automatically format code.** Just run `standardrb --fix` and say goodbye to messy or inconsistent code.
> 1. **Catch style issues & programmer errors early.** Save precious code review time by eliminating back-and-forth between reviewer & contributor.
>
> No decisions to make. It just works." ...

[Read more about standard](https://github.com/testdouble/standard). [Watch the standard lightning talk](https://www.youtube.com/watch?v=uLyV5hOqGQ8)

## Usage

The way we have knitted standard and rubocop together here, we are essentially using the standard style guide without being coupled to its linter tool. Otherwise put: we can (and probably should*) continue running `rubocop` for linting (vs. `standardrb`) and `rubocop -[aA]` for auto-fixing (vs. `standardrb --fix`).

* "should" because CI---via our ruby-rails orb, which uses the CircleCI ruby orb, which supports Rubocop but not standard---will continue using rubocop tooling, as will most/many of our IDEs. Better to be consistent with that practice in local dev than be surprised later, I reckon?

## RSpec Extension
  
Also includes:

* Extend standard style guide with rubocop-rspec (per https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard)

The `standard` gem does not include rubocop-rspec, and we find (most of) these cops valuable. This commit extends the prior commit's standard configuration by re-adding the rubocop-rspec dependency and adding more Rubocop configuration, including TODOs. The trade-off of using rubocop-rspec is apparent here.

## Rails Extension
  
Also includes:

* Extend standard style guide with rubocop-rails (per https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard)

The `standard` gem does not include rubocop-rails, and we find (most of) these cops valuable. This commit extends the prior commit's standard configuration by re-adding the rubocop-rails dependency and adding more Rubocop configuration, including TODOs. The trade-off of using rubocop-rails is apparent here.

## Auto-fix Style with Weekly Updates

This is now done, tested in https://github.com/sul-dlss/argo/pull/3887/commits/3137072cbf7fa6abcab9ffe531114be4a29352d2

CI


.rubocop.yml
```diff
@@ -1,496 +1,12 @@
-inherit_from: .rubocop_todo.yml
-require:
-  - rubocop-rspec
-  - rubocop-rails
+require: standard
```
mjgiarlo, line 1: Rubocop config shrinks to a 9-line file since standard now provides the bulk of rubocop configuration.

.rubocop_todo.yml
```diff
@@ -1,372 +0,0 @@
-# This configuration was generated by
```
mjgiarlo, line 1: No one misses this file.

.standard.yml
```diff
@@ -0,0 +1 @@
+parallel: true
```
mjgiarlo, line 1: This is literally the only configuration we need for standard. And we don't even need it; it just speeds things up.

Capfile
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 # Load DSL and Setup Up Stages
-require 'capistrano/setup'
+require "capistrano/setup"
```
mjgiarlo, line 4: Use double-quotes by default instead of fussing about single vs. double depending on whether a string needs interpolation. Works well with Rails which generates files with double-quotes anyway.

README.md
```diff
@@ -82,7 +82,7 @@ CI runs a series of steps;  this the sequence to do it locally, along with some
     bin/rails test:prepare
     ```
 
-5. **Run rubocop and the tests**
+5. **Run the linter and the tests**
```
mjgiarlo, line 85: Couple to the functionality, not the tool.

app/components/download_all_button_component.rb
```diff
@@ -8,10 +8,10 @@ def initialize(document:)
 
   def link_options
     # This onclick handler prevents the accordion which contains this element from folding.
-    options = { onclick: 'event.stopPropagation()' }
+    options = {onclick: "event.stopPropagation()"}
```
mjgiarlo, line 11: Omit spaces inside braces for hashes. Meant to make hashes visually distinct from blocks.

app/components/lazy_tag_facet_component.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
 class LazyTagFacetComponent < ViewComponent::Base
-  def initialize(**); end
```
mjgiarlo, line 4: I was always conflicted about this semi-colon usage...

app/components/show/controls_component.rb
```diff
@@ -69,57 +69,57 @@ def create_embargo
       return if !item? || embargoed?
 
       render ActionButton.new url: new_item_embargo_path(druid),
-                              label: 'Create embargo',
-                              open_modal: true,
-                              disabled: button_disabled?
+        label: "Create embargo",
+        open_modal: true,
+        disabled: button_disabled?
```
mjgiarlo, line 74: Prefer two-line indents in more contexts.

app/components/workflow_table_process_component.rb
```diff
@@ -10,22 +10,21 @@ def initialize(workflow_table_process:, name:, data:)
...
   end
 
   def workflow_item_count(status)
-    new_params = search_state.filter('wf_wps_ssim').add([name, process, status].compact.join(':')).params.merge(controller: 'catalog', action: 'index')
-    rotate_facet_params('wf_wps', 'wps', facet_order('wf_wps'), new_params)
+    new_params = search_state.filter("wf_wps_ssim").add([name, process, status].compact.join(":")).params.merge(controller: "catalog", action: "index")
+    rotate_facet_params("wf_wps", "wps", facet_order("wf_wps"), new_params)
     item_count = 0
-    if data[process] && data[process][status] && item = data[process][status][:_] # rubocop:disable Lint/AssignmentInCondition
+    if data[process] && data[process][status] && (item = data[process][status][:_])
```
mjgiarlo, line 21: Wrap conditional assignments in parens.

app/components/workflow_table_process_component.rb
```diff
@@ -34,43 +33,44 @@ def rotate_facet_params(prefix, from, to, p = params.dup)
     p[:f] = (p[:f] || {}).dup # the command above is not deep in rails3, !@#$!@#$
     p[:f][from_field] = (p[:f][from_field] || []).dup
     p[:f][to_field] = (p[:f][to_field] || []).dup
-    # rubocop:disable Style/Semicolon
-    p[:f][from_field].reject! { |v| p[:f][to_field] << rotate_facet_value(v, from, to); true }
-    # rubocop:enable Style/Semicolon
+
+    p[:f][from_field].reject! { |v|
```
mjgiarlo, line 37: Be OK with multi-line blocks using braces. (I'm meh on this change but can live with it.)

app/controllers/concerns/creates_bulk_actions.rb
```diff
@@ -8,12 +8,13 @@ module CreatesBulkActions
...
 
-  def new; end
+  def new
+  end
 
   def create
     begin
       bulk_action_job_params = job_params
-    rescue StandardError => e
+    rescue => e
```
mjgiarlo, line 17: `StandardError` is the default here.

app/components/lazy_tag_facet_component.rb
aaron-collier: same.

app/components/workflow_table_process_component.rb
edsu: This seems much more readable to me.

app/components/workflow_table_process_component.rb
edsu: I have no idea what it's doing in either case :-)

.rubocop_todo.yml
edsu: 🎉

.standard.yml
edsu: This is *awesome*.

app/components/lazy_tag_facet_component.rb
edsu: Agree!

app/components/show/controls_component.rb
edsu: 💯

edsu:
I fully support this change, but I defer to the people who have written the majority of the SDR code to decide what to do.

justinlittman (commented):
At the end of the day, I don't care as long as rubocop auto-corrects.

That being said, I have the slightest of preferences for these changes.

app/components/workflow_table_process_component.rb
mjgiarlo: @edsu haha! Yeah, this is rather impenetrable, isn't it?

ndushay:
I am ok with this but not with the loss of rubocop-rspec.  There is a way for us to do both.  If we want, we can even have rspec-rails too.  (Having mucked about with a lot of rubocop rspec violations in the last couple months, I will say that I prefer the specs following rubocop-rspec defaults to specs that don't.)

Hallelujah for this in standard:

>Standard does not take a stance on whether you should plaster a [frozen_string_literal magic comment](https://docs.ruby-lang.org/en/3.0/doc/syntax/comments_rdoc.html#label-Magic+Comments) directive at the top of every file. Enforcing use of the comment became popular when it was believed that string literals would be frozen by default in a future version of Ruby, but [according to Matz](https://bugs.ruby-lang.org/issues/11473#note-53) there are no (longer any) such plans.

I find that stupid magic comment ... stupid.  In fact, we hates it.

mjgiarlo:
@ndushay 💬 
> I am ok with this but not with the loss of rubocop-rspec. There is a way for us to do both. If we want, we can even have rspec-rails too.

Tossed another commit at this branch that re-adds `rubocop-rspec`. Should I re-add `rubocop-rails` as well?

ndushay:
I don't have strong feelings about rubocop-rails, AFAIR.  there are only 4 rules.  https://docs.rubocop.org/rubocop-rspec/cops_rspec_rails.html

mjgiarlo:
> I don't have strong feelings about rubocop-rails, AFAIR. there are only 4 rules. [docs.rubocop.org/rubocop-rspec/cops_rspec_rails.html](https://docs.rubocop.org/rubocop-rspec/cops_rspec_rails.html)

That looks like the Rails section of rubocop-rspec. Here's all the cops in rubocop-rails: https://docs.rubocop.org/rubocop-rails/cops.html Way more than I expected or remembered!

ndushay:
> @ndushay 💬
> 
> > I don't have strong feelings about rubocop-rails, AFAIR. there are only 4 rules. [docs.rubocop.org/rubocop-rspec/cops_rspec_rails.html](https://docs.rubocop.org/rubocop-rspec/cops_rspec_rails.html)
> 
> That looks like the Rails section of rubocop-rspec. Here's all the cops in rubocop-rails: https://docs.rubocop.org/rubocop-rails/cops.html Way more than I expected or remembered!

wow - let's keep using, then?  Some of those are useful, i believe.

mjgiarlo:
@ndushay :speech_balloon:
> > That looks like the Rails section of rubocop-rspec. Here's all the cops in rubocop-rails: [docs.rubocop.org/rubocop-rails/cops.html](https://docs.rubocop.org/rubocop-rails/cops.html) Way more than I expected or remembered!
> 
> wow - let's keep using, then? Some of those are useful, i believe.

Added!

ndushay (approved):

ndushay:
you might consider taking "[SPIKE]" out of the heading.  You might also consider a title of "use standard as a linter on top of rubocop"

PR #3900, opened 2022-12-12
Revert "Uncorrupted downloads from pres."

mjgiarlo:
Reverts sul-dlss/argo#3899

We have a hypothesis that this patch was working around a bug present in net-protocol 0.2.0: https://github.com/ruby/net-protocol/pull/14


ndushay (approved):
approving but unclear on the harm of leaving it in.

PR #3907, opened 2022-12-13
Correctly set collection rights when editing

### LINKED ISSUE #3901: It is not possible to change the rights on a collection object from the show page
**Describe the bug**
It is not possible to change the rights on a collection object from the show page. When trying to edit the collection rights, the menu display shows code in the UI:
Screen Shot 2022-12-13 at 12 00 51 PM

Then, when you change the rights and click "Save" you are sent to a 500 error page. There's also a related HB error:
https://app.honeybadger.io/projects/49894/faults/91719097

**User Impact**
It is not possible to change the rights for a collection object. For collections that were previously "dark", this prevents users from making them available to the public.

**To Reproduce**
1. Go to this example on stage: https://argo-stage.stanford.edu/view/druid:wm284bn5869
2. Click on the pencil next to the access rights and choose "world" from the dropdown
3. Click "save" and see the error

**Expected behavior**
Collection rights can be world or dark and it should be possible to switch between them.

**Additional context**
The set rights bulk action also fails for collections. The error message is different so I'm going to ticket that separately. The issue on the show page is higher priority.

aaron-collier:

Fixes #3901 

There is a small difference between argo and cocina proper where we refer to "view_access" that needs to be sent as "view" for the Cocina::Models::CollectionAccess value

This also fixes a small template error that displayed HTML instead of the cancel button.

Before:
Screen Shot 2022-12-13 at 3 55 40 PM

After:

Screen Shot 2022-12-13 at 3 53 12 PM

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -46,7 +46,7 @@ class CollectionsController < ApplicationController
     return unless enforce_versioning
 
     change_set = CollectionChangeSet.new(@cocina)
-    attributes = params.require(:collection).permit(:copyright, :use_statement, :license)
+    attributes = params.require(:collection).permit(:view_access, :copyright, :use_statement, :license)
     change_set.validate(**attributes)
     change_set.save
     Argo::Indexer.reindex_druid_remotely(@cocina.externalIdentifier)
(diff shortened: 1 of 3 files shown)


jmartin-sul (approved):

aaron-collier:
tested and signed off by @andrewjbtw

PR #3908, opened 2022-12-14
Authorize SDR API token endpoint using new workgroup

mjgiarlo:

Connects to sul-dlss/sdr-client#250

This commit limits the number of Argo users that are authorized to generate sdr-api tokens. The workgroup that is now used contains the Infra devs, JCoyne, Andrew, and Tony C, and the group is managed by Andrew.

CI + tested in QA with Andrew

### DIFF
--- a/spec/features/token_spec.rb
+++ b/spec/features/token_spec.rb
@@ -3,7 +3,6 @@
 require "rails_helper"
 
 RSpec.describe "retrieve a token for sdr-api", js: true do
-  let(:groups) { ["sdr:administrator-role", "dlss:dor-admin", "dlss:developers"] }
   let(:conn) { instance_double(SdrClient::Connection, post: response) }
   let(:response) { instance_double(Faraday::Response, status: 200, body:) }
   let(:body) do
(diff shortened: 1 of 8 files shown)


spec/features/token_spec.rb
```diff
@@ -16,12 +16,29 @@
...
-    click_button "Generate new token"
-    expect(find_field("Token").value).to eq body
-    expect(SdrClient::Login).to have_received(:run)
+    it "gets a token" do
+      sign_in create(:user), groups: groups
+
+      visit "/settings/tokens"
+      click_button "Generate new token"
+      expect(find_field("Token").value).to eq body
+      expect(SdrClient::Login).to have_received(:run)
```
mjgiarlo, line 28: This will fail, and I will adjust the test.

justinlittman (approved):

spec/models/user_spec.rb
```diff
@@ -273,31 +289,42 @@
...
+          expect(subject).not_to include User::MANAGER_GROUPS.first
         end
       end
 
       context "and the impersonating user is a viewer" do
         let(:webauth_groups) { User::VIEWER_GROUPS }
 
         it "returns only the impersonated groups" do
-          expect(subject).not_to include User::VIEWER_GROUPS
+          expect(subject).not_to include User::VIEWER_GROUPS.first
```
mjgiarlo, line 327: These tests were not working as expected before, as they used to check whether the array contained a subarray, and now I believe the tests work as intended (and still pass, but now in the expected way).

PR #3913, opened 2022-12-19
Adds file hierarchy display.

### LINKED ISSUE #3881: Better view of hierarchical file content in Argo
For objects accessioned with hierarchy (e.g. via pre-assembly), the file content display area in Argo currently shows a flat list of files with paths.

Here is one example of how it currently looks (https://argo-qa.stanford.edu/view/druid:dt568vs5903):

Screen Shot 2022-11-10 at 9 45 15 AM

Explore the possibility of a better display, e.g. something like this:

Unknown

We will need to deal with the challenge that additional metadata is currently displayed for each file in Argo -- unclear how this would work with a hierarchical folder display.

justinlittman:
closes #3881

Hierarchy!

Local, unit

image

### DIFF
--- a/app/assets/stylesheets/show.scss
+++ b/app/assets/stylesheets/show.scss
@@ -172,4 +172,29 @@
     background-color: $silver-sand;
     color: $body-color;
   }
+
+  #filehierarchy {
+    ul {
+      list-style-type: none;
+
+      a:before {
+        display: inline-block;
+        content: "";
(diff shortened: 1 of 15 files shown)


justinlittman:
Tested by @andrewjbtw .

ndushay (approved):
I gave this a somewhat cursory lookthru and nothing stood out for comment or concern.  Approving.

PR #3931, opened 2023-01-26
use the same regex to verify source_ids across the app

### LINKED ISSUE #3894: agreement registration does not accept source ids with dashes
**Describe the bug**
Valid source ids can include dashes. The "register agreement" form is rejecting source ids with dashes.

**User Impact**
It's not possible to use a dash in the source id for agreements. This can be worked around but it goes against some established naming conventions.

**To Reproduce**
1. Go to https://argo-stage.stanford.edu/agreements/new
2. Create an agreement with a source id with dashes

**Expected behavior**
The register agreement form should accept all valid source id formats. The register items form does not have this problem.

**Screenshots**
This screenshot shows a form filled out with values that should be accepted:
Screen Shot 2022-11-29 at 1 41 03 PM

**Additional context**
I thought this bug had been fixed before. There is an old issue about dashes in source ids but I don't know if the fix there would have fixed the agreement form. See https://github.com/sul-dlss/argo/issues/3753

peetucket:

Fixes #3894 - use the same regex to verify source_ids in both the agreement and the item registration forms ... and put it into the settings so we do not get out of sync again.

This was fixed for the item reg form in https://github.com/sul-dlss/argo/pull/3754/files ... this takes that regex and moves it to settings, and then uses it for agreements validation

Localhost

### DIFF
--- a/app/forms/agreement_form.rb
+++ b/app/forms/agreement_form.rb
@@ -7,7 +7,7 @@ class AgreementForm < Reform::Form
   property :agreement_file2, virtual: true
 
   validates :title, presence: true
-  validates :source_id, format: {with: /\A\w+:\w+\z/,
+  validates :source_id, format: {with: Regexp.new(Settings.source_id_regex),
                                  message: "must have a single colon in the middle"},
     presence: true
   validates :agreement_file1, presence: true
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -44,7 +44,7 @@ class RegistrationsController < ApplicationController
 
   # Allow the front end to check if a source Id already exists
   def source_id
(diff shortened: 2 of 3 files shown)


mjgiarlo (approved):

app/forms/agreement_form.rb
```diff
@@ -7,7 +7,7 @@ class AgreementForm < Reform::Form
   property :agreement_file2, virtual: true
 
   validates :title, presence: true
-  validates :source_id, format: {with: /\A\w+:\w+\z/,
+  validates :source_id, format: {with: Regexp.new(Settings.source_id_regex),
```
mjgiarlo, line 10: I'd ordinarily ask for a test here but this change makes the validation less strict (because `.` covers more characters than `\w`) so I don't think it is worth the time.

mjgiarlo (approved):

PR #3715, opened 2022-05-26
Remove solr bootstrapping

jcoyne:
## Why was this change made? 🤔
DSA no longer looks at solr to see if an admin policy exists, so we can remove the bootstrapping

## How was this change tested? 🤨
locally

PR #3954, opened 2023-02-14
remove solr bootstrapping

ndushay:

based on PR #3715;  removes Solr bootstrapping, as it is no longer used in DSA

unit tests (this only affects specs)


justinlittman (approved):

ndushay:
Hrmm - it appears to be needed as the same tests are failing and all on missing admin policy.  Whee!

ndushay:
3 tries, same 7 tests failing due to missing admin policy;  apparently we DO still need it here.

PR #3959, opened 2023-02-23
Adds folio feature flag.

### LINKED ISSUE #3946: Add feature switch for toggling between Folio HRID and Symphony ckey in Argo UI
Argo needs a feature switch in settings.yml and in shared_configs for all 3 environments (qa, stage, prod), in order to switch between Symphony and Folio identifiers in the UI.

The switch should be "catalog_record_id" and the values should be "folio" or "symphony" and should default to "symphony"

This will allow us to develop what is needed to use Folio instead of Symphony for the Argo UI

justinlittman:
closes #3946

Unblock many tickets.

Note that I deviated from the requirements in the ticket:
* Followed the conventions for feature flags from dor-services-app. 
* Used a boolean because it is less error prone.
* Did not add to shared_configs, as this can be done when rolling out feature to each environment.


justinlittman:
Further note that it may be necessary to create more nuanced feature flags as the feature work proceeds, but this should be good enough to get us started.

ndushay (approved):

PR #3981, opened 2023-03-09
Add `metadata_source_ssim` facet field behind feature flag

### LINKED ISSUE #3977: Add `metadata_source_ssim` index field behind feature flag
If a feature flag is enabled---e.g., `Settings.enabled_features.multivalued_metadata_source_facet` or some such---add a new facet field using the new `metadata_source_ssim` field. This flag should be disabled by default but may be turned on in QA and stage.

NOTE: Work from https://github.com/sul-dlss/dor_indexing_app/issues/946 will feed into this, but this work in Argo isn't blocked.

mjgiarlo:

Fixes #3977

This commit adds a new facet field to the Blacklight catalog if and only if the multi-valued metadata sources feature flag is on. It also flips this flag on in dev & test, and allows the user to flip it on for a given request by adding `?multivalued=true` as a query param in the URL.

It's not clear that there are any specs that could be extended to cover the two small changes to `app/controllers/` and I'm not sure it's worth the effort to write new specs for this when we'll be extensively testing this via the browser given we're dealing with Blacklight internals.

- [x] CI
- [x] QA (flag flipped on)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
 
   before_action :authenticate_user!
   before_action :set_folio_flag_per_param
+  before_action :set_multivalued_metadata_source_flag_per_param
 
   rescue_from CanCan::AccessDenied, with: -> { render status: :forbidden, plain: "forbidden" }
 
@@ -37,6 +38,10 @@ class ApplicationController < ActionController::Base
 
   def set_folio_flag_per_param
     Settings.enabled_features.folio = true if params[:folio] == "true"
+  end
+
+  def set_multivalued_metadata_source_flag_per_param
+    Settings.enabled_features.multivalued_metadata_sources = true if params[:multivalued] == "true"
   end
 
   protected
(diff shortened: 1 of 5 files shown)


justinlittman (commented):
Curious why this is behind a feature flag?

justinlittman:
OK. I see the context in the ticket. Might I suggest as an alternative that we just leave this as a branch rather than feature flag. Feature flagging this doesn't seem critical and it is yet more code that we will need to clean up later.

mjgiarlo:
@justinlittman 💬 
> OK. I see the context in the ticket. Might I suggest as an alternative that we just leave this as a branch rather than feature flag. Feature flagging this doesn't seem critical and it is yet more code that we will need to clean up later.

For what it's worth, I was thinking we'd push through this set of tickets fairly quickly (basically, as quickly as we can backfill Folio IDs and reindex), and there's minimal clean-up involved. This is a tiny PR after all. I'm planning to shepherd this through.

justinlittman (approved):
Yeah, but cleaning up feature flag is just one more place to accidentally introduce bugs or leave behind cruft.

PR #3984, opened 2023-03-10
Pretty and reduce JSON in DescriptionsGroup spec.

justinlittman:
The test was incomprehensible.

### DIFF
--- a/spec/services/descriptions_grouper_spec.rb
+++ b/spec/services/descriptions_grouper_spec.rb
@@ -17,9 +17,6 @@ RSpec.describe DescriptionsGrouper do
     JSON.parse(
       <<~JSON
       {
-        "title": [{
-          "value": "Fisherman (Peter Needham)"
-        }],
         "contributor": [{
           "name": [{
             "value": "Wong, Martin",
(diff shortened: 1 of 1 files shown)


spec/services/descriptions_grouper_spec.rb
```diff
@@ -183,7 +682,7 @@
       expect(run.dig(druid2, "form3.type")).to eq("genre")
       expect(run.dig(druid2, "form4.value")).to eq("electronic resource")
       expect(run.dig(druid2, "form4.type")).to eq("form")
-      expect(run.dig(druid2, "form5.value")).to eq("1 CD-ROM ; 4 3/4 in. + 1 user's guide (loose-leaf ; 23 cm.)")
+      expect(run.dig(druid2, "form5.value")).to eq("1 CD-ROM ; 4 3/4 in. + 1 users guide (loose-leaf ; 23 cm.)")
```
mjgiarlo, line 685: Can you say more about these quote changes in the expectations?

justinlittman: Yes. I got rid of all of the single and double quotes from the test JSON. They are irrelevant to the test and made the JSON harder to deal with.

mjgiarlo: OK, no problem.

mjgiarlo (approved):

PR #3990, opened 2023-03-15
Fix catkey editing.

### LINKED ISSUE #3985: Updating catalog record id using the pencil icon does NOT include the "previous [folio|symphony]" in the cocina
**Describe the bug**
1.  Find an object with a ckey  https://argo-stage.stanford.edu/view/druid:ct013cq3840
2.  Open a new version
3. click on the pencil by the ckey in the "Details" section
image
4. change the ckey by editing the textbox
image
5. click on cocina model button.
6.  observe that previous symphony catalogLink is not present in cocina:
image

The same is also true for editing folio HRIDs.

justinlittman:
closes #3985

Support editing catalog record ids.

Unit, QA

### DIFF
--- a/app/forms/catalog_record_id_form.rb
+++ b/app/forms/catalog_record_id_form.rb
@@ -73,28 +73,15 @@ class CatalogRecordIdForm < Reform::Form
 
   # this is overriding Reforms save method, since we are persisting catalog record IDs in cocina only
   def save_model
-    # fetch the existing previous catalog record ID values from the cocina object
-    existing_previous_catalog_record_ids = CatalogRecordId.new(model).previous_links
+    refresh_catalog_record_id_from_form = catalog_record_ids.find { |id| id.refresh && id._destroy != "1" }&.value
+    non_refresh_catalog_record_ids_from_form = catalog_record_ids.filter_map { |id| id.value if id.refresh != true && id._destroy != "1" }
+    catalog_record_ids_from_form = [refresh_catalog_record_id_from_form].compact + non_refresh_catalog_record_ids_from_form
 
-    # these are all of the existing catalog record ID values the user wants to remove (i.e. for which they clicked the trash icon)
-    removed_catalog_record_ids = catalog_record_ids.filter_map { |id| id.value if id._destroy == "1" }
-
-    # build an array of all previous catalog record IDs (which includes the existing previous catalog record IDs plus any newly removed ones)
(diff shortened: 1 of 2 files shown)


andrewjbtw:
I tested this on stage for Symphony editing and QA for Folio editing.

For Symphony I checked:

- Changing a catkey
- Removing a catkey
- Adding a second catkey and then removing just one of the catkeys

I checked both items and collections. Everything is working as expected for Symphony. There are no "previous Folio" but I assume that's expected because Folio isn't live yet, so the Folio HRIDs are syncing only with the current Symphony.

For Folio I checked the same things as for Symphony (on Argo-QA). In this case, I got all the expected results with "Previous Folio". However, I did not get "Previous Symphony" and Symphony catkeys remained in the record as if they were live:

```
 "identification": {
    "catalogLinks": [
      {
        "catalog": "previous folio",
        "refresh": false,
        "catalogRecordId": "a10065784"
      },
      {
        "catalog": "previous folio",
        "refresh": false,
        "catalogRecordId": "a1006578"
      },
      {
        "catalog": "symphony",
        "refresh": true,
        "catalogRecordId": "10065784"
      }
    ],
```
I'm not sure if this is actually an issue, or if things will just work after the migration. There shouldn't be any "current" Symphony catkeys following the migration so I don't think we would leave "orphaned" catkeys behind.

justinlittman:
@andrewjbtw There is no Folio to Symphony syncing (only the other direction). If there is a need to migrate "symphony" catalog links to "previous symphony" catalog links after the migration, that should be ticketed, please (though I'm not sure what the value of that is).

andrewjbtw:
Right, I don't think there would much value in turning all catkeys into "previous symphony" following the cutover. But it does raise a question about what we're going to do with all the catkeys once Symphony is retired. Or has that been ticketed already?

ndushay:
@arcadiafalcone - there's a question for you here.

arcadiafalcone:
We haven't decided yet what to do with Symphony catkeys post-migration. The two most likely options are a) leave the Symphony side as-is or b) delete all Symphony catkeys formerly-current and previous.

mjgiarlo:
@andrewjbtw @arcadiafalcone @justinlittman @aaron-collier  I'm not sure how much of a priority this is in the current WC, but I'm seeing this might be the last Argo issue/PR on the board.

justinlittman:
@mjgiarlo I think this just needs a review and then can be merged.

PR #3995, opened 2023-03-16
increase label size to improve readability

### LINKED ISSUE #3992: Label on APO creation page should be on one line
image

last text line should display as
Collection FOLIO Instance HRID

peetucket:

Fixes #3992 

Looks like this now:

Screen Shot 2023-03-16 at 10 14 20 AM

Localhost

### DIFF
--- a/app/views/apo/_form.html.erb
+++ b/app/views/apo/_form.html.erb
@@ -83,7 +83,7 @@
 
     <div id="create-collection-catalog_record_id" data-apo-form-target="catalogRecordIdFields" hidden>
       <div class="mb-3">
-        <%= collection.label :collection_catalog_record_id, "Collection #{CatalogRecordId.label}", class: "col-sm-2 control-label" %>
+        <%= collection.label :collection_catalog_record_id, "Collection #{CatalogRecordId.label}", class: "col-sm-4 control-label" %>
         <div class="col-sm-10">
           <%= collection.text_field :collection_catalog_record_id, class: 'form-control' %>
         </div>


ndushay (approved):

app/views/apo/_form.html.erb
```diff
@@ -83,7 +83,7 @@
 
     <div id="create-collection-catalog_record_id" data-apo-form-target="catalogRecordIdFields" hidden>
       <div class="mb-3">
-        <%= collection.label :collection_catalog_record_id, "Collection #{CatalogRecordId.label}", class: "col-sm-2 control-label" %>
+        <%= collection.label :collection_catalog_record_id, "Collection #{CatalogRecordId.label}", class: "col-sm-4 control-label" %>
```
jcoyne, line 86: Aren't the 4 + 10 supposed to = 12?

jcoyne: See https://getbootstrap.com/docs/5.2/forms/layout/#horizontal-form

jcoyne: If you want the label over the controls, I don't think you want column values at all.  If you want it horizontal, I think there's a missing "row" class on line 85.

peetucket: Yes, it is odd choice...even though they are in two columns in the HTML, they are always split into two rows for display.  So...I'm not sure how much is intentionally on different rows vs not and what the design should actually be, but I'm reluctant to make that one change here when the rest of the form is in the same style (having labels and values in two columns) of the same row.  Or go too far down a design review for a pretty small visual change on this form.  See #3996 for a small change

PR #4012, opened 2023-04-20
Order notes with no labels and types together, with no leading empty cells

### LINKED ISSUE #3978: Cocina descriptive metadata downloads: when sorting notes, null displayLabels and/or types should be treated as having the same value.
(Note: I've rewritten this ticket to reflect the discussion below.)

Per @arcadiafalcone in the comments below, when sorting note fields for the Cocina descriptive metadata download spreadsheet (in the Bulk actions, not related to the Modsulator)

>null displayLabels and/or types should be treated as having the same value.

The current sorting logic is leading to a really odd-looking display where each note field is placed in a new column:

Screenshot 2023-03-08 at 3 26 44 PM

**User Impact**
Oddly-organized spreadsheets make it even harder to work with descriptive metadata. It's very difficult to read over the notes fields when they sprawl horizontally like this.

**To Reproduce**

I was able to create a set of druids that reproduce this behavior in the staging environment.

1. Load the items under the [project tag "note field numbering issue"](https://argo-stage.stanford.edu/catalog?f%5Bproject_tag_ssim%5D%5B%5D=note+field+numbering+issue) 
3. Click on the "Bulk actions" link from the search result (not the link in the top red bar)
4. Create a new bulk action
5. Choose "Download descriptive metadata spreadsheet"
6. Open the CSV
7. Scroll to the right and see the weird note field alignment

**Expected behavior**
The numbering of notes should start at 1 for each druid.

**Additional context**
I did test re-importing a spreadsheet that looks like this in stage and fortunately it doesn't create 15 notes when the first note field is numbered "15", so the problem may be just in the export function. Importing metadata seems to be functioning as expected.

lwrubel:
Resolves #3978. 

Unit and deploy to stage.

Updated spreadsheet with export from [Andrew's sample tag](https://argo-stage.stanford.edu/catalog?f%5Bproject_tag_ssim%5D%5B%5D=note+field+numbering+issue):

[t3978-cocina-desc-export-notes.csv](https://github.com/sul-dlss/argo/files/11296310/t3978-cocina-desc-export-notes.csv)

### DIFF
--- a/app/services/notes_grouper.rb
+++ b/app/services/notes_grouper.rb
@@ -63,15 +63,6 @@ class NotesGrouper
                                    [hash["note#{num}.displayLabel"], hash["note#{num}.type"]] == [label_for_note_number, type_for_note_number]
                                end
 
-        when 0
-          # If there is no matching note number, increment the
-          # current highest note number and use it for note
-          # elements for this type
-          max_note = ordered_mapping.max_by { |k, _v| k[/\d+/].to_i }.first
-          field, number = max_note.scan(/(\D+)(\d+)/).first
-          new_note = "#{field}#{number.to_i.succ}"
-          ordered_mapping[new_note] = [label_for_note_number, type_for_note_number]
-          new_note
         when 1
           # If there is only one matching note number in the mapping, use it and move on.
           ordered_mapping.key([label_for_note_number, type_for_note_number])
@@ -79,7 +70,8 @@ class NotesGrouper
           # If there are multiple notes of this type, e.g., an
           # item with multiple notes of type "abstract" with a
           # nil display label, use the first note number not
-          # already used
+          # already used.
+          # Also applies when there are no displayLabels or types for the note
           ordered_mapping.find do |mapped_note_number, (label_value, type_value)|
             [label_value, type_value] == [label_for_note_number, type_for_note_number] && !description.key?(key.sub(/^old_note\d+/, mapped_note_number))
           end&.first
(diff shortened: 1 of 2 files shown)


app/services/notes_grouper.rb
```diff
@@ -64,14 +64,10 @@ def group
                                end
 
         when 0
-          # If there is no matching note number, increment the
-          # current highest note number and use it for note
-          # elements for this type
-          max_note = ordered_mapping.max_by { |k, _v| k[/\d+/].to_i }.first
```
lwrubel, line 70: This code seemed to only get used in the situation of no displayLabels or types and was the cause of the new notes columns getting created for notes that didn't have labels or types. It may be I'm not understanding something about the grouping logic, however. 

I left the `when 0` statement, even though the action is now the same as with the `else` statement, although I could remove this. Was thinking it might be helpful to have the `0` scenario not mixed with the `>1` scenario for clarity with this complicated logic, but open to removing it and just adding a comment.

spec/services/descriptions_grouper_spec.rb
```diff
@@ -858,4 +858,380 @@
       expect(run.dig(druid6, "note12.displayLabel")).to eq("Contents")
     end
   end
+
+  context "with notes containing multiple repeat values including those with no displayLabel or type" do
```
lwrubel, line 862: The grouping logic is complicated enough that it seemed helpful to create some further testing scenarios, acknowledging there is overlap with the existing ones.

justinlittman (approved):

app/services/notes_grouper.rb
mjgiarlo: Oooh, if this isn't needed, that'd be great.

mjgiarlo (approved):

app/services/notes_grouper.rb
lwrubel: Have dropped this.

arcadiafalcone:
@lwrubel Looks correct in stage!

PR #4018, opened 2023-05-02
Create administrative tags before workflow

mjgiarlo:

Fixes #1043

This commit makes a small tweak to the Argo registration service, creating administrative tags for an object before creating its workflow. This should *ensure* the problematic behavior of getting too far in a workflow (the `goobiWF` in particular) before administrative tags are created since the tag creation in DSA is synchronous.

- [x] CI

### DIFF
--- a/app/services/registration_service.rb
+++ b/app/services/registration_service.rb
@@ -11,9 +11,10 @@ class RegistrationService
     response = Dor::Services::Client.objects.register(params: model)
     druid = response.externalIdentifier
 
+    Dor::Services::Client.object(druid).administrative_tags.create(tags:) unless tags.empty?
+
     WorkflowClientFactory.build.create_workflow_by_name(druid, workflow, version: "1")
 
-    Dor::Services::Client.object(druid).administrative_tags.create(tags:) unless tags.empty?
     Success(response)
   rescue Dor::Services::Client::UnexpectedResponse => e
     Failure(e)
--- a/app/services/registration_service.rb
+++ b/app/services/registration_service.rb
@@ -11,6 +11,8 @@ class RegistrationService
     response = Dor::Services::Client.objects.register(params: model)
     druid = response.externalIdentifier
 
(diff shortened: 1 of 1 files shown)


app/services/registration_service.rb
```diff
@@ -11,9 +11,10 @@ def self.register(model:, workflow:, tags:)
     response = Dor::Services::Client.objects.register(params: model)
     druid = response.externalIdentifier
 
+    Dor::Services::Client.object(druid).administrative_tags.create(tags:) unless tags.empty?
+
     WorkflowClientFactory.build.create_workflow_by_name(druid, workflow, version: "1")
```
mjgiarlo, line 16: AFAIK, creating admin tags before the creation of a workflow should be safe. Thoughts, @sul-dlss/infrastructure-team ?

app/services/registration_service.rb
```diff
@@ -11,9 +11,10 @@ def self.register(model:, workflow:, tags:)
     response = Dor::Services::Client.objects.register(params: model)
     druid = response.externalIdentifier
 
+    Dor::Services::Client.object(druid).administrative_tags.create(tags:) unless tags.empty?
```
ndushay, line 14: please add a comment indicating WHY admin tags should be created first.

ndushay (dismissed):

app/services/registration_service.rb
mjgiarlo, line 14:
```suggestion
    # NOTE: Create administrative tags before the workflow is created, else workflows
    #       that rely on admin tags (e.g., `goobiWF`) could sporadically fail.
    Dor::Services::Client.object(druid).administrative_tags.create(tags:) unless tags.empty?
```

peetucket:
Good snooping and catch to find this Mike.

jmartin-sul (approved):

ndushay (approved):

PR #4019, opened 2023-05-03
Remove NewRelic configuration and dependency

mjgiarlo:

Argo does not use NewRelic and we do not want it.

CI


ndushay (approved):

andrewjbtw:
FWIW, Argo did use NewRelic to analyze performance a few years ago. I guess that stopped being useful?

mjgiarlo:
@andrewjbtw 💬 
> FWIW, Argo did use NewRelic to analyze performance a few years ago. I guess that stopped being useful?

Yes, the consensus was that it was not very useful then or now.

PR #4021, opened 2023-05-08
[SPIKE] Switch to the standard-rails ruleset

mjgiarlo:

To expand Argo's usage of `standard` which now has a pattern for building plugins, e.g., `standard-rails`.

Note the comment in the `Gemfile` regarding the need to continue carrying the `rubocop-rails` dependency as long as (1) the `standard` plugin setup stays the same, **and** (2) we continue using `rubocop` tooling over `standard` tooling (thinking specifically about linting in CI and local editor setup). See https://github.com/standardrb/standard-rails/issues/6#issuecomment-1548256212.

CI (& I believe the coverage ding is not worth worrying over)


.rubocop.yml
```diff
@@ -3,16 +3,19 @@ inherit_mode:
     - Exclude
 
 require:
+  - standard
```
mjgiarlo, line 6: First-in wins, so let `standard` have the 🏆

.rubocop/custom.yml
```diff
@@ -65,148 +65,6 @@ RSpec/Rails/HaveHttpStatus: # new in 2.12
...
-Performance/StringIdentifierArgument: # new in 1.13
-  Enabled: true
-Performance/StringInclude: # new in 1.7
-  Enabled: true
-Performance/Sum: # new in 1.8
-  Enabled: true
-Rails/ResponseParsedBody: # new in 2.18
-  Enabled: true
-Rails/ThreeStateBooleanColumn: # new in 2.19
-  Enabled: true
```
mjgiarlo, line 209: We get all these from the standard style guides (configured [above](https://github.com/sul-dlss/argo/pull/4021/files#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR13-R15))

.rubocop/custom.yml
```diff
@@ -231,3 +89,9 @@ RSpec/Rails/MinitestAssertions: # new in 2.17
   Enabled: true
 RSpec/Rails/TravelAround: # new in 2.19
   Enabled: true
+FactoryBot/AssociationStyle: # new in 2.23
+  Enabled: true
+FactoryBot/FactoryAssociationWithStrategy: # new in 2.23
+  Enabled: true
+FactoryBot/RedundantFactoryOption: # new in 2.23
+  Enabled: true
```
mjgiarlo, line 97: These are new.

.standard.yml
```diff
@@ -1 +1,4 @@
 parallel: true
+format: progress
+plugins:
+  - standard-rails
```
mjgiarlo, line 4: Here's how to pull in the `standard-rails` style guide.

app/controllers/application_controller.rb
```diff
@@ -22,10 +22,11 @@ def current_user
...
       cur_user.display_name = request.env["displayName"]
-      if request.env["eduPersonEntitlement"]
-        cur_user.webauth_groups = request.env["eduPersonEntitlement"].split(";")
-      elsif Rails.env.development?
-        cur_user.webauth_groups = ENV.fetch("ROLES", "").split(";")
+      cur_user.webauth_groups = if request.env["eduPersonEntitlement"]
+        request.env["eduPersonEntitlement"].split(";")
+      else
+        # NOTE: config.user_groups is only defined in the development environment
+        Rails.application.config.try(:user_groups) || []
```
mjgiarlo, line 29: Makes the linter happy.

.standard.yml
ndushay: add a comment?   also interesting we don't need ones for performance or custom plugins

Gemfile
```diff
@@ -97,10 +97,15 @@ group :test, :development do
   gem "http_logger", require: false # Change this to `true` to see all http requests logged
   gem "pry-remote" # allows you to attach remote session to pry
   gem "rspec-rails", "~> 5.0"
+  # NOTE: the `standard` way does not support using the rubocop CLI with
+  # extensions like standard-rails, so until that changes or we decide to switch
+  # to the standardrb CLI (which would need CI and editor support), we have to
+  # carry both this dependency and standard-rails.
```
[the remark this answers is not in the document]
ndushay: bummer.   that seems like a down vote to me.

.rubocop.yml
```diff
@@ -3,16 +3,19 @@ inherit_mode:
...
   - rubocop-performance
   - rubocop-rails
   - rubocop-rspec
-  - standard
 
 inherit_gem:
   standard: config/base.yml
+  standard-performance: config/base.yml
+  standard-custom: config/base.yml
```
ndushay, line 14: so were we not using these before?

.standard.yml
mjgiarlo: Thinking about Sandi Metz's advice about not adding comments where code already tells the story... the YAML here tells us this is how to load a `standard` plugin such as `standard-rails`, so what would the comment say?

Gemfile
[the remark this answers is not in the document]
mjgiarlo: It's a bummer for sure. It's still a good ruleset, one might argue. And once we get everything over to standard (including CI), we should be able to drop `rubocop-rails`.

.rubocop.yml
mjgiarlo: We were using the `standard-performance` rules, but the recommended way to include them changed. `standard-custom` is a new thing that we get for free with the `standard` gem, but requires opting in via config. I think it's literally one new rule. This change is keeping up with the latest prescribed practices.

jmartin-sul (approved):

PR #4024, opened 2023-05-15
Stop pinning reform-rails to pick up patch in 0.2.5

mjgiarlo:

Keep pace with reform-rails now that the bug we found was patched.

CI

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,10 +60,7 @@ gem "nokogiri", "~> 1.14"
 gem "prawn", "~> 1"
 gem "prawn-table"
 gem "rake"
-# TODO: Deal with this
-# pinned because 0.2.4 broke the build: https://app.circleci.com/pipelines/github/sul-dlss/argo/6018/workflows/fd256885-f9aa-4cf6-bb60-8bd2be232f49/jobs/13023/tests#failed-test-1
-# Specifically because of the `TagsForm` thinking a new tag was a tag to update
-gem "reform-rails", "< 0.2.4"
+gem "reform-rails"
 gem "retries"
 gem "ruby-prof"
 gem "rubyzip"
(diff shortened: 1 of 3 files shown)


jmartin-sul (commented):
it looks like this unpins, but doesn't actually bump the version?  i still see [0.2.3 in `Gemfile.lock`](https://github.com/sul-dlss/argo/blob/2ae0f1ca796eec1c7478993c995ebf47a4c891d6/Gemfile.lock#L452)?

mjgiarlo:
@jmartin-sul 💬 
> it looks like this unpins, but doesn't actually bump the version? i still see [0.2.3 in `Gemfile.lock`](https://github.com/sul-dlss/argo/blob/2ae0f1ca796eec1c7478993c995ebf47a4c891d6/Gemfile.lock#L452)?

Good eye! I've got that included now.

mjgiarlo:
@jmartin-sul Whaddya know? I'll take a look. Thanks for putting this through its paces.

mjgiarlo:
Wrote up a related bug issue in reform-rails: https://github.com/trailblazer/reform-rails/issues/101 I'll

mjgiarlo:
@jmartin-sul I made a change to the populator in the `TagsForm` that fixes the specs atop reform-rails 0.2.5 (and still works atop 0.2.3, FWIW). I tested the patch in QA by issuing a variety of tag changes (simple add, simple delete, simple update, combo of adds and updates and deletes), and all tests I performed had the expected results.

I also paired with @jcoyne to try and make sense of how this ever worked... and we were both stumped. Given the pattern Argo's `TagsForm` uses in its populator is not used *anywhere* else in our portfolio, I was not inclined to give grokking this any more time than I already have but check me on this.

Long story short, I think this is ready for another look.

aaron-collier (approved):

PR #4037, opened 2023-06-19
Add validation for catalog record IDs/barcodes before saving in bulk action

### LINKED ISSUE #4034: Bulk action to update catkeys and barcodes from a CSV should validate the catkey or barcode before opening the object
**Describe the bug**
The bulk action to update a catkey, barcode, or both will open an object if it is not already opened. When opening an object, the bulk action supplies a version description like

> Catkey updated to 333.

The issue here is that the object is opened and the version description is applied before the job actually makes the update. If the update fails, the version is left open with an inaccurate version description, as the update never took place.

Example: I ran into this today when I uploaded an incorrectly formatted spreadsheet. I made two mistakes:

- I meant to update barcodes but my spreadsheet header said "Catkey"
- I meant input the barcode but had copied a column that was source id

The result was that the bulk action:

- Opened the item
- Described the version as "Catkey updated to googlebooks:stanford_36105022210830."
- Correctly failed to update either the catkey or barcode because the input isn't valid for those fields

These objects are now left open with an inaccurate version description. Example: https://argo.stanford.edu/view/druid:hn600xx9248

**User Impact**
This issue means that objects are opened unnecessarily with incorrect version descriptions. There is no user-facing way to correct these version descriptions in bulk.

**To Reproduce**
1. Go to bulk actions
2. Select Edit catkeys and barcodes from CSV
3. Upload a CSV with invalid data for the catkey or barcode

**Expected behavior**
Validating the input first to check for a valid catkey or barcode format would save time. I don't think it's necessary to also check that the catkey or barcode exist in the catalog, as that would introduce a lot of complexity. The format validation should be built into Cocina.

peetucket:

Fixes #4034 - validate that catalog record IDs (either catkey or folio_instance_hrid) and/or barcodes are valid before attempting to version/save an object.  This stops opening the object and adding a message, only to have the save fail because it was invalid.

Note: even though the ticket describes the problem as updates from a CSV file, the same thing would happen for the bulk job that allows you paste in druids, barcodes and catalog record IDs.  This will fix both jobs (as they both use the same underlying update code, just different code for getting the input data loaded).

~~[HOLD] to test on QA or stage first~~

Added specs.  Tested on QA with both CSV file and pasted values (i.e. both related jobs)

### DIFF
--- a/app/jobs/set_catalog_record_ids_and_barcodes_job.rb
+++ b/app/jobs/set_catalog_record_ids_and_barcodes_job.rb
@@ -27,8 +27,12 @@ class SetCatalogRecordIdsAndBarcodesJob < GenericJob
         args[:refresh] = refresh[i] if refresh
         args[:barcode] = barcodes[i] if barcodes && cocina_object.dro?
         change_set = change_set_for(cocina_object)
-        change_set.validate(args)
-        update_catalog_record_id_and_barcode(change_set, args, log) if change_set.changed?
+        if change_set.validate(args)
+          update_catalog_record_id_and_barcode(change_set, args, log) if change_set.changed?
+        else
+          log.puts("#{Time.current} Invalid #{CatalogRecordId.label}/barcode for #{cocina_object.externalIdentifier}")
+          bulk_action.increment(:druid_count_fail).save
+        end
       end
     end
   end
--- a/app/models/item_change_set.rb
+++ b/app/models/item_change_set.rb
@@ -22,6 +22,7 @@ class ItemChangeSet < ApplicationChangeSet
     with: /\A(2050[0-9]{7}|245[0-9]{8}|36105[0-9]{9}|[0-9]+-[0-9]+)\z/,
     allow_blank: true
   }
+  validate :format_of_catalog_record_ids
 
   def self.model_name
     ::ActiveModel::Name.new(nil, nil, "Item")
@@ -45,6 +46,12 @@ class ItemChangeSet < ApplicationChangeSet
     self.use_statement = model.access.useAndReproductionStatement
     self.license = model.access.license
     setup_view_access_with_cdl_properties(model.access)
+  end
+
+  def format_of_catalog_record_ids
+    return if catalog_record_ids.blank? || CatalogRecordId.valid?(catalog_record_ids)
+
+    errors.add(:catalog_record_ids, "are not a valid format")
   end
 
   def sync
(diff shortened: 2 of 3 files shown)


app/jobs/set_catalog_record_ids_and_barcodes_job.rb
```diff
@@ -27,8 +27,12 @@ def perform(bulk_action_id, params)
...
         change_set = change_set_for(cocina_object)
-        change_set.validate(args)
-        update_catalog_record_id_and_barcode(change_set, args, log) if change_set.changed?
+        if change_set.validate(args)
+          update_catalog_record_id_and_barcode(change_set, args, log) if change_set.changed?
+        else
+          log.puts("#{Time.current} Invalid #{CatalogRecordId.label}/barcode for #{cocina_object.externalIdentifier}")
+          bulk_action.increment(:druid_count_fail).save
+        end
       end
```
peetucket, line 36: basically validate the change set before actually doing the update...if validation fails, log and skip

app/models/item_change_set.rb
```diff
@@ -47,6 +48,12 @@ def setup_properties!(_options)
     setup_view_access_with_cdl_properties(model.access)
   end
 
+  def format_of_catalog_record_ids
+    return if catalog_record_ids.blank? || CatalogRecordId.valid?(catalog_record_ids)
+
+    errors.add(:catalog_record_ids, "are not a valid format")
+  end
+
   def sync
```
peetucket, line 57: this adds validation of catalog record IDs to the item change set (it was missing before) ...

the logic difference between folio vs symphony already exists and is abstracted away inside the `CatalogRecordId.valid?` method

Note: This adds some extra validation to the `ItemChangeSet` which is used by other parts of the code.  I believe this is OK because as soon as you save the change set, the catalog record ID would be validated anyway, and would throw an error if invalid on save.  So I don't think there are places in the code that would have worked ok that will now fail this addition (i.e. because a change set validation should be followed by a save anyway).

spec/jobs/set_catalog_record_ids_and_barcodes_csv_job_spec.rb
```diff
@@ -65,7 +71,33 @@
...
+        allow(subject).to receive(:update_catalog_record_id_and_barcode)
+        subject.perform(bulk_action.id, {csv_file:})
+      end
+
+      it "only attempts to update the catalog_record_id/barcode for the one druids with valid barcode/catalog_record_id" do
+        expect(bulk_action.druid_count_total).to eq druids.length
+        expect(bulk_action.druid_count_fail).to eq 2
+        expect(subject).to have_received(:update_catalog_record_id_and_barcode).with(ItemChangeSet, Hash, buffer).exactly(druids.length - 2).times
+      end
     end
```
peetucket, line 101: adding new tests and extending existing tests so they will continue work regardless of if we have folio vs symphony enabled

mjgiarlo (approved):

PR #4068, opened 2023-07-13
Update standard gems and require standard-custom

### LINKED ISSUE #4067: figure out how to unpin standard gem
It was pinned as part of this week's dependency updates, because our approach to running it errors out on versions >= 1.30.  See:
* https://github.com/sul-dlss/argo/pull/4065
* https://github.com/standardrb/standard/issues/569

Possible fixes:
* waiting for a new version of standard -- the issue linked above links to a PR that may fix this for our usage
* invoke via `standard` instead of `rubocop` (though my vague recollection is that our particular setup necessitates running via `rubocop`, so this approach might be a pain to do, or unable to meet our rubocop/standard needs)

mjgiarlo:
# Why was this change made? 🤔

Fixes #4067

This commit updates the standard gems and uses the new pattern for requiring standardrb plugins in the `require` section of the rubocop configuration.

CI

No

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -97,14 +97,9 @@ group :test, :development do
   gem "http_logger", require: false # Change this to `true` to see all http requests logged
   gem "pry-remote" # allows you to attach remote session to pry
   gem "rspec-rails", "~> 5.0"
-  # NOTE: the `standard` way does not support using the rubocop CLI with
-  # extensions like standard-rails, so until that changes or we decide to switch
-  # to the standardrb CLI (which would need CI and editor support), we have to
-  # carry both this dependency and standard-rails.
-  gem "rubocop-rails"
   gem "rubocop-rspec"
   gem "sqlite3", "~> 1.4.2"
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -97,14 +97,9 @@ group :test, :development do
   gem "http_logger", require: false # Change this to `true` to see all http requests logged
   gem "pry-remote" # allows you to attach remote session to pry
   gem "rspec-rails", "~> 5.0"
-  # NOTE: the `standard` way does not support using the rubocop CLI with
-  # extensions like standard-rails, so until that changes or we decide to switch
-  # to the standardrb CLI (which would need CI and editor support), we have to
-  # carry both this dependency and standard-rails.
-  gem "rubocop-rails"
```
mjgiarlo, line 104: TIL that adding `standard-rails` to the `require` section of the rubocop config means we no longer have to carry this direct dependency. That's a minor win, at least.

jmartin-sul (approved):

PR #4083, opened 2023-08-10
Prevent catalog record IDs in item registration forms during ILS cutover

### LINKED ISSUE #4035: Folio cutover: when no ILS is available, reject registration form entry if it contains a catkey
During the Folio migration cutover week, I would like to keep registration available so that the digitization labs and others can continue their work. Since catkey/Folio HRID is an optional field during registration, it should be possible to continue to support registration as long as no one enters a catkey/Folio HRID.

Proposal: modify the validation that we already have for catkey/Folio HRID to reject all entries in that field. I'm not sure if that's implementable but it seems like it would require the least modification to the UI. We would still show the field but it would reject any value entered.

We would obviously warn users and ask them not to enter any catalog ids until Folio is live.

mjgiarlo:
# Why was this change made? 🤔

Fixes #4035

 This commit prevents users from providing catalog record IDs in both item registration forms (UI and CSV) when the new `ils_cutover_in_progress` feature flag is flipped on.

CI, local

No.


justinlittman (approved):

ndushay (approved):
oh no - failing tests!

spec/rails_helper.rb
```diff
@@ -16,6 +16,7 @@
 WebMock.disable_net_connect!(allow_localhost: true,
   allow: [
     "https://chromedriver.storage.googleapis.com",
+    "https://googlechromelabs.github.io",
```
mjgiarlo, line 19: Google added a new Chrome base URL that has been failing Argo builds for two days. This is a separate commit which I added to this PR so the build can pass.

config/settings.yml
```diff
@@ -100,3 +100,5 @@ redis_url: 'redis://localhost:6379/'
 
 enabled_features:
   folio: false
+
+ils_cutover_in_progress: false
```
mjgiarlo, line 104: Add the new feature flag.

spec/services/csv_upload_validator_spec.rb
```diff
@@ -3,31 +3,59 @@
 require "rails_helper"
 
 RSpec.describe CsvUploadValidator do
+  let(:catalog_record_id_header) { CatalogRecordId.label.downcase.tr(" ", "_") }
+  let(:csv) { "Druid,#{catalog_record_id_header},Barcode\ndruid:bb396kf5077,13157971,\ndruid:bb631ry3167,13965062,\n" }
+  let(:headers) { ["Druid", catalog_record_id_header] }
   let(:validator) { described_class.new(csv:, headers:) }
 
-  let(:csv) { "Druid,Catkey,Barcode\ndruid:bb396kf5077,13157971,\ndruid:bb631ry3167,13965062,\n" }
```
mjgiarlo, line 8: Somehow we were still hard-coding this instead of switching the value based on the Folio feature flag. 🤷🏻‍♂️

PR #4086, opened 2023-08-11
Disable manage release on cutover flag

### LINKED ISSUE #4030: Folio cutover: during transition week, turn off release to SW functionality
NOTE: If possible, implement this behind a "ils-cutover-in-progress" feature flag.

During the week when neither Symphony nor Folio will be available, we should "freeze" SDR with respect to releases to SearchWorks.

There are two ways this could be done:

- **Option 1**: Turn off the release bulk action entirely and gray out the release button on show pages
- **Option 2**: Turn off only the release path that updates a MARC record
  - This would allow release to EW and release without a MARC record to continue to function

Given the outage should only last for a week and SearchWorks will be largely frozen for other reasons, I think Option 1 should be ok. It's the simplest course of action.

aaron-collier:
# Why was this change made? 🤔

Fixes #4030 

Requires the flag introduced in #4085 

CI

### DIFF
--- a/app/components/bulk_actions_form_component.rb
+++ b/app/components/bulk_actions_form_component.rb
@@ -16,7 +16,7 @@ class BulkActionsFormComponent < ApplicationComponent
   def action_types
     grouped_options_for_select [
       ["Perform actions", [
-        ["Manage release", new_manage_release_job_path(search_of_druids)],
+        Settings.ils_cutover_in_progress ? ["DISABLED: Manage release", nil] : ["Manage release", new_manage_release_job_path(search_of_druids)],
         ["Reindex", new_reindex_job_path(search_of_druids)],
         ["Republish objects", new_republish_job_path(search_of_druids)],
         ["Purge", new_purge_job_path(search_of_druids)],


app/components/bulk_actions_form_component.rb
```diff
@@ -16,7 +16,7 @@ def search_of_druids
   def action_types
     grouped_options_for_select [
       ["Perform actions", [
-        ["Manage release", new_manage_release_job_path(search_of_druids)],
+        Settings.ils_cutover_in_progress ? ["DISABLED: Manage release", nil] : ["Manage release", nil] : new_manage_release_job_path(search_of_druids)],
```
justinlittman, line 19: I think you want:
```
Settings.ils_cutover_in_progress ? ["DISABLED: Manage release", nil] : ["Manage release", new_manage_release_job_path(search_of_druids)],
```

justinlittman (changes requested):

app/components/bulk_actions_form_component.rb
aaron-collier: Oh geez, copy pasta, thank you.

justinlittman (approved):

PR #4089, opened 2023-08-12
adding comments pertaining to folio cutover

ndushay:
# Why was this change made? 🤔

clarify comments;  avoid false matches on Symphony specific language

comments only

no


justinlittman (approved):
I don't feel strongly about this, but wondering if it wouldn't be better to use "Folio" instead of "ILS catalog", as in the future when we are grepping for Folio related code we are more likely to find it.

PR #4091, opened 2023-08-14
Disable metadata refresh on show page during ILS cutover

### LINKED ISSUE #4031: Folio cutover: Turn off metadata refresh on show page
NOTE: If possible, implement this behind a "ils-cutover-in-progress" feature flag.

During the period where both Symphony and Folio are unavailable, the "Refresh" option under the "Manage description" button on the Argo item and collection show pages should be grayed out, even when the object is opened for versioning.

Screenshot 2023-06-08 at 4 38 46 PM

mjgiarlo:
# Why was this change made? 🤔

Fixes #4031

This commit disables the metadata refresh button on the show page when the ILS cutover flag is flipped on. Not specified in the issue, I also included a small change to the "Refresh" label to make it clear to users *why* metadata refresh is not enabled. I pondered whether we should also include similar prevention code in the ItemsController, but since this operation is a POST, I figured it was very unlikely one of our users was going to bypass the UI controls and try to hack in a metadata refresh. Check me on this.

CI

No

### DIFF
--- a/app/components/show/controls_component.html.erb
+++ b/app/components/show/controls_component.html.erb
@@ -38,8 +38,8 @@
       </button>
       <ul class="dropdown-menu" aria-labelledby="dropdown-descriptive">
         <% if catalog_record? %>
-          <li><%= link_to 'Refresh', refresh_metadata_item_path(id: druid),
-                                     class: "dropdown-item #{'disabled' if button_disabled?}",
+          <li><%= link_to refresh_button_label, refresh_metadata_item_path(id: druid),
+                                     class: "dropdown-item #{'disabled' if button_disabled? || Settings.ils_cutover_in_progress}",
                                      data: { turbo_method: 'post' } %></li>
         <% end %>
         <li><%= link_to 'Download Cocina spreadsheet', item_descriptive_path(doc, format: :csv), class: 'dropdown-item' %></li>
--- a/app/components/show/controls_component.rb
+++ b/app/components/show/controls_component.rb
@@ -48,6 +48,12 @@ module Show
       !allows_modification?
     end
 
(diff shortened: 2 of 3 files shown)


app/components/show/controls_component.html.erb
```diff
@@ -38,8 +38,8 @@
       </button>
       <ul class="dropdown-menu" aria-labelledby="dropdown-descriptive">
         <% if catalog_record? %>
-          <li><%= link_to 'Refresh', refresh_metadata_item_path(id: druid),
-                                     class: "dropdown-item #{'disabled' if button_disabled?}",
+          <li><%= link_to refresh_button_label, refresh_metadata_item_path(id: druid),
+                                     class: "dropdown-item #{'disabled' if button_disabled? || Settings.ils_cutover_in_progress}",
```
justinlittman, line 42: Why not put this in `button_disabled?` method instead of the ERB?

justinlittman (approved):

app/components/show/controls_component.html.erb
mjgiarlo: I would have preferred that, but it looks like `button_disabled?` governs the behavior of a few buttons, and we only want to disable one of them.

andrewjbtw:
>since this operation is a POST, I figured it was very unlikely one of our users was going to bypass the UI controls and try to hack in a metadata refresh. Check me on this.

I think it's very unlikely someone would try to reverse engineer the process in order to make the request, given that the request still wouldn't work. I'm not even sure how to do this outside the UI, but I'm assuming you need to get your authorization into the POST so you'd have to figure out how to find/get the right token.

PR #4099, opened 2023-08-17
Disable SearchWorks release bulk action in the immediate pre-cutover period

### LINKED ISSUE #4097: Folio cutover:  turn off "release to SW" separately from other ils-cutover-in-progress features
I made a mistake in ticket #4030;  we need to turn off "release to SW" a day before the other ils_cutover_in_progress features.  We do, however, want to turn it back on at the same time as the other features.

Choices:
- we make a separate flag now 
- we deploy an argo update at end of workday Thursday Aug 17  that turns off "release to SW" manually based on PR #4086

mjgiarlo:
# Why was this change made?

Fixes #4097

This commit changes what was done in #4086 to disable releases to SearchWorks not *during* the cutover, but immediately *prior* to the cutover.

# How was this change tested?

CI

# Does your change introduce accessibility violations?

No.

### DIFF
--- a/app/components/bulk_actions_form_component.rb
+++ b/app/components/bulk_actions_form_component.rb
@@ -16,7 +16,7 @@ class BulkActionsFormComponent < ApplicationComponent
   def action_types
     grouped_options_for_select [
       ["Perform actions", [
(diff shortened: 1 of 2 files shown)


ndushay (changes requested):
ils_cutover_sw_release ?   Might be easier to understand next week.

ndushay (approved):

PR #4120, opened 2023-08-23
update maximum number of registered rows instructions

### LINKED ISSUE #4102: Update suggested maximum number of records on item registration form to 500
There's text on the item registration form suggesting that people limit their registration to 15 items at a time.
Screenshot 2023-08-17 at 4 46 38 PM

This limit has actually never been enforced and some users have tried close to 1000 records at a time successfully. The suggested limit was based on experience with the Fedora-based registration system, which was slower and less reliable than what we have now. I think we can increase it quite a bit.

My suggestion is to increase it to 500, based on conversations with users who have been registering in larger amounts. I could be convinced to increase it to a lower number, but 15 is definitely smaller than it needs to be.

peetucket:
# Why was this change made? 🤔

Fixes #4102

@andrewjbtw what do you think of this language?

Screen Shot 2023-08-23 at 11 22 21 AM

Localhost


jmartin-sul (approved):
LGTM if andrew likes it

andrewjbtw:
Looks good to me.

PR #4121, opened 2023-08-23
override styles for error and note text

### LINKED ISSUE #4066: Error message text is no longer red in "workflow view" modal
**Describe the bug**
The text of error messages used to be displayed in red on the workflow modal view. Starting in the past month or so, it started to appear without any color.

Screenshot 2023-07-11 at 12 36 12 PM

**User Impact**
It's a bit harder to pick out the error text when checking on errors.

**To Reproduce**
1. Go to an item that has a workflow error.
2. Click on the workflow where the error occurred.
3. See that all text is black.

**Expected behavior**
Error messages should be in red, as they are elsewhere in the Argo interface.

peetucket:
# Why was this change made? 🤔

Fixes #4066 - ensure error message and note colors do not have their colors overridden

Screen Shot 2023-08-23 at 11 50 56 AM

Localhost

### DIFF
--- a/app/assets/stylesheets/document.scss
+++ b/app/assets/stylesheets/document.scss
@@ -35,11 +35,13 @@ h1, h2, h3, h4, h5, h6 {
 }
 
 .errortext {
-  color: red;
+ color: red;
+ --bs-table-color: red; // override bootstrap table text color
 }
 
 .notetext {
   color: green;
+  --bs-table-color: green;  // override bootstrap table text color
 }
 
 .clear {


jcoyne (commented):
I don't like `!important` as it's never very clear why it's necessary. 

consider using `table .errortext` instead. You also may want to consider adding a comment here about why we're giving such a long descriptor `e.g. to override Bootstrap's .table`

I think the correct way might be to set `--bs-table-color-state: green;`

app/assets/stylesheets/document.scss
```diff
@@ -35,11 +35,11 @@ h1, h2, h3, h4, h5, h6 {
 }
 
 .errortext {
-  color: red;
+  color: red !important;
```
jcoyne, line 38:
```suggestion
  --bs-table-color-state: red;
```

app/assets/stylesheets/document.scss
```diff
@@ -35,11 +35,11 @@ h1, h2, h3, h4, h5, h6 {
...
 }
 
 .notetext {
-  color: green;
+  color: green !important;
```
jcoyne, line 42:
```suggestion
  --bs-table-color-state: green;
```

peetucket:
> I don't like `!important` as it's never very clear why it's necessary.
> 
> consider using `table .errortext` instead. You also may want to consider adding a comment here about why we're giving such a long descriptor `e.g. to override Bootstrap's .table`
> 
> I think the correct way might be to set `--bs-table-color-state: green;`

Ok, sounds fine.  I found the way to override bootstrap's text color and did that.  I also had to leave in the original `color: red` as well since we use that same class outside of tables (like indicating there is an WF error on the show page).

ndushay (approved):
LGTM.

jcoyne:
> . I also had to leave in the original color: red as well since we use that same class outside of tables 

I think that table that uses that should have the bootstrap `table` class on it too, then you wouldn't have to do it two ways.

PR #4131, opened 2023-08-28
add columns to registration job report output

### LINKED ISSUE #4104: Align registration bulk action report with output of registration form entry
When registering items using the registration form entry (no CSV), Argo displays five fields in the registration result list:

- druid
- barcode
- folio instance HRID
- source id
- label

Screenshot 2023-08-17 at 5 01 03 PM

The report provided by the registration CSV option should show the same data in the same order. The barcode and folio instance HRID columns should display even if blank.

Currently, the report only shows druid, source id, and label. We've gotten feedback from users who register with both methods that they'd like a consistent output. Often the output of Argo registration is pasted into tracking spreadsheets set up in the digitization labs that already have been set up with the columns listed above.

peetucket:
# Why was this change made? 🤔

Fixes #4104 - the output of the registration bulk job should include the same columns that are shown on the webpage you get when registering objects

Todo: 
- [x] fix tests

Localhost and CI

### DIFF
--- a/spec/jobs/register_druids_job_spec.rb
+++ b/spec/jobs/register_druids_job_spec.rb
@@ -11,10 +11,10 @@ RSpec.describe RegisterDruidsJob do
   let(:response) { Success(model) }
   let(:fake_log) { double("logger", puts: nil) }
   let(:catalog_link) do
-    instance_double(Cocina::Models::FolioCatalogLink, catalog: "folio", catalogRecordId: "in12345", refresh: true)
+    Cocina::Models::FolioCatalogLink.new(catalog: "folio", catalogRecordId: "in12345", refresh: true)
   end
   let(:identification) do
-    instance_double(Cocina::Models::Identification, barcode: "12345", catalogLinks: [catalog_link], sourceId: "foo:bar1")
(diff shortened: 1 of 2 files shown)


spec/jobs/register_druids_job_spec.rb
```diff
@@ -10,8 +10,11 @@
   let(:bulk_action) { create(:bulk_action) }
   let(:response) { Success(model) }
   let(:fake_log) { double("logger", puts: nil) }
+  let(:catalog_link) do
+    instance_double(Cocina::Models::FolioCatalogLink, catalog: "folio", catalogRecordId: "in12345", refresh: true)
```
justinlittman, line 14: This can probably be a plain-old Cocina::Models::FolioCatalogLink instead of an instance_double.

justinlittman (changes requested):

spec/jobs/register_druids_job_spec.rb
peetucket: Switched this and the identification as well to regular classes (and changed barcode value to satisfy constraints).

PR #4137, opened 2023-08-30
Export full Cocina json

### LINKED ISSUE #3847: Add bulk action to get full object cocina
Sometimes you need the full cocina for a set of objects in the original JSON because the flattening effect of CSV transformations makes a CSV version too hard to parse.

lwrubel:
# Why was this change made? 🤔
Resolves #3847. Download Cocina JSON

Unit, tested on QA. Checking with @andrewjbtw about whether he would like to test that. 

No

### DIFF
--- a/app/jobs/export_cocina_json_job.rb
+++ b/app/jobs/export_cocina_json_job.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+##
+# Job to export Cocina JSON
+class ExportCocinaJsonJob < GenericJob
+  ##
+  # A job that exports gzipped line-oriented JSON for one or more Cocina objects
+  # @param [Integer] bulk_action_id GlobalID for a BulkAction object
+  # @param [Hash] params additional parameters that an Argo job may need
+  def perform(bulk_action_id, params)
+    super
+
+    File.open(json_download_path, "w") do |jsonl_file|
+      with_items(params[:druids], name: "Export Cocina JSON") do |cocina_object, success, failure|
+        jsonl_file << "#{cocina_object.to_json}\n"
+        success.call("Exported full Cocina JSON")
+      end
+    end
+    gzip_file
+  end
+
+  private
+
+  def json_download_path
+    FileUtils.mkdir_p(bulk_action.output_directory)
+    File.join(bulk_action.output_directory, Settings.export_cocina_json_job.jsonl_filename)
+  end
+
+  def gzip_download_path
+    File.join(bulk_action.output_directory, Settings.export_cocina_json_job.gzip_filename)
+  end
+
+  def gzip_file
+    gzip = ActiveSupport::Gzip.compress(File.read(json_download_path))
+    File.write(gzip_download_path, gzip, mode: "wb")
+    FileUtils.rm_f(json_download_path)
+  end
+end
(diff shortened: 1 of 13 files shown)


app/jobs/export_cocina_json_job.rb
```diff
@@ -0,0 +1,42 @@
...
+# @param [Hash] params additional parameters that an Argo job may need
+class ExportCocinaJsonJob < GenericJob
+  def perform(bulk_action_id, params)
+    super
+
+    with_bulk_action_log do |log|
+      File.open(json_download_path, "w") do |jsonl_file|
+        with_items(params[:druids], name: "Export cocina JSON") do |cocina_object, success, failure|
+          if cocina_object.is_a?(NilModel)
+            failure.call("No JSON exported since not object not found (bulk_action.id=#{bulk_action_id})")
```
justinlittman, line 15: "No JSON exported since not object not found" <-- Wording off.

justinlittman (approved):
Approve pending fix of typo.

lwrubel:
Fixed the linting issue thanks to a pointer from @justinlittman.

mjgiarlo:
@lwrubel 💬 
> ...
> I am also stumped how to deal with this linting error:
> 
> ```
> spec/jobs/export_cocina_json_job_spec.rb:27:5: C: RSpec/SubjectStub: Do not stub methods of the object under test.
>     allow(job).to receive(:bulk_action).and_return(bulk_action)
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

Maybe I'm alone in this, but I tend not to care too much about this rule when the test is clear and useful, and I'd say that's true in this case. Maybe a good candidate for `# rubocop:disable RSpec/SubjectStub`

app/jobs/export_cocina_json_job.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+##
+# A job that exports line-oriented JSON for one or more objects
+# @param [Integer] bulk_action_id GlobalID for a BulkAction object
+# @param [Hash] params additional parameters that an Argo job may need
```
mjgiarlo, line 6: We might want these two lines immediately above `def perform` rather than on the class def

app/jobs/export_cocina_json_job.rb
```diff
@@ -0,0 +1,36 @@
...
+class ExportCocinaJsonJob < GenericJob
+  def perform(bulk_action_id, params)
+    super
+
+    File.open(json_download_path, "w") do |jsonl_file|
+      with_items(params[:druids], name: "Export cocina JSON") do |cocina_object, success, failure|
```
mjgiarlo, line 12: I can't remember how the `name` param is used, so maybe it's not worth doing here, but `cocina` is capitalized most other strings.

lwrubel: It shows up in the bulk action's log, mainly for failures, e.g. `2023-08-30 17:33:17 UTC Export cocina JSON failed Dor::Services::Client::NotFoundResponse ...`

Agree it's more consistent to capitalize, so doing that, thanks!

PR #4140, opened 2023-08-31
Change checksum report job to make multiple calls to pres cat.

### LINKED ISSUE #1997: Support longer lists of druids in checksum report bulk action
From @jcoyne in Slack:

> do you want to make a ticket so that the ChecksumReportJob breaks that into reasonable sized chunks so that no one request will timeout? 

Currently, checksum reports timeout if a batch is large enough. It can be hard to tell just how large a batch is too large, which is frustrating to users. There are use cases for generating large checksum reports, so the bulk action should support them.

justinlittman:
closes #1997

# Why was this change made? 🤔
To allow checksum report job to handle large numbers of druids.

Unit, QA, Andrew

⚡ ⚠ Please ensure this change does not introduce accessibility violations (at the WCAG A or AA conformance levels); if it does, include a rationale. See the [Infrastructure accessibility guide](https://github.com/sul-dlss/DeveloperPlaybook/blob/main/best-practices/infra-accessibility.md) for more detail. ⚡

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,7 +52,7 @@ gem "dor-services-client", "~> 12.5"
 gem "dor-workflow-client", "~> 6.0"
 gem "druid-tools"
 gem "mods_display", "~> 1.0"
-gem "preservation-client", "~> 6.0"
+gem "preservation-client", "~> 6.2"
 gem "rsolr"
 gem "sdr-client", "~> 2.0"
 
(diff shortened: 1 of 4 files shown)


peetucket (approved):
So this needs to go first presumably to avoid things breaking with the other pres cat changes?

justinlittman:
Yes, but I don't see any reason to not let this go out with dep updates next week.

peetucket:
Yeah, just thinking we wouldn't want to remove the endpoint from preservation catalog by deploying it before we deploy this version of argo.  As long as we don't deploy preservation catalog before dependency updates (very low risk), should be fine.

Issue #4133, opened 2023-08-29
"Download all files" option can crash Argo when invoked on large objects


andrewjbtw:
**Describe the bug**
It looks like downloads of large objects from Argo using the "download all files" option to create a zip can crash Argo.

The associated HB error is: https://app.honeybadger.io/projects/49894/faults/65734574

Multiple attempts to download [this object](https://argo.stanford.edu/view/sj917wc8897) (the one in the HB error) have resulted in one or both Argo nodes locking up, with Ops needing to reboot.

I replicated that object in stage and tried to download it and that crashed argo-stage: https://argo-stage.stanford.edu/view/druid:sc263ks6367

Total preserved size for that object is 20 GB.

**User Impact**
Argo becoming unstable is a problem for anyone. For the user who tries to download the files, they don't see that something has gone wrong. The browser just spins.

**To Reproduce**
1. Go to my test object: https://argo-stage.stanford.edu/view/druid:sc263ks6367
2. Try to download all files
3. Leave the browser tab open and attempting to download
4. Wait for Argo to crash

It may take more than one download attempt in close succession to cause a crash.

**Expected behavior**
We've always had problems downloading large zips but I don't remember Argo crashing if it timed out. In testing with hierarchical files back in January, I attempted to download large zips and don't remember Argo crashing. But I'm not sure I downloaded up to 20 GB.

**Additional context**
H2 had a similar problem with single-file downloads that we fixed recently. Argo single-file downloads don't cause crashes.

ndushay:
JLitt thinks there are more and less memory efficient ways to do this;   "zipline" and "zipxxxx" ... see what is used elsewhere.

For testing, Andrew has an object in stage that will "take down stage"

peetucket:
Possible gems referenced in discussion:

https://github.com/WeTransfer/zip_tricks  
https://github.com/fringd/zipline

PR #4153, opened 2023-09-07
Fixes download all.

justinlittman:
refs #4133

# Why was this change made? 🤔

Unit, @andrewjbtw test object in stage

⚡ ⚠ Please ensure this change does not introduce accessibility violations (at the WCAG A or AA conformance levels); if it does, include a rationale. See the [Infrastructure accessibility guide](https://github.com/sul-dlss/DeveloperPlaybook/blob/main/best-practices/infra-accessibility.md) for more detail. ⚡

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -2,7 +2,6 @@
 
 class FilesController < ApplicationController
   include ActionController::Live # required for streaming
-  include ZipTricks::RailsStreaming
 
   before_action :load_resource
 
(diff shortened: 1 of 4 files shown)


app/controllers/files_controller.rb
```diff
@@ -109,4 +104,45 @@ def filename
   def load_resource
     @cocina_model = Repository.find(params[:item_id])
   end
+
+  # Zip-tricks based streaming for files from preservation.
+  # Based on https://piotrmurach.com/articles/streaming-large-zip-files-in-rails/
+  class PresStreamer
```
edsu, line 110: It seems like one of the advantages of extracting this functionality as a class is that it can be tested individually? Maybe not though...

edsu (approved):

app/controllers/files_controller.rb
mjgiarlo: I was thinking the same, and wondering about giving this class its own file (in `app/services/` or something, but I'm also pondering whether testing this class in isolation might be more trouble than it's worth because:

1. Most of what's in the class is ZipTricks implementation details which---one hopes, lol---is already being tested by the ZipTracks maintainers.
2. What's not ZipTricks implementation details is in a fairly nested block structure which may make for difficult testing.

I'm on the fence but can be convinced either way.

justinlittman: I opted not to test in more detail for the reasons that @mjgiarlo gave. It didn't seem that additional testing would give greater assurance that "downloading all" actually worked.

PR #4154, opened 2023-09-07
[HOLD] use the previously accessioned version when downloading all files

### LINKED ISSUE #4141: "Download all files" hits 404 errors when downloading from "Opened" items
**Describe the bug**
The "Download all files" function in Argo gives users a zip containing all the files in preservation for an object. This function appears to always fail when an object is in an "Opened" state. There's an HB error that shows a 404 for the download:
https://app.honeybadger.io/projects/49894/faults/68070005

It looks like the problem has to do with versioning and version numbering. When an object is opened, it's version number has been incremented by one from the last accessioned version. This means that:

- Argo says the object is at version n+1
- Preservation only has versions up to n

I think what's happening is that the download all files function is requesting the version number that Argo has, so:

1. Argo says: send all files from version n+1
2. Preservation says: I only have files from version n
3. 404 error

Once the object is closed, the download will work without any changes to the files in the Moab.

**User Impact**
Users can't download preserved files from objects that are opened and it's not clear why it's not working.

**To Reproduce**
1. Go to this test object https://argo-stage.stanford.edu/view/druid:bb020zq0872
5. Try to download all files
6. See that it doesn't work. The browser will look like the download hangs and HB will show a 404 for an Argo action

**Expected behavior**
When an object is opened, download all files for the most recent accessioned version, which must exist because the "Opened" status is only possible if an object has at least one accessioned version.

**Additional context**
A "draft" state that does not immediately increment the version number in Cocina instead of "opened" state might help here but that's some ways off.

peetucket:
# Why was this change made?

Fixes #4141 - download all files fails when an object is opened for versioning (it looks in preservation for the newly bumped version which doesn't exist yet, instead of the last version in preservation).  This fixes it by using the last preserved version.  

This also removes the "Download all files" link unless there is an accessioned version.  This is because you can only download files once they are in preservation (this is current behavior too).  Currently, the link shows for an v1 object that has never been accessioned, but gives you a blank ZIP file.  This change removes the link (because we now have to ask for the previous version and we may as well just not show the link in this case).

HOLD to verify with Andrew

# How was this change tested?

- [x] Updated specs and localhost.
- [x] Test on stage
- [ ] Validate with Andrew

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -11,13 +11,13 @@ class FilesController < ApplicationController
   def index
     raise ArgumentError, "Missing file parameter" if filename.blank?
 
-    @has_been_accessioned = WorkflowClientFactory.build.lifecycle(druid: params[:item_id], milestone_name: "accessioned")
+    @has_been_accessioned = StateService.new(@cocina_model).accessioned?
     files = Array(@cocina_model.structural&.contains).map { |fs| fs.structural.contains }.flatten
     @file = files.find { |file| file.filename == params[:id] }
 
     if @has_been_accessioned
       begin
-        @last_accessioned_version = Preservation::Client.objects.current_version(params[:item_id])
+        @last_accessioned_version = last_accessioned_version(params[:item_id])
       rescue Preservation::Client::NotFoundError
         return render status: :unprocessable_entity, plain: "Preservation has not yet received #{params[:item_id]}"
       rescue Preservation::Client::Error => e
@@ -74,7 +74,7 @@ class FilesController < ApplicationController
         zip.write_deflated_file(filename) do |sink|
           Preservation::Client.objects.content(druid: @cocina_model.externalIdentifier,
             filepath: filename,
-            version: @cocina_model.version,
+            version: last_accessioned_version(@cocina_model.externalIdentifier),
             on_data: proc { |data, _count| sink.write data })
         rescue => e
           sink.close
@@ -88,6 +88,10 @@ class FilesController < ApplicationController
   end
 
   private
+
+  def last_accessioned_version(druid)
+    Preservation::Client.objects.current_version(druid)
+  end
 
   def preserved_files(cocina_model)
     resources = Array(cocina_model.structural&.contains)
--- a/app/services/state_service.rb
+++ b/app/services/state_service.rb
@@ -39,6 +39,13 @@ class StateService
     STATES[:unlock_inactive]
   end
 
+  ##
+  # Ported over logic from app/helpers/dor_object_helper.rb#LN133
+  # @return [Boolean]
+  def accessioned?
+    @accessioned ||= lifecycle("accessioned") ? true : false
+  end
+
   private
 
   attr_reader :druid, :version
@@ -57,13 +64,6 @@ class StateService
 
   def submitted?
     @submitted ||= active_lifecycle("submitted")
-  end
-
-  ##
-  # Ported over logic from app/helpers/dor_object_helper.rb#LN133
-  # @return [Boolean]
-  def accessioned?
-    @accessioned ||= lifecycle("accessioned") ? true : false
   end
 
   def active_assembly_wf?
--- a/spec/services/state_service_spec.rb
+++ b/spec/services/state_service_spec.rb
@@ -15,8 +15,6 @@ RSpec.describe StateService do
   let(:service) { described_class.new(cocina) }
 
   describe "#allows_modification?" do
-    subject(:allows_modification?) { service.allows_modification? }
-
     before do
       allow(service).to receive(:object_state).and_return(:unlock)
     end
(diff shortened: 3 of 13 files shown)


app/controllers/files_controller.rb
```diff
@@ -11,13 +11,13 @@ class FilesController < ApplicationController
   def index
     raise ArgumentError, "Missing file parameter" if filename.blank?
 
-    @has_been_accessioned = WorkflowClientFactory.build.lifecycle(druid: params[:item_id], milestone_name: "accessioned")
+    @has_been_accessioned = StateService.new(@cocina_model).accessioned?
     files = Array(@cocina_model.structural&.contains).map { |fs| fs.structural.contains }.flatten
```
peetucket, line 15: we already have a class that does this, let's use it

app/services/state_service.rb
```diff
@@ -59,13 +66,6 @@ def submitted?
...
   end
 
-  ##
-  # Ported over logic from app/helpers/dor_object_helper.rb#LN133
-  # @return [Boolean]
-  def accessioned?
-    @accessioned ||= lifecycle("accessioned") ? true : false
-  end
-
   def active_assembly_wf?
```
peetucket, line 69: this method is now made public so we can use it in our code elsewhere

app/controllers/files_controller.rb
```diff
@@ -74,7 +74,7 @@ def download
         zip.write_deflated_file(filename) do |sink|
           Preservation::Client.objects.content(druid: @cocina_model.externalIdentifier,
             filepath: filename,
-            version: @cocina_model.version,
+            version: last_accessioned_version(@cocina_model.externalIdentifier),
             on_data: proc { |data, _count| sink.write data })
```
peetucket, line 78: this is the key fix for the actual bug we are fixing

spec/services/state_service_spec.rb
```diff
@@ -139,4 +139,28 @@
...
+      before do
+        allow(workflow_client).to receive(:lifecycle).with(druid:, milestone_name: "accessioned").and_return(nil)
+      end
+
+      it "returns false" do
+        expect(subject).to be false
+      end
+    end
+  end
 end
```
peetucket, line 166: new tests for our now public method

spec/services/state_service_spec.rb
```diff
@@ -139,4 +139,28 @@
...
+  describe "#accessioned?" do
+    subject(:accessioned?) { service.accessioned? }
+
+    context "if the accessioned lifecycle exists" do
+      before do
+        allow(workflow_client).to receive(:lifecycle).with(druid:, milestone_name: "accessioned").and_return("2022-04-20 21:55:25 +0000")
+      end
+
+      it "returns true" do
+        expect(subject).to be true
```
mjgiarlo, line 152: We could also consider ditching line 144, since we don't appear to be referring to the `accessioned?` variable, and changing this line to use a predicate matcher:


```suggestion
        expect(service).to be_accessioned
```

spec/services/state_service_spec.rb
```diff
@@ -139,4 +139,28 @@
...
+      end
+    end
+
+    context "if the accessioned lifecycle does not exist" do
+      before do
+        allow(workflow_client).to receive(:lifecycle).with(druid:, milestone_name: "accessioned").and_return(nil)
+      end
+
+      it "returns false" do
+        expect(subject).to be false
```
mjgiarlo, line 162: Likewise above:

```suggestion
        expect(service).not_to be_accessioned
```

mjgiarlo (approved):
Approve pending consideration of admittedly not at all crucial test changes

spec/services/state_service_spec.rb
peetucket: fine with me either way, but if you look up at the other tests in this file, they all uses this `subject` type syntax

should I change them all to match?

mjgiarlo: If you're inclined, sure, that'd be swell 🙂

peetucket: Service with a smile complete

mjgiarlo (approved):

peetucket:
Sorry, had intended to HOLD this until checking in with Andrew about the change in behavior (hiding download link when object not accessioned).  But I also think this is a pretty minor concern, and if there is an issue with this, we can deal with a follow up.

mjgiarlo:
@peetucket Oh, crap, I totally missed that on the latest review (after noticing it earlier). Shall I revert?

peetucket:
Since the previous behavior of that download link with a non-accessioned item was weird (it existed but gave you an empty ZIP), I'd say we can keep it as is.  If Andrew objects on his return, we'd can always return to showing the link even in that situation, but we'd need to add some extra error handling since the logic to download all files now makes a call to Preservation to find the latest version (which will fail if there is nothing there yet).

PR #4178, opened 2023-09-21
Link to druids and purls in reports view

### LINKED ISSUE #923: Make the druids in the "Report View" active links to the objects
A link to the detail view of the object, like selecting an individual item from the Results View. From original JIRA ticket https://jirasul.stanford.edu/jira/browse/ARGO-262

"It would be helpful to be able to click on the druids in the report view and discovery report and be brought directly to that object (for example, so that edits can be made to that object). Currently, the user must copy and paste the druid in the search box in order to go to the object."

mjgiarlo:
# Why was this change made?

Fixes #923

# How was this change tested?

CI


ndushay (approved):
approving, but it would be great in future if you showed a before and after picture so I was looking at the UI as well as a the code.

PR #4203, opened 2023-10-02
Change configuration of date facets to add all query.

### LINKED ISSUE #4198: Show the "Last accessioned", "Earliest accessioned" and "Registered" facets regardless of date
Thanks to the analysis in https://github.com/sul-dlss/argo/issues/4165, I've learned that the reason these facets

- Last accessioned
- Earliest accessioned
- Registered

don't appear on many search results is that they are filtered to appear only if data matches certain criteria, like "at least one item has been accessioned within the last 365 days". 

I would like these facets to always be available so that they can be used to filter search results. My primary interest is actually having access to the date range search boxes, not the pre-generated values. I use these when analyzing data or identifying objects that might need remediation.

Screenshot 2023-09-29 at 4 16 23 PM

If we need to include the value "All" to make the facet always appear (see the "Last Opened" facet for an example), that would be ok, though it doesn't seem helpful to be able to select that value since it will just be the same set of search results. I don't see a clear need to change the lists of precomputed date range values otherwise.

justinlittman:
closes #4198

# Why was this change made?

So that date facets are always displayed.

# How was this change tested?

Prod

### DIFF
--- a/app/controllers/concerns/date_facet_configurations.rb
+++ b/app/controllers/concerns/date_facet_configurations.rb
@@ -14,6 +14,7 @@ module DateFacetConfigurations
       config.add_facet_field "registered_date", home: false, label: "Registered",
         component: DateChoiceFacetComponent,
         query: {
+          all: {label: "All", fq: "#{SolrDocument::FIELD_REGISTERED_DATE}:*"},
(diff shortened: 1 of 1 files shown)


justinlittman:
image

jmartin-sul (approved):

PR #4209, opened 2023-10-03
Use a date input for date facet.

### LINKED ISSUE #4199: In facets that use a date range search entry box, show the user the accepted formats
A number of facets have a pair of entry boxes where users can enter a date range. A number of people have asked what format is required in these boxes. If possible, we should give some date format guidance in the UI itself.

Screenshot 2023-09-29 at 4 28 01 PM

From what I've learned, it is acceptable to use ISO 8601 for YYYY-MM-DD. Less precise dates are also allowed:

- YYYY-MM (gets interpreted with day as 01)
- YYYY (gets interpreted with month-day as 01-01)

It is also permissible to enter data into just one box. So the range

- From: 2023
- Through: [left blank]

will be interpreted as "everything from 2023-01-01 until the present day".

To avoid overcomplicating the display, we shouldn't show the format down to hour, minute, or second as it's less likely people will be that precise.

justinlittman:
closes #4199

# Why was this change made?

# How was this change tested?

Andrew

image


justinlittman:
Screenshot 2023-10-03 at 2 53 48 PM

jmartin-sul (approved):

PR #4210, opened 2023-10-03
Adds human preserved size field.

### LINKED ISSUE #4205: Add new "Preservation size" column that shows human-readable sizes
After completing the rename of the existing "preservation size" column in https://github.com/sul-dlss/argo/issues/4204 create a new column called "Preservation size" that shows disk usage in human-readable units - KB, MB, GB, TB.

It's useful to have the raw bytes (in the existing column) for certain types of analysis, but for other types of reports or quick evaluations of the data in the browser, the human-readable units are more approachable.

justinlittman:
closes #4205

# Why was this change made?

# How was this change tested?

Stage

image

### DIFF
--- a/spec/requests/report_spec.rb
+++ b/spec/requests/report_spec.rb
@@ -4,11 +4,6 @@ require "rails_helper"
 
 RSpec.describe "Reports from a search" do
   before do
-    ResetSolr.reset_solr
-    solr_conn.add(id: "druid:xb482ww9999",
-      objectType_ssim: "item",
-      obj_label_tesim: "Report about stuff")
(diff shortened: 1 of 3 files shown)


spec/requests/report_spec.rb
```diff
@@ -4,11 +4,6 @@
 
 RSpec.describe "Reports from a search" do
   before do
-    ResetSolr.reset_solr
```
justinlittman, line 7: In looking for some way to test this, I realized this wasn't being used.

jmartin-sul (approved):

PR #4213, opened 2023-10-04
Move the sidebar above the content in the dom

### LINKED ISSUE #4161: Accessibility (major): Landing page - ensure that the reading order of content is logical [argo-003]
## Description

The programmatic (DOM) order does not match the visual reading order.

In DOM-order, the "Limit your search" facet region comes after "Search results" region, but appears before the "Search results" in visual L-R reading order.

Moreover, the "Search results" instructions direct the user to "select a facet on the left to begin" but this facet region comes *after* the "Search results" in DOM-order (i.e., the screen reader user must navigate to the *right* to find the "Limit your search" region).

## Suggestion

- [x] Put the `<section id=""sidebar">` before the `<section id=""content">` in DOM-order.
- [x] Rewrite the "Welcome to Argo!" instructions to avoid using visual/spatial references. For example, instead of: "Enter one or more search terms or select a facet on the left to begin", consider something like "To begin, enter one or more search terms, or select a facet under the 'Limit your search' panel."
- [ ] Consider moving the "Welcome to Argo!" message outside of the Search results content container (see issue ID argo-001, Level 1 Hierarchical Order).

https://docs.google.com/spreadsheets/d/1m3obE89_mBCZpRkWAjJAXEr2ZfrUirzjvsJWqYH-G6Y/edit#gid=0

aaron-collier:
# Why was this change made?

Fixes #4161 

This overrides the default blacklight layout to move the sidebar higher in the dom than content to improve the screenreader experience.

Note: I will put up a separate blacklight PR to address this upstream, at which time we can remove this override.

# How was this change tested?

### DIFF
--- a/app/views/layouts/blacklight.html.erb
+++ b/app/views/layouts/blacklight.html.erb
@@ -1,0 +1,17 @@
+<% content_for(:content) do %>
+  <% if content_for? :sidebar %>
+    <section id="sidebar" class="<%= sidebar_classes %> order-first" aria-label="<%= t("blacklight.search.documents.aria.limit_search") %>">
+      <%= content_for(:sidebar) %>
+    </section>
+
+    <section id="content" class="<%= main_content_classes %> order-last" aria-label="<%= t("blacklight.search.documents.aria.search_results") %>">
+      <%= yield %>
+    </section>
+  <% else %>
+    <section class="col-md-12">
+      <%= yield %>
+    </section>
+  <% end %>
+<% end %>
+
+<%= render template: "layouts/blacklight/base" %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/blacklight.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for(:content) do %>
```
mjgiarlo, line 1: Can you drop an ERB comment here saying why we decided to override this layout?

```
<%# We are overriding this Blacklight layout because blah blah %>
```

mjgiarlo (approved):

PR #4216, opened 2023-10-05
Tweak Rubocop config so it is closer to pre-standardrb state

### LINKED ISSUE #4214: .rubocop.yml needs appropriate defaults now that we're off standard.
With PR #4200, we ended up with a `.rubocop.yml` that does not exclude enough directories (it takes many minutes to run rubocop when you have no local rubocop cache for Argo (e.g. after updating the rubocop gem)).   It also put EVERYTHING into .rubocop_todo.yml, which may be less troubling:

```
require:
  - rubocop-performance
  - rubocop-rspec
  - rubocop-rails

AllCops:
  TargetRubyVersion: 3.2
  DisplayCopNames: true
  SuggestExtensions: false
  NewCops: disable
  Exclude:
    - bin/**
    - vendor/bundle/**/*

inherit_from:
  - .rubocop_todo.yml
```

That's it.  Any of our local changes to default settings don't show.

The work for this PR:
- [x] expand what files are excluded to be more reasonable.  tmp, logs, lock files, build directories.  
     ```
  Exclude:
    - bin/**
    - db/migrate/**
    - db/schema.rb
    - vendor/bundle/**/*
    - node_modules/**/*
    - tmp/**/*
    - log/**/*
    - fedora_conf/**/*
    - public/**/*
    - coverage/**/*
    - postgres-data/**/*
    - spec/fixtures/**/*
   ``` from https://github.com/sul-dlss/argo/pull/3877/files
- [ ] add back rules we know we don't like, e.g. LineLength.   

You might 
a) look at rubocop.yml from before we went to standard:  https://github.com/sul-dlss/argo/pull/3877/files
b) see https://github.com/sul-dlss/DeveloperPlaybook/blob/main/style/infra_template.rubocop.yml
c) look at another app with a UI, such as ETD or PreAssembly and take appropriate settings from that. (Note that h2 uses standard.)

mjgiarlo:
# Why was this change made?

Fixes #4214

- Remove vestigial rubocop custom config from the standardrb days
- Tweak Rubocop configs so they are closer to the pre-standardrb state

# How was this change tested?

CI


.rubocop.yml
```diff
@@ -1,16 +1,503 @@
...
   - rubocop-rails
+  - rubocop-rspec
+
+inherit_from: .rubocop_todo.yml
 
 AllCops:
   TargetRubyVersion: 3.2
   DisplayCopNames: true
   SuggestExtensions: false
-  NewCops: disable
```
ndushay, line 10: if we don't disable these, then rubocop errors can overwhelm the FR.

.rubocop.yml
```diff
@@ -1,16 +1,503 @@
...
+    - coverage/**/*
+    - db/**/*
+    - log/**/*
+    - node_modules/**/*
+    - postgres-data/**/*
+    - public/**/*
+    - solr_conf/**/*
+    - spec/fixtures/**/*
+    - tmp/**/*
+    - vendor/**/*
```
ndushay, line 23: this alone will greatly speed up running rubocop.   Thank you!

ndushay (commented):
I mentioned one concern.   Is there anything you can put in the other 4 tickets for removing standard from our repos?   See https://docs.google.com/spreadsheets/d/1tAmXrcRmyoImqGSVpkMW-QzVgi-AoLIUl8DA5VbuWOQ/edit#gid=0.  Also note one of the gems was de-standard-ized also already, and may benefit from your expertise.

.rubocop.yml
mjgiarlo: @ndushay They're disabled by default.

justinlittman (approved):

.rubocop.yml
mjgiarlo: Yep. Took the runtime from 20s down to 10s.

PR #4229, opened 2023-10-11
Switch to direct report download.

### LINKED ISSUE #4226: Support reports of enormous size (500,000 druids or more)
Now that Argo has much more capable reporting than before, we've gotten requests to push the capability even further: to the level of downloading reports that could have more than 500,000 druids.

Example query: [a report on all items that contain a PDF](https://argo.stanford.edu/report?f%5Bcontent_file_mimetypes_ssim%5D%5B%5D=application%2Fpdf&f%5BobjectType_ssim%5D%5B%5D=item)

I've been able to download reports of around 100,000 items. It appears that there is a limit above that where the CSV download is never generated but I haven't seen any HB errors or error messages in the UI to indicate a report failure.

**Uses of large reports**

Certain activities, including planning for accessibility work and metadata remediation, involve repository-wide analyses where some queries will return large numbers of druids.

justinlittman:
closes #4226

# Why was this change made?
Per ticket.

# How was this change tested?

Stage, prod

### DIFF
--- a/app/javascript/controllers/object_reporter.js
+++ b/app/javascript/controllers/object_reporter.js
@@ -43,48 +43,7 @@ export default class extends Controller {
   }
 
   downloadCSV (event) {
-    event.target.disabled = true
-    this.table.alert('Your CSV report is being generated. Please hold...')
-
-    fetch(`${this.downloadUrlValue}?${this.downloadQueryParams()}`)
-      .then((response) => {
-        const reader = response.body.getReader()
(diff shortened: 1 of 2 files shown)


justinlittman:
image

ndushay:
does this need PO review, or is it ready to go?

ndushay (approved):

justinlittman:
PO review wasn't requested on original ticket so merging.

Issue #4225, opened 2023-10-10
Bulk action registration from CSV is not handling rights correctly


andrewjbtw:
**Describe the bug**
Argo has a bulk action that handles registration from a CSV. This is distinct from the registration form and uses its own CSV. I noticed while adding stage sample data that the bulk action is not handling rights correctly.

There are two issues here:

- some druids are given the rights of dark when that's not what's specified on the spreadsheet
- some druids are failing registration with an error message saying that the rights are not valid

I'm not sure if the problem is in my data entry or if the spreadsheet isn't parsing the rights model accurately.

**User Impact**
This spreadsheet is not as widely used as the registration form, but does get used in some projects. It's possible to work around the issue by changing writes after registration but one of the benefits of this bulk action over the registration form is that it allows you to choose different rights settings for each druid being registered. There's no other way to do that.

Errors in the log look like this:
```
2023-10-10 20:43:40 UTC RegisterDruidsJob: {"download"=>"stanford"} isn't one of in #/components/schemas/Access
2023-10-10 20:43:40 UTC RegisterDruidsJob: {"download"=>"world"} isn't one of in #/components/schemas/Access
```

**To Reproduce**

In stage, use this CSV (but change the source IDs to be unique) [test-bulk-registration.csv](https://github.com/sul-dlss/argo/files/12861747/test-bulk-registration.csv)

1. Go to New Bulk actions
2. Choose Register Druids
3. Upload the CSV
4. See that some druids get registered and some don't, and the druids that get registered may not have the rights specified on the spreadsheet.

**Expected behavior**
All the druids should be registered with the rights specified on the spreadsheet.

**Additional context**
I just noticed this yesterday when starting to populate stage with new test data after the reset.

justinlittman:
@andrewjbtw You have a column wrong.

 [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3d0QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--aeff9ed792e1cb31cb07bfe5261dc35164d0013d/image.png)

It looks like we're missing validation on some required CSV fields, so I will add.

andrewjbtw:
Thanks for catching the mistake. Closing this since the extra validation that would have caught the header mistake is also merged.

PR #4232, opened 2023-10-12
Adds additional required fields to registration bulk action CSV valid…

justinlittman:
…ation.

refs #4225

# Why was this change made?

To match:
image

# How was this change tested?


ndushay (approved):
i'm not quite with the context for this, so I'll let you merge it, assuming it's as simple as it looks?

PR #4238, opened 2023-10-16
removing vestigial references to Fedora

ndushay:
# Why was this change made?

die fedora die

# How was this change tested?

specs.


config/initializers/activefedora_solr_updates.rb
```diff
@@ -1,9 +0,0 @@
-# frozen_string_literal: true
-
-# Cannot just do:
-#   ::ENABLE_SOLR_UPDATES = false
-# Because "warning: already initialized constant ENABLE_SOLR_UPDATES".
-# ActiveFedora assigns a default (if unassigned) and we cannot pre-empt it from here.
-
-self.class.send(:remove_const, 'ENABLE_SOLR_UPDATES') if self.class.const_defined?(:ENABLE_SOLR_UPDATES)
-self.class.const_set(:ENABLE_SOLR_UPDATES, false)
```
ndushay, line 9: I searched sul-dlss for ENABLE_SOLR_UPDATES and came up with nothing, so this can go.  It was also last meaningfully changed in about 2016.

justinlittman (approved):

PR #4240, opened 2023-10-16
add copy/paste icon to facet modal

### LINKED ISSUE #4234: Enable users to copy the facet data that shows values and counts
We've received user feedback on Argo discovery and reporting that

- some users would like to be able to copy the value and count data that appears for each facet
- copying and pasting that data doesn't work well because it doesn't include a delimiter between the value and the count

For example, copying value and count data here:
> Screenshot 2023-10-12 at 11 01 51 AM

Comes out like this in a text file:

```
A Test Collection for Honors Theses Revised2Africana Librarians Council (ALC) Information Resources1Archive of Recorded Sound/Music Library - Supplemental Materials4Ashizawa family collection1BIOHOPK 47: Introduction to Research in Ecology and Ecological Physiology8BIOHOPK185: student videos12BIOHOPK185: time-series data4Berman Archive - test collection

```

And like this in a spreadsheet program

> 
Screenshot 2023-10-12 at 11 07 28 AM

Would it be possible to insert a tab before the count numbers? That would allow people to paste into spreadsheets and separate the values and counts into different columns.

peetucket:
# Why was this change made?

Fixes #4234 - add a copy/paste icon to the facet model to allow users to grab the facet list values onto their clipboard.  Note: copy/paste icon only shows on the modal window (for when you click "more >>").  We can adapt later if we need to shows this for all facets on the main argo page.

~~HOLD for testing on stage~~

Screen Shot 2023-10-17 at 9 26 18 AM

# How was this change tested?

Localhost and stage

### DIFF
--- a/app/views/catalog/facet.html.erb
+++ b/app/views/catalog/facet.html.erb
@@ -6,7 +6,7 @@
   <%= render BlacklightModalComponent.new do |component| %>
     <% component.header do %>
       <%= facet_field_label(@facet.key) %>
-      <a class="copy-button" data-controller="copy" data-action="copy#copy" aria-label="Copy facet values" href=""><i class="bi bi-clipboard"></i></a>
+      <a class="copy-button" data-controller="copy" data-action="copy#copy" aria-label="Copy this list" href="#"><i class="bi bi-clipboard" title="Copy this list"></i></a>
      <% end %>
     <% component.body do %>
 
--- a/app/views/catalog/facet.html.erb
+++ b/app/views/catalog/facet.html.erb
@@ -6,7 +6,7 @@
   <%= render BlacklightModalComponent.new do |component| %>
     <% component.header do %>
       <%= facet_field_label(@facet.key) %>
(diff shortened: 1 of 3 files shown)


app/views/catalog/facet.html.erb
```diff
@@ -4,8 +4,12 @@
 
 <div data-controller="facet-filter">
   <%= render BlacklightModalComponent.new do |component| %>
-    <% component.header { facet_field_label(@facet.key) } %>
+    <% component.header do %>
+      <%= facet_field_label(@facet.key) %>
+      <a class="copy-button" data-controller="copy" data-action="copy#copy" aria-label="Copy this list" href="#"><i class="bi bi-clipboard" title="Copy this list"></i></a>
```
mjgiarlo, line 9: It looks like other Bootstrap icons use `<span>` instead of `<i>` in Argo currently. Worth standardizing on that here or does using `<i>` have affordances here?

peetucket: Thanks for finding that.  I don't any benefit to `<i>` ... was just what I copied from website example.  Will change to `<span>`

mjgiarlo (approved):

Issue #4245, opened 2023-10-18
Speed up downloads of large reports


andrewjbtw:
When downloading reports of over about 75,000 druids, the download speed diminishes over time. This results in multi-hour download times for large reports.

Example: I started a report download for 500,000 druids this morning. It never finished because my wifi connection dropped two hours later.

I didn't screenshot the speed right away, but I think it started at about 60 KB/sec

After about 15 minutes this was the speed:

><img width="593" alt="Screenshot 2023-10-18 at 10 03 13 AM" src="https://github.com/sul-dlss/argo/assets/8975583/826cb453-727d-40bc-9f48-36b5990e0ca8">

And then it kept dropping:

> <img width="593" alt="Screenshot 2023-10-18 at 10 22 19 AM" src="https://github.com/sul-dlss/argo/assets/8975583/8b77461c-c440-41d8-b315-e386292cd0b6">

> <img width="593" alt="Screenshot 2023-10-18 at 10 41 57 AM" src="https://github.com/sul-dlss/argo/assets/8975583/9aaa2f68-41f8-4695-8524-d779c65d1c75">

> <img width="593" alt="Screenshot 2023-10-18 at 11 48 17 AM" src="https://github.com/sul-dlss/argo/assets/8975583/b2cee5e7-ddd1-42df-8a06-ce6a53059e2b">

At this point my wifi dropped, about 460,000 rows into the report.

If the report maintained the speed it started with, it still would have taken a while but I think the download would have worked. The drop in speed means the last rows take much longer than the first. I did the same download on Monday and it took over 3 hours.

justinlittman:
I'm fairly sure that deep pagination is problematic for Lucene (and hence Solr). See, for example, https://solr.apache.org/guide/8_11/pagination-of-results.html#performance-problems-with-deep-paging

andrewjbtw:
The report is being generated as if a browser is paging through results? Not through a query that asks for larger numbers of rows?

justinlittman:
If my understanding is correct, behind the scenes the reporter is iterating over pages of results from Solr. This is the standard way of dealing with a large result set (instead of just asking for all of the results in a single request).

There might be some performance gains by increasing the page size when iterating for a download.

justinlittman:
I set the page size to 1M and the download timed out after 5 minutes.

justinlittman:
@andrewjbtw The dynamic fields are:
* PURL
* Release To (returns "Not released" when not released)
* Citation
[...]

PR #4246, opened 2023-10-19
Increase solr page size for csv download.

justinlittman:
refs #4245

# Why was this change made?
Speedier reports

# How was this change tested?
stage, prod

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -20,6 +20,7 @@ class Report
 
   COLUMN_SELECTOR_COLUMN_SIZE = 3 # Helps display report columns in neatly lined up columns.
   ROWS_PER_PAGE = 100
+  ROWS_PER_PAGE_CSV = 10_000
 
   REPORT_FIELDS = [
     {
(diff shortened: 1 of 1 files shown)


app/models/report.rb
```diff
@@ -20,6 +20,7 @@ class << self
 
   COLUMN_SELECTOR_COLUMN_SIZE = 3 # Helps display report columns in neatly lined up columns.
   ROWS_PER_PAGE = 100
```
ndushay, line 22: Because we are only getting back from Solr the specific fields needed for the report, this could be bumped up to 200 or 250 for even better performance for large reports

ndushay (approved):

app/models/report.rb
justinlittman: If I understand correctly, this is the number of results when a user scrolls through the report in Argo. A larger number might make this laggier since a larger response is going to be slower (though they could scroll through more results before a new page is requested). Since there hasn't been a request to change the report in Argo and I don't know why this number was selected, I'm not inclined to make a change at this time.

justinlittman: @mjgiarlo looks like this is your code. Feel free to chime in if you feel this is worth further consideration.

mjgiarlo: You re correct, @justinlittman. Are your changes deployed anywhere? I'd be happy to do a quick test of the UI view to ensure the 10K change doesn't affect it, unless you've already done this.

justinlittman: I just checked (in prod). The Report UI only pulls back 100 records.

mjgiarlo: Super, thank you!

PR #4248, opened 2023-10-20
upgrade to solr 8

ndushay:
# Why was this change made?

Use same version of Solr for testing that we do in production

# How was this change tested?

tests, CI


ndushay:
closing this in favor of PR #4249

PR #4252, opened 2023-10-23
Build solr image in circleci

justinlittman:
# Why was this change made?
Less ci config.

# How was this change tested?


mjgiarlo (approved):
LGTM but deferring to @ndushay

ndushay (approved):
works for me (why didn't we think of this?)  Agree about Mike's cruft comment.

PR #4276, opened 2023-11-15
Reduce unnecessary facet querying to improve home page performance. C…

justinlittman:
…hange WPS workflow facet to lazy.

# Why was this change made?
Make Argo usable.

# How was this change tested?
Andrew

### DIFF
--- a/app/components/lazy_nonproject_tag_facet_component.html.erb
+++ b/app/components/lazy_nonproject_tag_facet_component.html.erb
@@ -2,7 +2,7 @@
   <div class="card facet-limit blacklight-exploded_nonproject_tag_ssim">
     <h3 class="card-header p-0 facet-field-heading" id="facet-exploded_nonproject_tag_ssim-header">
       <button type="button" disabled style="pointer-events: auto; cursor: wait" class="btn w-100 d-block btn-block p-2 text-start collapse-toggle collapsed" aria-expanded="false">
-        Tag
+        Tag  <div class="spinner-border spinner-border-sm" role="status"><span class="visually-hidden">Loading...</span></div>
       </button>
     </h3>
   </div>
--- a/app/components/lazy_project_tag_facet_component.html.erb
+++ b/app/components/lazy_project_tag_facet_component.html.erb
@@ -2,7 +2,7 @@
   <div class="card facet-limit blacklight-exploded_project_tag_ssim">
     <h3 class="card-header p-0 facet-field-heading" id="facet-exploded_project_tag_ssim-header">
       <button type="button" disabled style="pointer-events: auto; cursor: wait" class="btn w-100 d-block btn-block p-2 text-start collapse-toggle collapsed" aria-expanded="false">
-        Project Tag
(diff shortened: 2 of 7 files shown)


app/components/lazy_nonproject_tag_facet_component.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="card facet-limit blacklight-exploded_nonproject_tag_ssim">
     <h3 class="card-header p-0 facet-field-heading" id="facet-exploded_nonproject_tag_ssim-header">
       <button type="button" disabled style="pointer-events: auto; cursor: wait" class="btn w-100 d-block btn-block p-2 text-start collapse-toggle collapsed" aria-expanded="false">
-        Tag
```
justinlittman, line 5: Adding spinner because it makes the async state much clearer.

app/components/lazy_wps_workflow_facet_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<turbo-frame id="wps-workflow-facet" src="<%= lazy_wps_workflow_facet_catalog_path(helpers.search_state.params_for_search) %>" target="_top">
```
justinlittman, line 1: Making WPS workflow async because it is slow and hierarchical.

ndushay: but just this one for now as an experiment?   It's two brethren will be jealous

app/components/lazy_project_tag_facet_component.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="card facet-limit blacklight-exploded_project_tag_ssim">
```
ndushay, line 2: i'm wondering if we could have a generic lazy_facet_component and just feed in the field name or something similar.

app/controllers/catalog_controller.rb
```diff
@@ -6,6 +6,29 @@ class CatalogController < ApplicationController
   include DateFacetConfigurations
 
   before_action :limit_facets_on_home_page, only: [:index]
+  before_action :adjust_lazy_limits, only: [:index]
+
+  # The subset of facets that are displayed on the home page.
+  # (Before a user clicks "Show more facets")
+  HOME_FACETS = [
```
ndushay, line 13: if the only other slow ones are the wps and its brethren, maybe Andrew would want them to display on the home page?

app/controllers/catalog_controller.rb
```diff
@@ -277,6 +291,7 @@ def index
   end
 
   def lazy_nonproject_tag_facet
```
ndushay, line 293: similarly, wondering if these methods could be a single method taking a field argument.

app/components/lazy_wps_workflow_facet_component.html.erb
justinlittman: Only this one shows on the home page.

app/controllers/catalog_controller.rb
justinlittman: These are controller endpoints.

ndushay (approved):

PR #4278, opened 2023-11-15
Increase the limit for project and tag results.

### LINKED ISSUE #4273: New "Project" and "Tag" facets do not provide access to full facet hierarchy
**Describe the bug**
Changing the Project and Tag facets has changed the behavior of the facets in a subtle way that is preventing users from accessing all top-level Project and Tag values from the Argo home page.

The problem is that the facet only shows values up to a certain point and then shows a "more" link:

Project facet:
> Screenshot 2023-11-13 at 8 55 24 AM

Tag facet:
>Screenshot 2023-11-13 at 8 56 18 AM

When you click on the "more" link, you get the same facet list in a modal with the same "more" link at the same place:

Screenshot 2023-11-13 at 9 08 57 AM

Clicking on that sends you to the same modal with the same problem, forever. You can never get past the point in each list where the "more" link appears.

I'm not sure exactly what the previous cut-off point was for the "more" link but @oceanofsound reports that he used to be able to get at least as far as the SMPL tag, and that he never previously clicked on the "more" link. 

From what I remember, I also never clicked on the "more" link outside of looking at it when testing something. Unfortunately, stage does not have enough tags to trigger the "more" link so we didn't see the before and after there.

**User Impact**
Users have to find other strategies to get to the tag (Project or nonproject) they want. That might include searching for an item that has the appropriate tag, then clicking on the tag from there.

**To Reproduce**
1. See above: go to the Argo home page and expand the facets

**Expected behavior**
It should be possible to get to the full first level of the tag hierarchy. I know that in the previous implementation, you couldn't find all the tags from the Argo home page if you descended into a hierarchy but you could get to all (or almost?) all of the first-level tags. 

**Additional context**
Is there a depth first vs. breadth first setting? That is, is there a choice between loading the first N tags regardless of hierarchy or the first N tags measured by level of nesting. If there is, we should prefer loading all the first-level tags, then the next level, then the next, until the limit of tags to load is reached.

justinlittman:
closes #4273

# Why was this change made?

See ticket

# How was this change tested?

image

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -56,12 +56,12 @@ class CatalogController < ApplicationController
 
     # exploded_project_tag_ssim indexes all project tag prefixes for hierarchical facet display, whereas
     #   project tag_ssim only indexes whole tags
-    config.add_facet_field 'exploded_project_tag_ssim', label: 'Project', limit: 9999,
+    config.add_facet_field 'exploded_project_tag_ssim', label: 'Project', limit: 100_000,
                                                         component: LazyProjectTagFacetComponent,
                                                         unless: ->(controller, _config, _response) { controller.params[:no_tags] }
     # exploded_nonproject_tag_ssim indexes all tag prefixes, except project tags, for hierarchical facet display,
     #   whereas tag_ssim only indexes whole tags.
-    config.add_facet_field 'exploded_nonproject_tag_ssim', label: 'Tag', limit: 9999,
+    config.add_facet_field 'exploded_nonproject_tag_ssim', label: 'Tag', limit: 100_000,
                                                            component: LazyNonprojectTagFacetComponent,
                                                            unless: ->(controller, _config, _response) { controller.params[:no_tags] }
     config.add_facet_field 'objectType_ssim', label: 'Object Type', component: true, limit: 10


app/controllers/catalog_controller.rb
```diff
@@ -56,12 +56,12 @@ class CatalogController < ApplicationController
 
     # exploded_project_tag_ssim indexes all project tag prefixes for hierarchical facet display, whereas
     #   project tag_ssim only indexes whole tags
-    config.add_facet_field 'exploded_project_tag_ssim', label: 'Project', limit: 9999,
-                                                        component: LazyProjectTagFacetComponent,
+    config.add_facet_field 'exploded_project_tag_ssim', label: 'Project', limit: 100_000,
+                                                        component: Blacklight::Hierarchy::FacetFieldListComponent,
                                                         unless: ->(controller, _config, _response) { controller.params[:no_tags] }
```
peetucket, line 61: we aren't using the `LazyProjectTagFacetComponent` anymore ?

justinlittman: No, that was something I was testing that I forgot to roll back. Good catch.

peetucket (approved):

Issue #4265, opened 2023-11-03
tags should be searchable


ndushay:
All our tag fields are strings (no tokens).  Exploded tag fields are for hierarchical facets.

We need a new tag field that is for searching ... perhaps similar to the source id field for searching?

<img width="750" alt="image" src="https://github.com/sul-dlss/argo/assets/96775/1797a801-bb21-47cd-887f-3e791f67b1e5">

ndushay:
Example tags
sul : SC1248
Registered By : llam813
DPG : Workflow : OS_Widetek
Remediated By : 4.15.4
Project : Catalhöyük : Images : Batch 1
Process : Content Type : Book (image-only)
Process : Content Type : Book (flipbook, ltr)
accession : 2018-037
exhibits : ai-cs-feigenbaum
Project : Revs : EuroMoto_BOX3

ndushay:
- [x] argo Solr config files updated for tokenized unstemmed field 
    - [ ] sul-dlss/sul-solr-configs/pull/283
    - [ ] sul-dlss/argo/pull/4286
- dor_indexing_app populates new field `tag_text_unstemmed_sim`
    - [ ] sul-dlss/dor_indexing_app/pull/1068
    - [ ] dor_indexing_app qa 
    - [ ] dor_indexing_app stage 
    - [ ] dor_indexing_app prod
- [ ] argo solr prod `tag_text_unstemmed_sim` has begun populating
- [ ] argo solr prod field has been populated
- update Argo solr and/or app (add it to blacklight_config_helper) to use new field(s) for searching by default (and add weighting and to pf, etc.)
   - [ ] argo solrconfig.xml file PR
   - [ ] solr config changes deployed to qa
   - [ ] solr config changes deployed to stage
   - [ ] solr config changes deployed to prod
   - [ ] argo PR for replacing usage of `tag_ssim`
   - [ ] argo changes deployed to qa 
   - [ ] argo changes deployed to stage 
   - [ ] argo changes deployed to prod 
- [ ] remove `tag_ssim`
   - [ ] argo
   - [ ] dor_indexing_app
   - [ ] indexes
       - [ ] qa
       - [ ] stage
       - [ ] prod

ndushay:
as of 2024-01-04, prod has 2,287,601 for `tag_text_unstemmed_im` but 6,719,330 for `tag_ssim`.

ndushay:
closing this in anticipation of Argo deploy with next dep updates (presumably on Monday).

PR #4287, opened 2023-12-09
searches use searchable tag field

ndushay:
HOLD for prod index to have new field fully populated.

# Why was this change made?

part of #4265 

# How was this change tested?

on stage and qa,  with all the objects reindexed


lwrubel (approved):

ndushay:
This is done;  deployed as of 2024-01-21

PR #4290, opened 2023-12-13
Remove symphony links.

### LINKED ISSUE #4289: Removing a Folio HRID from a druid should also remove any corresponding catkey that may still be there
**Describe the bug**
When a user removes a Folio HRID from a druid, it moves the previous Folio HRID to "previous folio". But if the item also still has a catkey, it leaves the catkey behind. This means the catkey still gets published to the public XML, which means that it's still treated in Access systems like it has a catkey. That in turn, prevents some updates from going through, such as changes to SW release status.

Example: https://argo.stanford.edu/view/wz486hk4528

This item had a Folio HRID but it was recently removed. The Cocina now shows:

```
"catalogLinks": [
      {
        "catalog": "previous folio",
        "refresh": false,
        "catalogRecordId": "a11760291"
      },
      {
        "catalog": "symphony",
        "refresh": true,
        "catalogRecordId": "11760291"
      }
    ],
```
In the past, moving an identifier from "folio" to "previous folio" would move the corresponding "symphony" to "previous symphony". We could still do that. Or we could remove the catkey completely since the "previous folio" tracks the information. What we can't do is leave the catkey in place.

**User Impact**
This has a subtle but noticeable effect on how things appear in SW, including in some cases preventing items from appearing in SW at all.

**To Reproduce**
It doesn't look like this can be reproduced in stage because there are no more catkeys there.

1. Go to an item with a catkey and Folio HRID
2. Remove the Folio HRID
3. See the catkey is left behind

**Expected behavior**
Removing an HRID should remove a corresponding catkey.

**Additional context**
I wonder if we removed the syncing code that used to handle this.

Also, before anyone asks, removing all the catkeys will require coordination with the Access portfolio. So I think it's better in the short run to address this issue rather than wait for a mass cleanup.

justinlittman:
closes #4289

# Why was this change made?

Keeping symphony links is causing mayhem in access systems.

# How was this change tested?
Unit

### DIFF
--- a/app/services/catalog_record_id.rb
+++ b/app/services/catalog_record_id.rb
@@ -43,10 +43,14 @@ class CatalogRecordId
     def previous_type
       "previous #{type}"
     end
+
+    def other_type
+      'symphony'
+    end
   end
 
   def self.links(model)
-    new(model).links
+    new(model).links.map(&:catalogRecordId)
   end
 
   def self.serialize(model, catalog_record_ids, refresh: true)
@@ -66,8 +70,8 @@ class CatalogRecordId
   # @param [boolean] refresh first catalog record ID
   # @return [Array<Hash>] a list of catalog links
   def serialize(new_catalog_record_ids, refresh: true)
-    removed_links = links - new_catalog_record_ids
-    links = (previous_links + removed_links).map do |record_id|
+    removed_links = links.map(&:catalogRecordId) - new_catalog_record_ids
+    links = (previous_links.map(&:catalogRecordId) + removed_links).map do |record_id|
       { catalog: self.class.previous_type, catalogRecordId: record_id, refresh: false }
     end.uniq
     links +
@@ -82,7 +86,8 @@ class CatalogRecordId
   end
 
   def other_links
-    find_not(self.class.type)
+    # Remove symphony links. See https://github.com/sul-dlss/argo/issues/4289
+    find_not(self.class.type) - find(self.class.other_type)
   end
 
   def link_refresh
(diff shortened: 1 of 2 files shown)


app/services/catalog_record_id.rb
```diff
@@ -98,7 +103,7 @@ def find_not(type)
   end
 
   def find(type)
```
justinlittman, line 105: I found it confusing that some find methods returned catalog record ids and others returned catalog records. This normalizes to returning catalog records.

app/services/catalog_record_id.rb
```diff
@@ -43,10 +43,14 @@ def type
...
     end
+
+    def other_type
+      'symphony'
+    end
   end
 
   def self.links(model)
-    new(model).links
+    new(model).links.map(&:catalogRecordId)
```
jmartin-sul, line 53: i found it a little confusing that this class method `links` does the `.map(&:catalogRecordId)`, but instance method `links` doesn't.

jmartin-sul (approved):
comment about one thing that i found confusing, but otherwise makes sense to me, and unit test change looks good, so ✅

app/services/catalog_record_id.rb
jmartin-sul: ah, i see that the same is true for the instance and class method versions of `other_links`, so at least it's consistently inconsistent 🙂, and keeps from breaking an interface with the rest of the codebase, so that makes sense too.

also weird, now that i think about it, that some class methods are defined using the `class << self` approach, and others the `self.` approach, but that's not something i'd ask to see changed in this PR, since this PR didn't introduce that.

PR #4297, opened 2024-01-08
Upgrade view component.

justinlittman:
# Why was this change made?

# How was this change tested?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -606,8 +606,8 @@ GEM
       unf_ext
     unf_ext (0.0.9.1)
     unicode-display_width (2.5.0)
-    view_component (2.74.1)
-      activesupport (>= 5.0.0, < 8.0)
+    view_component (2.82.0)
+      activesupport (>= 5.2.0, < 8.0)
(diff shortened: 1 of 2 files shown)


ndushay (approved):

Gemfile.lock
```diff
@@ -606,8 +606,8 @@ GEM
       unf_ext
     unf_ext (0.0.9.1)
     unicode-display_width (2.5.0)
-    view_component (2.74.1)
-      activesupport (>= 5.0.0, < 8.0)
+    view_component (2.82.0)
+      activesupport (>= 5.2.0, < 8.0)
```
mjgiarlo, line 610: It looks like in H2, VC updated to the latest. Do you know why it was held to 2.x in Argo? Some other dependency, presumably.

justinlittman: I suspect that in the early, unstable days of VC there was an update that caused a problem and we pinned a few apps and forgot to unpin.

mjgiarlo: In this case, judging by the dependabot PR I closed, I think it is blacklight that may be holding us back.

Issue #4292, opened 2023-12-20
Provide user feedback when a registration CSV doesn't have the required headers


andrewjbtw:
**Describe the bug**
Recently a user tried to register with a CSV that didn't have the proper headers. The problem was that they had "title" as a field instead of "label". The form did not report an error but no druids were registered. There should be a visible message in Argo when the CSV is not valid.

**User Impact**
When this happens, it's difficult to tell what went wrong. The user tried a couple of times before getting in touch with me and then we eventually worked out what was wrong.

**To Reproduce**
1. Go to argo-stage registration form: https://argo-stage.stanford.edu/registration
2. Try to register with this CSV: [registration.csv](https://github.com/sul-dlss/argo/files/13721902/registration.csv)
3. See no error
4. See that registration has not worked

In stage, what I'm seeing is that a bulk action runs but gets caught in retries.

In prod, the user who reported this was taken to the page that shows registration results for form-based registration, but it was blank. This screenshot is what they saw: ![blank-registration-results](https://github.com/sul-dlss/argo/assets/8975583/da3e515f-ecd0-4e74-a022-1269b11d06fa)

**Expected behavior**
A registration CSV should not be processed when it's not valid and the user should see an error message.

**Additional context**
It's a bit confusing that the field is called label when it really means title, but that's a legacy label that we have not changed from Fedora days. If we could accept "label" or "title" that might actually be better but I don't know if that option is available to us.

PR #4301, opened 2024-01-09
Validate that CSV registration form has label or catalog record id field.

justinlittman:
refs #4292

# Why was this change made?
So that users can be notified of the errors in their ways.

# How was this change tested?
Unit, local

image


app/controllers/concerns/creates_bulk_actions.rb
```diff
@@ -64,7 +64,8 @@ def validate_job_params(_job_params)
   end
 
   def validate_csv_headers(csv, headers)
-    validator = CsvUploadValidator.new(csv:, headers:)
+    header_validators = headers.present? ? CsvUploadValidator::RequiredHeaderValidator.new(headers:) : []
+    validator = CsvUploadValidator.new(csv:, header_validators:)
```
mjgiarlo, line 68: Is the idea here that `CsvUploadValidator`'s collaborators should decide what sort of header validations they want instead of delegating that to the `CsvUploadValidator`?

app/controllers/registrations_controller.rb
```diff
@@ -77,7 +77,7 @@ def spreadsheet
     respond_to do |format|
       format.csv do
         csv_template = CSV.generate do |csv|
-          csv << ['barcode', CatalogRecordId.label.downcase.tr(' ', '_'), 'source_id', 'label']
+          csv << ['barcode', CatalogRecordId.csv_header, 'source_id', 'label']
```
mjgiarlo, line 80: Great catch!

app/controllers/concerns/creates_bulk_actions.rb
justinlittman: Yes. Previously the only need for validation was to check required headers. However, resolving this issue required additional validation which was only needed by some collaborators.

app/services/csv_upload_validator.rb
```diff
@@ -2,11 +2,60 @@
...
+      missing_headers = required_headers.select { |required_header| csv.headers.exclude?(required_header) }
+      return ["missing headers: #{missing_headers.join(', ')}."] if missing_headers.present?
+
+      []
+    end
+
+    private
+
+    attr_reader :required_headers
+  end
```
mjgiarlo, line 21: Given the validator classes here are so small, it seems perfectly fine to leave them all in the same file and test them together as a unit.

mjgiarlo (approved):
Approve pending consideration of typo fix comment

PR #4303, opened 2024-01-09
Remove profiling tools.

justinlittman:
# Why was this change made?
They're not used.

# How was this change tested?


jmartin-sul (approved):
needs a rebase, `config/settings.yml` has conficts, but ✅

PR #4322, opened 2024-02-20
csv registration ignores blank lines at end

### LINKED ISSUE #4316: Ignore CSV trailing lines that are blank except for commas and whitespace
**Describe the bug**
When a CSV has rows that contain commas but no content other than whitespace, Argo reports an error at registration. Argo should ignore these blank lines if they come at the bottom of a CSV registration import.

The problem on the user side is that spreadsheet programs don't do a good job visually indicating when a spreadsheet has empty rows at the bottom of the sheet. For example, a spreadsheet with rows 1-10 filled in could:

- have only 10 rows
- have 15 rows but rows 11-15 are blank

In many common spreadsheet programs (Excel, Google sheets, etc.) those spreadsheets will be visually equivalent. The user only finds out they have extra rows if they save as CSV and then inspect the CSV in a text editor.

**User Impact**
This issue causes seemingly mysterious registration errors. To the user, the spreadsheet looks complete. But Argo reports an error message that suggests data is missing.

**To Reproduce**
1. Go to https://argo-stage.stanford.edu/registration
2. Select "Upload CSV" on the lower part of the page
3. Upload this CSV
[registration-stage-blanks.csv](https://github.com/sul-dlss/argo/files/14186373/registration-stage-blanks.csv)

Screenshot 2024-02-06 at 1 49 06 PM

**Expected behavior**
Argo should ignore rows with no content at the end of the CSV. It makes sense to catch blanks that are in the middle of the CSV because that should also be a visible problem for the user creating the CSV. Because of the nature of how people copy and paste from CSVs, it's important to catch blanks in the middle of the CSV because it will throw off row alignments if the user copies from the registration report back into their original CSV.

ndushay:
# Why was this change made?

Fixes #4316

# How was this change tested?

Wrote a test and it failed without the fix, and passes with the fix.

Also, deployed to stage and Andrew exercised it and signed off.

### DIFF
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -49,7 +49,17 @@ class RegistrationCsvConverter
   #  14: tags (optional, may repeat)
 
   def convert
-    CSV.parse(csv_string, headers: true).map { |row| convert_row(row) }
+    csv_table = CSV.parse(csv_string, headers: true)
+    remove_blank_rows_at_end(csv_table)
+    csv_table.map { |row| convert_row(row) }
+  end
+
+  def remove_blank_rows_at_end(csv_table)
+    last = csv_table.size - 1
+    while csv_table[last].fields.all?(&:blank?) && last != 0
+      csv_table.delete(last) if csv_table[last].fields.all?(&:blank?)
+      last -= 1
+    end
   end
 
   def convert_row(row)
--- a/app/services/csv_upload_validator.rb
+++ b/app/services/csv_upload_validator.rb
@@ -72,6 +72,18 @@ class CsvUploadValidator
   attr_reader :raw_csv, :header_validators
 
   def csv
-    @csv ||= CSV.parse(raw_csv, headers: true)
+    @csv ||= begin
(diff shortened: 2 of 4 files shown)


app/services/registration_csv_converter.rb
```diff
@@ -49,7 +49,17 @@ def initialize(csv_string:, params:)
...
+    remove_blank_rows_at_end(csv_table)
+    csv_table.map { |row| convert_row(row) }
+  end
+
+  def remove_blank_rows_at_end(csv_table)
+    last = csv_table.size - 1
+    while csv_table[last].fields.all?(&:blank?) && last != 0
+      csv_table.delete(last) if csv_table[last].fields.all?(&:blank?)
+      last -= 1
+    end
```
ndushay, line 62: AFAICT, we can get here without going through the upload validator so it is needed.  However, if a csv file is uploaded on the registration form, the validation needs to happen in THAT class or there will be an error on the last (blank) lines that is spurious.

aaron-collier (approved):

PR #4328, opened 2024-02-27
replace solr search field objectId_tesim with druid_prefixed_ssi and druid_bare_ssi

ndushay:
# Why was this change made?

Part of sul-dlss/dor_indexing_app/issues/1031

Search the newly populated fields `druid_prefixed_ssi` and `druid_bare_ssi` instead of the old `objectId_tesim` which had both a poor name and a poor field type.

# How was this change tested?

I deployed to qa and searches for druids still worked, with and without prefixes.

### DIFF
--- a/solr_conf/conf/solrconfig.xml
+++ b/solr_conf/conf/solrconfig.xml
@@ -62,7 +62,8 @@
         collection_title_tesim
 
         id
-        objectId_tesim
+        druid_bare_ssi
+        druid_prefixed_ssi
         obj_label_tesim
         identifier_ssim
         identifier_tesim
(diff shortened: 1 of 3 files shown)


solr_conf/conf/solrconfig.xml
```diff
@@ -85,7 +86,6 @@
 
         collection_title_tesim^5
 
-        objectId_tesim^5
```
ndushay, line 88: there is no point in boosting any sort of phrase or n-gram matches in a single token string.

mjgiarlo (approved):

PR #4340, opened 2024-03-08
CollectionTitleService is new name of CollectionNameService

ndushay:
# Why was this change made?

It's really the title for the collection;  we use "name" in speech but it's a bit weird to have the "CollectionNameService" get the data in the title field.

Or maybe I'm too close to the descriptive metadata at the moment.   It's fine if you close this PR and delete the branch if you think CollectionName makes more sense.

# How was this change tested?

CI

### DIFF
--- a/app/controllers/apo_controller.rb
+++ b/app/controllers/apo_controller.rb
@@ -34,18 +34,18 @@ class ApoController < ApplicationController
     }.compact)
     @access_template = AccessTemplate.new(access_template:, apo_defaults_template: administrative.accessTemplate)
 
-    @collections = Array(administrative.collectionsForRegistration).filter_map do |col_id|
-      name = CollectionNameService.find(col_id)
-      unless name
-        Honeybadger.notify("The APO #{params[:id]} asserts that #{col_id} is a collection for registration, but we don't find that collection in solr")
+    @collections = Array(administrative.collectionsForRegistration).filter_map do |coll_id|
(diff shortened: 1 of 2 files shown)


app/controllers/apo_controller.rb
```diff
@@ -34,18 +34,18 @@ def registration_options
     }.compact)
     @access_template = AccessTemplate.new(access_template:, apo_defaults_template: administrative.accessTemplate)
 
-    @collections = Array(administrative.collectionsForRegistration).filter_map do |col_id|
-      name = CollectionNameService.find(col_id)
-      unless name
-        Honeybadger.notify("The APO #{params[:id]} asserts that #{col_id} is a collection for registration, but we don't find that collection in solr")
+    @collections = Array(administrative.collectionsForRegistration).filter_map do |coll_id|
```
justinlittman, line 37: Minor request: Get rid of the abbreviations in this method.

justinlittman (approved):

app/controllers/apo_controller.rb
ndushay: I confess I was slightly confused by the original `col_id` because I read that as column, but `collection_id` is so much longer and the context of use is small here.

ndushay: I'm afraid I'm going to merge, and if you want to refactor, go for it.

PR #4356, opened 2024-03-20
Remove the version significance selector from open/close

jcoyne:
# Why was this change made?
Fixes https://github.com/sul-dlss/argo/issues/4363
Fixes https://github.com/sul-dlss/argo/issues/4364

# How was this change tested?


justinlittman (approved):
Approve pending rubo fix.

PR #4357, opened 2024-03-20
Use ruby debug rather than byebug

jcoyne:
# Why was this change made?

Byebug is obsolete and unmaintained

# How was this change tested?


justinlittman (approved):

README.md
```diff
@@ -147,10 +147,10 @@ To enable interactive debugging, invoke `bin/dev` as follows:
 REMOTE_DEBUGGER=byebug bin/dev
```
lwrubel, line 147: We can remove this now, right?

lwrubel: Meaning-- just `bin/dev`

Issue #4331, opened 2024-02-28
add integration tests for indexing


ndushay:

jcoyne:
The two tests that have been added so far have proved to be flaky and are detrimental to the developer experience.  I propose that instead, a new integration test suite (that is just run in CI) is developed.  This can test all of the integrations without making Argo hard to work on and without impacting the state of the stage servers.

lwrubel:
For the short term, we'll mark these tests to be skipped in CI. https://github.com/sul-dlss/argo/issues/4380.

PR #4361, opened 2024-03-22
add title_indexing_integration feature spec

ndushay:
# Why was this change made?

Part of https://github.com/sul-dlss/argo/issues/4331

As discussed in Infra Dev Planning meeting on Tues 3/5/24, this PR introduces some integration tests that allow us to put arbitrary cocina objects into dor-services-app via factory_bot (see spec/factories/items.rb), index them using the Argo indexer (which calls the dor-indexing-app), and then use the Argo UI to search for the objects, ensure the search results are as expected, ditto the facets, blah blah.

# How was this change tested?

CI

### DIFF
--- a/spec/features/indexing_other_spec.rb
+++ b/spec/features/indexing_other_spec.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):

mjgiarlo:
Thank you, @ndushay!

PR #4376, opened 2024-03-27
Update test to work without accessionWF:descriptive-metadata

jcoyne:
# Why was this change made?

This was removed from the accessionWF in the workflow-service

# How was this change tested?


justinlittman (approved):
Approve pending green.

PR #4390, opened 2024-04-04
Enhance version service

justinlittman:
# Why was this change made?
First step in normalizing versioning in Argo.

# How was this change tested?

Unit

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -20,7 +20,7 @@ class VersionsController < ApplicationController
   end
 
   def open
-    VersionService.open(identifier: @cocina_object.externalIdentifier,
+    VersionService.open(druid: @cocina_object.externalIdentifier,
                         description: params[:description],
                         opening_user_name: current_user.to_s)
     msg = "#{@cocina_object.externalIdentifier} is open for modification!"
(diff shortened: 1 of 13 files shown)


app/controllers/versions_controller.rb
```diff
@@ -20,7 +20,7 @@ def close_ui
   end
 
   def open
-    VersionService.open(identifier: @cocina_object.externalIdentifier,
+    VersionService.open(druid: @cocina_object.externalIdentifier,
```
justinlittman, line 23: `identifier:` was renamed to `druid:` here and below.

app/services/version_service.rb
```diff
@@ -4,30 +4,59 @@
...
+      new(...).version
     end
   end
 
-  attr_reader :identifier
+  attr_reader :druid
 
-  delegate :close, :open, :openable?, to: :version_client
+  delegate :close, :open, to: :version_client
+  delegate :open?, :openable?, :assembling?, :accessioning?, :closed?, :closeable?, :version, to: :status
```
justinlittman, line 47: This is the big change: Adding methods from version status.

lwrubel (approved):

PR #4392, opened 2024-04-04
Refactor bulk action jobs to use version service.

### LINKED ISSUE #4391: Bulk action jobs should use VersionService
This will get status from DSA version status endpoint, instead of using local logic.

justinlittman:
closes #4391

# Why was this change made?

To get rid of local versioning logic.

# How was this change tested?

Unit

### DIFF
--- a/spec/jobs/set_license_and_rights_statements_job_spec.rb
+++ b/spec/jobs/set_license_and_rights_statements_job_spec.rb
@@ -11,7 +11,6 @@ RSpec.describe SetLicenseAndRightsStatementsJob do
   let(:logger) { instance_double(File, puts: nil) }
 
   describe '#perform' do
-    let(:open) { true }
     let(:params) do
       {
         copyright_statement_option: '1',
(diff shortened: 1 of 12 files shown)


spec/jobs/set_source_ids_csv_job_spec.rb
```diff
@@ -38,13 +38,14 @@
   let(:object_client1) { instance_double(Dor::Services::Client::Object, find: item1, update: true) }
   let(:object_client2) { instance_double(Dor::Services::Client::Object, find: item2, update: true) }
   let(:object_client3) { instance_double(Dor::Services::Client::Object, find: item3, update: true) }
-  let(:state_service) { instance_double(StateService, open?: true) }
+  # let(:state_service) { instance_double(StateService, allows_modification?: true) }
```
jmartin-sul, line 41: cruft?

spec/jobs/set_source_ids_csv_job_spec.rb
```diff
@@ -38,13 +38,14 @@
...
 
   before do
     allow(subject).to receive(:bulk_action).and_return(bulk_action)
     allow(BulkJobLog).to receive(:open).and_yield(log_buffer)
     allow(Ability).to receive(:new).and_return(ability)
-    allow(StateService).to receive(:new).and_return(state_service)
+    # allow(StateService).to receive(:new).and_return(state_service)
```
jmartin-sul, line 47: cruft?

spec/jobs/set_license_and_rights_statements_job_spec.rb
jmartin-sul, line 1: i think the `open` let var can be removed now also, doesn't seem to be used in this file anymore

jmartin-sul (approved):
approved, found some spots that might be cruft, but overall LGTM

PR #4395, opened 2024-04-04
[REFACTOR] Move accessioned? from state service to workflow service

justinlittman:
# Why was this change made?

The workflow service should answer workflow questions.

# How was this change tested?

### DIFF
--- a/app/services/state_service.rb
+++ b/app/services/state_service.rb
@@ -39,16 +39,16 @@ class StateService
     STATES[:unlock_inactive]
   end
 
+  private
+
+  attr_reader :druid, :version
+
   ##
   # Ported over logic from app/helpers/dor_object_helper.rb#LN133
   # @return [Boolean]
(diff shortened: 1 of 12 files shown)


app/services/state_service.rb
```diff
@@ -39,17 +39,17 @@ def object_state
...
   end
 
+  private
+
+  attr_reader :druid, :version
+
   ##
   # Ported over logic from app/helpers/dor_object_helper.rb#LN133
   # @return [Boolean]
   def accessioned?
```
aaron-collier, line 49: Should this be removed after (or as part of) this PR?

aaron-collier (approved):

app/services/state_service.rb
justinlittman: Yes, it will be removed once StateService is refactored. (Forthcoming PR.)

PR #4396, opened 2024-04-04
[REFACTOR] Move published

justinlittman:
# Why was this change made?

It's already handled by the WorkflowService.

# How was this change tested?

Unit


andrewjbtw:
@justinlittman what does this do to the "publish" button in Argo that functions outside of workflows?

justinlittman:
It should do nothing. Same call to workflow system, just shuffling around the code.

aaron-collier (approved):

PR #4401, opened 2024-04-08
Refactor state service open

justinlittman:
# Why was this change made?

Determining if an item is open is the responsibility of VersionService, not StateService.

# How was this change tested?

### DIFF
--- a/app/components/show/agreement/details_component.html.erb
+++ b/app/components/show/agreement/details_component.html.erb
@@ -11,7 +11,7 @@
       <tr>
         <th class="col-3" scope="row">Content type</th>
         <td>
-          <%= render Show::ContentTypeComponent.new(document: @solr_document, state_service:) %>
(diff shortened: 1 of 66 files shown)


app/components/show/agreement/details_component.html.erb
```diff
@@ -11,7 +11,7 @@
       <tr>
         <th class="col-3" scope="row">Content type</th>
         <td>
-          <%= render Show::ContentTypeComponent.new(document: @solr_document, state_service:) %>
+          <%= render Show::ContentTypeComponent.new(document: @solr_document, version_service: @presenter.version_service) %>
```
justinlittman, line 14: This PR is most replacing StateService usage with VersionService.

app/services/state_service.rb
```diff
@@ -14,10 +14,6 @@ def initialize(cocina)
     @version = cocina.version
   end
 
-  def open?
```
justinlittman, line 17: And this is the method that we got rid of.

mjgiarlo (approved):

PR #4404, opened 2024-04-10
Use VersionService in ApplyModsMetadata

### LINKED ISSUE #4403: Update versioning in the ApplyModsMetadata
The logic should be:
1. If the item is not open and is not openable? then skip it.
2. If the item is not open then open a new version.
3. Update the cocina.

The item is NOT closed after updating the cocina.

The VersionService should be used for getting status and performing versioning operations; local logic for determining version status based on workflows should be removed.

lwrubel:
# Why was this change made?
Resolves #4403.

# How was this change tested?
Unit. And passes integration tests for bulk_metadata_update. Andrew did some testing, but there is an HB error to investigate: https://app.honeybadger.io/projects/50568/faults/106837203 and which should be resolved with https://github.com/sul-dlss/dor-services-app/pull/4863.

### DIFF
--- a/app/services/apply_mods_metadata.rb
+++ b/app/services/apply_mods_metadata.rb
@@ -24,12 +24,7 @@ class ApplyModsMetadata
       return false
     end
 
-    if in_accessioning?
-      log.puts("argo.bulk_metadata.bulk_log_skipped_accession #{cocina.externalIdentifier}")
-      return false
-    end
-
-    return false unless status_ok?
+    return false unless updatable?
 
     return false unless ability.can? :update, cocina
 
@@ -76,15 +71,9 @@ class ApplyModsMetadata
     log.puts(exception.backtrace.to_s)
   end
 
-  # Open a new version for the given object if it is in the accessioned state.
+  # Open a new version for the given object if not already open
   def version_object
-    return unless accessioned?
-
-    unless DorObjectWorkflowStatus.new(cocina.externalIdentifier, version: cocina.version).can_open_version?
-      log.puts("argo.bulk_metadata.bulk_log_unable_to_version #{cocina.externalIdentifier}") # totally unexpected
-      return
-    end
-    commit_new_version
+    open_version unless VersionService.open?(druid: cocina.externalIdentifier)
   end
 
   def update_metadata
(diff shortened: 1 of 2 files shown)


app/services/apply_mods_metadata.rb
```diff
@@ -93,39 +82,18 @@ def update_metadata
...
-  # Checks whether or not a DOR object's status is OK for a descMetadata update. Basically, the only times we are
-  # not OK to update is if the object is currently being accessioned and if the object has status unknown.
-  #
-  # @return [Boolean] true if the object's status allows us to update the descripitive metadata, false otherwise
-  def status_ok?
-    [1, 6, 7, 8, 9].include?(status)
-  end
+  # Item must be open or openable? to be updated
+  def updatable?
+    return true if VersionService.open?(druid: cocina.externalIdentifier) || VersionService.openable?(druid: cocina.externalIdentifier)
```
justinlittman, line 93: Using the class methods will result in separate calls to DSA for .open? and .openable? If you instantiate a VersionService and call the instance methods it will be only a single call to DSA.

lwrubel: Thanks for the advice, will implement that.

lwrubel:
@justinlittman I did some refactoring of this and it could use a review (even though still draft). However, I'm stuck getting the bulk_update_metadata integration test to pass on QA, and trying to figure out what is going on there. So there may be something still needing fixing in this. I have not been able to determine yet what is failing in the integration test.

app/services/apply_mods_metadata.rb
```diff
@@ -93,39 +82,22 @@ def update_metadata
...
 
-  # Checks whether or not a DOR object is in accessioning or not.
-  #
-  # @return [Boolean] true if the object is currently being accessioned, false otherwise
-  def in_accessioning?
-    (2..5).cover?(status)
-  end
+  # Item must be open or openable? to be updated
+  def updatable?
+    return true if version_service.open?(druid: cocina.externalIdentifier) || version_service.openable?(druid: cocina.externalIdentifier)
```
justinlittman, line 93: I don't think you need to call open? and openable? with an args here.

app/services/apply_mods_metadata.rb
```diff
@@ -76,15 +71,9 @@ def log_error!(exception)
...
+  # Open a new version for the given object if not already open
   def version_object
-    return unless accessioned?
-
-    unless DorObjectWorkflowStatus.new(cocina.externalIdentifier, version: cocina.version).can_open_version?
-      log.puts("argo.bulk_metadata.bulk_log_unable_to_version #{cocina.externalIdentifier}") # totally unexpected
-      return
-    end
-    commit_new_version
+    open_version unless version_service.open?(druid: cocina.externalIdentifier)
```
justinlittman, line 76: I would just put the code from `open_version` in `version_object`. I'd suggest that it will it clearer what is happening.

justinlittman (commented):
Looks good, with a few comments.

app/services/apply_mods_metadata.rb
lwrubel: Fixed, thanks.

justinlittman (approved):

PR #4407, opened 2024-04-12
Fix tracking-sheets to use string based title field

### LINKED ISSUE #4406: Tracking sheets are showing only the first letter where they should be showing the title
**Describe the bug**
The "object label" field in a tracking sheet should show the title of an object. Currently, tracking sheets are only showing the first letter of the title.

Example: [tracking_sheet.pdf](https://github.com/sul-dlss/argo/files/14951871/tracking_sheet.pdf)
Screenshot 2024-04-11 at 3 26 31 PM

**User Impact**
Tracking sheets are used to connect physical objects with druids. Tracking sheets generally accompany physical objects throughout the digitization process, so having the full title helps everyone make sure the correct sheet is with the correct item.

**To Reproduce**
1. Go to Argo bulk actions
2. Choose the "Download tracking sheet" bulk action
3. Enter this druid ns724rz6542 (same as attached to this ticket)
4. See only the letter 'S' under object label

**Expected behavior**
The title of the object should be printed on the sheet.

**Additional context**
Cocina has a "label" field but that should not be used for the tracking sheet. We should use the title. The Cocina object label field comes from the Fedora object label field and we are trying to deprecate it. I think the tracking sheet is already using the descriptive title but I'm adding this note just in case it isn't yet.

aaron-collier:
# Why was this change made?

Fixes #4406 

# How was this change tested?

### DIFF
--- a/app/models/track_sheet.rb
+++ b/app/models/track_sheet.rb
@@ -95,8 +95,8 @@ class TrackSheet
   def doc_to_table(solr_doc)
     doc = solr_doc.with_indifferent_access # solr doc always has string keys, SolrDocument::FIELDS can be strings or symbols
     table_data = []
-    labels = doc[SolrDocument::FIELD_TITLE]
-    label = labels.blank? ? '' : labels.first.truncate(110)
+    label = doc[SolrDocument::FIELD_TITLE].is_a?(Array) ? doc[SolrDocument::FIELD_TITLE].first : doc[SolrDocument::FIELD_TITLE]
+    label = label.truncate(110)
     table_data.push(['Object Label:', label])
     table_data.push(['Project Name:', doc['project_tag_ssim'].to_s]) if doc['project_tag_ssim']
 
--- a/app/models/track_sheet.rb
+++ b/app/models/track_sheet.rb
@@ -93,6 +93,7 @@ class TrackSheet
   # @param [Hash] doc Solr document or to_solr Hash
   # @return [Array<Array<String>>] Complex array suitable for pdf.table()
   def doc_to_table(solr_doc)
+    solr_doc = SolrDocument.new(solr_doc) unless solr_doc.is_a?(SolrDocument)
     table_data = []
     label = solr_doc.title_display.truncate(110)
     table_data.push(['Object Label:', label])
--- a/app/models/track_sheet.rb
+++ b/app/models/track_sheet.rb
@@ -93,23 +93,18 @@ class TrackSheet
   # @param [Hash] doc Solr document or to_solr Hash
   # @return [Array<Array<String>>] Complex array suitable for pdf.table()
   def doc_to_table(solr_doc)
-    doc = solr_doc.with_indifferent_access # solr doc always has string keys, SolrDocument::FIELDS can be strings or symbols
     table_data = []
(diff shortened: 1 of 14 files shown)


mjgiarlo (approved):

justinlittman (approved):

app/models/track_sheet.rb
```diff
@@ -95,8 +95,8 @@ def generate_tracking_sheet(druid, pdf)
   def doc_to_table(solr_doc)
     doc = solr_doc.with_indifferent_access # solr doc always has string keys, SolrDocument::FIELDS can be strings or symbols
     table_data = []
-    labels = doc[SolrDocument::FIELD_TITLE]
-    label = labels.blank? ? '' : labels.first.truncate(110)
+    label = doc[SolrDocument::FIELD_TITLE].is_a?(Array) ? doc[SolrDocument::FIELD_TITLE].first : doc[SolrDocument::FIELD_TITLE]
```
aaron-collier, line 98: While in practice I'm seeing this value as a string, the tests indicate that an array may be possible. Update the logic to account for that possibility.

jcoyne: That is just bad test setup. It will never be multivalued. See https://github.com/sul-dlss/argo/blob/cd655c7e56afc4d9dce41e3a765039ae001c126c/solr_conf/conf/schema.xml#L22

The `TrackSheet` class should be using `doc.title_display` (https://github.com/sul-dlss/argo/blob/main/app/models/solr_document.rb#L67C14-L67C27) and not `SolrDocument::TitleField`

jcoyne: It looks like the `title_display` attribute was never changed from `Array` to `String` when the `SolrDocument::TitleField` was changed.  That should be done too.

aaron-collier: This is helpful @jcoyne thank you. Updating.

jcoyne: @aaron-collier I think it also makes sense to remove `SolrDocument#title` as that is now just the same as `title_display`

jcoyne (approved):

app/models/track_sheet.rb
```diff
@@ -93,23 +93,19 @@ def generate_tracking_sheet(druid, pdf)
   # @param [Hash] doc Solr document or to_solr Hash
   # @return [Array<Array<String>>] Complex array suitable for pdf.table()
   def doc_to_table(solr_doc)
-    doc = solr_doc.with_indifferent_access # solr doc always has string keys, SolrDocument::FIELDS can be strings or symbols
+    solr_doc = SolrDocument.new(solr_doc) unless solr_doc.is_a?(SolrDocument)
```
mjgiarlo, line 96: As far as I can tell, this method is only ever called here: https://github.com/sul-dlss/argo/blob/12018e207868fc6e8ef85f11c1d3f0aaca0899d9/app/models/track_sheet.rb#L63

Then `doc` is set to the return value of this method: https://github.com/sul-dlss/argo/blob/12018e207868fc6e8ef85f11c1d3f0aaca0899d9/app/models/track_sheet.rb#L123C11-L127

And that method only returns what comes back from `SearchService.query(%(id:"#{namespaced_druid}"), rows: 1)['response']['docs'].first` which as far as I can tell is only ever a hash. Can you confirm there's some context where a SolrDocument instance is passed to this method?

aaron-collier: Yeah, it was sort of a mix in the tests from what I could tell but a hash in the code. It's now getting a SolrDocument

PR #4412, opened 2024-04-15
Disable Create collection button only when APO not open and not openable

### LINKED ISSUE #4410: "Create collection" button/modal/function should automatically handle APO versioning
Argo users should be able to use the "Create collection" function without having to open an APO first.
Screenshot 2024-04-12 at 2 28 34 PM

But the function updates the APO Cocina because it adds the new collection druid to the `collectionsForRegistration` list in the administrative metadata. Under the new approach to versioning, the APO object will need to be opened for this change to take place.

Request:

- Create collection button remains active even when object is locked
- Create collection modal allows user to create collection as is done now
- Submitting the modal functions the same way to the user
- Somewhere in that sequence, either while the user is editing or after they submit the form, the APO versioning is taken care of on the system side
- Version description should be "Created new collection: ${collection_druid}"

If it's not possible to insert the collection druid into the version description, just go with "Created new collection".

lwrubel:
# Why was this change made?
Resolves #4410 to adjust Create Collection button.

# How was this change tested?
Unit and testing on QA. Passed the `apo_creation_spec` and `collection_create_spec` integration tests. Will require PO testing.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -21,12 +21,12 @@ class CollectionsController < ApplicationController
     return render 'new' unless form.validate(params.merge(apo_druid: params[:apo_id]))
 
     form.save
+    collection_druid = form.model.externalIdentifier
     # open version for APO if not already open
     version_service = VersionService.new(druid: cocina_admin_policy.externalIdentifier)
-    version_service.open unless version_service.open?
+    cocina_admin_policy = version_service.open(description: "Created new collection: #{collection_druid}") unless version_service.open?
 
     # update APO
-    collection_druid = form.model.externalIdentifier
     collections = Array(cocina_admin_policy.administrative.collectionsForRegistration).dup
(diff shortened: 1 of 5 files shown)


app/controllers/collections_controller.rb
```diff
@@ -35,6 +39,8 @@ def create
       )
     )
     Repository.store(updated_cocina_admin_policy)
+    # Close the APO version and reindex
+    version_service.close(description: "Created new collection: #{collection_druid}")
```
justinlittman, line 43: You should be able to skip the description on close.

justinlittman: I wouldn't overwrite an existing version description, but I don't think it really matters much.

justinlittman (approved):

app/controllers/collections_controller.rb
lwrubel: Fixed to not overwrite an existing version description.

PR #4422, opened 2024-04-24
Switch to using DSA reindex endpoint.

justinlittman:
# Why was this change made?

To allow DIA to be retired.

# How was this change tested?

Unit, integration


app/controllers/apo_controller.rb
```diff
@@ -61,16 +61,16 @@ def create
 
     @form.save
 
-    # Index imediately, so that we have a page to send the user to. DSA indexes asynchronously.
-    Argo::Indexer.reindex_druid_remotely(@form.model.externalIdentifier)
+    # Index immediately, so that we have a page to send the user to. DSA indexes asynchronously.
+    Dor::Services::Client.object(@form.model.externalIdentifier).reindex
```
justinlittman, line 65: The entire PR is basically this ^^.

PR #4429, opened 2024-05-01
Fix stacks and preservation links to filenames with hierarchy

### LINKED ISSUE #4427: Argo links to files are not working when files are in a hierarchy
**Describe the bug**
Argo shows "file not found" errors for links to files that are in a hierarchical arrangement. This is a problem for both the stacks and the preservation links.

**User Impact**
Argo users can't download individual files from items that use hierarchy. The link to download a zip of the preserved files does work.

**To Reproduce**
1. Go to https://argo-stage.stanford.edu/view/druid:fx702cp4356
2. Click on any of the file download links
3. Try the stacks and preservation links
4. See file not found errors

**Expected behavior**
The links to individual files should work.

**Additional context**
We needed to do something with the URLs to make these file links work when we first launched hierarchy. This seems like a regression.

mjgiarlo:
# Why was this change made?

Fixes #4427

# How was this change tested?

- [x] CI
- [x] QA/stage

### DIFF
--- a/app/helpers/items_helper.rb
+++ b/app/helpers/items_helper.rb
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
 module ItemsHelper
-  def stacks_url_full_size(druid, file_name)
-    "#{Settings.stacks_file_url}/#{druid}/#{ERB::Util.url_encode(file_name)}"
+  def stacks_url_full_size(druid, filename)
+    # Allow literal slashes in the filename (do not encode them)
+    encoded_filename = filename.split('/').map { |filename_part| ERB::Util.url_encode(filename_part) }.join('/')
+    "#{Settings.stacks_file_url}/#{druid}/#{encoded_filename}"
   end
 
   # Overriding blacklight so we can pass @cocina to the presenter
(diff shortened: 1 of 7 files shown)


app/helpers/items_helper.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
 module ItemsHelper
-  def stacks_url_full_size(druid, file_name)
-    "#{Settings.stacks_file_url}/#{druid}/#{ERB::Util.url_encode(file_name)}"
+  def stacks_url_full_size(druid, filename)
+    # Allow literal slashes in the filename (do not encode them)
+    encoded_filename = filename.split('/').map { |filename_part| ERB::Util.url_encode(filename_part) }.join('/')
```
mjgiarlo, line 6: The same approach is already used in [sul-embed](https://github.com/sul-dlss/sul-embed/pull/1376/files#diff-bbb9f89625df7768bb24af8d2af0b559798c229837899cbaf3c67d7091c506feR39)

spec/helpers/items_helper_spec.rb
```diff
@@ -6,7 +6,7 @@
   describe '#stacks_url_full_size' do
     subject { helper.stacks_url_full_size('druid:999', 'foo/bar.txt') }
 
-    it { is_expected.to eq 'https://stacks-test.stanford.edu/file/druid:999/foo%2Fbar.txt' }
+    it { is_expected.to eq 'https://stacks-test.stanford.edu/file/druid:999/foo/bar.txt' }
```
mjgiarlo, line 9: I can't explain the coverage dings given this is a whole-ass test for the helper.

justinlittman (approved):

PR #4437, opened 2024-05-13
show workflow context in workflow view

### LINKED ISSUE #4436: Show workflow context in Argo workflow display modal
It would be useful to see the workflow context in this view in Argo (when available), particular for troubleshooting OCR workflows:

Image

Can be fetched with the dor-workflow-client via the `context` method on any process name for a given workflow

peetucket:
# Why was this change made?

Fixes #4436 - show workflow context in workflow detail view (if it's there).

Here is what it would look like if there is context:

Screenshot 2024-05-13 at 2 24 11 PM

And no context:

Screenshot 2024-05-13 at 2 24 32 PM

# How was this change tested?

Localhost

### DIFF
--- a/app/controllers/workflows_controller.rb
+++ b/app/controllers/workflows_controller.rb
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
 class WorkflowsController < ApplicationController
-  # Called from "Add Workflow" button. This is content for a modal invoked via XHR
-  # so we don't want a layout.
+  ##
+  # Renders a view with process-level state information for a given object's workflow.
+  #
+  # @option params [String] `:item_id` The druid for the object.
+  # @option params [String] `:id` The workflow name. e.g., accessionWF.
   def show
     workflow = WorkflowClientFactory.build.workflow(pid: params[:item_id], workflow_name: params[:id])
     respond_to do |format|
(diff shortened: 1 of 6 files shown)


app/controllers/workflows_controller.rb
```diff
@@ -57,7 +51,13 @@ def create
...
 
-  # add a workflow to an object if the workflow is not present in the active table
+  ##
+  # Updates the status of a specific workflow process step to a given status.
+  #
+  # @option params [String] `:item_id` The druid for the object.
+  # @option params [String] `:id` The workflow name. e.g., accessionWF.
+  # @option params [String] `:process` The workflow step. e.g., publish.
+  # @option params [String] `:status` The status to which we want to reset the workflow.
   def update
```
peetucket, line 61: all changes above are just moving the comments around since I think they were all in the wrong place for some reason

edsu (approved):
It's kind of bleh display to just dump stringified ruby in the interface, but ok.

peetucket:
> It's kind of bleh display to just dump stringified ruby in the interface, but ok.

Agreed, this is definitely developer focused for debugging.  it does have the advantage of being compact as opposed to something like a table (which would be pretty easy to do).  I figured at some point we could display more nicely on the primary item detail page itself it it proved useful (like near where the workflow list is).

PR #4454, opened 2024-05-24
add okcomputer check for ocrWF queue size

peetucket:
# Why was this change made?

Fixes sul-dlss/common-accessioning#1242

This monitors the size of the ocrWF:ocr-create queue (by looking at the number of objects waiting). If this gets too big, it likely means ABBYY has stopped processing.  The check is fast (hits solr) and simple (has no dependency other than what is on argo).  

Downsides: it doesn't tell you why the queue is big -- this could also mean the robots are not picking up items, the folder listener is down etc., but we have monitoring in place for other possibilities, so this adds an extra check for something outside our control (ABBYY itself).

~~HOLD until we discuss potential other solutions~~

# How was this change tested?

Localhost

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -33,6 +33,22 @@ class TablesHaveDataCheck < OkComputer::Check
   end
 end
 
+class OcrQueueDepthCheck < OkComputer::Check
+  MAX_OCR_WAITING = 20 # alert if we have more than this number of objects waiting in ocrWF:ocr-create
+
+  def check
+    num_ocr_waiting = SearchService.query('wf_wps_ssim:ocrWF:ocr-create:waiting', fl: 'id')['response']['numFound']
+
+    msg = "ocrWF:ocr-create step has #{num_ocr_waiting} waiting"
+    if num_ocr_waiting > MAX_OCR_WAITING
+      mark_failure
+      msg += " (more than #{MAX_OCR_WAITING})"
+    end
+
+    mark_message msg
+  end
+end
+
 # REQUIRED checks, required to pass for /status/all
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -33,6 +33,22 @@ def table_check(klass)
   end
 end
 
+class OcrQueueDepthCheck < OkComputer::Check
+  MAX_OCR_WAITING = 20 # alert if we have more than this number of objects waiting in ocrWF:ocr-create
+
+  def check
+    num_ocr_waiting = SearchService.query('wf_wps_ssim:ocrWF:ocr-create:waiting', fl: 'id')['response']['numFound']
```
edsu, line 40: I guess there's not a way to get this from Workflow Service?

peetucket: we could, but i think the local solr query is faster than making an API call to an external service

edsu: Ok. I don't think speed is an issue here, as much as using the microservice architecture that has been built? But that's your call.

I can't decide whether Solr as a scratch space for everything is magic I should be grateful for or an anti-pattern to be avoided at all costs. I kind of lean towards the latter.

jcoyne: I'm with @edsu on this one.  This should be a db query in WFS.

peetucket: okcomputer check via db query in WFS makes sense ... i'll do that instead

peetucket:
Closing in favor of check in WFS

peetucket:
new PR: https://github.com/sul-dlss/workflow-server-rails/pull/774

PR #4457, opened 2024-05-29
add ocr modal and bulk job to argo

### LINKED ISSUE #4423: Add OCR action/modal on item detail page (argo)
Allow user to start OCR for a single object in Argo.

Users need to provide some OCR related information, which will set workflow variables and/or update cocina.

See https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.dphvindpiqxy

Only allow OCR to be started if it should be possible (i.e. limit by object type and object content type)

(Workflow list: https://github.com/sul-dlss/argo/blob/main/app/helpers/apo_helper.rb)

But started OCR will not just require starting `ocrWF`, it will also require setting the appropriate workflow variables, as shown in pre-assembly here: https://github.com/sul-dlss/pre-assembly/blob/5cc6ad439e3447408cdee37bc0aafae57158b7f4/app/services/start_accession.rb#L9-L16

~~One of the things that will need to be selectable/recordable is the language selection a la preassembly.  This will probably go in a workflow context when ocrWF is applied?~~

~~If there's no language selector, this might just be adding `ocrWF` to the list of apply-able workflows in Argo (using existing Add Workflow UI).~~

### LINKED ISSUE #4449: Add new OCR bulk action (argo)
Bulk action to initialize OCR for a set of objects.

Users need to provide some OCR related information, which will set workflow variables and/or update cocina.

See https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.bafslbs3xknn

Only allow OCR to be started if it should be possible (i.e. limit by object type and object content type)

(Workflow list: https://github.com/sul-dlss/argo/blob/main/app/helpers/apo_helper.rb)

But started OCR will not just require starting `ocrWF`, it will also require setting the appropriate workflow variables, as shown in pre-assembly here: https://github.com/sul-dlss/pre-assembly/blob/5cc6ad439e3447408cdee37bc0aafae57158b7f4/app/services/start_accession.rb#L9-L16

### LINKED ISSUE #4469: Argo text extraction modal and bulk job should work on open versions
See discussion https://stanfordlib.slack.com/archives/C06EVDJV45N/p1717692358883249

It will become ok for the `ocrWF:start-ocr` to operate on open versions, so we should not disable the text extraction button or block text extraction jobs from running on open versions.

Currently the work to edit is on an open PR: https://github.com/sul-dlss/argo/pull/4457

### LINKED ISSUE #4471: Disable "Text extraction" button during accessioning
As implemented in https://github.com/sul-dlss/argo/pull/4457, the "Text Extraction" button is always available.

It should be disabled while accessionWF is running (or rather, the “In Accessioning” status, however that is determined).

See also #4472

Blocked by #4457

dnoneill:
# Why was this change made?
closes #4423 and closes #4449 and closes #4469 and closes #4471

This adds a button to OCR that allows text extraction along with a form that provides the user with the ability to OCR.
It also adds a bulk job that performs the same actions

It is enabled via settings.yml `Settings.features.ocr_workflow` (already set in QA)

Note that to avoid undo code complexity, the bulk action job is still visible in the menu even if disabled in the settings, but if selected, it will tell you it's not available and not let you start it.  The button the item detail page will be invisible though.

Todo:
- [x] add test for create in the text_extraction controller
- [x] add language/warnings to modal - see https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.dphvindpiqxy

~~HOLD because it probably makes sense to bundle this with bulk action work, since the logic for what to do to start the text extraction workflow is the same and should be centralized~~

~~HOLD for testing on QA~~

~~HOLD for PO review~~

Note that the bulk action form shares some common elements with all other jobs (the druid entry box and the job description), so placement of text around/in these areas is restricted a bit.

**Bulk action view:**

Screenshot 2024-06-05 at 11 17 26 AM

**Single item model view:**

Screenshot 2024-06-05 at 11 17 49 AM

# How was this change tested?
QA + locally + rspec tests


app/components/show/controls_component.rb
```diff
@@ -83,6 +83,32 @@ def create_embargo
...
     end
 
+    def create_text_extraction
+      return unless Settings.features.ocr_workflow
+      return unless ocrable? && !registered_only?
+
+      render ActionButton.new url: new_item_text_extraction_path(druid),
+                              label: 'Text extraction',
+                              open_modal: true,
+                              disabled: !open? && openable?
```
dnoneill, line 93: Not totally sure about this. I think this is what we want. If it is open? we don't want to allow but I am having trouble understanding what open means.

peetucket: I think we want just `!openable?` here and I pushed a commit to change this.  Basically, the button is disabled if the object cannot be opened (which could be because it is already open and just sitting or because it happens to be  in the middle of being accessioning after it was recently closed).

app/helpers/text_extraction_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+module TextExtractionHelper
+end
```
peetucket, line 4: probably don't need to commit this file it it's empty?

peetucket:
Added a line of text, warnings from the google doc (dependent on content type of the object) and changed the name of the submit button to make it more clear: 

Screenshot 2024-06-03 at 2 32 42 PM

peetucket:
his is an image object which renders the button and should successfully start ocrWF in a new version of the object : https://argo-qa.stanford.edu/view/druid:sj104cm1759    

This is a book object which renders the button but since the book has no OCRable files, it should just skip all ocrWF steps after they start and not open a version (a separate common accessioning PR I am testing):  https://argo-qa.stanford.edu/view/druid:bv915yp9350    

This is a file object, which does not render the button at all: https://argo-qa.stanford.edu/view/druid:gz167yz5017

This is an image object which renders the button but which is disabled since the object is not in a state where OCR can be started: https://argo-qa.stanford.edu/view/druid:mq536zc6278

app/controllers/workflows_controller.rb
```diff
@@ -86,13 +86,6 @@ def history
 
   delegate :can_update_workflow?, to: :current_ability
 
-  # Fetches the workflow from the workflow service and checks to see if it's active
-  def workflow_active?(wf_name, druid, version)
-    client = WorkflowClientFactory.build
-    workflow = client.workflow(pid: druid, workflow_name: wf_name)
-    workflow.active_for?(version:)
-  end
-
```
peetucket, line 95: this method was added to the existing `WorkflowService` class for easy testing and re-use elsewhere

app/jobs/add_workflow_job.rb
```diff
@@ -17,25 +17,13 @@ def perform(bulk_action_id, params)
...
-
-  # Fetches the workflow from the workflow service and checks to see if it's active
-  def workflow_active?(druid, version)
-    workflow = client.workflow(pid: druid, workflow_name:)
-    workflow.active_for?(version:)
-  end
-
-  def client
-    @client ||= WorkflowClientFactory.build
-  end
```
peetucket, line 40: we can use our new `WorkflowService` method instead of duplicating this code

app/services/text_extraction.rb
```diff
@@ -0,0 +1,51 @@
...
+      'https://cocina.sul.stanford.edu/models/document' => 'ocrWF',
+      'https://cocina.sul.stanford.edu/models/image' => 'ocrWF'
+    }
+  end
+
+  # the workflow context to set
+  def context
+    { runOCR: true, manuallyCorrectedOCR: false, ocrLanguages: languages }
+  end
+end
```
peetucket, line 51: This class is responsible for determining if text extraction is possible and for how to start it.  It is used in both the bulk action job and the single action modal.

app/assets/stylesheets/text-extraction.scss
```diff
@@ -0,0 +1,53 @@
...
+    cursor: pointer;
+  }
+
+  label:hover {
+    background-color: #ddd;
+  }
+}
+
+.text-cardinal {
+  color: $color-cardinal-red;
```
jcoyne, line 37:
```suggestion
  color: var(--bs-primary);
```

After this change is made, can this just be a `css` file (e.g. not a scss)?

jcoyne (changes requested):

app/helpers/application_helper.rb
```diff
@@ -48,4 +48,8 @@ def workflow_grid_view?
   def profile_view?
     current_page?(search_profile_path)
   end
+
+  def avalaible_ocr_languages
```
jcoyne, line 52: Can you put this in a component? That way it isn't in the global view scope.  Also, typo on the method name.

jcoyne (changes requested):

app/javascript/application.js
```diff
@@ -1,16 +1,16 @@
 // Entry point for the build script in your package.json
 
-import "jquery"; // Blacklight 7 needs jQuery. Remove when we upgrade to Blacklight 8?
-import bootstrap from "bootstrap/dist/js/bootstrap";
-global.bootstrap = bootstrap; // Required for Blacklight 7 so it can manage the modals
+import 'jquery' // Blacklight 7 needs jQuery. Remove when we upgrade to Blacklight 8?
```
jcoyne, line 3: JSLint prefers double quotes.  Why did you change all of these to single?

jcoyne (changes requested):

app/assets/stylesheets/text-extraction.scss
peetucket: Made the code change, though every other file in the stylesheets folder is .scss, so not sure if changing the type of file makes sense (for the sake of consistency)

app/helpers/application_helper.rb
peetucket: It was used in a couple spots (bulk job view and modal), so I made a new component, moved the method there, fixed the typo, and used the component in both spots instead of the partial.

app/javascript/application.js
peetucket: Probably Niqui's editor did this without her noticing.  I changed it back.

alundgard:
Nitpicky, non-blocking style issues:
#### Bold style on "Any uncorrected OCR files will be overwritten" differs between the Item view and the Bulk action view
argo-bulk

#### Sentence case consistency
Item view modal title: "Text Extraction" :arrow_right:  "Text extraction"
Item view modal button: "Start Text Extraction" :arrow_right:  "Start text extraction"
Bulk action menu option: "Run Text Extraction" :arrow_right:  "Run text extraction" (We could change this to "Start text extraction" to be consistent with the item view modal.)

peetucket:
> Nitpicky, non-blocking style issues:
> 
> #### Bold style on "Any uncorrected OCR files will be overwritten" differs between the Item view and the Bulk action view
> argo-bulk
> #### Sentence case consistency
> Item view modal title: "Text Extraction" ➡️ "Text extraction" Item view modal button: "Start Text Extraction" ➡️ "Start text extraction" Bulk action menu option: "Run Text Extraction" ➡️ "Run text extraction" (We could change this to "Start text extraction" to be consistent with the item view modal.)

Styling and sentence cases updated and made consistent.

edsu (approved):
🚀

app/javascript/application.js
peetucket: Huh, maybe the preference changed in newer versions of jslint?

app/javascript/application.js
```diff
@@ -1,16 +1,16 @@
 // Entry point for the build script in your package.json
 
-import "jquery"; // Blacklight 7 needs jQuery. Remove when we upgrade to Blacklight 8?
-import bootstrap from "bootstrap/dist/js/bootstrap";
-global.bootstrap = bootstrap; // Required for Blacklight 7 so it can manage the modals
+import "jquery" // Blacklight 7 needs jQuery. Remove when we upgrade to Blacklight 8?
+import bootstrap from "bootstrap/dist/js/bootstrap" // Required for Blacklight 7 so it can manage the modals
```
jcoyne, line 4: This comment was previously referring to what is now on line 11.

jcoyne (changes requested):

app/javascript/controllers/text_extraction_controller.js
```diff
@@ -0,0 +1,91 @@
...
+    'selectedLanguages', 'languageWarning', 'dropdownContent']
+
+  static values = { languages: Array }
+
+  connect () {
+    this.element.querySelector('form').reset()
+  }
+
+  languageDropdown (event) {
+    const ishidden = Array.from(this.dropdownContentTarget.classList).includes('d-none')
```
jcoyne, line 14:
```suggestion
    const ishidden = this.dropdownContentTarget.classList.contains('d-none')
```

app/javascript/controllers/text_extraction_controller.js
```diff
@@ -0,0 +1,91 @@
...
+    this.dropdownContentTarget.classList.toggle('d-none')
+    this.textExtractionDropdownTarget.querySelector('#caret').innerHTML = `<i class="bi bi-caret-${ishidden ? 'up' : 'down'}">`
+    event.preventDefault()
+  }
+
+  clickOutside (event) {
+    const isshown = !Array.from(this.dropdownContentTarget.classList).includes('d-none')
```
jcoyne, line 21:
```suggestion
    const isshown = !this.dropdownContentTarget.classList.contains('d-none')
```

app/javascript/controllers/text_extraction_controller.js
```diff
@@ -0,0 +1,91 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static targets = ['textExtractionLanguages', 'textExtractionDropdown',
+    'selectedLanguages', 'languageWarning', 'dropdownContent']
+
+  static values = { languages: Array }
+
+  connect () {
+    this.element.querySelector('form').reset()
```
jcoyne, line 10: Can you drop a comment about why the form needs to be reset when the component initializes?

app/javascript/controllers/text_extraction_controller.js
```diff
@@ -0,0 +1,91 @@
...
+
+    if (this.languagesValue.length > 8) {
+      this.languageWarningTarget.classList.remove('d-none')
+    } else {
+      this.languageWarningTarget.classList.add('d-none')
+    }
+  }
+
+  search (event) {
+    console.log('dalfkjdaljfldskj')
```
jcoyne, line 56: Cruft?

jcoyne (changes requested):

app/javascript/controllers/text_extraction_controller.js
dnoneill: Yeah, if you refresh the page and you previously clicked on checkbox or radio buttons they will remained selected which I assume isn't something you would want.

jcoyne: I mean, can you add a comment into the js file about this.

peetucket:
@jcoyne ok to release this PR?

jcoyne (approved):

PR #4460, opened 2024-05-31
Restrict editing while assembling.

### LINKED ISSUE #4458: Don't allow Argo edits to objects that have an active assemblyWF
When an item has an active assemblyWF, it should be treated the same as if the object is in accessioning. The assemblyWF always hands off directly to the accessionWF, so users should not edit the object because they could potentially change Cocina metadata that assemblyWF has been updating. They can also run into errors because the object may move on the accessionWF before they're done making edits, which will then prevent them from saving those edits.

Currently, Argo allows editing of objects while assemblyWF is active. This means that users can potentially update descriptive, structural, etc. metadata while assemblyWF runs. We should prevent that by making those features inactive and/or hiding them, similar to what we do for items "in accessioning". 

If possible, we should also display a new status like "in assembly" since Argo is currently displaying "Opened" during the assemblyWF. I suspect that people don't realize the assemblyWF is running when they edit objects in this state because the show page makes it look like it's Opened and editable.

justinlittman:
closes #4458

# Why was this change made?
Per @andrewjbtw keep users from getting in trouble.

# How was this change tested?

Unit, integration, Andrew

### DIFF
--- a/app/components/contents_component.html.erb
+++ b/app/components/contents_component.html.erb
@@ -8,7 +8,7 @@
     <div class="accordion-body" data-controller="structural">
       <div class="mb-3">
         <%= render DownloadAllButtonComponent.new(cocina: @cocina, document: @document) %>
-        <% if open? %>
+        <% if open_and_not_assembling? %>
           <%= button_tag class: 'btn btn-link float-end', data: { action: 'click->structural#open' } do %>
             <span class="bi-upload"></span> Upload CSV
           <% end %>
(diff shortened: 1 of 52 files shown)


andrewjbtw:
This looks good to me functionally. I started an accession from Preassembly and couldn't edit it in Argo as soon as assemblyWF started.

Minor stylistic comment: the icon that shows in place of the lock during assemblyWF is different than the one that shows during accessionWF.

assemblyWF:
>  Screenshot 2024-06-10 at 2 37 59 PM

accessionWF:
>  Screenshot 2024-06-10 at 2 38 16 PM

For consistency, I'd like to keep the lock icon as a lock icon while assemblyWF is active.

jcoyne (approved):

PR #4462, opened 2024-05-31
update context workflow to display as table

### LINKED ISSUE #4456: Display better view of workflow context variables
The argo workflow view currently shows any available workflow context as a hash at the bottom of the workflow display. Improve the display (tabular?)

Image

dnoneill:
# Why was this change made?
closes #4456 

Screenshot 2024-05-31 at 1 26 53 PM

# How was this change tested?
Viewing on Argo QA.

### DIFF
--- a/app/views/workflows/_show.html.erb
+++ b/app/views/workflows/_show.html.erb
@@ -30,5 +30,22 @@
   </tbody>
 </table>
 <% if @presenter.workflow_context.present? %>
-  <div id="workflow-context">Context: <%= @presenter.workflow_context %></div>
+  <div id="workflow-context">
+  <table class="table table-bordered">
+    <thead>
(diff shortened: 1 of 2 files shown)


justinlittman:
Will workflow settings always be scalar?

dnoneill:
I checked. Peter pointed to this https://github.com/sul-dlss/dor-workflow-client/blob/main/lib/dor/workflow/response/process.rb#L54-L59

justinlittman:
But that doesn't stop a workflow setting from being an array or another hash or similar?

peetucket:
@alundgard and @andrewjbtw FYI if you want to see how nicer looking workflow context variables look in this PR, see the screenshot

peetucket (approved):

PR #4475, opened 2024-06-11
add sdrGeneratedText and correctedForAccessibility to structural csv

### LINKED ISSUE #4473: Add new text extraction fields to the Argo structural CSV
Users should be able to edit the new text extraction Cocina fields via the Argo structural CSV. The fields should not be required.

I can see pros and cons to exposing these fields to editing but if we don't, any updates to the values would need to go through a request for developer help.

peetucket:
# Why was this change made?

Fixes #4473 - add new OCR cocina attributes to structural CSV download and upload

~~HOLD for test on QA~~

# How was this change tested?

CI and argo QA

### DIFF
--- a/app/services/structure_updater.rb
+++ b/app/services/structure_updater.rb
@@ -76,6 +76,8 @@ class StructureUpdater
       hasMimeType: row['mimetype'],
       use: row['role'],
       languageTag: row['file_language'],
+      sdrGeneratedText: ActiveModel::Type::Boolean.new.cast(row['sdr_generated_text']) || false,
+      correctedForAccessibility: ActiveModel::Type::Boolean.new.cast(row['corrected_for_accessibility']) || false,
       administrative: existing_file.administrative.new(
         publish: row['publish'] == 'yes',
         shelve: row['shelve'] == 'yes',
--- a/app/services/structure_serializer.rb
+++ b/app/services/structure_serializer.rb
@@ -3,7 +3,7 @@
 class StructureSerializer
   HEADERS = %w[druid resource_label resource_type sequence filename file_label publish
                shelve preserve rights_view rights_download rights_location mimetype
-               role file_language].freeze
+               role file_language sdr_generated_text corrected_for_accessibility].freeze
 
   def self.as_csv(druid, structural)
     new(druid, structural).as_csv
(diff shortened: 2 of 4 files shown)


app/services/structure_updater.rb
```diff
@@ -76,6 +76,8 @@ def update_file(existing_file, row)
       hasMimeType: row['mimetype'],
       use: row['role'],
       languageTag: row['file_language'],
+      sdrGeneratedText: ActiveModel::Type::Boolean.new.cast(row['sdr_generated_text']) || false,
+      correctedForAccessibility: ActiveModel::Type::Boolean.new.cast(row['corrected_for_accessibility']) || false,
```
peetucket, line 80: this ensures "true" --> `true`, "false" --> `false` and `nil` --> `false`

edsu (changes requested):
I've tried testing this in QA and it doesn't appear to be working with Preassembly. Here's what I did:

I ran the OCR image integration test to get an Item, and looked that `ocrGeneratedText` was set correctly to `true` for `testfile.xml` and `testfile2.xml`.

Then I used Preassembly to update the `testocr.tiff` file in the item using a file manifest I downloaded from Argo (not selecting to do any OCR processing). I could see that `sdrGeneratedText` was set to `true` for `testocr.xml` in the `file_manifest.csv`.

However when the Preassembly job ran the `sdrGeneratedText` value got flipped from `true` to `false`.  I guess this might be more of a problem in Preassembly perhaps than a problem with this PR?

edsu (approved):

peetucket:
> I've tried testing this in QA and it doesn't appear to be working with Preassembly. Here's what I did:
> 
> I ran the OCR image integration test to get an Item, and looked that `ocrGeneratedText` was set correctly to `true` for `testocr.xml` and `testocr2.xml`.
> 
> Then I used Preassembly to update the `testocr.tiff` file in the item using a file manifest I downloaded from Argo (not selecting to do any OCR processing). I could see that `sdrGeneratedText` was set to `true` for `testocr.xml` in the `file_manifest.csv`.
> 
> However when the Preassembly job ran the `sdrGeneratedText` value got flipped from `true` to `false`. I guess this might be more of a problem in Preassembly perhaps than a problem with this PR?
> 
> Here's the Item's Cocina after I followed these steps: https://argo-qa.stanford.edu/items/druid:zf055ry5227.json

I'm not 100% understanding what may have happened here, but I concur with your assessment, that pre-assembly may have done some amiss more than Argo.  Pre-assembly has some built in assumptions about incoming files...

I guess one question is if you gave it all of the files again in pre-assembly or just the single tiff.  If just the single tiff, I would not expect it to have updated cocina for any existing files you had not supplied.  If it did, that would likely be an issue in pre-assembly.

PR #4477, opened 2024-06-12
[HOLD] Disable text extraction and editing if OCR workflow is running

### LINKED ISSUE #4470: Items should not be editable while workflows are running
Argo should disable any editing (opening/closing/pencil icons) while accessioning is running, i.e. while ocrWF, assemblyWF, accessionWF are running.  This prevents you from making changes to an object that is about to be closed, which could introduce problems with you editing files, etc. from under running processes or causing race conditions.

May be already in progress in versioning workcycle check with other team.

see https://github.com/sul-dlss/argo/pull/4460 and https://github.com/sul-dlss/argo/pull/4458

### LINKED ISSUE #4472: Disable "Text extraction" if active ocrWF
As implemented in https://github.com/sul-dlss/argo/pull/4457, the "Text Extraction" button is always available.

It should be disabled if there is an active ocrWF in the current version.

See also #4471 

Blocked by #4457

peetucket:
# Why was this change made?

Fixes #4470 and fixes #4472

Disable text extraction button and editing editing if a text extraction workflow is running, similar to if assembly is running.

HOLD for https://github.com/sul-dlss/dor-services-app/pull/5080 for the DSA related change

# How was this change tested?

CI (fix the build)
and QA (eventually)

### DIFF
--- a/app/components/contents_component.html.erb
+++ b/app/components/contents_component.html.erb
@@ -8,7 +8,7 @@
     <div class="accordion-body" data-controller="structural">
       <div class="mb-3">
         <%= render DownloadAllButtonComponent.new(cocina: @cocina, document: @document) %>
-        <% if open_and_not_assembling? %>
+        <% if open_and_not_processing? %>
           <%= button_tag class: 'btn btn-link float-end', data: { action: 'click->structural#open' } do %>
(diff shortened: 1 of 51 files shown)


peetucket:
Closing in favor of a much smaller change in Argo to come.

PR #4478, opened 2024-06-13
disable text extraction button during assembling or not open

peetucket:
# Why was this change made?

Fixes https://github.com/sul-dlss/argo/issues/4470 and fixes https://github.com/sul-dlss/argo/issues/4472

1. Disable text extraction button and editing editing if a text extraction workflow is running, similar to if assembly is running.
2. Disable text extraction button unless object is open, the same as other actions which edit an object.

HOLD for https://github.com/sul-dlss/dor-services-app/pull/5083 for the DSA related change

NOTE: due to the issue fixed here: https://github.com/sul-dlss/common-accessioning/pull/1290 most existing ocrWF objects in QA will have a mismatched version between SDR and workflow service which leads to the incorrect "assembling?" status being reported.  This means the "Text Extraction" button in Argo will be enabled when it shouldn't be.  After the common accessioning is merged, the behavior of this button on new objects will be as described here.

# How was this change tested?

CI

### DIFF
--- a/app/components/show/controls_component.rb
+++ b/app/components/show/controls_component.rb
@@ -43,7 +43,7 @@ module Show
     end
 
     delegate :admin_policy?, :agreement?, :item?, :collection?, :embargoed?, to: :doc
-    delegate :open?, :openable?, :open_and_not_assembling?, :assembling?, to: :presenter
+    delegate :open?, :openable?, :open_and_not_assembling?, to: :presenter
 
     def button_disabled?
       !open_and_not_assembling?
@@ -90,7 +90,7 @@ module Show
       render ActionButton.new url: new_item_text_extraction_path(druid),
                               label: 'Text extraction',
                               open_modal: true,
-                              disabled: button_disabled? || in_accessioning?
+                              disabled: in_accessioning? || button_disabled?
     end
 
     def edit_apo
--- a/app/components/show/controls_component.rb
+++ b/app/components/show/controls_component.rb
@@ -90,7 +90,7 @@ module Show
       render ActionButton.new url: new_item_text_extraction_path(druid),
                               label: 'Text extraction',
                               open_modal: true,
-                              disabled: in_accessioning? || workflow_errors? || assembling?
+                              disabled: button_disabled? || in_accessioning?
     end
 
     def edit_apo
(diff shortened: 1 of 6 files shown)


app/components/show/controls_component.rb
```diff
@@ -90,7 +90,7 @@ def create_text_extraction
       render ActionButton.new url: new_item_text_extraction_path(druid),
                               label: 'Text extraction',
                               open_modal: true,
-                              disabled: in_accessioning? || workflow_errors?
+                              disabled: in_accessioning? || workflow_errors? || assembling?
     end
```
peetucket, line 94: the actual change (this now includes ocrWF check in sul-dlss/dor-services-app#5083)

everything else is spec changes

app/components/show/controls_component.rb
```diff
@@ -157,10 +157,6 @@ def in_accessioning?
       doc['processing_status_text_ssi']&.include?('In accessioning')
     end
 
-    def workflow_errors?
-      doc[SolrDocument::FIELD_WORKFLOW_ERRORS].present?
-    end
-
```
peetucket, line 163: added for this button earlier, but we don't need it now that we have incorporated ocrWF into the existing check for assembling

app/controllers/text_extraction_controller.rb
```diff
@@ -6,6 +6,8 @@ class TextExtractionController < ApplicationController
   def new; end
 
   def create
+    return redirect_to solr_document_path(@cocina_object.externalIdentifier), flash: { error: 'Object is not open' } unless VersionService.open?(druid: @cocina_object.externalIdentifier)
+
```
peetucket, line 10: we disable the button if not open, this just guards against on the server side too

spec/components/show/controls_component_spec.rb
```diff
@@ -294,28 +294,4 @@
...
-    end
-
-    context 'when workflow errors do not exist' do
-      let(:doc) do
-        SolrDocument.new(id: item_id)
-      end
-
-      it { is_expected.to be false }
-    end
-  end
```
peetucket, line 320: this method is now gone

edsu (approved):

PR #4480, opened 2024-06-14
text extraction job needs to determine if object is opened or not

peetucket:
# Why was this change made?

We need this for the same reason we need https://github.com/sul-dlss/common-accessioning/pull/1290 

If you start a text extraction workflow like ocrWF that can operate on both currently closed and currently opened versions, you need to know what the state is.  If the object is *not* opened yet, you need to start ocrWF with the next version up (since this is what it will be once opened by `ocrWF:start-ocr`).  If already opened, the current version is correct.

Argo can handle both situations ("text extraction" button in UI only works on already open objects but the job can handle either situation).  This allows for both to work.

# How was this change tested?

CI

### DIFF
--- a/app/jobs/text_extraction_job.rb
+++ b/app/jobs/text_extraction_job.rb
@@ -16,11 +16,13 @@ class TextExtractionJob < GenericJob
     with_items(params[:druids], name: 'Start Text Extraction') do |cocina_object, success, failure|
       next failure.call('Not authorized') unless ability.can?(:update, cocina_object)
 
-      text_extraction = TextExtraction.new(cocina_object, languages: params[:text_extraction_languages])
+      version_service = VersionService.new(druid: cocina_object.externalIdentifier)
+
+      text_extraction = TextExtraction.new(cocina_object, languages: params[:text_extraction_languages], already_opened: version_service.open?)
 
       next failure.call('Text extraction is not possible for this object') unless text_extraction.possible?
 
-      next failure.call("#{text_extraction.wf_name} already exists for this version") if WorkflowService.workflow_active?(druid: cocina_object.externalIdentifier, version: cocina_object.version, wf_name: text_extraction.wf_name)
+      next failure.call('Object is currently assembling') if version_service.assembling?
 
       text_extraction.start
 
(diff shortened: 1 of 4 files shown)


app/jobs/text_extraction_job.rb
```diff
@@ -16,11 +16,13 @@ def perform(bulk_action_id, params)
...
 
-      text_extraction = TextExtraction.new(cocina_object, languages: params[:text_extraction_languages])
+      version_service = VersionService.new(druid: cocina_object.externalIdentifier)
+
+      text_extraction = TextExtraction.new(cocina_object, languages: params[:text_extraction_languages], already_opened: version_service.open?)
 
       next failure.call('Text extraction is not possible for this object') unless text_extraction.possible?
 
-      next failure.call("#{text_extraction.wf_name} already exists for this version") if WorkflowService.workflow_active?(druid: cocina_object.externalIdentifier, version: cocina_object.version, wf_name: text_extraction.wf_name)
+      next failure.call('Object is currently assembling') if version_service.assembling?
```
peetucket, line 26: this check can now be simpler because of https://github.com/sul-dlss/dor-services-app/pull/5083  ... `assembling?` will now indicate if there is an active text extraction workflow

app/services/text_extraction.rb
```diff
@@ -2,23 +2,30 @@
...
 class TextExtraction
-  attr_reader :cocina_object, :languages
+  attr_reader :cocina_object, :languages, :already_opened
 
-  # param [Cocina::Models::DRO] cocina_object the object to start text extraction for
+  # @param [Cocina::Models::DRO] cocina_object the object to start text extraction for
   # @param [Array<String>] languages the languages to extract text for, default to empty
-  def initialize(cocina_object, languages: [])
+  # @param [Boolean] already_opened whether the object has already been opened, default to true
+  def initialize(cocina_object, languages: [], already_opened: true)
```
peetucket, line 10: default to `true` ... this is what the use of the `Text Extraction` button in the UI requires (only possible for open objects like all of the other actions)

edsu (approved):

PR #4494, opened 2024-06-27
Stub out initial user version detail page.

### LINKED ISSUE #4491: Stub user version detail page
1. Add a user version select to the item detail page.
  * The select only shows when the item has user versions and if the item only has a single user version that user version does not correspond with the current object version (i.e., don't provide a user version view of the current is that user version).
  * When the user selects a user version, the user version details page should show.
  * The user version select options should be formatted "Public version X" where X is the version number.
  * The placement of this select doesn't matter.
2. Stub the user version detail page.
  * The page should display the user version number.
  * The page should have a user version select.
  * The page should have a "Back to latest" button that returns to the item detail page.
  * The formatting of this page doesn't matter.

A feature flag isn't required. (There are no user versions in production.)

justinlittman:
closes #4491

# Why was this change made?
Support additional work on version detail page.

# How was this change tested?
Unit, QA

### DIFF
--- a/app/components/version_milestones_component.rb
+++ b/app/components/version_milestones_component.rb
@@ -1,11 +1,28 @@
 # frozen_string_literal: true
 
 class VersionMilestonesComponent < ViewComponent::Base
-  def initialize(version:, title:, steps:)
+  def initialize(version:, milestones_presenter:)
     @version = version
-    @title = title
-    @steps = steps
+    @milestones_presenter = milestones_presenter
   end
 
-  attr_reader :version, :title, :steps
+  def title
+    @title ||= milestones_presenter.version_title(version)
+  end
+
+  def steps
+    @steps ||= milestones_presenter.steps_for(version)
+  end
+
+  def user_version
+    @user_version ||= milestones_presenter.user_version_for(version)
+  end
+
+  def user_version_path
+    item_user_version_path(item_id: druid, id: user_version)
+  end
+
+  attr_reader :version, :milestones_presenter
+
+  delegate :druid, to: :milestones_presenter
 end
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -353,11 +353,15 @@ class CatalogController < ApplicationController
     render partial: 'lazy_wps_workflow_facet'
   end
 
+  # rubocop:disable Metrics/PerceivedComplexity
+  # rubocop:disable Metrics/CyclomaticComplexity
   def show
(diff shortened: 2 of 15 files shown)


justinlittman:
image

image

app/components/version_milestones_component.rb
```diff
@@ -1,11 +1,28 @@
 # frozen_string_literal: true
 
 class VersionMilestonesComponent < ViewComponent::Base
-  def initialize(version:, title:, steps:)
+  def initialize(version:, milestones_presenter:)
     @version = version
```
mjgiarlo, line 5: Is this instance variable no longer used? Seems like we use the version already set on the presenter currently, if I'm reading this correctly.

mjgiarlo: Oh, nevermind. I see it's an attr_reader below!

app/controllers/catalog_controller.rb
```diff
@@ -353,29 +353,36 @@ def lazy_wps_workflow_facet
...
+    @workflows = WorkflowService.workflows_for(druid: @druid)
+    milestones = MilestoneService.milestones_for(druid: @druid)
+    object_client = Dor::Services::Client.object(@druid)
     versions = object_client.version.inventory
+
     user_versions = object_client.user_version.inventory
-    @milestones_presenter = MilestonesPresenter.new(milestones:, versions:, user_versions:)
+    raise ActionController::RoutingError, 'Not Found' if @user_version && user_versions.none? { |version| version.userVersion.to_s == @user_version }
+
+    @milestones_presenter = MilestonesPresenter.new(milestones:, versions:, user_versions:, druid: @druid)
```
mjgiarlo, line 380: Take this suggestion or leave it: it looks like this method is now significantly more complex, and the milestones presenter is still on the slim side. If you were interested in re-balancing the complexity, I think you could shovel much of lines 372-377 into the presenter itself, who can take more responsibility for its own setup, given `@druid` and `@user_version`?

app/components/version_milestones_component.rb
justinlittman: Version is the system version. User version is from the presenter. Sigh.

app/controllers/catalog_controller.rb
justinlittman: That makes sense. Would you be OK if I did this in a follow-up PR? I want to unblock some of the additional work.

mjgiarlo: @justinlittman 💯

mjgiarlo (approved):

PR #4504, opened 2024-06-27
Move some initialization from CatalogController to MilestonesPresenter.

justinlittman:
# Why was this change made?

# How was this change tested?

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -353,8 +353,6 @@ class CatalogController < ApplicationController
     render partial: 'lazy_wps_workflow_facet'
   end
 
-  # rubocop:disable Metrics/PerceivedComplexity
-  # rubocop:disable Metrics/CyclomaticComplexity
   def show
     # If showing a user version, druid will be :item_id and user_version will be :id.
     @druid = Druid.new(params[:item_id] || params[:id]).with_namespace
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -370,22 +370,15 @@ def show
     authorize! :read, @cocina
 
     @workflows = WorkflowService.workflows_for(druid: @druid)
-    milestones = MilestoneService.milestones_for(druid: @druid)
-    object_client = Dor::Services::Client.object(@druid)
-    versions = object_client.version.inventory
 
     user_versions = object_client.user_version.inventory
     raise ActionController::RoutingError, 'Not Found' if @user_version && user_versions.none? { |version| version.userVersion.to_s == @user_version }
```
mjgiarlo, line 375: Looking good. I was wondering about also doing this (plus the suggestion below):

```suggestion
    @milestones_presenter = MilestonesPresenter.new(user_version: @user_version, druid: @druid)
    raise ActionController::RoutingError, 'Not Found' if @milestones_presenter.valid_user_version?
```

app/presenters/milestones_presenter.rb
```diff
@@ -2,13 +2,9 @@
...
 class MilestonesPresenter
-  # @param [Hash<String,Hash>] milestones the milestone data
-  # @param [Array<Dor::Services::Client::ObjectVersion::Version>] versions the version tag data
   # @param [Array<Dor::Services::Client::UserVersion::UserVersion>] user_versions the user versions data
   # @param [String] druid the druid of the object
-  def initialize(milestones:, versions:, user_versions:, druid:)
-    @milestones = milestones
-    @versions = versions
+  def initialize(user_versions:, druid:)
     @user_versions = user_versions
```
mjgiarlo, line 8: Building on what I wrote above:

```suggestion
  def initialize(user_version:, druid:)
    @user_version = user_version
```

app/controllers/catalog_controller.rb
justinlittman: Sneaky.

app/presenters/milestones_presenter.rb
```diff
@@ -36,5 +32,13 @@ def user_version_for(version)
 
   private
 
-  attr_reader :milestones, :versions, :user_versions
+  attr_reader :user_versions
+
```
mjgiarlo, line 36:
```suggestion
  attr_reader :user_version

  def valid_user_version?
    @user_version && versions.any? { |version| version.userVersion.to_s == user_version }
  end
```

mjgiarlo (approved):
The suggestions may have come through wonky, but either way this is an improvement and I approve. Consider the other changes or not, no biggie.

jcoyne:
This goes contrary to the presenter pattern. Because presenters should not make API calls, they are a data object for presentation. Maybe you should rename it?

PR #4513, opened 2024-07-01
Change description button on item show

### LINKED ISSUE #4495: Display descriptive metadata for user version
On the user version details page, the MODS button should display the descriptive metadata for the user version.

image

lwrubel:
# Why was this change made?
Resolves #4495 to show user version descriptive metadata. Changes the "MODS" button to "Description". 

Screenshot 2024-07-02 at 10 29 28 AM

# How was this change tested?
Unit and stage.

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -15,6 +15,10 @@ class MetadataController < ApplicationController
   private
 
   def cocina
-    Repository.find(params[:item_id])
+    if params.key?(:user_version_id)
+      Repository.find_user_version(params[:item_id], params[:user_version_id])
+    else
+      Repository.find(params[:item_id])
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/metadata_controller.rb
```diff
@@ -15,6 +15,10 @@ def descriptive
   private
 
   def cocina
-    Repository.find(params[:item_id])
+    if params.key?(:user_version_id)
+      Repository.find_user_version(params[:item_id], params[:user_version_id])
```
lwrubel, line 19: I'm not sure why this is showing as not covered in tests. Is there something I could change in the spec below?

mjgiarlo: I think you'd need a request spec to cover this change, like https://github.com/sul-dlss/argo/pull/4514/files#diff-723822a1f53de8c8a1789d62ffa3b5c2b773e72b24881c5faf229189f1864a19

lwrubel: Thanks, I'll work on that.

mjgiarlo (approved):

PR #4514, opened 2024-07-01
Make Cocina CSV download path user version-aware

### LINKED ISSUE #4498: Download cocina spreadsheet should download cocina for user version
When on a user version details page, the download cocina spreadsheet button should download the cocina for the user version:

image

mjgiarlo:
# Why was this change made?

Fixes #4498

# How was this change tested?

CI, local

### DIFF
--- a/app/controllers/descriptives_controller.rb
+++ b/app/controllers/descriptives_controller.rb
@@ -2,15 +2,13 @@
 
 # Download the CSV descriptive metadata
 class DescriptivesController < ApplicationController
+  before_action :maybe_find_user_version, only: :show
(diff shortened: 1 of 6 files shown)


app/controllers/descriptives_controller.rb
```diff
@@ -2,15 +2,13 @@
 
 # Download the CSV descriptive metadata
 class DescriptivesController < ApplicationController
+  before_action :maybe_find_user_version, only: :show
```
mjgiarlo, line 5: Decided on this approach due to https://github.com/CanCanCommunity/cancancan/blob/develop/docs/changing_defaults.md#override-loading, but if we already have a pattern we like better, I'd be happy to follow it.

justinlittman (approved):

PR #4516, opened 2024-07-02
Update structural display for user_versions

### LINKED ISSUE #4500: Display files for user version
On a user version detail page, show the files for the current user version. This is in both the content and file hierarchy sections. Also, the files should link to the appropriate versions.

image

aaron-collier:
# Why was this change made?

Fixes #4500 

# How was this change tested?

### DIFF
--- a/app/components/contents_component.rb
+++ b/app/components/contents_component.rb
@@ -23,6 +23,6 @@ class ContentsComponent < ApplicationComponent
   end
 
   def structural_link_path
-    user_version_view? ? item_user_version_structure_path(@document.id, user_version) : item_structure_path(@view_token)
+    user_version_view? ? item_user_version_structure_path(@view_token, user_version) : item_structure_path(@view_token)
   end
 end
--- a/app/components/contents_component.rb
+++ b/app/components/contents_component.rb
@@ -12,7 +12,7 @@ class ContentsComponent < ApplicationComponent
     @cocina.respond_to?(:structural)
   end
 
-  delegate :open_and_not_assembling?, :user_version_view?, to: :@presenter
+  delegate :open_and_not_assembling?, :user_version_view?, :user_version, to: :@presenter
 
   def upload_csv?
     !user_version_view? && open_and_not_assembling?
@@ -20,5 +20,9 @@ class ContentsComponent < ApplicationComponent
 
   def download_csv?
     !user_version_view?
+  end
+
+  def structural_link_path
+    user_version_view? ? item_user_version_structure_path(@document.id, user_version) : item_structure_path(@view_token)
   end
 end
--- a/app/controllers/structures_controller.rb
+++ b/app/controllers/structures_controller.rb
@@ -16,7 +16,7 @@ class StructuresController < ApplicationController
       end
       format.html do
         # Lazy loading of the structural part of the show page
-        @cocina_item = fetch_cocina
+        @cocina_item = fetch_cocina if decrypted_token
       end
     end
   end
@@ -32,6 +32,8 @@ class StructuresController < ApplicationController
   end
 
   def hierarchy
+    return unless decrypted_token
+
     @cocina_item = fetch_cocina
     @root_directory = FileHierarchyService.to_hierarchy(cocina_object: @cocina_item)
   end
--- a/app/controllers/structures_controller.rb
+++ b/app/controllers/structures_controller.rb
@@ -9,14 +9,14 @@ class StructuresController < ApplicationController
     respond_to do |format|
       format.csv do
         # Download the structural spreadsheet
-        cocina = fetch_cocina(item_id: params[:item_id])
+        cocina = fetch_cocina(item_id: params[:item_id], user_version_id: params[:user_version_id])
         authorize! :update, cocina
         filename = "structure-#{Druid.new(cocina).without_namespace}.csv"
         send_data StructureSerializer.as_csv(cocina.externalIdentifier, cocina.structural), filename:
       end
       format.html do
         # Lazy loading of the structural part of the show page
-        @cocina_item = fetch_cocina(item_id: decrypted_token.fetch(:druid), user_version_id: decrypted_token&.fetch(:user_version_id, nil))
+        @cocina_item = fetch_cocina(item_id: decrypted_token.fetch(:druid), user_version_id: decrypted_token.fetch(:user_version_id, nil))
       end
     end
   end
@@ -32,7 +32,7 @@ class StructuresController < ApplicationController
   end
 
   def hierarchy
-    @cocina_item = fetch_cocina(item_id: decrypted_token.fetch(:druid), user_version_id: decrypted_token&.fetch(:user_version_id, nil))
+    @cocina_item = fetch_cocina(item_id: decrypted_token.fetch(:druid), user_version_id: decrypted_token.fetch(:user_version_id, nil))
     @root_directory = FileHierarchyService.to_hierarchy(cocina_object: @cocina_item)
   end
 
--- a/app/controllers/structures_controller.rb
+++ b/app/controllers/structures_controller.rb
@@ -9,14 +9,14 @@ class StructuresController < ApplicationController
     respond_to do |format|
       format.csv do
         # Download the structural spreadsheet
-        cocina = fetch_cocina
+        cocina = fetch_cocina(item_id: params[:item_id])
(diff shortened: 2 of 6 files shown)


app/controllers/structures_controller.rb
```diff
@@ -51,4 +59,10 @@ def display_error(messages)
   def decrypted_token
     Argo.verifier.verified(params[:item_id], purpose: :view_token)
   end
+
+  def fetch_cocina(item_id:, user_version_id: nil)
```
mjgiarlo, line 63: I wonder if we need these params passed from above, given `#item_id` and `#user_version_id` can be called from `#fetch_cocina` without being passed in.

app/controllers/structures_controller.rb
```diff
@@ -32,12 +32,20 @@ def update
...
 
   private
 
+  def item_id
+    decrypted_token&.fetch(:druid, nil) || params[:item_id]
+  end
+
+  def user_version_id
+    decrypted_token&.fetch(:user_version_id, nil) || params[:user_version_id]
+  end
```
mjgiarlo, line 47: If we combine these two modes (get a value from a token, else get it from params), is there a risk of going around the token-based method? Presumably we had a good reason for getting values from encrypted tokens. If we're already doing this elsewhere and we dismissed this as a risk already, feel free to disregard!

app/controllers/structures_controller.rb
```diff
@@ -9,14 +9,14 @@ def show
...
-        cocina = Repository.find(params[:item_id])
+        cocina = fetch_cocina(item_id:, user_version_id:)
         authorize! :update, cocina
         filename = "structure-#{Druid.new(cocina).without_namespace}.csv"
         send_data StructureSerializer.as_csv(cocina.externalIdentifier, cocina.structural), filename:
       end
       format.html do
         # Lazy loading of the structural part of the show page
-        @cocina_item = Repository.find(decrypted_token.fetch(:druid))
+        @cocina_item = fetch_cocina(item_id:, user_version_id:) # decrypted_token.fetch(:druid)
```
mjgiarlo, line 19: Cruft comment?

app/controllers/structures_controller.rb
aaron-collier: Good catch, I think I had this in place before cleaning up the methods above.

app/controllers/structures_controller.rb
aaron-collier: Good question, I could revert the change at line 12 that uses param for the csv, but I wonder if it should use the token and then remove the use of params.

mjgiarlo: I don't have a good handle on why/how we're using the token here, so maybe?

app/components/contents_component.rb
```diff
@@ -21,4 +21,8 @@ def upload_csv?
   def download_csv?
     !user_version_view?
   end
+
+  def structural_link_path
+    user_version_view? ? item_user_version_structure_path(@document.id, user_version) : item_structure_path(@view_token)
```
justinlittman, line 26: I think @view_token is needed for item_user_version_structure_path, probably the second parameter but including both parameters.

app/controllers/structures_controller.rb
justinlittman: Yeah, I agree with Mike. I think the DRYing allows bypassing the tokens. I think it would be easier to understand if getting cocina from params and tokens were separate (or if `find_cocina` took druid and version parameters).

app/controllers/structures_controller.rb
```diff
@@ -9,14 +9,14 @@ def show
...
-        cocina = Repository.find(params[:item_id])
+        cocina = fetch_cocina(item_id: params[:item_id])
         authorize! :update, cocina
         filename = "structure-#{Druid.new(cocina).without_namespace}.csv"
         send_data StructureSerializer.as_csv(cocina.externalIdentifier, cocina.structural), filename:
       end
       format.html do
         # Lazy loading of the structural part of the show page
-        @cocina_item = Repository.find(decrypted_token.fetch(:druid))
+        @cocina_item = fetch_cocina(item_id: decrypted_token.fetch(:druid), user_version_id: decrypted_token&.fetch(:user_version_id, nil))
```
mjgiarlo, line 19: Do you think we still need the safe navigation in `decrypted_token&.fetch`?

app/controllers/structures_controller.rb
```diff
@@ -32,7 +32,7 @@ def update
   end
 
   def hierarchy
-    @cocina_item = Repository.find(decrypted_token.fetch(:druid))
+    @cocina_item = fetch_cocina(item_id: decrypted_token.fetch(:druid), user_version_id: decrypted_token&.fetch(:user_version_id, nil))
```
mjgiarlo, line 35: Same ❓ as ☝🏻

app/controllers/structures_controller.rb
```diff
@@ -9,14 +9,14 @@ def show
     respond_to do |format|
       format.csv do
         # Download the structural spreadsheet
-        cocina = Repository.find(params[:item_id])
+        cocina = fetch_cocina(item_id: params[:item_id])
```
mjgiarlo, line 12: Maybe this goes beyond the scope of the issue you're working on and verges into a different one, but I'm wondering if this call should also have `, user_version_id: params[:user_version_id]`

mjgiarlo: Yeah, probably beyond the scope since this change will also need a path helper change to use the user_version_aware path?

mjgiarlo: @justinlittman do we need a ticket for ☝🏻 or is this not in the requirements? "This" being user version-aware structure CSV downloads.

justinlittman: I'm not sure I'm following. We already add version-aware routes for other endpoints. Is this what you mean? https://github.com/sul-dlss/argo/blob/main/config/routes.rb#L221-L232

app/controllers/structures_controller.rb
aaron-collier: Oh no, probably not, copy-pasta.

app/controllers/structures_controller.rb
aaron-collier: I think it probably should, and was the crux of the original DRYing, so adding it is a good catch.

justinlittman (approved):

app/components/contents_component.rb
```diff
@@ -21,4 +21,8 @@ def upload_csv?
   def download_csv?
     !user_version_view?
   end
+
+  def structural_link_path
+    user_version_view? ? item_user_version_structure_path(@document.id, @view_token) : item_structure_path(@view_token)
+  end
```
mjgiarlo, line 27: @jcoyne I think you made the initial change a couple years ago where we're passing an encrypted token to path helpers (to allow us to lazy-load turbo frames). At the time, the token was used in place of an item ID. Now we have user version-aware path helpers that require two arguments, an item ID and a user version ID, and that makes me wonder what the args to the path helper ought to be. Can you advise? E.g., should both args be a token? the same token or diff tokens? should just one arg be a token? if so, which one, and why? I'm not sure any of us have a better handle on this than you (but am happy to be wrong about that), and I think we ought to understand it more before we proceed.

jcoyne: The purpose of the token was to make the assurance that the token holder had access to view the resource, not only to be an identifier of the resource.  This allowed us to just show the resource without having to authorize it again.

app/controllers/structures_controller.rb
```diff
@@ -51,4 +51,10 @@ def display_error(messages)
   def decrypted_token
     Argo.verifier.verified(params[:item_id], purpose: :view_token)
```
mjgiarlo, line 52: I wonder if the path above, `item_user_version_structure_path(@document.id, @view_token)`, will work with this line unchanged. This path evaluates to `/items/druid:bb000zn0114/user_versions/skret-t0k3n/structure`, which means in the controller `params[:item_id]` is `druid:...` and `params[:user_version_id]` is the token, so I don't understand how `Argo.verified.verified(params[:item_id], purpose: :view_token)` is verifying a non-token.

mjgiarlo (approved):

app/controllers/structures_controller.rb
mjgiarlo: We changed this.

PR #4517, opened 2024-07-02
add additional header title validation

### LINKED ISSUE #4492: Prevent malformed titles in Cocina spreadsheet uploads
See: https://app.honeybadger.io/projects/50568/faults/108793688

```
"title" => [{
      "structuredValue" => [
        {
          "value" => "Bulletin of the Monterey Audubon Society",
        },
        {
          "type" => "subtitle",
        }
      ],
```

It is currently possible if you have a typo in a column header in your spreadsheet to generate malformed titles. One example is above, where some structured title parts have a value without a type and a type without a value. These spreadsheet uploads succeed, and then items get stuck in accessioning and cannot be indexed (see HB alert above). The one time we saw this happen, it was in a single collection and 130 items were affected, so we hand-remediated the Cocina JSON in DSA prod, but ideally we would be able to flag this earlier. Look into how we might do this before accessioning begins, ideally before Argo accepts the spreadsheet. If you want a sample bad spreadsheet, you can ask Amanda W.

See https://stanfordlib.slack.com/archives/C0EAZ09QF/p1718988097815579

peetucket:
# Why was this change made?

Fixes #4492 - prevent malformed title header columns from being considered valid

~~WIP, as the validation may change~~

~~HOLD for testing on QA~~

# How was this change tested?

- New specs
- Example bad spreadsheet runs ok on stage, but shows this on QA with this PR deployed:

Screenshot 2024-07-02 at 4 23 35 PM


spec/fixtures/files/bulk_upload_descriptive.csv
```diff
@@ -1,2 +1,2 @@
...
....type,event1.note1.value,event2.date1.type,event2.date1.value,form1.source.value,form1.type,form1.value,form2.source.code,form2.type,form2.value,form3.type,form3.value,language1.code,language1.source.code,note1.type,note1.value,note2.type,note2.value,purl,relatedResource1.contributor1.name1.structuredValue1.type,relatedResource1.contributor1.name1.structuredValue1.value,relatedResource1.contributor1.name1.structuredValue2.type,relatedResource1.contributor1.name1.structuredValue2.value,relatedResource1.contributor1.type,relatedResource1.title1.structuredValue1.type,relatedResource1.title1.structuredValue1.value,relatedResource1.title1.structuredValue2.type,relatedResource1.title1.structuredValue2.value,subject1.source.code,subject1.structuredValue1.type,subject1.structuredValue1.value,subject1.structuredValue2.type,subject1.structuredValue2.value,title1.status,title1.value,title2.structuredValue1.type,title2.structuredValue1.value,title2.structuredValue2.structuredValue1.type,title2.structuredValue2.structuredValue1.value,title2.structuredValue2.structuredValue2.type,title2.structuredValue2.structuredValue3.value,title2.structuredValue2.type,title2.structuredValue2.value,title2.type
```
peetucket, line 1: according to the new validation, this existing fixture file was invalid causing other tests to fail.

i made it valid by adding a `title2.structuredValue2.value` column

spec/fixtures/files/bulk_upload_descriptive.csv
```diff
@@ -1,2 +1,2 @@
...
+druid:rb369bd9414,foo:12345,CSt,marcorg,original cataloging agency,organization,marc,790730,creation,iso8601,19900820141050.0,modification,SIRSI,a111,record origin,Converted from MARCXML to MODS version 3.7 using MARC21slim2MODS3-7_SDR_v2-5.xsl (SUL 3.7 version 2.5 20210421; LC Revision 1.140 20200717),name,"Harvey, Robert Gibson",life dates,1951-,primary,person,xx,marccountry,MODS issuance terms,issuance,monographic,publication,1978,MODS resource types,resource type,notated music,marcform,form,print,extent,score (77 leaves) : music ; 28 cm.,und,iso639-2b,statement of responsibility,by Robert Harvey.,thesis,"D.M.A. project - Department of Music, Stanford University.",https://sul-purl-stage.stanford.edu/bb041bm1345,name,Stanford University,name,Department of Music,organization,main title,Projects,part name,D.M.A. Final,lcsh,topic,Concertos (Piano),genre,Scores,primary,"Concerto for piano, strings, brass and percussion!",title,"Concertos, piano, orchestra",name,"Harvey, Robert Gibson",life dates,1951-,name,uniform,
```
mjgiarlo, line 2: It looks like `title2.type` changed from `uniform` to a blank string or nil, and `title2.structuredValue2.value` took the former `title2.type`'s value of `uniform`? Or am I misreading this? 👀

peetucket: nope i goofed, though it turns out not to matter for the tests

but i will fix

app/services/description_validator.rb
```diff
@@ -47,6 +48,19 @@ def validate_title_headers
...
+    # first find all the values of X as array: the structured title numbers (e.g. title1.XXX title2.XXX, etc.)
+    structured_title_numbers = @headers.filter_map { |header| header&.match(/title(\d+).structuredValue\d+\.(type|value)/) }.pluck(1).uniq
+
+    # for each structured title number, find the values for Y (e.g. the value/type number) and ensure they are same for each title
+    structured_title_numbers.each do |title_number|
+      num_structured_title_types = @headers.filter_map { |header| header&.match(/title#{Regexp.quote(title_number)}.structuredValue(\d+)\.type/) }.pluck(1).sort
+      num_structured_title_values = @headers.filter_map { |header| header&.match(/title#{Regexp.quote(title_number)}.structuredValue(\d+)\.value/) }.pluck(1).sort
+      errors << "Mismatched title#{title_number}.structuredValue column: ensure each structured title has a matching type and value column." unless num_structured_title_types == num_structured_title_values
+    end
+  end
```
mjgiarlo, line 62: With the combination of multiple iterations and regexes, I wondered if something like the following would be worth considering:

```suggestion
  # verify that each titleX.structuredValueY.type has a corresponding titleX.structuredValueY.value (where X and Y are any integer).
  def validate_matching_structured_title_headers
    @headers.each do |header|
      next unless header&.match?(/\Atitle(\d+).structuredValue\d+\.(type|value)\z/)

      expected_corresponding_header = header.ends_with?('type') ? header.sub('type', 'value') : header.sub('value', 'type')
      next if @headers.include?(expected_corresponding_header)

      errors << "Unexpected or missing title structuredValue columns: found #{header} but not #{expected_corresponding_header}"
    end
  end
```

This gets the validation down to one full iteration and one regex, and the error message is perhaps a bit more explicit about what it found and what it didn't.

peetucket: I like this approach -- latest commit takes it and just adds a check for blank header column to avoid the safe operator

peetucket:
Not sure what the deal is with the test drop coverage.  Every new line of code is covered.

mjgiarlo (approved):

PR #4518, opened 2024-07-03
Show user version info in Overview and Details sections.

### LINKED ISSUE #4515: Use user version data for overview and details section when displaying user version details page
image

Also, all edit buttons should be removed.

----
Differences from current version show page:

- "Released to": show a message "Older versions are not released"
Screenshot 2024-07-03 at 8 58 50 AM

- Do show "processing status": it will be the "{system version} Accessioned"
- Show the same Project/tags that are show on all versions.

justinlittman:
closes #4515

# Why was this change made?
Per ticket.

# How was this change tested?
Unit

### DIFF
--- a/app/components/show/agreement/details_component.html.erb
+++ b/app/components/show/agreement/details_component.html.erb
@@ -11,7 +11,7 @@
       <tr>
         <th class="col-3" scope="row">Content type</th>
         <td>
-          <%= render Show::ContentTypeComponent.new(document: @solr_document, version_service: @presenter.version_service) %>
+          <%= render Show::ContentTypeComponent.new(document: @solr_document, presenter: @presenter) %>
         </td>
       </tr>
 
@@ -25,7 +25,7 @@
       <tr>
         <th class="col-3" scope="row">Source IDs</th>
         <td>
-          <%= render Show::SourceIdComponent.new(document: @solr_document, version_service: @presenter.version_service) %>
+          <%= render Show::SourceIdComponent.new(document: @solr_document, presenter: @presenter) %>
         </td>
       </tr>
 
@@ -46,7 +46,7 @@
       <tr>
         <th class="col-3" scope="row">Tags</th>
         <td>
-          <%= render Show::TagsComponent.new(document: @presenter.document) %>
+          <%= render Show::TagsComponent.new(document: @presenter.document, presenter: @presenter) %>
         </td>
       </tr>
     </tbody>
--- a/app/components/show/barcode_component.rb
+++ b/app/components/show/barcode_component.rb
@@ -2,16 +2,20 @@
 
 module Show
   class BarcodeComponent < ApplicationComponent
-    def initialize(change_set:, version_service:)
-      @change_set = change_set
-      @version_service = version_service
+    def initialize(presenter:)
+      @presenter = presenter
     end
 
     def barcode
-      @change_set.barcode || 'Not recorded'
+      change_set.barcode || 'Not recorded'
     end
 
-    delegate :open_and_not_assembling?, to: :@version_service
-    delegate :id, to: :@change_set
+    def edit?
+      !user_version_view? && open_and_not_assembling?
+    end
+
+    delegate :version_service, :user_version_view?, :change_set, to: :@presenter
+    delegate :open_and_not_assembling?, to: :version_service
+    delegate :id, to: :change_set
   end
 end
--- a/app/components/show/item/details_component.html.erb
+++ b/app/components/show/item/details_component.html.erb
@@ -11,7 +11,7 @@
       <tr>
         <th class="col-3" scope="row">Content type</th>
         <td>
-          <%= render Show::ContentTypeComponent.new(document: @solr_document, version_service: @presenter.version_service) %>
+          <%= render Show::ContentTypeComponent.new(document: @solr_document, presenter: @presenter) %>
         </td>
       </tr>
 
@@ -25,7 +25,7 @@
       <tr>
         <th class="col-3" scope="row">Source IDs</th>
         <td>
-          <%= render Show::SourceIdComponent.new(document: @solr_document, version_service: @presenter.version_service) %>
+          <%= render Show::SourceIdComponent.new(document: @solr_document, presenter: @presenter) %>
         </td>
       </tr>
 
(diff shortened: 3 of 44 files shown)


app/components/show/agreement/details_component.html.erb
```diff
@@ -11,7 +11,7 @@
       <tr>
         <th class="col-3" scope="row">Content type</th>
         <td>
-          <%= render Show::ContentTypeComponent.new(document: @solr_document, version_service: @presenter.version_service) %>
+          <%= render Show::ContentTypeComponent.new(document: @solr_document, presenter: @presenter) %>
```
justinlittman, line 14: Here and lots of places below: Passing in presenter instead of multiple arguments that are provided by the presenter.

app/components/show/barcode_component.rb
```diff
@@ -2,16 +2,20 @@
...
 
     def barcode
-      @change_set.barcode || 'Not recorded'
+      change_set.barcode || 'Not recorded'
     end
 
-    delegate :open_and_not_assembling?, to: :@version_service
-    delegate :id, to: :@change_set
+    def edit?
+      !user_version_view? && open_and_not_assembling?
```
justinlittman, line 14: Here and many places below: Not showing edit pencil when a user version view.

app/controllers/catalog_controller.rb
```diff
@@ -355,18 +355,24 @@ def lazy_wps_workflow_facet
 
   def show
     # If showing a user version, druid will be :item_id and user_version will be :id.
-    @druid = Druid.new(params[:item_id] || params[:id]).with_namespace
-    @user_version = params.key?(:item_id) ? params[:id] : nil
-    _deprecated_response, @document = search_service.fetch(@druid)
-
-    @cocina = Repository.find_lite(@druid, structural: false)
+    if params.key?(:item_id)
```
justinlittman, line 358: Getting druid, cocina, document, user_version based on whether this is an item detail page or a user version detail page.

app/controllers/catalog_controller.rb
```diff
@@ -355,18 +355,24 @@ def lazy_wps_workflow_facet
...
+      @druid = Druid.new(params[:item_id]).with_namespace
+      @user_version = params[:id]
+      @cocina = Repository.find_user_version(@druid, @user_version)
+      @document = SolrDocument.new(object_client.user_version.solr(@user_version))
```
justinlittman, line 362: DSA is creating the solr document on the fly. It is not added to the solr index.

app/controllers/items_controller.rb
```diff
@@ -243,6 +233,13 @@ def authorize_manage!
     authorize! :update, @cocina
   end
 
+  def build_argo_show_presenter
+    ArgoShowPresenter.new(nil, nil, nil).tap do |presenter|
```
justinlittman, line 237: This is just enough of an ArgoShowPresenter to render the components.

spec/support/selenium_webdriver.rb
```diff
@@ -4,7 +4,7 @@
 
 Capybara.register_driver :headless_chrome do |app|
   options = Selenium::WebDriver::Chrome::Options.new.tap do |opts|
-    opts.add_argument('--headless')
+    opts.add_argument('--headless') unless ENV.key?('NO_HEADLESS')
```
justinlittman, line 7: Makes working on feature tests easier.

app/components/show/barcode_component.rb
mjgiarlo: This branch adds nine identical definitions of `#edit?`. I wonder if this method should move to the presenter. Then `Show::SourceIdComponent#edit?` could look like 

```ruby
def edit?
  presenter.edit? && item?
end
```

❓

app/components/show/item/details_component.html.erb
```diff
@@ -39,7 +39,7 @@
       <tr>
         <th class="col-3" scope="row">Released to</th>
         <td>
-            <%= released_to %>
+            <%= user_version_view? ? 'Older versions are not released' : released_to %>
```
mjgiarlo, line 42: Will this message also show on the current version?

app/controllers/items_controller.rb
mjgiarlo: Hmmm, is the change set no longer needed within e.g. `Show::Item::AccessRightsComponent`? That component looks like it needs the change set to for `:view_access, :download_access, :access_location, :controlled_digital_lending`

app/components/show/item/details_component.html.erb
justinlittman: Yes. I think this is OK for now, as @andrewjbtw is going to be reconsidering the display of non-version specific fields like "Status".

app/components/show/barcode_component.rb
justinlittman: In the context of a component, a method named `edit?` makes sense, as you can easily tell what is to be edited. In the context of the presenter, it isn't clear what `edit?` could apply to (edit what?) as there are many parts of the page that support editing.

Can you think of a name for the presenter method that would be clearer than `edit?`? Otherwise, I'd be inclined to leave it the way it is.

app/controllers/items_controller.rb
justinlittman: Change set is constructed from the cocina here https://github.com/sul-dlss/argo/blob/main/app/presenters/argo_show_presenter.rb#L12-L14

app/components/show/barcode_component.rb
mjgiarlo: A name like `editable?` might make the intention behind this method clearer, namely that it is intended to determine whether or not to present edit controls to the user. As for making clearer what is to be edited, I think a developer could reasonably conclude it's an object in Argo (given that the ArgoShowPresenter is used to show objects in Argo).

justinlittman: Totally, but of all of the things on the show page that can update an object, which do `editable?` refer to (because it isn't all of them)?

mjgiarlo: If it isn't all of them, then I'm fine with the repetition. A misleading abstraction is worse than repetition.

mjgiarlo (approved):

PR #4523, opened 2024-07-05
Show user version banner.

### LINKED ISSUE #4520: Show banner indicating when the user is viewing an old version of an item
The text should be: "You are viewing an older version."

No need to specify the version number or distinguish between user or system version. The actual version number should be indicated below in the History list (separate ticket).

Screenshot 2024-07-03 at 4 48 41 PM

([Figma link](https://www.figma.com/proto/XjubCohES3rE8xSQpiNghq/Versioning?page-id=0%3A1&node-id=320-782&viewport=-2707%2C299%2C0.25&t=qxaxfTWyA5QqTX1p-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=320%3A782&show-proto-sidebar=1))

justinlittman:
closes #4520

# Why was this change made?

# How was this change tested?
Unit

### DIFF
--- a/app/components/document_title_component.rb
+++ b/app/components/document_title_component.rb
@@ -11,4 +11,10 @@ class DocumentTitleComponent < Blacklight::DocumentTitleComponent
   def object_type_class
     object_type_label.tr(' ', '-')
   end
+
+  def version_banner
+    return 'You are viewing the latest version.' if @presenter.head_user_version_view?
+
+    'You are viewing an older version.' if @presenter.user_version_view?
+  end
 end
--- a/app/components/document_title_component.html.erb
+++ b/app/components/document_title_component.html.erb
@@ -1,4 +1,9 @@
 <header class="document-header mb-5 pb-3">
+  <% if version_banner %>
+    <div class="alert alert-warning" role="alert">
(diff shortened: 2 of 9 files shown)


justinlittman:
image

justinlittman:
@andrewjbtw This is ready for testing.

app/components/document_title_component.rb
```diff
@@ -11,4 +11,10 @@ def object_type_label
   def object_type_class
     object_type_label.tr(' ', '-')
   end
+
+  def version_banner
+    return 'You are viewing the latest version.' if @presenter.head_user_version_view?
+
+    'You are viewing an older version.' if @presenter.user_version_view?
+  end
```
mjgiarlo, line 19: What would you think about this method living in the presenter itself? Currently, how and whether the banner is presented depends entirely on presenter-based concerns, so it seems like we could make a case for it.

mjgiarlo (approved):

app/components/document_title_component.rb
justinlittman: Since it is only used in one place, I'd prefer to see the logic closer to where the rendering is performed.

PR #4533, opened 2024-07-11
Do not render lock on user version detail page.

### LINKED ISSUE #4519: Don't show lock icon on version pages (outside of the HEAD version)
We are not using the lock/unlock icon/metaphor for old user versions, so don't show the lock icon on those show pages.

The lock/unlock icon is for managing the opening and closing of versions. If I understand the model correctly, it will only apply to the HEAD version or the current system version.

([Figma link](https://www.figma.com/proto/XjubCohES3rE8xSQpiNghq/Versioning?page-id=0%3A1&node-id=320-782&viewport=-2707%2C299%2C0.25&t=qxaxfTWyA5QqTX1p-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=320%3A782&show-proto-sidebar=1))

justinlittman:
closes #4519

# Why was this change made?

The user can't use the lock on a user version.

# How was this change tested?

Unit, QA

### DIFF
--- a/app/assets/stylesheets/show.scss
+++ b/app/assets/stylesheets/show.scss
@@ -82,15 +82,12 @@
   .open-close {
     width: 2rem;
     float: left;
+    margin-right: 2rem;
 
     .btn {
       font-size: x-large;
       color: $color-cardinal-red;
     }
-  }
-
-  .document-title-heading {
-    margin-left: 2rem;
   }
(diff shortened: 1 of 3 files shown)


justinlittman:
image

mjgiarlo (approved):

PR #4534, opened 2024-07-11
[HOLD] Update button visual design.

### LINKED ISSUE #4530: Update button visual design in Argo (this is a design component)
Current buttons
1. Primary filled
image

2. Secondary outline 
image

3. Primary disabled
image

4. Secondary disabled
image

Proposed buttons
1. Primary filled buttons (digital blue base 006CB8, on hover = digital blue dark 00548F)
They have a default state and hover state. (the mockup below shows an icon which is optional)
image

2. Secondary outline buttons (digital blue base 006CB8, on hover = digital blue dark 00548F)They have a default state and hover state. (the mockup below shows an icon which is optional) 
image

3. Primary disabled: Same button as described in 1 above, but no interactive elements (no hover or while pressing states). Background color is 50% black.
image

4. Secondary outline buttons: Same button as described in 2 above, but no interactive elements (no hover or while pressing states). Outline is 50% black and text is 50% black, no fill. 
image

justinlittman:
closes #4530

# Why was this change made?

# How was this change tested?

image

### DIFF
--- a/app/assets/stylesheets/bootstrap-overrides.scss
+++ b/app/assets/stylesheets/bootstrap-overrides.scss
@@ -3,6 +3,36 @@
   background-color: $floral-white;
 }
 
+.btn-primary {
+  @include button-variant($background: $digital-blue-base, $border: $digital-blue-base, $hover-background: $digital-blue-dark, $hover-border: $digital-blue-dark);
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+.btn-outline-primary {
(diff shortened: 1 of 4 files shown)


jcoyne:
This should not be done until the component is completed in the component library: https://github.com/sul-dlss/component-library/issues/10

jcoyne:
Buttons in the component library is now complete.  Can we switch to using those styles?

justinlittman:
What does that mean?

jcoyne:
@justinlittman something like #4539

PR #4539, opened 2024-07-12
Use component-library styles

jcoyne:
# Why was this change made?
Use the component-library

# How was this change tested?

PR #4542, opened 2024-07-16
[DRAFT] Update argo header to match redesign from component library

### LINKED ISSUE #4529: Update visual design of header
Current header:
image

New header (for reference, this is the same style as the [Preassembly header](https://sul-preassembly-prod.stanford.edu/) that was just updated):

image

- Background color: Stone/Dark 544948
- Stanford University in cardinal red 8C1515 at the top on white background: links to www.stanford.edu
- Stanford Libraries logo (white text w/full color rosette): links to library.stanford.edu
- Top right nav: Feedback and Logout (goes to the same place they do today); white links with underline on hover
- Title: Argo (Semibold 36pt)
- Bottom right nav: Impersonate   •.  All workflows   •.  Bulk action   •   Register (Bold with white link on hover, there should be about 3 spaces between each link and the bullet point separating the 4 links)

aaron-collier:
# Why was this change made?

Fixes #4529 

Screenshot 2024-07-16 at 4 19 04 PM

I have an outstanding question on the header font. I will make this a ready PR once that is complete.

# How was this change tested?

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -23,3 +23,6 @@
 @import "tags-autocomplete";
 @import "registration";
 @import "text-extraction";
+
+// SUL Styles from the component library
+@import url("https://sul-dlss.github.io/component-library/styles/sul.css");
--- a/app/assets/stylesheets/argo.scss
+++ b/app/assets/stylesheets/argo.scss
@@ -116,3 +116,23 @@ legend {
     margin-bottom: 5px;
   }
 }
+
+.palo-alto-dark {
+  --bs-dark-rgb: var(--stanford-stone-dark-rgb);
+}
+
+.su-brand-bar__logo {
+  color: var(--stanford-cardinal);
+}
+
+.masthead {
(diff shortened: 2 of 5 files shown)


jcoyne:
@aaron-collier I think this PR should wait until https://github.com/sul-dlss/argo/pull/4539 is reviewed.

app/assets/stylesheets/application.scss
```diff
@@ -23,3 +23,6 @@
 @import "tags-autocomplete";
 @import "registration";
 @import "text-extraction";
+
+// SUL Styles from the component library
+@import url("https://sul-dlss.github.io/component-library/styles/sul.css");
```
jcoyne, line 28: This is not a stable URL. So the approach taken by #4539 will cause less unexpected change.

aaron-collier: Sounds good, part of why I wanted this as DRAFT is that I wasn't sure about the component library approach, so it makes sense to wait on your PR and rebase from that approach. Thanks.

PR #4547, opened 2024-07-26
Adds Withdraw / Restore buttons.

### LINKED ISSUE #4528: Add "withdraw" button to previous user version page
When an item has multiple user versions, we need to support the ability to withdraw an older version. An older version may be withdrawn for a number of reasons: author's request, third-party takedown, etc. The mechanics of the approval of a withdrawal request is outside the scope of SDR applications. This will not be self-service for depositors and won't be supported in H2. Instead, similar to other takedowns of ETD or H2 items, the request will go through email and be approved that way.

We are supporting this in Argo because there is the potential to add user version creation to Argo in the future. We do not want to implement it only in H2 and then find that we need to implement it again in Argo. This also allows H2 to operate only on latest versions rather than on the whole history of an item.

**The withdraw interface**

Argo users will be able to handle withdrawal requests through a "Withdraw" button on the Argo interface. ([Figma](https://www.figma.com/proto/XjubCohES3rE8xSQpiNghq/Versioning?page-id=0%3A1&node-id=320-782&viewport=-2707%2C299%2C0.25&t=qxaxfTWyA5QqTX1p-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=320%3A782&show-proto-sidebar=1))

Screenshot 2024-07-09 at 5 10 02 PM

 The button should only display under the following conditions:

- The page is an "old" user version page. The latest user version cannot be withdrawn. 
- The item has not already been withdrawn. If the item has already been withdrawn, see "Restore" below
- If there is only one user version, there should not be a "Withdraw" button
- If the page is not a user version page, there should not be a "Withdraw" button

When a user clicks on the "Withdraw" button, pop up a confirmation box. In the box, display this message:

> Are your sure you want to withdraw this version? Once you withdraw this version, the Purl will no longer display it.

**What happens when you withdraw**

When a version is withdrawn, 

- a tombstone page should replace that version's page on the Purl. (This is ticketed separately.)
- links to files (versions of files) specific to that version should no longer return the files (404 or Access forbidden)
- files (versions of files) specific to that version can remain on Stacks as long as they can't be accessed

**Argo show page for a withdrawn version**

Once a version has been withdrawn, the show page for that version should:

- Display a modified banner indicating the version has been withdrawn
- Show a "Restore" button where the "Withdraw" button was displayed

([Figma](https://www.figma.com/proto/XjubCohES3rE8xSQpiNghq/Versioning?page-id=0%3A1&node-id=541-1014&viewport=-2707%2C299%2C0.25&t=qxaxfTWyA5QqTX1p-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=541%3A1014&show-proto-sidebar=1))

Screenshot 2024-07-09 at 5 23 55 PM

**Restoring a version, aka un-withdrawing**

We should also support restoring a withdrawn version. Especially while the feature is new, we may find that we need to restore a version that had been withdrawn. Putting this in the Argo UI will make it easier to correct possible mistakes.

Restoring a version will be like withdrawing a version. When the user clicks the restore button, show a dialog box with this message:

> Are your sure you want to restore this version? Once you restore this version, it will become visible on the Purl again.

Restoring the version should mean:

- the tombstone is replaced by the original Purl page
- the links to files should be re-enabled
- since the files will have remained on Stacks, no additional shelving should be required

**Additional context**

Withdraw/restore apply only to old user versions. The following cases are different in significant ways:

***Full takedown of all versions + a message about the item on the latest version.***

In this case, we would withdraw all old versions and create a new latest version with a message. I've been creating tombstones like this for content with DOIs since DataCite requests that we do not take pages down completely. Instead, the DOI resolves to a page that indicates that content has been removed (or moved to a new URL).

***Actual file deletion and other exceptional requests***

For example, if we need to remove files for privacy reasons. This will likely take some manual work, as it does now in these situations. We also have had one or two takedowns where we went as far as making an effort to de-list items from search engines. We should not even show a tombstone purl under that circumstance.

***Takedowns of items that are not using "user" versioning***

We will continue to have collections materials (not H2, ETD, etc) where we will take the items down by changing the rights to dark. That operation should continue to remove the Purl and not leave a tombstone behind.

justinlittman:
closes #4528

# Why was this change made?
So that users can hide / restore user versions.

# How was this change tested?
Unit, Stage

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -450,7 +450,7 @@ Rails/I18nLazyLookup: # new in 2.14
 Rails/I18nLocaleAssignment: # new in 2.11
   Enabled: true
 Rails/I18nLocaleTexts: # new in 2.14
-  Enabled: true
+  Enabled: false
 Rails/IgnoredColumnsAssignment: # new in 2.17
   Enabled: true
 Rails/Inquiry: # new in 2.7
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -236,24 +236,6 @@ Rails/HelperInstanceVariable:
   Exclude:
     - 'app/helpers/items_helper.rb'
 
-# Offense count: 18
-Rails/I18nLocaleTexts:
-  Exclude:
-    - 'app/controllers/agreements_controller.rb'
-    - 'app/controllers/apo_controller.rb'
-    - 'app/controllers/bulk_actions_controller.rb'
-    - 'app/controllers/catalog_controller.rb'
-    - 'app/controllers/content_types_controller.rb'
-    - 'app/controllers/descriptives_controller.rb'
-    - 'app/controllers/embargos_controller.rb'
-    - 'app/controllers/items_controller.rb'
-    - 'app/controllers/publishes_controller.rb'
-    - 'app/controllers/registrations_controller.rb'
-    - 'app/controllers/uploads_controller.rb'
-    - 'app/forms/agreement_form.rb'
-    - 'app/forms/has_view_access.rb'
-    - 'app/forms/registration_form.rb'
-
 # Offense count: 5
 Rails/OutputSafety:
   Exclude:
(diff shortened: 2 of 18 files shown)


justinlittman:
image

jcoyne (approved):

PR #4553, opened 2024-07-29
Change to user version links in version history list.

### LINKED ISSUE #4548: Version history list should link to the user versions not currently displayed
In the links to "public versions" in the version history list, link to the public versions that are not currently displayed. In the screenshot below, we are viewing version 1 (see the URL):

Screenshot 2024-07-26 at 1 31 35 PM

The links should be present for public versions 1 and 3. If possible, the link to the latest system version should be the "regular" (non-versioned) show page.

Currently, the links are always being generated for the older public versions and never generated for the most recent version.

justinlittman:
closes #4548

# Why was this change made?
Per @andrewjbtw 

# How was this change tested?
Unit, Andrew

### DIFF
--- a/app/components/version_milestones_component.html.erb
+++ b/app/components/version_milestones_component.html.erb
@@ -1,6 +1,6 @@
 <tr class="version<%= version %>">
  <td colspan="3">-
-   <%= title %><% if user_version %> (<%= current_version? ? user_version_label : link_to(user_version_label, user_version_path) %>)<% end %>
+   <%= title %><% if user_version %> (<%= user_version_link_or_label %>)<% end %>
  </td>
 </tr>
 
--- a/app/components/version_milestones_component.rb
+++ b/app/components/version_milestones_component.rb
@@ -15,12 +15,28 @@ class VersionMilestonesComponent < ViewComponent::Base
     @steps ||= milestones_presenter.steps_for(version)
   end
 
-  def current_version?
-    version.to_i == milestones_presenter.current_version
+  def link_user_version?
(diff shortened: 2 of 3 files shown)


app/components/version_milestones_component.rb
```diff
@@ -15,12 +15,28 @@ def steps
...
+     && user_version == user_versions_presenter.head_user_version&.to_i \
+     && user_versions_presenter.user_version.present?
+  end
+
+  def user_version_link
+    return link_to(user_version_label, user_version_path) if link_user_version?
+    return link_to(user_version_label, solr_document_path(milestones_presenter.druid)) if link_to_document?
+
+    user_version_label
   end
```
mjgiarlo, line 36: It looks like this can return either a link or a label. Worth considering a rename to `user_version_link_or_label` or some such?

app/components/version_milestones_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <tr class="version<%= version %>">
  <td colspan="3">-
-   <%= title %><% if user_version %> (<%= current_version? ? user_version_label : link_to(user_version_label, user_version_path) %>)<% end %>
+   <%= title %><% if user_version %> (<%= user_version_link %>)<% end %>
```
mjgiarlo, line 3: Good call moving this logic into the Ruby side of the component.

mjgiarlo (approved):

app/components/version_milestones_component.rb
justinlittman: Good suggestion. Fixed.

PR #4557, opened 2024-07-30
Use css variables for colors

jcoyne:
# Why was this change made?
Allows us to move away from sass.


justinlittman (approved):
Approve pending green.

PR #4566, opened 2024-07-30
Rename sulFooter.scss to layout.css

jcoyne:
# Why was this change made?
Move away from sass


justinlittman (approved):
Approve pending green.

PR #4577, opened 2024-07-31
Decouple the validation from the presentation of invalid fields

jcoyne:
# Why was this change made?

We have css classes sprinkled into the validation code. We can just use the validation API instead.


justinlittman (approved):

andrewjbtw:
The validation is now running immediately on page load. This is making the required fields light up in red and the cursor is placed in the label field. Sometimes my browser is taken to the label field and sometimes (as in the recording below) it's not taken there, but the validation message still appears on the screen.

https://github.com/user-attachments/assets/33504f7a-3b98-4218-a672-62021a6fb9b9

jcoyne:
Andrew approved changes on Slack.

PR #4579, opened 2024-07-31
Make file links specific to version.

### LINKED ISSUE #4522: On old version pages, file links should link to the version of the file that matches the version being shown
The Argo contents list contains links to each file in the item. Those links could be to Stacks and/or Preservation. If more than one version of a file exists, the link should be to the version of the file that matches the version of the item being viewed.

justinlittman:
closes #4522

# Why was this change made?
Allow users to get files for previous versions.

# How was this change tested?
Unit, Andrew

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class FilesController < ApplicationController
+class FilesController < ApplicationController # rubocop:disable Metrics/ClassLength
   include ActionController::Live # required for streaming
 
   before_action :load_resource, except: [:download]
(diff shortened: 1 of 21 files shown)


app/controllers/files_controller.rb
```diff
@@ -10,20 +10,24 @@ class FilesController < ApplicationController
...
-        Honeybadger.notify(message)
-        return render status: :internal_server_error, plain: message
+    @file = find_file
+    @user_version = params[:user_version_id]
+    if @user_version
+      @has_been_accessioned = WorkflowService.accessioned?(druid: @cocina_model.externalIdentifier)
+      @version = @cocina_model.version
+    else
+      @has_been_accessioned = true
+      if @has_been_accessioned
```
lwrubel, line 20: Won't this always be true since it's assigned immediately above?

justinlittman: Yes, this seems wrong.

PR #4592, opened 2024-08-05
Update dependencies

sul-devops-team:


jcoyne:
There was a major version bump of view-component here.

jcoyne (approved):

app/components/application_component.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationComponent < ViewComponent::Base
-  delegate :can?, to: :controller
+  delegate :can?, :current_user, to: :controller
```
jcoyne, line 4: You can remove this addition.

spec/rails_helper.rb
```diff
@@ -84,4 +84,8 @@
   config.include SigninHelper, type: :view
   config.include Devise::Test::ControllerHelpers, type: :controller
   config.include ViewComponent::TestHelpers, type: :component
+
+  config.before(:each, type: :component) do
+    @request = vc_test_controller.request
```
jcoyne, line 89: This is unnecessary.

app/components/application_component.rb
mjgiarlo: @jcoyne If I remove the delegation, there are failures due to https://github.com/sul-dlss/argo/pull/4592/files#diff-6c1bb310d0eac46c88fec511636f41bd4577cb38e6dbf2455a6f4db8112c4632R13-R15

If I revert *that* change so that we go back to getting the current user from `helpers`, that leads to this:

Screenshot from 2024-08-05 13-36-36

mjgiarlo (approved):
Approved by automated merge script

PR #4596, opened 2024-08-05
Simplify testing of components

jcoyne:
# Why was this change made?
 Simpler.


mjgiarlo:
Pulled these changes into #4592. Thanks!

PR #4612, opened 2024-08-08
Revert "Use the facet styles from the component library"

jcoyne:
This reverts commit f2a23aeeec8ac4df94219cbafadb034151d5009b.

# Why was this change made?

Andrew doesn't like the new designs.


justinlittman (approved):
Approve pending green.

PR #4618, opened 2024-08-09
Refactor file controller index.

justinlittman:
# Why was this change made?
@lwrubel pointed out a logic error that prompted a refactoring.

# How was this change tested?
Unit


lwrubel (approved):
Approved pending green

PR #4619, opened 2024-08-09
Rename docker-compose.yml to compose.yaml

jcoyne:
# Why was this change made?

This is what the docker documentation indicates the current practice is

# How was this change tested?
Ran `docker compose up`


justinlittman:
Yes, but all of our other repos use `docker-compose.yml`. Can you change it everywhere or no where?

jcoyne:
Actually there are some repos that have it. https://github.com/sul-dlss/SearchWorks/blob/main/compose.yaml  I'm not interested in changing repos that I don't have downloaded and don't regularly contribute to.

aaron-collier:
Just dropping a link to the ref indicating this is the current default: https://docs.docker.com/compose/compose-application-model/#the-compose-file

aaron-collier (approved):

PR #4620, opened 2024-08-09
Update version banner with link to latest

### LINKED ISSUE #4608: Banner for old versions in Argo
We are requesting a few refinements to the banner that appears for old versions of items in Argo:

1. Specify if version is a system or public version in the text:
-Current: You are viewing an older version.
-Proposed: You are viewing an older {system, public} version.

2. Add a link to the current version in the banner:
- Current: There is a link by the title to get back to the latest version, but we think it will be easier to locate if it's in the banner. 
image

- Proposed: Add link to latest version into banner-- "You are viewing an older {system, public} version. [View latest version.](https://argo.stanford.edu)" --- this link should go to the latest version
image

3. Visual design:
- Current: Looks like the banner is in the old visual style.
image

- Proposed: Use the visual style from the component library
image

lwrubel:
# Why was this change made?
Resolves #4608. 

Screenshot 2024-08-09 at 4 13 13 PM

Screenshot 2024-08-09 at 4 13 29 PM

# How was this change tested?
Unit and deploy to QA.

### DIFF
--- a/app/components/document_title_component.rb
+++ b/app/components/document_title_component.rb
@@ -15,8 +15,20 @@ class DocumentTitleComponent < Blacklight::DocumentTitleComponent
   def version_banner
     return 'You are viewing the latest version.' if @presenter.head_user_version_view? || @presenter.current_version_view?
 
-    'You are viewing an older version.' if version_or_user_version_view?
+    "You are viewing an older #{version_type} version." if version_or_user_version_view?
   end
 
-  delegate :version_or_user_version_view?, to: :@presenter
+  def show_latest_link?
+    return previous_user_version_view? if @presenter.user_version_view?
+
+    !current_version_view?
+  end
+
+  def version_type
+    return 'public' if @presenter.user_version_view?
+
+    'system' if @presenter.version_view?
+  end
+
+  delegate :version_or_user_version_view?, :previous_user_version_view?, :current_version_view?, to: :@presenter
 end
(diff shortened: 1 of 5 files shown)


app/components/document_title_component.rb
```diff
@@ -15,8 +15,20 @@ def object_type_class
   def version_banner
     return 'You are viewing the latest version.' if @presenter.head_user_version_view? || @presenter.current_version_view?
 
-    'You are viewing an older version.' if version_or_user_version_view?
+    "You are viewing an older #{version_type} version." if version_or_user_version_view?
```
justinlittman, line 18: Is the `if` needed here?

app/components/document_title_component.rb
```diff
@@ -15,8 +15,20 @@ def object_type_class
...
+  def show_latest_link?
+    return previous_user_version_view? if @presenter.user_version_view?
+
+    !current_version_view?
+  end
+
+  def version_type
+    return 'public' if @presenter.user_version_view?
+
+    'system' if @presenter.version_view?
```
justinlittman, line 30: Is the `if` needed here?

justinlittman (approved):
Approve pending consideration of comments.

app/components/document_title_component.rb
lwrubel: Will remove it here, thanks.

app/components/document_title_component.rb
lwrubel: The `if` is here so that the function returns `nil` when we're on the latest version and there should not be a banner displaying. That's the `<% if version_banner %>`. Would it be better to move the check for whether a banner is needed into a separate method?

justinlittman: No, fine as is.

PR #4627, opened 2024-08-13
Upgrade blacklight-hierarchy to 6.4.0

jcoyne:
# Why was this change made?
Keep dependencies up to date.

# How was this change tested?


aaron-collier:
@jcoyne are there any visible changes related to this that need to be signed off on? I wouldn't necessarily think so with it being a dot release, but maybe. Thanks.

jcoyne:
Not that I know of.

aaron-collier (approved):

PR #4628, opened 2024-08-13
Upgrade to Rails 7.2

jcoyne:
# Why was this change made?

Keep up to date with dependencies 

# How was this change tested?


bin/brakeman
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env ruby
+require "rubygems"
+require "bundler/setup"
+
+ARGV.unshift("--ensure-latest")
+
+load Gem.bin_path("brakeman", "brakeman")
```
mjgiarlo, line 7: Interesting addition. When should we be running `brakeman`? Should it be included in any workflows?

mjgiarlo (approved):
LGTM. Should we deploy to stage or QA and run a couple integration tests to be safe?

bin/setup
```diff
@@ -30,4 +30,8 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Restarting application server =="
   system! "bin/rails restart"
+
+  # puts "\n== Configuring puma-dev =="
+  # system "ln -nfs #{APP_ROOT} ~/.puma-dev/#{APP_NAME}"
+  # system "curl -Is https://#{APP_NAME}.test/up | head -n 1"
```
aaron-collier, line 36: Do we want to maintain this comment as documentation or is it cruft?

aaron-collier: I see this in all of the other 7.2 PRs, so seems OK to keep.

PR #4632, opened 2024-08-15
[HOLD] Link version to user version path when version has user version.

### LINKED ISSUE #4631: In Argo version history list, only link the public version if there's a public version
We're showing system version and public version links in the Argo version history. 

Screenshot 2024-08-14 at 2 08 27 PM

This is functioning but it's a bit odd because the system version link doesn't show the features that are available for the public version:

- no downloadable files
- no withdraw/restore

Since each public version is identical (in the data) to a system version, show only the public version view. The system version view is redundant and more limited.

justinlittman:
closes #4631

# Why was this change made?
So that the user is linked to the more useful user version page instead of the less useful version page.

# How was this change tested?
Unit, Andrew

### DIFF
--- a/app/components/version_milestones_component.rb
+++ b/app/components/version_milestones_component.rb
@@ -32,6 +32,7 @@ class VersionMilestonesComponent < ViewComponent::Base
   end
 
   def version_link_or_label
+    return link_to(title, user_version_path) if user_version && link_user_version?
     return link_to(title, version_path) if link_version?
     return link_to(title, solr_document_path(druid)) if version_link_to_document?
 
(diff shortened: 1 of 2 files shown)


app/components/version_milestones_component.rb
```diff
@@ -32,6 +32,7 @@ def link_version?
   end
 
   def version_link_or_label
+    return link_to(title, user_version_path) if user_version && link_user_version?
```
justinlittman, line 35: I beg to differ with CodeClimate.

mjgiarlo (approved):

PR #4669, opened 2024-10-30
Upgrade to Rails 7.2.1.2

mjgiarlo:
# Why was this change made?

- **Upgrade to Rails 7.2**
- **Ignore rubocop violations in generated files**

Closes #4628

# How was this change tested?

- [x] CI
- [x] qa/stage


config/environments/test.rb
```diff
@@ -39,13 +35,26 @@
...
+  # Disable caching for Action Mailer templates even if Action Controller
+  # caching is enabled.
   config.action_mailer.perform_caching = false
 
+  # Set the ActiveJob queue adapter for tests
+  # NOTE: the test adapter does *not* automatically perform jobs. To perform jobs,
+  #       call `perform_enqueued_jobs` from the ActiveJob TestHelper at the point
+  #       you want them run. See `spec/rails_helper.rb` for an example of how to
+  #       enable this helper for a class of tests (e.g., feature specs).
+  # config.active_job.queue_adapter = :test
```
mjgiarlo, line 47: I added this explicitly. I found that using the `:test` adapter (now default) caused failures with some feature specs and the `:async` adapter caused failures with some request specs. This is because some feature specs need queued jobs to actually run and the `:test` adapter does not run them by default. The `:async` adapter does. However, some request specs check that a job has been enqueued, which the `:async` adapter does not support; only the `:test` adapter does.

Using the `:test` adapter in the test env and then manually performing jobs (via `perform_enqueued_jobs` which comes from `ActiveJob::TestHelper`) where needed appears to be the Rails happy path, so that's what you'll see in this PR.

I don't know how we didn't see this with the prior version of Rails. 🤷🏻‍♂️

spec/support/page_helpers.rb
```diff
@@ -2,22 +2,26 @@
 
 # Used for testing background jobs in integration tests where we need to reload the page until the job is done.
 module PageHelpers
-  def reload_page_until_timeout(timeout: 10)
+  TRIES = 10
+
+  def reload_page_until_timeout
     browser = page.driver.browser
     content = nil
-    Timeout.timeout(timeout) do
```
mjgiarlo, line 8: The `Timeout` library caused non-deterministic test failures, and sometimes got stuck in endless loops. `Timeout` is bad because it doesn't play nicely in threaded environments (see, e.g., https://adamhooper.medium.com/in-ruby-dont-use-timeout-77d9d4e5a001). So I ditched the old approach and came up with a simpler one that plays nice with threads.

peetucket (approved):
Thanks for doing battle with rspec and congrats on the victory.

PR #4678, opened 2024-11-18
do not allow sdr-ingest steps to be set to skip or complete

### LINKED ISSUE #4670: do not show options to set sdr-ingest-transfer or sdr-ingest-received steps to 'skip' or 'complete'
The `sdr-ingest-transfer` step is the step that starts ultimately starts `preservationIngestWF`, and `preservationIngestWF` is how objects are ultimately sent to preservation. 

The `sdr-ingest-received` step is the step that starts ultimately records that the `preservationIngestWF` has been completed successfully.

If these steps are set to `skipped` or 'completed`, we can end up with items that are out of sync between DSA and preservation because the accessionWF will continue if these steps are skipped/completed, resulting in a situation where it looks like the latest version of an item was accessioned when in fact nothing for that version was sent to preservation. I believe skipping/completing this step in the past has been the source of some of the preservation sync errors we've seen.

There are very limited circumstances where these steps must be marked completed after-the-fact. But if we can limit the ability to do that to the workflow server, it should reduce the number of times it's done mistakenly.

**User impact**

I don't think skipping this step happens too often, but it's a risk we shouldn't be taking.

peetucket:
# Why was this change made?

Fixes #4670 - do not show the skip/complete options for workflow steps sdr-ingest-transfer or sdr-ingest-received.  Other steps remain unchanged.

Set to complete button is also hidden for those steps if they are in waiting state.  See screenshots below for examples

Screenshot 2024-11-18 at 10 57 29 AM

Screenshot 2024-11-18 at 10 57 35 AM

# How was this change tested?

Localhost and specs

### DIFF
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -7,4 +7,8 @@ class ApplicationComponent < ViewComponent::Base
   def inspect
     "#<#{self.class.name}:#{object_id}>"
   end
+
+  def allow_skip_or_complete?(workflow_step)
+    %w[sdr-ingest-transfer sdr-ingest-received].exclude?(workflow_step)
+  end
 end
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -8,6 +8,10 @@ class ApplicationComponent < ViewComponent::Base
     "#<#{self.class.name}:#{object_id}>"
   end
 
+  # this is used by the WorkFlowUpdateButton and the WorklowStepStatusSelect components
+  # to determine if the user is allowed to skip or complete a particular workflow step
+  # (because we do not want to allow the user to manually skip or complete some steps)
+  # see https://github.com/sul-dlss/argo/issues/4670
   def allow_skip_or_complete?(workflow_step)
     %w[sdr-ingest-transfer sdr-ingest-received].exclude?(workflow_step)
   end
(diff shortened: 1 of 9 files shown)


spec/system/apo_spec.rb
```diff
@@ -50,12 +51,12 @@
...
-    accession_step_count.times do
-      # Ensure every step of accessionWF is completed, this will allow us to open a new version.
-      click_link 'accessionWF'
-      click_button 'Set to completed', match: :first
+    # Ensure every step of accessionWF is completed, this will allow us to open a new version.
+    druid = page.current_url.split('/').last
+    accession_processes.each do |process|
+      workflow_client.update_status(druid:, workflow: 'accessionWF', process:, status: 'completed')
     end
+    click_link 'Reindex'
```
peetucket, line 59: these changes in the tests are needed because we can no longer click the "Set to completed" button in the UI for all steps - so just do it via the workflow client (same change in a different spec below)

app/components/application_component.rb
```diff
@@ -7,4 +7,8 @@ class ApplicationComponent < ViewComponent::Base
   def inspect
     "#<#{self.class.name}:#{object_id}>"
   end
+
+  def allow_skip_or_complete?(workflow_step)
```
justinlittman, line 11: Can you put a comment explaining what this is and where it is used?

peetucket: good idea, adding comment

PR #4714, opened 2025-02-06
Filter out bulk action form params

### LINKED ISSUE #4710: Bulk action parameters persist in Argo URL after running a bulk action, causing problems when running multiple bulk actions
**Describe the bug**

This was a difficult bug to isolate and hopefully I've described it clearly. The short summary is that under some conditions, the parameters from a bulk action persist into subsequent bulk actions. This causes those subsequent bulk actions to behave in unexpected ways, including silent failures and cocina validation errors.

This issue happens only when a bulk action is initiated via the following steps:

1. Run a search
2. Start the bulk action from the search by clicking the bulk action button on this part of the screen above and to the right of the search results (not the link in the Argo header)

> Image

3. Click on "new bulk action"

Once you've started a bulk action that way, the parameters from that first bulk action will persist if you run a sequence of bulk actions in a row in that same tab.

Example, [starting from this search](https://argo.stanford.edu/catalog?f%5Bcontent_type_ssim%5D%5B%5D=image&f%5Bis_governed_by_ssim%5D%5B%5D=info%3Afedora%2Fdruid%3Atx440hx7665&f%5BobjectType_ssim%5D%5B%5D=item):

1. Run a bulk action to set rights on objects to `view: world`, `download: none`
2. See this URL in Argo after submitting the bulk action: `https://argo.stanford.edu/bulk_actions?controlled_digital_lending=false&download_access=none&f%5Bcontent_type_ssim%5D%5B%5D=image&f%5Bis_governed_by_ssim%5D%5B%5D=info%3Afedora%2Fdruid%3Atx440hx7665&view_access=world`
3. Click on New bulk action to run another bulk right change
4. See this URL with those parameters still present in Argo when viewing the new bulk action form`https://argo.stanford.edu/bulk_actions/new?controlled_digital_lending=false&download_access=none&f%5Bcontent_type_ssim%5D%5B%5D=image&f%5Bis_governed_by_ssim%5D%5B%5D=info%3Afedora%2Fdruid%3Atx440hx7665&view_access=world`
5. Submit a bulk action changing the rights on those same objects to `view: world`, `download: world`
6. See that the URL is actually still the one that has `world` and `none` in the parameters:`https://argo.stanford.edu/bulk_actions?controlled_digital_lending=false&download_access=none&f%5Bcontent_type_ssim%5D%5B%5D=image&f%5Bis_governed_by_ssim%5D%5B%5D=info%3Afedora%2Fdruid%3Atx440hx7665&view_access=world`
7. See that the objects were not given the correct `world`, `world` rights but remain `world`, `none`
8. See that the log reports success even though the job failed to change the rights to the correct values.

This problem does not seem to happen if you initiate a bulk action with a "clean" URL, such as when you start from the "Bulk action" link in the Argo header. It only seems to happen if you start from an Argo search.

**User Impact**
Users may not realize that their bulk action failed to make the correct changes. In some cases, the parameters from the first bulk action will conflict with parameters from the second one and result in cocina validation errors. 

**To Reproduce**
See above for sequence.

**Expected behavior**
The bulk action function should not save the parameters from a previous bulk action. But it should save the list of druids from a previous search.

**Additional context**
I've used the example of a rights change but I think this is a problem with any bulk action that submits data modifications via the form.

lwrubel:
# Why was this change made?
Resolves #4710 to strip bulk action form params from future bulk actions.

# How was this change tested?
Request spec; tested by PO

### DIFF
--- a/app/components/new_bulk_action_button_component.rb
+++ b/app/components/new_bulk_action_button_component.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class NewBulkActionButtonComponent < ApplicationComponent
+  def link_path
+    new_bulk_action_path(search_state.reset_search.to_h.except(*bulk_action_params))
+  end
+
+  def bulk_action_params
+    %i[tag
+       who
+       what
+       workflow
+       to
+       version_description
+       new_apo_id
+       download_access
+       view_access
+       copyright_statement
+       copyright_statement_option
+       license
+       license_option
+       use_statement
+       use_statement_option
+       barcodes
+       use_barcodes_option
+       catalog_record_ids
+       use_catalog_record_ids_option
+       current_resource_type
+       new_resource_type
+       new_content_type
+       new_collection_id]
+  end
+end
--- a/app/components/new_bulk_action_button_component.rb
+++ b/app/components/new_bulk_action_button_component.rb
@@ -2,6 +2,8 @@
 
 class NewBulkActionButtonComponent < ApplicationComponent
   def link_path
+    # Since reset_search does not remove all the bulk action form params from the search state,
+    # remove them as specified
     new_bulk_action_path(search_state.reset_search.to_h.except(*bulk_action_params))
   end
 
--- a/app/components/new_bulk_action_button_component.html.erb
+++ b/app/components/new_bulk_action_button_component.html.erb
@@ -1,0 +1,1 @@
+<%= link_to 'New Bulk Action', link_path, class: 'btn btn-primary' %>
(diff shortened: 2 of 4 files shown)


app/components/new_bulk_action_button_component.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class NewBulkActionButtonComponent < ApplicationComponent
+  def link_path
+    new_bulk_action_path(search_state.reset_search.to_h.except(*bulk_action_params))
```
justinlittman, line 5: Can you document what is going on here and why?

justinlittman (approved):

app/components/new_bulk_action_button_component.rb
lwrubel: Added a comment

PR #4724, opened 2025-02-14
show workflow name on modal

peetucket:
# Why was this change made?

I noticed the workflow name itself was not showing in the modal that shows workflow details.  I think this is an improvement.

**Current**

Screenshot 2025-02-14 at 10 29 32 AM

**Updated**

Screenshot 2025-02-14 at 10 28 59 AM

# How was this change tested?

Localhost/CI

### DIFF
--- a/app/views/workflows/_show.html.erb
+++ b/app/views/workflows/_show.html.erb
@@ -1,10 +1,10 @@
 <span class="section-head-link">
+  <h4><%= @presenter.workflow_name %> - <%= @presenter.druid %></h4>
   View:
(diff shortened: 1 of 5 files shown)


app/views/workflows/_show.html.erb
```diff
@@ -1,10 +1,10 @@
 <span class="section-head-link">
+  <h4><%= @presenter.workflow_name %> - <%= @presenter.druid %></h4>
```
justinlittman, line 2: I'd make this `<div class="h4">` to avoid accessibility issues.

justinlittman (approved):

PR #4737, opened 2025-03-21
bootstrap item/apo/agreement creation to better link together

peetucket:
# Why was this change made?

Allows you to work on localhost without hitting errors where the fixture APO is not connected to an actual agreeement.  Instead create each object as it is needed, so it can be properly wired together.

# How was this change tested?

Localhost

### DIFF
--- a/spec/factories/items.rb
+++ b/spec/factories/items.rb
@@ -25,7 +25,12 @@ FactoryBot.define do
       Dor::Services::Client.objects.register(params: builder.cocina_model)
     end
 
-    admin_policy_id { 'druid:hv992ry2431' }
+    admin_policy_id do
+      FactoryBot.create_for_repository(:persisted_apo,
+                                       roles: [{ name: 'dor-apo-manager',
+                                                 members: [{ identifier: 'sdr:administrator-role',
+                                                             type: 'workgroup' }] }]).externalIdentifier
+    end
     label { 'test object' }
     type { Cocina::Models::ObjectType.object }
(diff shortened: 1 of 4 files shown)


spec/factories/items.rb
```diff
@@ -38,6 +43,7 @@
     factory :agreement do
       type { Cocina::Models::ObjectType.agreement }
       label { 'Test Agreement' }
+      admin_policy_id { 'druid:hv992ry2431' }
```
peetucket, line 46: since the agreement is created on demand, it must be linked to an existing apo

mjgiarlo (approved):
LGTM

spec/requests/apo_count_spec.rb
```diff
@@ -40,22 +40,20 @@
...
-      expect(rendered).to have_link '2', href: search_catalog_path(
+      expect(rendered).to have_link '1', href: search_catalog_path(
         f: {
-          is_governed_by_ssim: ["info:fedora/#{ur_apo_id}"],
+          is_governed_by_ssim: ["info:fedora/#{admin_policy_id}"],
           objectType_ssim: ['item']
         }
       )
-      expect(rendered.find_link('2')[:target]).to eq '_top'
+      expect(rendered.find_link('1')[:target]).to eq '_top'
```
peetucket, line 56: since we are now creating new APOs each time we create an item, we can't rely on the UR APO to have items associated with it for testing purposes

PR #4756, opened 2025-04-18
Switch default database from MySQL to Postgres

mjgiarlo:
# Why was this change made?

Connects to #4743

Includes:

* Change DB config to use Postgres for all envs
* Remove MySQL-ism from DB migration and allow db:migrate to function with Postgres underneath
* Change DB seeding so db:prepare functions like it does everywhere else
* Make web container depend on Postgres in docker compose
* Reflect above changes in README

# How was this change tested?

CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,14 +26,6 @@ cd argo
 bundle install
 ```
 
-Note that `bundle install` may complain if MySQL isn't installed.  You can either comment out the `mysql2` inclusion in `Gemfile` and come back to it later (you can develop using `sqlite3`), or you can install MySQL.
-
-Install foreman for local development (foreman is not supposed to be in the Gemfile, See this [wiki article](https://github.com/ddollar/foreman/wiki/Don't-Bundle-Foreman)):
-
-```bash
-gem install foreman
-```
-
 ## Local Development TL;DR
 
 Brings up app at localhost:3000 with some test data:
@@ -43,6 +35,7 @@ gem install foreman
 yarn install
 docker compose up -d
 docker compose stop web
+bin/rails db:prepare
 bin/dev
 bin/rake argo:seed_data # run in separate terminal window
 rdbg -A # run in separate terminal window
@@ -79,7 +72,7 @@ CI runs a series of steps;  this the sequence to do it locally, along with some 
 4. **Prepare rails for testing**
 
     ```
-    bin/rails test:prepare
+    bin/rails db:prepare test:prepare
     ```
 
 5. **Run the linters and the tests**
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -26,14 +26,6 @@ cd argo
 bundle install
 ```
 
-Note that `bundle install` may complain if MySQL isn't installed.  You can either comment out the `mysql2` inclusion in `Gemfile` and come back to it later (you can develop using `sqlite3`), or you can install MySQL.
-
-Install foreman for local development (foreman is not supposed to be in the Gemfile, See this [wiki article](https://github.com/ddollar/foreman/wiki/Don't-Bundle-Foreman)):
-
-```bash
-gem install foreman
-```
```
mjgiarlo, line 35: Newer `bin/dev` scripts (such as is in Argo) automatically install `foreman` if missing.

db/seeds.rb
```diff
@@ -3,19 +3,18 @@
...
-unless Rails.env.production?
+# Prevent executing this code in the production Rails env, which is what we use in deployed environments.
+if Rails.env.local?
   require_relative '../spec/support/create_strategy_repository_pattern'
   require_relative '../spec/support/item_method_sender'
   require_relative '../spec/support/apo_method_sender'
   require_relative '../spec/support/reset_solr'
 
-  $stdout.puts 'This will clear the Solr repo. Are you sure? [y/n]:'
-  if gets.chomp == 'y'
```
mjgiarlo, line 13: Forcing the `db:seed` task to be used in an interactive mode breaks tasks that depend on it, such as `db:prepare`. I could be convinced that automatically clearing Solr in dev/test is not the best idea, in which case I'd probably take a different approach like sniffing for an opt-in env variable that communicates a desire to clear Solr. Open to other ideas too.

db/migrate/20160318203331_increase_delayed_job_column.rb
```diff
@@ -1,8 +1,11 @@
 class IncreaseDelayedJobColumn < ActiveRecord::Migration[4.2]
   def change
-    # Ensure that the handler column in the delayed_job table is MySQL longtext
-    # so that we can pass in lots of druids in the params when a new Delayed
-    # Job is created.
-    change_column :delayed_jobs, :handler, :text, limit: 4_294_967_295
+    # Only run this migration if MySQL is the database. The :text datatype in Postgres is unlimited.
+    if ActiveRecord::Base.connection.adapter_name.match?(/mysql/i)
```
aaron-collier, line 4: Do we need this if we're configuring postgres and using it in docker? Is this just an "in-case someone is using it" or do we still have mysql out there?

aaron-collier (approved):

db/migrate/20160318203331_increase_delayed_job_column.rb
mjgiarlo: @aaron-collier Now that the migration's done in all three envs, I think I'll be inclined to drop this in a follow-up PR. I forgot I even made this branch backwards-compatible. 😆

PR #4763, opened 2025-04-28
Switch containerland over to Postgres and drop MySQL from the deps

mjgiarlo:
# Why was this change made?

Finish the migration away from MySQL/MariaDB

# How was this change tested?

CI, ran container & checked localhost locally

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -110,7 +110,7 @@ services:
       RAILS_LOG_TO_STDOUT: 'true'
 
   db:
-    image: postgres:11 # aligns the pg version with what is supported by dor-services-app
+    image: postgres:15 # use the highest version that works with all the services herein
     ports:
       - 5432:5432
     environment:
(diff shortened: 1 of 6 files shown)


lwrubel:
Should we update the `compose.yaml` to use Postgres 13 for the argo db container now? I see that's what DSA is now using in its compose.yaml, if that makes a difference.

mjgiarlo:
@lwrubel 💬 
> Should we update the `compose.yaml` to use Postgres 13 for the argo db container now? I see that's what DSA is now using in its compose.yaml, if that makes a difference.

It looks like Argo's on 16 (TIL). I'll bump that in compose, good idea!

compose.yaml
```diff
@@ -110,7 +110,7 @@ services:
       RAILS_LOG_TO_STDOUT: 'true'
 
   db:
-    image: postgres:11 # aligns the pg version with what is supported by dor-services-app
+    image: postgres:15 # use the highest version that works with all the services herein
```
mjgiarlo, line 113: sdr-api barked at 16.

PR #4767, opened 2025-05-01
Validate title structured value type on spreadsheet upload

### LINKED ISSUE #4279: Descriptive spreadsheet import should not accept title1.structuredValue1.value without title*.structuredValue*.type
**Describe the bug**
Descriptive metadata spreadsheet import is allowing imports of spreadsheets where there is a value in `title1.structuredValue1.value` but no type in `title1.structuredValue1.type`. This is causing indexing and publishing failures.

Example spreadsheet to reproduce the issue on stage: 
[descriptive-bx947ty9936.csv](https://github.com/sul-dlss/argo/files/13385406/descriptive-bx947ty9936.csv)
https://argo-stage.stanford.edu/view/druid:bx947ty9936

**User Impact**
Users should be notified of metadata validation errors at the time of import. Without that, hidden errors cause problems later in accessioning including preventing items from being made accessible on the purl.

**Expected behavior**
Validation should prevent invalid titles from being created.

aaron-collier:
# Why was this change made?

Fixes #4279 

There is existing validation the `title*.structuredValue*.value` has a cooresponding `title*.structuredValue*.type` exists. This adds an additional validation that the value of `title*.structuredValue*.type` is one of the valid types as defined in cocina and stops the processing of the spreadsheet such that the value isn't allowed to be updated on the object without a valid type.

Screenshot 2025-05-01 at 1 54 42 PM

Note: while this is relatively straight forward, we may want to follow this up with an update to cocina-models so that `Cocina::Models::Mapping::FromMods::Title::TYPES` is available somewhere that isn't `....Mapping::FromMods...`

# How was this change tested?


app/services/description_validator.rb
```diff
@@ -62,6 +63,16 @@ def validate_matching_structured_title_headers
     end
   end
 
+  def validate_structured_title_types
```
justinlittman, line 66: Can you put in a comment describing what this is checking?

justinlittman (approved):

PR #4772, opened 2025-05-05
Validate that part label is present if sort key is present in serials form

### LINKED ISSUE #4771: Serials form should validate that if there is a part label if a sort key is provided
Otherwise the update will fail with: `Cocina::Models::ValidationError: partLabel must also be present if a sortKey is used in catalog link`

See: https://app.honeybadger.io/projects/49894/faults/119939738#notice-context

mjgiarlo:
# Why was this change made?

Fixes #4771

Screenshot from 2025-05-05 09-30-40

# How was this change tested?

CI

### DIFF
--- a/app/controllers/serials_controller.rb
+++ b/app/controllers/serials_controller.rb
@@ -9,12 +9,14 @@ class SerialsController < ApplicationController
   end
 
   def update
-    form = SerialsForm.new(@cocina)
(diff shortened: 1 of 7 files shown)


lwrubel (approved):

lwrubel:
Looks good getting the validation to display, thanks!

PR #4792, opened 2025-06-12
Fix Argo spreadsheet validation for titles.

### LINKED ISSUE #4789: Argo spreadsheet title validation is too strict
**Describe the bug**

@arcadiafalcone alerted me to a problem with Argo's cocina descriptive spreadsheet title validation. This is spreadsheet-specific validation, not general Cocina validation. We recently added this validation in an attempt to address a common problem we've seen with cocina spreadsheet uploads: titles should have both a `value` and a matching `type` but if they are missing one or the other of `value`/`type` then that causes problems in accessioning/publishing/indexing.

I believe these are the two PRs that added the validation: https://github.com/sul-dlss/argo/pull/4767 and https://github.com/sul-dlss/argo/pull/4768

In an effort to catch this issue in the spreadsheet itself - because Cocina-level title validation seems to be a problem requiring a longer time frame to solve - we added a validation that checks:

- if `title#.value` has data, is there a corresponding `title#.type`? (and vice versa)
- if `title#.structuredValue#.value` has data, is there a corresponding `title#.structuredValue#.type` (and vice versa)

What I did not realize is that there is not always a one-to-one correspondence between the Cocina spreadsheet headers. For example, it is valid to have this combination of `type` and `value`:

- `title#.structuredValue#.value` with `title#.type`

To take a prod example: in [this item](https://argo.stanford.edu/view/nt361pq4141):

- there's a `title2` that has a `title2.type` of `alternative` in the cocina CSV. 
- there is no `title2.value` but 
- instead, the `title2` `value` is actually a set of multiple fields under `structuredValue`

The Cocina spreadsheet looks like this for `title2`:

title2.note1.type | title2.note1.value | title2.structuredValue1.type | title2.structuredValue1.value | title2.structuredValue2.type | title2.structuredValue2.value | title2.type
-- | -- | -- | -- | -- | -- | --
nonsorting character count | 2 | nonsorting characters | A | main title | Family in Sunset Park Brooklyn | alternative

This is valid cocina but Argo is currently rejecting spreadsheet import because there is no data under the header `title2.value`.

**User Impact**
This is preventing spreadsheet upload for valid metadata that matches metadata already in SDR. This affects some of our higher visibility collections in particular, including the Martin Wong exhibit.

This is ultimately a blocker for descriptive metadata uploads. We either need to revert the validation changes we put in last month or update the validation to handle this use case.

**Expected behavior**
It should always be possible to export a Cocina spreadsheet from Argo, update the data without changing the headers or entering invalid values, and import it again without hitting a validation error.

justinlittman:
closes #4789

# Why was this change made?
Validation was too strict.

# How was this change tested?
Unit, Andrew

### DIFF
--- a/app/services/description_validator.rb
+++ b/app/services/description_validator.rb
@@ -9,12 +9,13 @@ class DescriptionValidator
     @errors = []
   end
 
-  attr_reader :errors
-
   def valid?
     validate_duplicate_headers
     validate_title_headers
-    validate_matching_structured_title_headers
+    validate_title_value_for_type
+    validate_title_value_for_structured_type
+    validate_title_type_for_value
+    validate_title_type_for_structured_value
     validate_structured_title_types
     validate_header_paths
     validate_cell_values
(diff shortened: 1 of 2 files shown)


lwrubel (approved):
Approved, assuming I understood the requirements correctly.

PR #4833, opened 2025-08-20
Fix embargo location setting in access rights

### LINKED ISSUE #4830: Embargo rights menu does not correctly set view/download/location
**Describe the bug**
The embargo menu is not sending the correct values for download and location. Many combinations are not functioning: https://app.honeybadger.io/projects/49894/faults/122771087

The problems I've seen:

- citation-only doesn't send "none" for the download
- dark doesn't send "none" for the download
- if a location was set, the location is not removed when choosing world/stanford/citation/dark

All of these result in cocina validation errors and the user is taken to a "Bad Gateway" message.

**User Impact**
It's not possible to modify embargoes for most rights settings. In this case, the user wants the embargo removed and for the rights to be maintained as "citation-only." But I can't choose citation-only on the embargo menu.

**To Reproduce**
1. Go to https://argo.stanford.edu/view/druid:gq369xb7697
2. Try to change embargo to citation-only set rights to citation only when the embargo expires.

**Expected behavior**
The embargo rights menu should function like all of the other rights menus.

**Additional context**
I have not had time to confirm that the other rights menus are functioning properly. I suspect only the embargo menu is broken.

### LINKED ISSUE #4832: Remove "controlled digital lending" from the access rights menus
I had held off ticketing this because it sounded like CDL would be revisited after the FOLIO migration. The previous implementation was tied to Symphony and was turned off at the migration time. No items currently have CDL rights, but it's still possible to choose CDL as a rights option.

However, a lot of the old CDL code in the Access systems has been or is being removed, so I think we should also remove CDL as a choice in the Argo rights menu. That way no one can select it as an option.

The rights menu appears in the following places:

- item show page rights form
- bulk action rights
- registration form
- embargo form

**Additional information**

We may want to consider removing it from cocina models altogether but that would be a separate ticket.

aaron-collier:
# Why was this change made?

Fixes #4830 
Fixes #4832 

Andrew has verified this fix in stage. (As of 9/30/25)

This actually fixes 2 bugs related to this ticket. First, it resets the location selection to the blank option when the location menu is disabled -- this forces the location attribute to change on submission, otherwise we continue to send the existing location which creates invalid cocina (the main bug).

Second - this removes the cdl related fields and javascript as they are no longer used but because they were "disabled" in one path, a javascript error  was raised on the console, breaking the form. For now, this always sets `controlledDigitialLending` to false until we can remove the cdl related code (which is somewhat extensive).

# How was this change tested?

### DIFF
--- a/app/javascript/controllers/access_rights_controller.js
+++ b/app/javascript/controllers/access_rights_controller.js
@@ -73,24 +73,10 @@ export default class extends Controller {
     }
   }
 
-  downloadDropDown () {
-    const locationDropDown = document.getElementById('item_download_access') ||
-                             document.getElementById('embargo_download_access') ||
-                             document.getElementById('registration_download_access')
-
-    return locationDropDown
-  }
-
   disableDownload () {
     // ** Reset the download dropdown
     // * - This forces the download to change to blank when selecting disabled
-    const downloadDropDown = this.downloadDropDown()
-    const selectedDownloadIndex = downloadDropDown.options.selectedIndex
-    if (selectedDownloadIndex >= 0) {
-      downloadDropDown.options[selectedDownloadIndex].removeAttribute('selected')
-    }
-    downloadDropDown.options[3].setAttribute('selected', 'selected')
-    this.downloadTarget.value = 'none'
+    this.downloadTarget.selectedIndex = -1
     this.downloadRowTarget.hidden = true
     this.downloadTarget.disabled = true
   }
@@ -134,24 +120,10 @@ export default class extends Controller {
 
   disableLocation () {
     // ** Reset the location dropdown
-    // * - This forces the location to change to blank when selecting non location-based rights
-    const locationDropDown = this.locationDropDown()
-    const selectedLocationIndex = locationDropDown.options.selectedIndex
-    if (selectedLocationIndex >= 0) {
-      locationDropDown.options[selectedLocationIndex].removeAttribute('selected')
-    }
-    locationDropDown.options[0].setAttribute('selected', 'selected')
+    this.locationTarget.selectedIndex = -1
     this.locationTarget.value = null
     this.locationRowTarget.hidden = true
     this.locationTarget.disabled = true
-  }
-
-  locationDropDown () {
-    const locationDropDown = document.getElementById('item_access_location') ||
-                             document.getElementById('embargo_access_location') ||
-                             document.getElementById('registration_access_location')
-
-    return locationDropDown
   }
 
   enableLocation () {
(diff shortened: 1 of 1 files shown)


andrewjbtw:
In further testing of the other places the rights menu appears, I saw some issues that I think should prevent this from going to prod as it is. Transferring some of my comments [from Slack](https://stanfordlib.slack.com/archives/C09M7P91R/p1755811901613819):

- the "access location" is now showing in the rights menu even if the rights don't involve a location. This is on argo-stage on the registration form:
Screenshot 2025-08-22 at 12 28 10 PM

- also, the form behavior isn't adding or removing the location selector at the expected times, which means that the form can end up showing invalid combinations of rights:

https://github.com/user-attachments/assets/7f98128f-3c99-483e-b1ba-105ed8d49bce

Sometimes the location selector is visible when neither view nor download includes a location-based setting. And sometimes the location selector does not appear when both view and download are location-based.

app/javascript/controllers/access_rights_controller.js
```diff
@@ -84,7 +73,24 @@ export default class extends Controller {
     }
   }
 
+  downloadDropDown () {
+    const locationDropDown = document.getElementById('item_download_access') ||
```
justinlittman, line 77: These should be set as targets instead of doing lookups by id.

app/javascript/controllers/access_rights_controller.js
```diff
@@ -84,7 +73,24 @@ export default class extends Controller {
...
+
   disableDownload () {
+    // ** Reset the download dropdown
+    // * - This forces the download to change to blank when selecting disabled
+    const downloadDropDown = this.downloadDropDown()
+    const selectedDownloadIndex = downloadDropDown.options.selectedIndex
+    if (selectedDownloadIndex >= 0) {
+      downloadDropDown.options[selectedDownloadIndex].removeAttribute('selected')
+    }
+    downloadDropDown.options[3].setAttribute('selected', 'selected')
```
justinlittman, line 92: Setting the value by index is fragile.

app/javascript/controllers/access_rights_controller.js
```diff
@@ -128,25 +133,30 @@ export default class extends Controller {
...
 
   disableLocation () {
+    // ** Reset the location dropdown
+    // * - This forces the location to change to blank when selecting non location-based rights
+    const locationDropDown = this.locationDropDown()
+    const selectedLocationIndex = locationDropDown.options.selectedIndex
+    if (selectedLocationIndex >= 0) {
+      locationDropDown.options[selectedLocationIndex].removeAttribute('selected')
+    }
+    locationDropDown.options[0].setAttribute('selected', 'selected')
```
justinlittman, line 143: Same here.

justinlittman (changes requested):

PR #4841, opened 2025-09-18
Exclude etdSubmitWF from workflow grid

justinlittman:
# Why was this change made?
This wf is deprecated.

# How was this change tested?


mjgiarlo (approved):
LGTM. We should double-check with Cathy, Emma, and Andrew that this does not have historical value before we merge & deploy. (Unless that is already clear / known to you.)

justinlittman:
Already blessed by aforementioned.

PR #4856, opened 2025-10-03
Get apo collection for registration information from DSA instead of SOLR

### LINKED ISSUE #4824: Registration form takes a long time to populate fields after selecting an APO
**Describe the bug**
The Argo item registration form requires you to first choose an APO and then populates some of the form fields using that APO's default settings. Populated fields include collection, access rights, and initial workflow. The problem is that it can take a long time - minutes - for the form values to populate when choosing certain APOs. 

The Special Collections and University Archives APOs are good examples because many of them contain many collections, and there seems to be a direct relationship between number of collections and length of time to populate the form.

I recorded an example below:

https://github.com/user-attachments/assets/4cfeed4a-e748-40c4-ba50-9c5450f01eb4

**User Impact**
Users are having to wait a long time to be able to fill out the registration form. There's also a risk that if the form is submitted before all the fields are populated, incorrect data could be used for registration as the initial values in the form may not match the APO the user selected.

This issue may also be causing problems with a feature in registration that allows users to retain the form settings for multiple registration batches. I'm ticketing that as a separate issue. (Now ticketed: https://github.com/sul-dlss/argo/issues/4825)

**To Reproduce**
1. Go to the item registration form
2. Choose an APO like "Stanford University Libraries - Special Collections"
3. Wait for the collection list to be filled in. I think the other fields (like rights) also only fill in with a delay.

**Expected behavior**
The registration form should load the data for an APO without a lengthy delay. 

**Additional context**
I'm not sure if this has always been a problem. Some users report that it's been a problem for quite a while but had not reported it until it was reported on the #dlss-aaas channel today. I don't recall waiting this long for the collections list when we first launched the redesigned registration form around 2022.

aaron-collier:
# Why was this change made?

Fixes #4824 

# How was this change tested?


app/controllers/apo_controller.rb
mjgiarlo, line 42: We could also consider chaining `.sort_by { |collection_title| collection_title.first.downcase.delete('[]') }` (no `!`) to the `end` of the prior call and use its return value which should be marginally speedier than mutating what could be a large array. 

Also, not code you touched, but I think the block param here is the array holding the collection title and collection druid, so perhaps `collection_title` is misleading and maybe `collection_entry` or smth would be more intention revealing? 

☝🏻 is all non-blocking

PR #4889, opened 2025-10-09
Invalidate data upload if it wont index

### LINKED ISSUE #4871: When importing cocina descriptive from a spreadsheet, check if the resulting data can be indexed before applying it to the cocina
Gaps in cocina validation make it possible to upload cocina descriptive spreadsheets that create functionally invalid data. By "functionally invalid", I mean data that cannot be parsed by key SDR functions, including Argo indexing and purl publishing. This data has to be remediated before any further work can be done on the affected druids, and often the steps required to remediate the data are quite time-consuming.

As a stopgap to prevent invalid data from entering the system, I am proposing that we run a check to see if the data to be imported can be indexed in Argo before we commit the data to the cocina. 

I'm basing this request on the fact that in essentially every case I've seen of functionally invalid druids, the Argo reindex function has thrown an error on those druids. Preventing non-indexable data from entering the system may not catch all validation issues, but it will save user and repository manager time that would otherwise go towards remediation. The most common way invalid data enters the system is via bulk action, which means the number of druids to remediate is non-trivial.

**Illustration of the problem**

I've created a test druid and test data to illustrate this issue:

- [stage druid](https://argo-stage.stanford.edu/view/druid:bd196mw3407)
- CSV that will create invalid data [generate-indexing-error.csv](https://github.com/user-attachments/files/22752898/generate-indexing-error.csv)

I also created a CSV with valid data: 

[fix-bd196mw3407.csv](https://github.com/user-attachments/files/22752907/fix-bd196mw3407.csv) When testing this issue, we'll want to reupload valid data to prevent HB from repeatedly alerting on the indexing problem.

I've also made a screen recording of what it looks like when non-indexable data is uploaded on an Argo show page. In this sequence, the "Bad Gateway" screen seems to be the result of Argo attempting to reindex after importing the data. Ideally, we would be able to catch that issue before importing the data and then display a message on the show page modal. In the bulk action context, the same check should result in an error written to the log.

https://github.com/user-attachments/assets/02419a8e-2a52-4c83-8fda-6f12b7acfb22

I'm going to write a separate ticket for showing the reindex error details in Argo. HB has more detail than what Argo displays. In the past, when working with users I've had to relay the detail from HB to users. Showing the error in Argo, even if we haven't customized it to be more readable - would at least give users an idea of what's wrong. In my test data, the problem has to do with the parallel contributor values.

aaron-collier:
# Why was this change made?

Fixes #4871 

I'm not sure how to improve the test coverage report on these as they seem very opaque, but are tested.

Locally, this creates a log with:

```
2025-10-09 21:17:40 UTC Starting DescriptiveMetadataImportJob for BulkAction 2
CSV filename: generate-indexing-error.csv
Line 2: indexing failed for druid:sp239qc6098: undefined method 'valueLanguage' for an instance of Array for druid:sp239qc6098 (2025-10-09 21:17:40 UTC)
2025-10-09 21:17:40 UTC Finished DescriptiveMetadataImportJob for BulkAction 2
```

# How was this change tested?

### DIFF
--- a/spec/jobs/descriptive_metadata_import_job_spec.rb
+++ b/spec/jobs/descriptive_metadata_import_job_spec.rb
@@ -103,9 +103,10 @@ RSpec.describe DescriptiveMetadataImportJob do
 
     context 'when index validation fails' do
       let(:ability) { instance_double(Ability, can?: true) }
+      let(:response) { instance_double(Faraday::Response, status: 422, body: nil, reason_phrase: 'Example field error') }
 
       before do
-        allow(Dor::Services::Client.objects).to receive(:indexable).and_raise(StandardError, 'Simulated index failure')
+        allow(Dor::Services::Client.objects).to receive(:indexable).and_raise(Dor::Services::Client::UnprocessableContentError.new(response:))
         allow(Ability).to receive(:new).and_return(ability)
         subject.perform(bulk_action.id, { csv_file:, csv_filename: filename })
       end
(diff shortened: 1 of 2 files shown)


spec/jobs/descriptive_metadata_import_job_spec.rb
```diff
@@ -100,6 +100,33 @@
       end
     end
 
+    context 'when index validation fails' do
```
justinlittman: Seems like (1) DSC should be mocked and (2) that the error is written to the log should be tested.

PR #4905, opened 2025-10-14
Refactor bulk actions

justinlittman:
# Why was this change made?
So that bulk actions are more consistent, and therefore more maintainable.

# How was this change tested?
Unit


app/jobs/add_workflow_job.rb
```diff
@@ -2,26 +2,29 @@
 
 ##
 # Job to add a workflow to and object
-class AddWorkflowJob < GenericJob
-  attr_reader :workflow_name
+class AddWorkflowJob < BulkActionJob
```
justinlittman: Most of the boilerplate is in BulkActionJob (or BulkActionCsvJob for bulk actions that take CSV as input).

Each bulk action has to implement a subclass of BulkActionItem which performs processing for a single item.

app/jobs/add_workflow_job.rb
```diff
@@ -2,26 +2,29 @@
...
+  def workflow_name
+    params.fetch(:workflow)
```
justinlittman: A bulk action might have parameters, which are made available to BulkActionItems (see delegate below).

app/jobs/add_workflow_job.rb
```diff
@@ -2,26 +2,29 @@
...
+  end
+
+  class BulkActionItem < BulkActionItem
+    delegate :workflow_name, to: :job
+
+    def perform
+      return unless check_update_ability?
```
justinlittman: BulkActionItem provides re-usable methods like success!, failure!, and check_update_ability?

app/jobs/add_workflow_job.rb
```diff
@@ -2,26 +2,29 @@
...
 ##
 # Job to add a workflow to and object
-class AddWorkflowJob < GenericJob
-  attr_reader :workflow_name
+class AddWorkflowJob < BulkActionJob
+  def workflow_name
+    params.fetch(:workflow)
+  end
+
+  class BulkActionItem < BulkActionItem
```
aaron-collier: `BulkActionItem < BulkActionItem` seems incorrect. Is this expected?

aaron-collier (approved):

PR #4918, opened 2025-10-16
Add header to checksum report CSV

### LINKED ISSUE #4912: Add headers to the checksum report
The checksum report is a bulk action that returns filenames, file sizes, and checksums for a set of druids. This report has historically never included headers and was originally used only in certain workflows where people had scripts to parse the report.

We now have more users interested in checksum reports but the lack of headers makes it hard to understand what's in the report.

The headers should be:

- druid
- filename
- md5
- sha256
- sha512
- size

mjgiarlo:
# Why was this change made?

Fixes #4912

# How was this change tested?

- [x] CI
- [x] QA

### DIFF
--- a/app/jobs/checksum_report_job.rb
+++ b/app/jobs/checksum_report_job.rb
@@ -1,15 +1,7 @@
 # frozen_string_literal: true
 
-##
-# Job to run checksum report
+# A job that, given list of druids of objects, runs a checksum report using a presevation_catalog API endpoint and returns a CSV file to the user
 class ChecksumReportJob < BulkActionJob
-  ##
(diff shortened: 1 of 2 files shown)


app/jobs/checksum_report_job.rb
```diff
@@ -1,31 +1,28 @@
 # frozen_string_literal: true
 
-##
-# Job to run checksum report
```
mjgiarlo: No need to document a class in two places. Move the more descriptive comment below up to serve as the class comment.

aaron-collier (approved):

PR #4923, opened 2025-10-20
Adds support for closing to bulk actions.

### LINKED ISSUE #4863: Close versions after running bulk action depending on user input
In the bulk action modal, for every action that requires an open version, allow the user to choose whether to close the version after the action is completed. 

#3683 will standardize bulk actions so that versions are opened as needed before the bulk action starts; we want similar functionality for version-closing, but we want to allow the user to opt out of automatic version-closing in case there are additional bulk actions they want to run before closing the version. If user chooses to close versions, then Argo should run the close-version task after completing the bulk action.

Design: User could make their choice via radio button (similar to the one in the "Manage release" bulk action modal) or checkbox. Options should be labeled something like "Close open versions" or "Don't close open versions". If the user chooses not to close open versions, there should be a message saying something like "Versions must be closed for changes to take effect." 

Possible questions for discussion:
* Version closure is not tracked in the downloadable bulk action log file, or evaluated as part of the success of the action. Should it be? Are there cases where closure might fail and the user would want to know about it, or can we assume it will be successful?
* Will each object be closed as soon as it's acted upon, or will all objects be closed at the very end of the action? (I'd prefer the former.)
* In the UI, should the pre-selected option be "close" or "don't close"?
* If the user has run multiple bulk actions on the open version, but is closing the version as part of the current bulk action, do we need to specify what the version name should be?

**Requires prod deployment coordination and user communication.**

justinlittman:
closes #4863

# Why was this change made?
* User joy (the close checkbox)
* Developer joy (refactored specs)

# How was this change tested?


app/controllers/bulk_actions/apply_apo_defaults_jobs_controller.rb
```diff
@@ -5,5 +5,9 @@ class ApplyApoDefaultsJobsController < ApplicationController
     include CreatesBulkActions
 
     self.action_type = 'ApplyApoDefaultsJob'
+
+    def job_params
+      super.merge(close_version: params[:close_version])
```
justinlittman: This needs to be added to the controller for each job.

app/jobs/checksum_report_job.rb
```diff
@@ -2,19 +2,16 @@
...
   end
 
   def export_file
-    @export_file ||= CSV.open(report_filename, 'w').tap do |csv|
-      # The CSV header needs to be injected before any item-level CSV is
-      # generated. If this is moved into the job item's `#perform` method, the
-      # header will be repeated once for every druid.
-      csv << %w[druid filename md5 sha256 sha512 size]
-    end
+    @export_file ||= CSV.open(report_filename, 'w', write_headers: true, headers: HEADERS)
```
justinlittman: A wee bit more idiomatic.

app/jobs/open_version_job.rb
```diff
@@ -20,27 +20,4 @@ def perform
 
     delegate :description, to: :job
   end
-
```
justinlittman: Cruft

app/jobs/register_druids_job.rb
```diff
@@ -62,59 +62,4 @@ def row
       ]
     end
   end
-  # queue_as :default
```
justinlittman: Cruft

app/jobs/register_druids_job.rb
```diff
@@ -62,59 +62,4 @@ def row
...
-
```
mjgiarlo: What's going on with all the commented-out code here? Cruft? WIP? etc.

mjgiarlo (approved):
Approve pending consideration yada yada

app/jobs/set_catalog_record_ids_and_barcodes_csv_job.rb
```diff
@@ -58,7 +59,7 @@ def log_catalog_record_id_update
     end
 
     def log_barcode_update
-      return unless change_set.changed?(:barcode)
+      return unless change_set.is_a?(ItemChangeSet) && change_set.changed?(:barcode)
```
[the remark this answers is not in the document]
justinlittman: Yes, CollectionChangeSets don't have a barcode attribute. Adding comment.

PR #4926, opened 2025-10-20
Expand/collapse event hash like cocina models hash

### LINKED ISSUE #4913: Enhancements to Events section on show item page
1. Change the date to this format: yyyy-mm-dd 
2. Instead of the 'view more/view less' for the sections that can be expanded/collapsed, use the [+] / [-] icons

Image

aaron-collier:
# Why was this change made?

Fixes #4913 

https://github.com/user-attachments/assets/399e6f55-efe0-4a4b-9eac-729594044495

# How was this change tested?

### DIFF
--- a/app/presenters/event_hash_presenter.rb
+++ b/app/presenters/event_hash_presenter.rb
@@ -15,17 +15,17 @@ class EventHashPresenter
     DeepCompactBlank.run(enumerable: data)
   end
 
-  # Determine if the data has nested structures that can be expanded/collapsed
+  # Determine if the event has a nested structure.
   #
-  # If the data structure isn't an expandable hash, return false
-  # Otherwise, check if any value is a hash or an array containing hashes/arrays.
+  # If the event has any values that are arrays or hashes,
+  # render the expand/collapse controls. If not, e.g. an
+  # event with only string values, do not render the controls
+  # because they are noisy and misleading.
   #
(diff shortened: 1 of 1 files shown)


app/views/events/show.html.erb
```diff
@@ -6,16 +6,28 @@
...
+        <td>
+          <div data-controller="events-json-renderer">
+            <div class="row">
+              <div class="col-2">
+                <a class="btn btn-primary" data-action="events-json-renderer#toggle">
+                  <span data-events-json-renderer-target="toggle" class="bi-plus" />
+                </a>
+                <script data-events-json-renderer-target="events">
+                <%= EventHashPresenter
+                      .new(event: [event.data])
```
aaron-collier: As is, this needs to be an array in the innerText (see note in the javascript controller)

config/locales/en.yml
```diff
@@ -29,6 +29,7 @@ en:
       am: am
       formats:
         default: "%Y-%m-%d %I:%M%p"
+        event: "%Y-%m-%d"
```
mjgiarlo: It looks like Rails has a couple built-in date formats that are identical: `db` and `inspect`. Should we use a built-in or is `event` more descriptive?

mjgiarlo: Date formats are cheap, so I buy that we want this in our domain language.

PR #4963, opened 2025-10-29
Log instead of alert on Turbo fetch error

edsu:
# Why was this change made?

These alerts were causing problems for integration tests that needed to handle them when the speed at which the tests were operating caused the `turbo:fetch-request-error` to be triggered. Removing the alert also made a workaround in one of the Argo unit tests no longer necessary.

The alternative is peppering multiple `sleep` calls into 7 different feature tests. There may be some way to instrument rspec's `visit` to do something to avoid the alerts, since it seems most of the `sleep` calls needed to be added before those...

# How was this change tested?

Unit tests & integration tests.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -27,7 +27,7 @@ document.addEventListener('turbo:before-fetch-response', async (event) => {
 })
 
 document.addEventListener('turbo:fetch-request-error', async (event) => {
+  console.log("refreshing page after turbo:fetch-request-error")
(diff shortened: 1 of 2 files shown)


edsu:
The reload still happens, it just doesn't alert about it. Also, the alert about stale Authentication is still in place:

https://github.com/sul-dlss/argo/blob/39b982f407f57c76ecfa0c8ba8376adba1410f02/app/javascript/application.js#L18-L27

PR #4966, opened 2025-10-29
Stream reports.

### LINKED ISSUE #4245: Speed up downloads of large reports

justinlittman:
closes #4245

# Why was this change made?
Faster reports.

# How was this change tested?


mjgiarlo (approved):

app/models/report.rb
```diff
@@ -28,23 +28,17 @@ class << self
...
-      field: 'druid', label: 'Druid',
-      proc: ->(doc) { doc.druid },
+      field: SolrDocument::FIELD_BARE_DRUID, label: 'Druid',
       sort: true, default: true, width: 100, formatter: 'linkToArgo',
       category: FREQUENTLY_USED_CATEGORY
     },
     {
-      field: 'purl', label: 'Purl',
-      proc: ->(doc) { "#{Settings.purl_url}/#{doc.druid}" },
-      solr_fields: %w[id],
```
justinlittman: solr_fields is no longer supported since each field corresponds with a single solr field.

app/models/report.rb
```diff
@@ -306,46 +274,52 @@ def druids(opts = {})
...
-  def to_csv
-    @params[:page] = 1
-    @params[:per_page] = ROWS_PER_PAGE_CSV
-    Enumerator.new do |yielder|
-      yielder << CSV.generate_line(@fields.map { |field| field.fetch(:label) }, force_quotes: true) # header row
-      (@response,) = search_results(params)
-      until @response.documents.empty?
-        report_data.each do |record|
-          yielder << CSV.generate_line(@fields.map { |field| record[field.fetch(:field).to_sym].to_s }, force_quotes: true)
+  def stream_csv(stream:)
```
justinlittman: This is where the streaming magic happens.

PR #4968, opened 2025-10-29
[DRAFT] Add initial decommission service

aaron-collier:
# Why was this change made?

Fixes #5599

DRAFT notes:

- I'm working through adding an un-release endpoint to DSA in order to be able to add all `release: false` tags before reindexing and running the `releaseWF` in order to avoid triggering those more than once at a time for the same object.
- I need to add the bulk option here. My intention is to allow a csv file with individual reasons/tickets to allow bulk decommissioning of multiple tickets at once that can be overridden on the command line

# How was this change tested?

### DIFF
--- a/app/services/decommission_service.rb
+++ b/app/services/decommission_service.rb
@@ -20,7 +20,10 @@ class DecommissionService
                          description: "Decommissioned: #{reason}",
                          opening_user_name: sunetid)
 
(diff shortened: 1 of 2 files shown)


justinlittman:
Good news, bad news. Bad: This service should be in DSA, not Argo. Good: That should make it easier.

PR #5010, opened 2026-01-08
Set the Solr URL in one place and use it everywhere

mjgiarlo:
# Why was this change made?

This will reduce the number of steps when performing maintenance such as Solr cluster upgrades and will increase clarity about where changes need to be made.

# How was this change tested?

CI and will be tested in stage

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,7 +149,7 @@ jobs:
       SETTINGS__DOR_SERVICES__URL: http://dor-services-app:3000
       SETTINGS__REDIS_URL: redis://redis:6379/
       SETTINGS__SDR_API__URL: http://sdr-api:3000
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -149,7 +149,7 @@ jobs:
       SETTINGS__DOR_SERVICES__URL: http://dor-services-app:3000
       SETTINGS__REDIS_URL: redis://redis:6379/
       SETTINGS__SDR_API__URL: http://sdr-api:3000
-      SETTINGS__SOLRIZER_URL: http://solr:8983/solr/argo
```
mjgiarlo: Solrizer is no longer in any SDR stacks, including Argo. Standardize on `Settings.solr.url`.

Issue #5023, opened 2026-02-06
cocina descriptive spreadsheet upload: if a type is supplied but no main value is supplied that corresponds to the type, drop the type


andrewjbtw:
The title of this issue is a mouthful, but this comes out of discussion of https://github.com/sul-dlss/cocina-models/issues/835

We want to start validating that when a type is supplied, a main value is supplied to go with that type. There are many examples of what this means in the linked ticket, a relatively straightforward one is:

- if there is a contributor, and a contributor name (as a type), there should be a contributor (as the main value to go with the name)

However, we can't turn this validation on unless we also address how data with this problem gets into SDR. And the main route is the cocina spreadsheet upload. If we turned on validation without addressing spreadsheet uploads there would be a user experience problem because many spreadsheets that users currently upload would be rejected as invalid, and they would likely need to spend significant time re-editing those spreadsheets. 

The problem comes down to using a spreadsheet structure (and UI) for batches of items where the metadata can vary quite a bit on a per-item basis. A spreadsheet may need enough columns to support 4 event types because there are items that need 4 event types. But that can leave items with only 1 event type having "empty" values for the other 3 event types. A user could edit row by row but we've gotten feedback that this would make it very difficult to work on large projects. Longstanding practice has been to fill entire columns with a type, and add a main value to go with it only on rows where necessary. 

The current SDR spreadsheet upload generally allows this for many types - title may be an exception - but the result is that we've been storing a lot of messy data: types without main values. No filtering or validation takes place to prevent this data from entering the system. We then have to work around this messy data when parsing cocina. This has come up a lot with respect to cocina display.

**Requested behavior**

If there is a descriptive type without a corresponding main value, drop the type when turning the spreadsheet into JSON. Cocina validation would come after this step.

Also, we do not need to log when this filtering takes place, neither as an error nor a warning. This is a judgement call based on our understanding that when a type is supplied without a corresponding value, that is usually the correct metadata: there is no "missing" value, and the type is only present as an artifact of bulk spreadsheet creation. If a user did mean to supply a value to go with a type, we have no way to reliably detect that. Catching that kind of error is ideally a part of metadata QC, and not left up to the system.

**Why not reject these cases as invalid?**

We've considered this option. If we rejected as invalid, the uploader would have to go back and adjust each invalid row in their spreadsheet. Reasons we are not in favor of going this route:

- It adds to the user's burden in creating a spreadsheet, something many users already find challenging.
[...]

PR #5025, opened 2026-02-09
Do not import languages without a value

jcoyne:
# Why was this change made?

Ref https://github.com/sul-dlss/cocina-models/issues/835
Ref #5023

# How was this change tested?
CI


lwrubel (approved):
approved with edit

PR #5039, opened 2026-02-17
Render descriptive metadata from CocinaDisplay

### LINKED ISSUE #5036: Use CocinaDisplay to render Description modal instead of via ModsDisplay
Modal contents are rendered via: https://github.com/sul-dlss/argo/blob/main/app/views/metadata/descriptive.html.erb

This will bring the modal in line with PURL display and help with reviewing display before publishing.

jcoyne:
# Why was this change made?

Fixes #5036 

# How was this change tested?
ci

### DIFF
--- a/app/components/cocina_descriptive_component.rb
+++ b/app/components/cocina_descriptive_component.rb
@@ -10,20 +10,20 @@ class CocinaDescriptiveComponent < ApplicationComponent
   def cocina_display_methods
     %i[
       title_display_data
-      contributor_display_data
-      language_display_data
+      form_display_data
+      form_note_display_data
       publication_display_data
       event_date_display_data
       event_note_display_data
-      subject_display_data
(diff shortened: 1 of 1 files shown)


app/components/cocina_descriptive_component.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class CocinaDescriptiveComponent < ApplicationComponent
+  def initialize(cocina_display:)
+    @cocina_display = cocina_display
+  end
+
+  attr_reader :cocina_display
+
+  def cocina_display_methods
+    %i[title_display_data contributor_display_data language_display_data
```
jcoyne: add `subject_display_data`

jcoyne:
This is waiting on a release of cocina-display so we can add `publication_display_data`

PR #5060, opened 2026-03-02
Update Solr dockerfile to use latest Solr version

mjgiarlo:
# Why was this change made?

To fix the CircleCI build.

# How was this change tested?

CI

### DIFF
--- a/docker/Dockerfile-solr
+++ b/docker/Dockerfile-solr
@@ -1,5 +1,7 @@
-FROM solr:9.6.1
+FROM solr:9.10.0
 
 COPY solr_conf/conf/* /myconfig/
 
-CMD ["solr-create", "-c", "argo", "-d", "/myconfig"]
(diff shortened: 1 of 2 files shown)


docker/Dockerfile-solr
```diff
@@ -1,5 +1,7 @@
-FROM solr:9.6.1
+FROM solr:9.10.0
 
 COPY solr_conf/conf/* /myconfig/
 
-CMD ["solr-create", "-c", "argo", "-d", "/myconfig"]
+ENV SOLR_OPTS="-Dsolr.config.lib.enabled=true"
+
+CMD ["solr-precreate", "argo", "/myconfig"]
```
mjgiarlo: Also not the fix. I thought we should align the built Solr version & config with what's running in production and what's setup in compose.yaml.

PR #5071, opened 2026-03-16
Do not remove blank columns for ImportTagJob

### LINKED ISSUE #5065: Bulk action tag import skips some tags
**Describe the bug**
Bulk action tag import is not reliably importing all of the tags in a CSV. I think there may be an issue with how the import handles spreadsheets of variable column width.

**User Impact**
Some users have ended up manually updating items when they've noticed missing tags. Also, users may not even see that certain tags have not been applied.

**To Reproduce**

I made a spreadsheet to use in argo-stage that has two columns of tags where the tags in the second column do not get imported:  [test-missed-tags.csv](https://github.com/user-attachments/files/25727371/test-missed-tags.csv)

1. Go to new bulk actions
2. Run the bulk action with that spreadsheet
3. Go to an item that should have two tags
4. See that it only has the tag in the first column

**Expected behavior**
Tag import should be able to handle tags where some rows have more tags than others.

**Possible source of the problem**
I think the tag import may be treating the first row as a header and determining the number of columns from just that row. So if the first row is narrower than subsequent rows, the columns that are blank on the first row are ignored.

To test that theory, I created another tag spreadsheet for stage where I put two tag columns in the first row: [test-included-tags.csv](https://github.com/user-attachments/files/25850425/test-included-tags.csv)

Other than that one change, it's similar to the other test CSV that reproduces the problem. When importing this spreadsheet, both columns of tags get applied on import.

aaron-collier:
Fixes #5065 

Sometimes, we do not want to remove columns if there are blanks. This is currently true for the ImportTagJob where we don't actually have headers/fields list we do in other CSV based imports, we often have just a druid column and potentially multiple tag columns (variable between rows).

Log output from a local shows this now includes the columns even if they are not the first object:

```
2026-03-17 20:43:29 UTC Starting ImportTagsJob for BulkAction 22
2026-03-17 20:43:29 UTC  - line 2 - Replaced tags (tags : test) for druid:bc903cd4962
2026-03-17 20:43:29 UTC  - line 3 - Replaced tags (tags : test, tags : another tag) for druid:cc466fh7891
2026-03-17 20:43:29 UTC  - line 4 - Replaced tags (tags : test) for druid:gm706hs4976
```


aaron-collier:
NOTE: As of right now, this probably isn't the best approach - I will be reworking ImportTagsCsvConverter into this job.

https://github.com/sul-dlss/argo/blob/main/app/jobs/import_tags_job.rb#L43

PR #5086, opened 2026-03-24
Colorize JSON (e.g., Cocina), shift disclosure controls, and make all links non-selectable

mjgiarlo:
Here's a little spike that re-styles the Cocina JSON (and the event JSON) making the following tweaks:

* Bump the JSON font size up slightly (<1em)
* Colorize the JSON using the Stanford palette
* Move expand/collapse controls to a fixed location (left-most within the containing div)
* Make sure expand/collapse controls and collapse text (e.g., `3 items`) are not user-selectable, which keeps them out of copy/pastes (ensuring the copied JSON is still valid JSON)
* Tweak the style of the expand/collapse controls (from +/- icons to caret right/down icons)

Happy to get your impressions on these changes, @andrewjbtw and @astridu. Of course any/all of this is subject to change. We could even dump the whole thing. Only a spike.

# Before

argo_cocina_before

# After

argo_cocina_after

### DIFF
--- a/app/assets/stylesheets/renderjson.css
+++ b/app/assets/stylesheets/renderjson.css
@@ -1,12 +1,60 @@
 .renderjson {
-  /* Hide top level disclosure */
-   > span > .object > .disclosure:first-of-type {
+  /* Bump up the font size slightly to make the JSON more readable. The default
+   font size is 0.875em. */
+  font-size: 0.95em;
+
+  /* Allow elements to be positioned absolutely within the container. This has the
+   effect of shifting the expand/collapse controls so that their positioning is
+   consistent. */
+  position: relative;
+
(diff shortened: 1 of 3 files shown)


mjgiarlo:
@astridu 💬 
> Looking good! I see why you would want to add some color to the JSON. Can I suggest using our teal and orange so it's not so Christmas-y?
> 
> Lagunita: 007C92 Dark Orange: C15600

Yes! Which thingies (strings, numbers, booleans, braces, commas, yada yada) should change to these colors?
 
> Regarding the arrows:
> 
> 1. can you change the gray of the arrows to: 767674 so it's accessible?
> 2. Is it possible to make the arrows just a tiny bit bigger so they are easier to click?

Definitely, yes, thank you :)

astridu:
String and number are similar, but they both contrast to key so I think it should work.

PR #5131, opened 2026-05-22
Upgrade to Blacklight 9

jcoyne:
# Why was this change made?

# How was this change tested?


justinlittman:
Just to say out loud: We're going to be working on the successor to Argo, which will not be on Blacklight.

jcoyne:
This resolves the 2 security errors found in https://github.com/sul-dlss/argo/security/dependabot

PR #5148, opened 2026-06-16
Refactor CsvUploadNormalizer to use proper CSV.read for raw csv files

aaron-collier:
# Why was this change made?

This resolved the issue found in this thread: https://stanfordlib.slack.com/archives/C09M7P91R/p1781562951022659

I've confirmed by running the `zero-druids-processed.csv` file in the thread into `CsvUploadNormalizer.read` on main producing only the header line, and then through this branch produced the full data. Tests still pass.

# How was this change tested?


app/services/csv_upload_normalizer.rb
```diff
@@ -59,18 +55,31 @@ def normalized_file_extension
...
-    csv = File.read(path, encoding: 'bom|utf-8')
-    lines = csv.split(/\r?\n/)
-    lines.shift until lines.first&.match?(/\Adruid,/i) || lines.empty?
-    return csv if lines.empty?
-
-    lines.join("\n")
+  # 1. Reads a CSV with the standard library if the extention is CSV
+  # 2. Otherwise reads the file with ROO and parses into a CSV
+  def normalized_csv_file
+    return CSV.read(path, skip_blanks: true, converters: whitespace_only_nullifier) if normalized_file_extension == '.csv' # .map { |line| line.map { |field| field }.join(",") }.flatten.join("\n")
```
justinlittman: Cruft?

app/services/csv_upload_normalizer.rb
```diff
@@ -59,18 +55,31 @@ def normalized_file_extension
...
-  def normalized_csv_string
-    csv = File.read(path, encoding: 'bom|utf-8')
-    lines = csv.split(/\r?\n/)
-    lines.shift until lines.first&.match?(/\Adruid,/i) || lines.empty?
-    return csv if lines.empty?
-
-    lines.join("\n")
+  # 1. Reads a CSV with the standard library if the extention is CSV
+  # 2. Otherwise reads the file with ROO and parses into a CSV
+  def normalized_csv_file
```
justinlittman, line 82: What is this returning?

app/services/csv_upload_normalizer.rb
```diff
@@ -59,18 +55,31 @@ def normalized_file_extension
...
+  # Given raw CSV data from normalized_csv_file remove any preamble
+  # rows before the proper header row based on DRUID_HEADER
+  def normalized_csv_data
+    raw = normalized_csv_file
+
+    # Find the first row that starts with DRUID_HEADER
+    druid_index = raw.index { |line| line.first.casecmp?(DRUID_HEADER) }
+
+    # DROP all preamble rows leading up to the proper header row
+    csv_string = CSV.generate do |csv|
```
justinlittman, line 70: Why does this have to go to a string and back?

aaron-collier: I tried to avoid this, but it seems the best approach to determining which line in the file starts with `druid` and delete all lines up to that point, then re-parse WITH headers so that we get a CSV table with headers instead of just an array of unlabeled fields. Open to suggestion, but this seemed like the best approach (which was original done in `normalized_csv_string`.

app/services/csv_upload_normalizer.rb
aaron-collier: Add param comment

justinlittman (approved):

PR #5152, opened 2026-06-19
drop empty form note if no value present

### LINKED ISSUE #6144: Cocina descriptive spreadsheet upload: drop form.note if main value not present
Follow the same rules as for a top-level note field - see https://github.com/sul-dlss/argo/issues/5051.

Example spreadsheet with empty form notes present (displayLabel only in form13): 

[descriptive-267.csv](https://github.com/user-attachments/files/29144990/descriptive-267.csv)

(These objects will be remediated in prod.)

peetucket:
# Why was this change made?

Fixes https://github.com/sul-dlss/dor-services-app/issues/6144

# How was this change tested?

Spec

### DIFF
--- a/app/services/description_import_filter.rb
+++ b/app/services/description_import_filter.rb
@@ -70,6 +70,7 @@ class DescriptionImportFilter
   end
 
   def remove_form_without_value(forms)
+    Array(forms).each { |form| remove_note_without_value(form[:note]) }
     Array(forms).delete_if { !descriptive_value_sufficient?(it) }
   end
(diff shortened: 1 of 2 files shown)


justinlittman (approved):
Approve pending green.

peetucket:
> Approve pending green.

due to a cocina-models mismatch between apps currently.  should be sorted by dependency-updates (just fyi, since I am out monday-wednesday next week if you want to re-run the test and move this forward before I am back)

PR #5160, opened 2026-06-25
use different folioclient method to check for instance record

### LINKED ISSUE #5149: Check for instance not MARC record when registering
Use `fetch_instance_info` instead of `fetch_marc_hash` when checking whether a record exists in FOLIO for an HRID. 

https://github.com/sul-dlss/argo/blob/main/app/controllers/registrations_controller.rb#L60C8-L60C9

Make sure the integration test `/spec/features/accessioning/item_creation_with_folio_hrid_spec.rb` still passes as well.

peetucket:
# Why was this change made?

Fixes #5149 

~~HOLD until verify that /spec/features/accessioning/item_creation_with_folio_hrid_spec.rb test passes~~
Integration tests passed

# How was this change tested?

- Updated spec
- Integration test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,6 +180,9 @@ jobs:
           key: gems-v2
       - node/install-packages:
           pkg-manager: yarn
+      - run:
+          name: Build JS assets
+          command: yarn build
       # Check DB status
       - run:
           name: Wait for DB
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -59,7 +59,7 @@ class RegistrationsController < ApplicationController
   # Allow the front end to check if a catalog record ID exists
   def catalog_record_id
     resp = begin
-      FolioClient.fetch_marc_hash(instance_hrid: params[:catalog_record_id])
+      FolioClient.fetch_instance_info(hrid: params[:catalog_record_id])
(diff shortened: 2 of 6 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -59,7 +59,7 @@ def source_id
   # Allow the front end to check if a catalog record ID exists
   def catalog_record_id
     resp = begin
-      FolioClient.fetch_marc_hash(instance_hrid: params[:catalog_record_id])
+      FolioClient.fetch_instance_info(hrid: params[:catalog_record_id])
```
peetucket, line 62: (1) different method signature, updated it
(2) different response structure, but it doesn't matter, because we only care about existence of the record
(3) confirmed we get the same exception if record not found, so no need to change the rescue clause

.circleci/config.yml
```diff
@@ -180,6 +180,9 @@ jobs:
           key: gems-v2
       - node/install-packages:
           pkg-manager: yarn
+      - run:
+          name: Build JS assets
+          command: yarn build
```
peetucket, line 185: some flaky tests were resolved by forcing JS assets to rebuild when running tests (also did it locally to fix my local flakiness)

justinlittman (approved):

PR #5166, opened 2026-06-29
Bumped pres client and unpinned other deps.

justinlittman:
# Why was this change made?

# How was this change tested?


mjgiarlo (approved):

mjgiarlo:
I wonder if this change might be worth running a few integration tests?

justinlittman:
I don't think integration tests will touch the actual dependencies that are updated, so deploying to QA and doing some focused, manual testing.

justinlittman:
Manual testing successful.

PR #5169, opened 2026-06-29
marc availability feedback when registering

### LINKED ISSUE #5150: Provide feedback about MARC unavailability when registering
Provide user with a message concerning druids that do not have a FOLIO MARC record available. Currently, the page which lists the druid(s) hardcodes the message: https://github.com/sul-dlss/argo/blob/main/app/views/registrations/create_status.html.erb. The idea is to add client-side checking of the record source. 

 - [ ] Design: https://github.com/sul-dlss/argo/issues/4922#issuecomment-4685002245 (and see comment below it about druid info placement). 
- [ ] Add client-side check. To determine which druids do not have MARC records yet, consider the same approach as in the registrations page, with a stimulus controller pointing to a controller action to look up the HRID and see if a MARC record is found. (FolioClient.fetch_marc_hash).  https://github.com/sul-dlss/argo/blob/5483e19d0b9a4bde1d034e7697ec9bdbfeb593c2/app/controllers/registrations_controller.rb#L60. Note that you will probably need to create a new controller action since the existing catalog_record_id method was changed to look up a FOLIO instance record, not MARC record, in [#5149](https://github.com/sul-dlss/argo/pull/5160). The message in this ticket is specifically for warning the user about a MARC record not existing.

peetucket:
# Why was this change made?

Fixes #5150 - feedback to users on MARC availability for Folio records when registering

Check happens after registering on the server, and is loaded via turbo after the page is rendered.  This allows the registration table to render immediately instead of being blocked by the Folio checks.  After the checks for MARC records are complete, the results will be rendered via turbo into a warning dialog box.

If there are any items to check, a "loading" type message is shown so the user knows something is happening and this is either removed once the check is complete if there are no errors or replaced with a warning dialog if there are problems.

When running:

Screenshot 2026-06-30 at 3 00 34 PM

If problems:
 
Screenshot 2026-06-29 at 3 34 07 PM

Note that if there are many items to check, the server call may take a while.  We are also passing the payload consisting of druids/hrids back to the server to run the check async, for a very large of objects, this may be a large payload (but no larger than the one we passed to register the objects in the first place).

Note: Claude assisted code.

# How was this change tested?

Specs, localhost


app/controllers/registrations_controller.rb
```diff
@@ -73,6 +73,23 @@ def catalog_record_id
...
+      false
+    rescue FolioClient::Error => e
+      # In production, this will prevent false negatives when ILS (Folio) is not available.
+      # In other environments, where Folio is never available, it will assume MARC exists.
+      return render plain: e.message, status: :bad_gateway if Rails.env.production?
+
+      true
+    end
+    render json: resp.to_json, layout: false
+  end
```
peetucket, line 101: this is basically the endpoint method that was recently changed in https://github.com/sul-dlss/argo/pull/5160

this creates a new endpoint with similar logic that we will use here

peetucket:
Note: test records on folio-test-api.stanford.edu

**HRID without MARC:**
in00000206878
in00000997501
in00000208217

**HRID with MARC:**
in00000378391
a11971843
a8391050

app/views/registrations/create_status.html.erb
```diff
@@ -1,6 +1,11 @@
 <h1>DOR item(s) submitted</h1>
 
 <p>Items successfully registered.</p>
+
+<% if @folio_items_to_check.any? %>
+  <turbo-frame id="marc_warnings" src="<%= marc_warnings_registration_path(items: @folio_items_to_check) %>"></turbo-frame>
+<% end %>
```
peetucket, line 9: only need to embed the turbo-frame and fire off the check if there are any objects with folio instance hrids

app/views/registrations/marc_warnings.html.erb
```diff
@@ -0,0 +1,17 @@
...
+        <p class="mb-1">The following druids do not have MARC records. You can work with these items in Argo, but to release these items to SearchWorks, the MARC record must be created first.</p>
+        <ul class="mb-0">
+          <% @warnings.each do |warning| %>
+            <li><%= link_to warning[:druid], solr_document_path("druid:#{warning[:druid]}"), target: '_top' %></li>
+          <% end %>
+        </ul>
+      </div>
+      <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
+    </div>
+  <% end %>
```
peetucket, line 16: only need to render the warning box if there are any objects with problems

lwrubel (approved):
Thanks for thinking through a server-side way to do this to handle larger numbers of objects in the form. Approved pending PO testing in a deployed environment.

Issue #5174, opened 2026-06-30
Update Consul documentation for: registration form CSV option should use "title" instead of "label" as a column header


andrewjbtw:
- [ ] Update [Consul documentation on registration form](https://consul.stanford.edu/spaces/SDR/pages/139654960/Filling+out+the+registration+form)

The other CSV updates/downloads below will be handled in https://github.com/sul-dlss/argo/pull/5176, once completed.
- - -
We are removing the cocina top-level "label" field in favor of the descriptive title: see https://github.com/sul-dlss/argo/issues/5172

Since the `label` field will be renamed `title` in the form, so we should also rename the CSV column to match. 

The registration form CSV option is available by toggling to the "Upload CSV" option below the top-level form fields.

<img width="307" height="243" alt="Image" src="https://github.com/user-attachments/assets/e93db43d-62f7-47ef-ad30-512d384cff63" />

Note that users can also download a blank template, which we will also need to update.
- [ ] upload should accept `title` instead of `label`
- [ ] template CSV should use `title` instead of `label`

lwrubel:
Update RegisterDruidsJob as part of this workflow. (It is used by the registration CSV and the bulk action.)

https://github.com/sul-dlss/argo/blob/7978ae6b2a3e8a99899a9bd07c61fccf57b19646/app/jobs/register_druids_job.rb#L6
https://github.com/sul-dlss/argo/blob/7978ae6b2a3e8a99899a9bd07c61fccf57b19646/app/jobs/register_druids_job.rb#L61

lwrubel:
@andrewjbtw looks like this work will be covered in @mjgiarlo's held PR https://github.com/sul-dlss/argo/pull/5176, with the exception of the Consul docs update. I'm going to update this ticket to just be the Consul update. How would you want that update timed in relation to the deploy of the registration changes?

andrewjbtw:
Does that PR also cover the downloadable template CSV served from Argo?

lwrubel:
@andrewjbtw The PR did not; I have requested that addition.

lwrubel:
@andrewjbtw Can this now be closed as Complete?

andrewjbtw:
Yes.

PR #5176, opened 2026-07-01
Switch registration from label to title, update Solr field

### LINKED ISSUE #5158: Use title instead of label for Collection and APO display purposes
Anywhere we are displaying a label instead of a title, we should switch to using the title. We should already be using title (via cocina_display) instead of label for DROs, so this work should mainly involve Collections and APO display.

mjgiarlo:
HOLD on merging and deploying to prod until Tuesday, July 7, for PO communication. 

# Why was this change made?

Fixes #5158

This commit severs Argo's dependency on the top-level cocina-models `label`
property. To that end, current uses of label, e.g. in indexing and searching,
have been replaced with uses of appropriate title fields `display_title_ss` for
display and `main_title_tenim` for searching.

HOLD until registration changes within have been communicated to users.

# How was this change tested?

- [x] CI
- [ ] QA/stage

### DIFF
--- a/app/forms/registration_form.rb
+++ b/app/forms/registration_form.rb
@@ -45,7 +45,7 @@ It's legal to have more than one colon in a hierarchy, but at least one colon is
                      prepopulator: ->(*) { (1 - items.count).times { items << VirtualModel.new } } do
     property :source_id, virtual: true
     property :catalog_record_id, virtual: true
-    property :label, virtual: true
+    property :title, virtual: true
     property :barcode, virtual: true
     validates :source_id, format: { with: /\A.+:.+\z/, message: 'ID is invalid' }
     validates :barcode, allow_blank: true, format: { with: VALID_BARCODE_REGEX, message: 'is invalid' }
@@ -106,10 +106,10 @@ It's legal to have more than one colon in a hierarchy, but at least one colon is
   def cocina_model(item)
     model_params = {
       type: content_type,
-      label: item.label,
       version: 1,
       administrative:,
       identification: identification(item),
+      description: { title: [{ value: item.title }] },
       structural:,
       access:
     }
--- a/app/jobs/register_druids_job.rb
+++ b/app/jobs/register_druids_job.rb
@@ -3,7 +3,7 @@
 ##
 # job to register a list of druids
 class RegisterDruidsJob < BulkActionJob
-  HEADERS = ['Druid', 'Barcode', CatalogRecordId.label, 'Source Id', 'Label'].freeze
+  HEADERS = ['Druid', 'Barcode', CatalogRecordId.label, 'Source Id', 'Title'].freeze
 
   def perform_bulk_action
     convert_results.each.with_index do |convert_result, index|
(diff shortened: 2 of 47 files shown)


justinlittman (approved):

lwrubel:
Note this needs to be held until @andrewjbtw has a chance to communicate about registration changes.

app/forms/registration_form.rb
```diff
@@ -106,10 +106,11 @@ def registered_by_tag
   def cocina_model(item)
     model_params = {
       type: content_type,
-      label: item.label,
+      label: '',
```
lwrubel: OK to remove this now, since the cocina-models release just now will create a default value.

mjgiarlo: Got it, thx!

spec/jobs/register_druids_job_spec.rb
```diff
@@ -19,12 +19,16 @@
...
+  let(:description) do
+    Cocina::Models::Description.new(title: [{ value: 'My object' }], purl: 'https://purl.stanford.edu/123')
+  end
 
   let(:cocina_object) do
     instance_double(Cocina::Models::DROWithMetadata,
                     externalIdentifier: 'druid:123',
-                    label: 'My object',
-                    identification:)
+                    label: 'Unused Label',
```
lwrubel: Also can remove here.

mjgiarlo: Got it

spec/system/create_collection_from_apo_spec.rb
```diff
@@ -16,7 +16,8 @@
   let(:result) { { 'response' => { 'numFound' => 1 } } }
   let(:apo_druid) { 'druid:vt333hq2222' }
   let(:cocina_model) do
-    instance_double(Cocina::Models::AdminPolicyWithMetadata, label: 'hey', externalIdentifier: apo_druid)
+    instance_double(Cocina::Models::AdminPolicyWithMetadata, label: '', externalIdentifier: apo_druid,
```
lwrubel: label is not needed here, either. But maybe it would be better if I ticketed making a pass through all of the specs / factories for labels?

mjgiarlo: @lwrubel Thanks! I'm flexible. This PR is likely to hang out for a little bit, so I'll defer to you whether we add the changes to this PR or ticket it.

app/jobs/register_druids_job.rb
lwrubel, line 21: `RegistrationCsvConverter` will also need updating in a few places to take a title not label and include a title not label when creating model_params.

lwrubel:
Also update the downloadable template spreadsheet to use title not label (see https://github.com/sul-dlss/argo/issues/5174)

https://github.com/sul-dlss/argo/blob/affa487fc9f5cc5042910c23a95e39b139f05812/app/controllers/registrations_controller.rb#L80

lwrubel (changes requested):

app/jobs/register_druids_job.rb
mjgiarlo: Good eye!

lwrubel:
It looks like this needs to be updated, too (see https://github.com/sul-dlss/argo/issues/5184):
 https://github.com/sul-dlss/argo/blob/b57a515b483dd0cff989c80d5b38f4cc0db53c64/app/views/bulk_actions/register_druid_jobs/_new.html.erb#L21

lwrubel (approved):
Approved pending final testing in deployed env and Codecov ding.

PR #5193, opened 2026-07-14
skip pagination when downloading reports

### LINKED ISSUE #5192: Report view download csv doesn't include all druids if report is loaded from result page > 1
**Describe the bug**
If you enter reports view from a search result page where the page number > 1, the report view in the browser will show all druids in the results list, but the report download will include only the druids starting from the page you were on when you entered reports view.

For example, if you have 

- 39 results for your search
- navigate to page 4 (results numbering 31-39) in "results" view
- enter "report view"
- download a report

The report will include only the druids that were search results 31-39.

**To Reproduce**
1. Go to [this search that currently returns 39 results](https://argo.stanford.edu/catalog?f%5Bcontent_type_ssimdv%5D%5B%5D=3d&f%5Brights_descriptions_ssimdv%5D%5B%5D=stanford&page=4&sort=id+asc) 
2. Click on page 4 ([direct link](https://argo.stanford.edu/catalog?f%5Bcontent_type_ssimdv%5D%5B%5D=3d&f%5Brights_descriptions_ssimdv%5D%5B%5D=stanford&page=4&sort=id+asc))
3. Go to "select view" on the right and select "report view"
4. See that the report list in the browser shows all results
5. Download a CSV
6. See only results 31-39 in the report

**Expected behavior**
The report should always include the entire result list.

**Additional context**
There was a similar problem with bulk actions in the past: https://github.com/sul-dlss/argo/issues/3349 The issue seems to be including the page parameter in the query that produces the list of druids.

peetucket:
# Why was this change made?

Fixes #5192 - report view should download all results by skipping pagination

Note: Codex written spec.

# How was this change tested?

Tested on stage with https://argo-stage.stanford.edu/catalog?f%5Bwf_wps_ssimdv%5D%5B%5D=accessionWF%3Astage%3Acompleted&page=53 and then switching to report view and downloading.

Before fix, only get 2 results.  After fix, get all results.

Also, new spec

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -300,9 +300,10 @@ class Report # rubocop:disable Metrics/ClassLength
     # Get the underlying Faraday connection to use its streaming API
     connection = search_service.repository.connection.connection
     fl = @fields.collect { |f| f[:solr_fields] || f[:field] }.flatten.uniq.join(',')
+    search_params = @params.except(:page, :per_page, :rows)
     # Setting wt=csv tells solr to return CSV data
     data = { wt: :csv, rows: 10_000_000, fl:, 'csv.mv.separator' => ';' }
-           .reverse_merge(search_service.search_builder.with(@params)).to_h
+           .reverse_merge(search_service.search_builder.with(search_params)).to_h
 
     first_chunk = true
     connection.post blacklight_config.solr_path do |req|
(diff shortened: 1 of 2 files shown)


app/models/report.rb
```diff
@@ -300,9 +300,10 @@ def stream_csv(stream:)
     # Get the underlying Faraday connection to use its streaming API
     connection = search_service.repository.connection.connection
     fl = @fields.collect { |f| f[:solr_fields] || f[:field] }.flatten.uniq.join(',')
+    search_params = @params.except(:page, :per_page, :rows)
```
peetucket, line 303: skip page and per_page params, we don't want them... rows are already hardcoded below

justinlittman (approved):

PR #5200, opened 2026-07-17
Further label removal and changing to title

### LINKED ISSUE #5199: Remove "label" from registration instructions
Should say Title not Label

Image

lwrubel:
# Why was this change made?
Fixes #5199 and further removes label references as much as possible.

# How was this change tested?
Tested registration on stage and ran argo-related integration tests.


mjgiarlo (approved):
🎉

PR #5205, opened 2026-07-21
Drop nested, valueless notes uploaded via descriptive spreadsheet

mjgiarlo:
# Why was this change made?

Fixes #5191 
Fixes #5204 

- **Fix nested date dropping behavior in DescriptionImportFilter**
  - This commit changes description import filtering from a breadth-first approach to a depth-first one, which allows for clearing out a value-less property nested within another value-less property. Before, only the innermost value-less property was dropped, and parents were left alone due to the breadth-first approach. This work should set up the follow-up commit that directly addresses the changes requested in https://github.com/sul-dlss/argo/issues/5191.
- **Drop valueless nested notes and expand filter coverage**
  - Add support for dropping valueless nested notes under access, adminMetadata, contributor, event, subject, title, and structuredValue to ensure only meaningful notes are retained
  - Update MODELS_WITH_NESTED_ATTRIBUTES to include contributor, date, structuredValue, subject, and title for more comprehensive filtering
  - Change filter to use compact_blank! for more aggressive removal of empty attributes, preventing retention of blank hashes or arrays
  - Expand and revise specs to cover new nested note-dropping behavior, verifying that parent nodes are preserved when only the note is valueless
  - Adjust test expectations for access and adminMetadata to reflect new logic of dropping only the empty node rather than rejecting the entire item
- **Drop valueless contributors within `adminMetadata` and drop valueless `structuredValue`s**
  - Consolidate filtering methods for descriptive values to reduce duplication
  - Add filtering for 'name' and update subject, form, identifier, digitalLocation
  - Refactor contributor filtering to use a new helper for sufficiency
  - Add test to ensure valueless adminMetadata contributors are dropped
  - Update spec context descriptions for clarity and consistency
  - Improve maintainability by centralizing sufficiency checks for each type

# How was this change tested?

- [x] CI
- [x] QA/Stage

PR #5207, opened 2026-07-23
Refactor DescriptionImportFilter

mjgiarlo:
# Why was this change made?

As I was working on #5205, I made note of various rough edges or head-scratchers as I went along. Now that #5205 has been shipped, I revisited this class to perform a series of refactors. I would recommend reviewing this commit-by-commit to see the progression. Also, if some of these commits make more sense than others, I'm happy to cherry-pick the stuff we like and ditch the stuff we don't.

## Includes

- **Rename empty property removal methods in DescriptionImportFilter**
- **Remove date-specific handling in DIF class. Handle dates as descriptive values.**
- **Remove one layer of abstraction between attrs and their removal**

# How was this change tested?

CI

### DIFF
--- a/app/services/description_import_filter.rb
+++ b/app/services/description_import_filter.rb
@@ -14,7 +14,7 @@ class DescriptionImportFilter
 
   ATTRIBUTES_TO_FILTER = {
     contributor: :remove_empty_contributors,
-    date: :remove_empty_dates,
+    date: :remove_empty_descriptive_values,
     digitalLocation: :remove_empty_descriptive_values,
     event: :remove_empty_events,
     form: :remove_empty_descriptive_values,
@@ -81,11 +81,6 @@ class DescriptionImportFilter
 
   def remove_empty_languages(languages)
     Array(languages).delete_if { !language_sufficient?(it) }
-  end
-
-  # @param [Array<Hash>] dates an array of hashes that each represent a DescriptiveValue.
-  def remove_empty_dates(dates)
-    Array(dates).delete_if { !descriptive_value_sufficient?(it) }
   end
 
   # Ignore DescriptiveValue that is just "type" or "source"
--- a/app/services/description_import_filter.rb
+++ b/app/services/description_import_filter.rb
@@ -13,17 +13,17 @@ class DescriptionImportFilter
   end
 
   ATTRIBUTES_TO_FILTER = {
-    contributor: :remove_contributors_without_value,
-    date: :remove_dates_without_value,
-    digitalLocation: :remove_descriptive_values_without_value,
-    event: :remove_events_without_value,
-    form: :remove_descriptive_values_without_value,
-    identifier: :remove_descriptive_values_without_value,
-    language: :remove_languages_without_value,
-    name: :remove_descriptive_values_without_value,
-    note: :remove_notes_without_value,
-    structuredValue: :remove_descriptive_values_without_value,
-    subject: :remove_descriptive_values_without_value
+    contributor: :remove_empty_contributors,
+    date: :remove_empty_dates,
+    digitalLocation: :remove_empty_descriptive_values,
+    event: :remove_empty_events,
+    form: :remove_empty_descriptive_values,
+    identifier: :remove_empty_descriptive_values,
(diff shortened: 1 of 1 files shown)


app/services/description_import_filter.rb
```diff
@@ -12,112 +12,44 @@ def self.filter(compacted_params)
     Cocina::Models::Description.new(compacted_params)
   end
 
+  SUFFICIENT_DESCRIPTIVE_VALUE_KEYS = %i[value code uri identifier note valueAt structuredValue parallelValue groupedValue].freeze
```
mjgiarlo: I removed two or three layers of indirection to expose immediately that what distinguishes these attributes is the list of properties that make them sufficient, so we put that right up front in the developer's face.

app/services/description_import_filter.rb
```diff
@@ -12,112 +12,44 @@ def self.filter(compacted_params)
...
-    geographic: Cocina::Models::DescriptiveGeographicMetadata,
-    name: Cocina::Models::DescriptiveValue,
-    relatedResource: Cocina::Models::RelatedResource,
-    structuredValue: Cocina::Models::DescriptiveValue,
-    subject: Cocina::Models::DescriptiveValue,
-    title: Cocina::Models::DescriptiveValue
-  }.freeze
+  MODELS_WITH_NESTED_ATTRIBUTES = %i[
+    access adminMetadata contributor date event form geographic name relatedResource structuredValue subject title
+  ].freeze
```
mjgiarlo: We don't actually *need* to record the Ruby class name associated with an attribute. It is arguably more efficient to do so, at the cost of more stuff the developer has to consult and pack into their brain and remember to update later.

app/services/description_import_filter.rb
```diff
@@ -12,112 +12,44 @@ def self.filter(compacted_params)
...
 
-  # recursive, depth first search for incomplete nodes
-  def filter(compacted_params, model: Cocina::Models::Description)
-    MODELS_WITH_NESTED_ATTRIBUTES.each do |attribute, model|
+  # Recursive, depth-first search for insufficient nodes
+  def filter(compacted_params) # rubocop:disable Metrics/CyclomaticComplexity
```
mjgiarlo: The trade-off of inlining more logic into `#filter` is that we reveal the complexity inherent in this work rather than stuffing it in private methods dozens of lines below.

app/services/description_import_filter.rb
```diff
@@ -53,37 +53,17 @@ def filter(compacted_params, model: Cocina::Models::Description)
     end
 
     ATTRIBUTES_TO_FILTER.each do |attribute, method|
-      next unless model.attribute_names.include?(attribute)
+      next if model.attribute_names.exclude?(attribute)
```
justinlittman, line 57: A comment describing what is happening here would be appreciated.

justinlittman (commented):
I was with you up to the final commit, which for me decreased clarity.

lwrubel:
In the first commit, I wasn't sure about renaming methods to checking if a property is "empty". That makes me think of `empty?` which is not really the case here. While empty values would get filtered out, the primary intent was to avoid incomplete properties that lack "main" values. There could very well be multiple values already in a property, but none are a main value.

But where this ends up by the third commit, just using the `_sufficient?` methods, makes the previous point moot and is a good landing place from my point of view. I'm imagining trying to troubleshoot questions about what happened with a spreadsheet, and I could jump in and figure it out at that point.  Agree the very last refactor gets a bit harder to follow.

app/services/description_import_filter.rb
mjgiarlo: Added a comment and dropped the fourth commit.

mjgiarlo:
@lwrubel 💬 
> But where this ends up by the third commit, just using the `_sufficient?` methods, makes the previous point moot and is a good landing place from my point of view. I'm imagining trying to troubleshoot questions about what happened with a spreadsheet, and I could jump in and figure it out at that point. Agree the very last refactor gets a bit harder to follow.

🙏🏻 I've dropped the fourth commit.

justinlittman (approved):

PR #5208, opened 2026-07-24
derive purl when doing metadata descriptive upload

peetucket:
# Why was this change made?

The user shouldn't need to have a `purl` column when uploading a descriptive metadata spreadsheet, this column can be derived from the druid, which is part of the cocina_object, which we have when updating metadata.

see https://stanfordlib.slack.com/archives/C5RGK4PKM/p1784932056509249

This change passes in the druid to the descriptive import class, and the druid is dervied from the cocina_object, which is available in the job that uses the import class.

HOLD for testing on stage with actual data

See also https://github.com/sul-dlss/argo-b3/pull/234

Note: codex assisted code and spec updates

# How was this change tested?

Specs

### DIFF
--- a/spec/jobs/descriptive_metadata_import_job_spec.rb
+++ b/spec/jobs/descriptive_metadata_import_job_spec.rb
@@ -6,11 +6,11 @@ RSpec.describe DescriptiveMetadataImportJob do
   subject(:job) { described_class.new(bulk_action.id, csv_file:) }
 
   let(:druid) { 'druid:bc123df4567' }
+  let(:purl) { "#{Settings.purl_url}/#{druid.delete_prefix('druid:')}" }
   let(:cocina_object) do
     object = build(:dro_with_metadata, id: druid)
     object.new(description: object.description.new(purl:))
   end
-
   let(:expected_cocina_object) do
     cocina_object.new(description: cocina_object.description.new(title: [{ value: 'new title 1' }], purl:))
   end
@@ -27,7 +27,6 @@ RSpec.describe DescriptiveMetadataImportJob do
     end
   end
 
-  let(:purl) { Cocina::Models::Mapping::Purl.for(druid:) }
   let(:csv_file) do
     [
       'druid,source_id,title1:value',
--- a/spec/jobs/descriptive_metadata_import_job_spec.rb
+++ b/spec/jobs/descriptive_metadata_import_job_spec.rb
@@ -6,10 +6,13 @@ RSpec.describe DescriptiveMetadataImportJob do
   subject(:job) { described_class.new(bulk_action.id, csv_file:) }
 
   let(:druid) { 'druid:bc123df4567' }
-  let(:cocina_object) { build(:dro_with_metadata, id: druid) }
+  let(:cocina_object) do
+    object = build(:dro_with_metadata, id: druid)
+    object.new(description: object.description.new(purl:))
+  end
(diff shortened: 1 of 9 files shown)


spec/jobs/descriptive_metadata_import_job_spec.rb
```diff
@@ -6,10 +6,13 @@
   subject(:job) { described_class.new(bulk_action.id, csv_file:) }
 
   let(:druid) { 'druid:bc123df4567' }
-  let(:cocina_object) { build(:dro_with_metadata, id: druid) }
+  let(:cocina_object) do
+    object = build(:dro_with_metadata, id: druid)
+    object.new(description: object.description.new(purl:))
+  end
```
peetucket: i'm not sure why this change was made by codex... i asked it to explain and it said this, which i don't fully understand either. I vaguely get it, but not enough to explain it to someone else.

```
The Cocina test factory hardcodes a production-style PURL, while the importer now derives the PURL using the environment-configured Cocina::Models::Mapping::Purl. In tests, those values can differ.

Because Cocina models are immutable:
* object.description.new(purl:) copies the description and replaces its PURL.
* object.new(description: ...) copies the object with that updated description.

This ensures the stored test object already has the same canonical PURL the importer will derive, which is especially important for the “Description unchanged” test.
```

justinlittman: I would not get rid of the hardcoding that is already present. It tests that everything is wired together correctly.

spec/services/description_import_spec.rb
```diff
@@ -3,7 +3,32 @@
...
+  let(:purl) { Cocina::Models::Mapping::Purl.for(druid:) }
+
+  context 'with a druid' do
+    let(:druid) { 'druid:bc123df4567' }
+    let(:csv) { CSV.parse("title1.value\nA title\n", headers: true) }
+
+    it 'derives the purl from the druid' do
+      result = described_class.import(csv_row: csv.first, druid:)
+
+      expect(result.value!.purl).to eq Cocina::Models::Mapping::Purl.for(druid:)
```
justinlittman: This should be hardcoded. How is this different from L10.

spec/system/upload_descriptive_metadata_spec.rb
```diff
@@ -22,7 +22,7 @@
 
   context 'with a csv file' do
     let(:csv) do
-      "title1.value,purl\nmy title,https://purl.stanford.edu/#{Druid.new(item.externalIdentifier).without_namespace}\n"
+      "title1.value\nmy title\n"
```
justinlittman: Why are you getting rid of the purl from this spec?

peetucket: I put back, test is green both ways (purl column is just ignored)

spec/services/description_import_spec.rb
peetucket: made `purl` value hardcoded in line 10, reused value here

spec/jobs/descriptive_metadata_import_job_spec.rb
peetucket: Reverted back to hardcoded with a small refactor to avoid duplication (and used Settings.purl_url to construct expected purl url)

justinlittman (approved):

PR #5214, opened 2026-07-29
Render tracking sheets with a UTF-8 capable font

### LINKED ISSUE #5210: Tracking sheets bulk action should support UTF-8
**Describe the bug**
The bulk action to "download tracking sheets" creates a PDF that contains the druids, titles, tags, and a few other metadata fields for a set of items. Titles may contain characters that require UTF-8 support but the tracking sheet function does not appear to support UTF-8.

Example: The title of [hz570fs7419](https://argo.stanford.edu/view/druid:hz570fs7419) is `Saūdi Arabia : special strategic map`
`
The tracking sheet bulk action for this item generates the following error in the log:
```
2026-07-27 21:41:52 UTC Starting TrackingSheetReportJob for BulkAction 41656
2026-07-27 21:41:52 UTC TrackingSheetReportJob creation failed Prawn::Errors::IncompatibleStringEncoding Your document includes text that's not compatible with the Windows-1252 character set.
If you need full UTF-8 support, use external fonts instead of PDF's built-in fonts.
2026-07-27 21:41:52 UTC Finished TrackingSheetReportJob for BulkAction 41656
```
The problem is almost certainly the `ū` character.

**User Impact**
This prevents generating tracking sheets for items with titles outside of the one supported encoding (apparently Windows-1252). Since the tracking sheet bulk action creates a single PDF for a set of items, this prevents generating tracking sheets for any set of items that includes a character outside the supported encoding.

**To Reproduce**
See the example above. Run the "download tracking sheets" bulk action with druid hf409jk1888.

**Expected behavior**
The tracking sheet bulk action should support UTF-8. Any valid title should be supported.

**Additional context**
It's not clear what changed here, as SDR has many titles that likely use UTF-8 and never resulted in an error before. It's possible this problem came in via dependency updates. The tracking sheet function is nearly as old as SDR.

edsu:
# Why was this change made?

Fixes #5210

TrackSheet used Prawn's built-in Courier (AFM) font, which (according to Claude) only supports the Windows-1252 charset. Titles with characters outside it (e.g. Arabic) raised Prawn::Errors::IncompatibleStringEncoding, causing the TrackingSheetReportJob to fail.

Gnu Unifont was selected as a replacement since it is also monospace but has good coverage across entire Basic Multilingual Plane (Latin, Cyrillic, Greek, Arabic, CJK, Hangul, Kana, etc). 

Screenshot 2026-07-29 at 6 18 42 PM

# How was this change tested?

New unit test and acceptance testing in Stage.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


justinlittman (approved):
Interesting.

edsu:
It seems like CJK text isn't rendered at all, so DejaVu Sans may not be sufficient.

edsu:
After some testing @andrewjbtw approved GNU Unifont since its coverage across Unicode was good. It's a bit retro dot-matrix, which isn't entirely inappropriate for the tracking sheets :-)

PR #5226, opened 2026-09-04
Handle CSV registration bulk action errors with no title or catalog record id

lwrubel:
## Why was this change made?
If a CSV was used for registration bulk action and it had a label field, and no title or catalog record id, the job would get stuck at "Processing" and a difficult-to-diagnose error would be raised. See: https://app.honeybadger.io/projects/49894/faults/134080271

Now, if there is a label column and no title (or an empty label column), the following appears in Argo (tested in QA):

Screenshot 2026-09-04 at 10 08 59 AM

Used Claude to analyze and find a solution to provide better feedback:
Here's the full summary covering all 16 files in the branch:

A CSV registration bulk action was submitted with a `label` column where `title` was
expected. The job appeared to hang forever: the Bulk Actions page showed "Processing" with
counts of 0 / 0 / 0, the downloadable log contained only the "Starting RegisterDruidsJob"
line, and Honeybadger reported a misleading
`NoMethodError: undefined method 'output_directory' for nil`.

Three separate problems combined to produce that:

1. **A missing column killed the whole job.** `RegistrationCsvConverter` used
   `row.fetch('title')` (and `fetch` for five other columns), which raises `KeyError`.
   `convert_row` only rescued `Cocina::Models::ValidationError`, so the `KeyError` escaped
   `convert`. It surfaced from `druid_count` inside `BulkActionJob#update_druid_count!` —
   before `perform_bulk_action` runs — so the per-row rescue never got a chance and the bulk
   action was left on `status: 'Processing'` with nothing useful logged. Sidekiq then retried
   it, failing identically every time.

2. **Nothing rejected the CSV at upload.** `RegisterDruidJobsController::REQUIRED_HEADERS` did
   not require a title, so a title-less CSV was accepted. The registration page
   (`CsvRegistrationForm`) did check for `title` or a catalog record id; the bulk action route
   did not. Neither path said anything about `label` specifically, which is the actual mistake
   people make — `label` has been removed from Cocina, but template files referencing it are
   still in circulation.

3. **A keyword mismatch broke row-level error reporting.** `RegisterDruidsJob#perform_bulk_action`
   called `failure!(druid:, message:, index:)`, but `RegisterDruidsJob` inherits from
   `BulkActionJob`, whose `failure!` takes only `druid:` and `message:` — the `index:` keyword
   only exists on `BulkActionCsvJob`. Any exception raised while registering a row therefore
   raised `ArgumentError: unknown keyword: :index` from inside the rescue, stranding the bulk
   action on "Processing" with `druid_count_fail: 0` — the same symptom as (1), from a
   different trigger.

The `NoMethodError` in Honeybadger was a masking error, not the underlying cause. When
`perform` fails before `@bulk_action` is assigned, its `ensure` block calls `export_file&.close`
→ `report_filename` → `bulk_action.output_directory` on `nil`. An exception raised in `ensure`
replaces the in-flight exception, so the real error survives only as `#cause`. (Not addressed
here — see follow-ups.)

## Changes

**`RegistrationCsvConverter`** — added a `MissingColumnError` and a `fetch_column` helper that
raises it, replacing all six `row.fetch` calls (`title`, `source_id`, `content_type`,
`administrative_policy_object`, `initial_workflow`, `rights_location`). `convert_row` rescues it
alongside `ValidationError`, so a missing column is now a per-row `Failure` reading
`Missing required column: title` rather than an exception that kills the job. A column that is
present but empty still returns `nil` as before; only an absent header fails.

**Upload validation, in both paths** — `RegisterDruidJobsController` and `CsvRegistrationForm`
now share two checks, worded identically so the same mistake reports the same way regardless of
which upload path is used:

- `label_header_errors` — flags a `label` column, case-insensitively:
  `has a "label" column, which is not valid (titles must be in a column named "title")`
- `missing_title_errors` — requires `title` or `CatalogRecordId.csv_header` to hold data in
  every row: `is missing a title (each row needs a value in title or folio_instance_hrid)`.
  Because a missing column reads as blank for every row, one check covers both the
  absent-header and blank-row cases; this replaces the form's previous if/elsif pair and lets
  its `Metrics/CyclomaticComplexity` / `PerceivedComplexity` disable comment come off.

To support the controller side, `CreatesBulkActions#validate_csv_headers` now forwards an
optional block to `CsvUploadValidator#valid?`, which already accepted one for additional checks.
The block is optional, so the other nine callers are unchanged.

**Error message rendering** — both messages are phrased as clauses rather than sentences,
because `bulk_actions/_errors.html.erb` renders `@errors.to_sentence`; full sentences produced
`". and "` mid-message. For the same reason, `CsvUploadValidator`'s missing-headers error no
longer ends in a period, and `CsvRegistrationForm` now joins with `to_sentence` instead of
`join(' ')` so the registration page and the bulk action page read the same way. A CSV with a
`label` column and no title now reports:

> has a "label" column, which is not valid (titles must be in a column named "title") and is
> missing a title (each row needs a value in title or folio_instance_hrid)

**`RegisterDruidsJob`** — `success!`/`failure!` now accept `index:` and prefix log messages with
`" - line #{index} - "`, following `BulkActionCsvJob`/`BulkActionCsvJobItem`.
`RegisterDruidsJobItem` passes `index:` through, so item-level failures get line numbers too
(they previously had none). Iteration uses `each.with_index(2)` so the index is the real CSV line
number. Since every row logs as `Unregistered`, the line number is the only thing distinguishing
one failure from another.

# How was this change tested?

New and updated specs, QA

### DIFF
--- a/app/controllers/bulk_actions/register_druid_jobs_controller.rb
+++ b/app/controllers/bulk_actions/register_druid_jobs_controller.rb
@@ -25,14 +25,24 @@ module BulkActions
 
     def validate_job_params(job_params)
       validate_csv_headers(job_params.fetch(:csv_file), REQUIRED_HEADERS) do |csv|
-        if ADDITIONAL_HEADERS.none? { |header| csv.headers.include?(header) }
-          ["missing header. One of these must be provided: #{ADDITIONAL_HEADERS.join(', ')}"]
-        elsif csv.any? { |row| ADDITIONAL_HEADERS.none? { |header| row[header].present? } }
-          ["missing data. For each row, one of these must be provided: #{ADDITIONAL_HEADERS.join(', ')}"]
-        else
-          []
-        end
+        label_header_errors(csv) + missing_title_errors(csv)
       end
+    end
+
+    private
+
+    def label_header_errors(csv)
+      # "label" has been removed from Cocina but users may still have template files referencing it.
+      return [] if csv.headers.none? { |header| header&.casecmp?('label') }
+
+      ['has a "label" column, which is not valid. Titles must be in a column named "title".']
+    end
+
+    # A missing column reads as blank data for every row, so this covers an absent header too.
+    def missing_title_errors(csv)
+      return [] if csv.all? { |row| ADDITIONAL_HEADERS.any? { |header| row[header].present? } }
+
+      ["missing title. For each row, one of these must be provided: #{ADDITIONAL_HEADERS.join(', ')}"]
     end
   end
 end
--- a/app/controllers/bulk_actions/register_druid_jobs_controller.rb
+++ b/app/controllers/bulk_actions/register_druid_jobs_controller.rb
@@ -15,12 +15,24 @@ module BulkActions
       rights_download
     ].freeze
 
+    # At least one of these is required: a title, or a catalog record id to derive one from.
+    # NOTE: Equivalent validation for the registration page is in CsvRegistrationForm.
+    ADDITIONAL_HEADERS = ['title', CatalogRecordId.csv_header].freeze
+
     def job_params
       { groups: current_user.groups, csv_file: CsvUploadNormalizer.read(params[:csv_file].path) }
     end
 
     def validate_job_params(job_params)
-      validate_csv_headers(job_params.fetch(:csv_file), REQUIRED_HEADERS)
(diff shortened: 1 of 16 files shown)


mjgiarlo (approved):

justinlittman:
@lwrubel Does this need to be ported to argo-b3?

